PR #3, opened 2018-08-08
Discover user's home directory properly

sidharthsurana:
The current way "~" does not work on a mac


sidharthsurana:
Aborting this PR as this change actually then breaks the linux env since `user.Current` needs cgo on Linux. Since we cross compile the binaries, cgo on max is not available causing `Current not implemented on linux/amd64` error. Although this can be remediated by building the binaries on a linux env, but that is not a great solution either.

PR #14, opened 2018-08-21
Update the CNI to weave

### LINKED ISSUE #25: Change the default CNI to weave
Request to change the default CNI to weave as it does not need an additional etcd, and it supports much richer network constructs.

The choice of weave for CNI is based on some available comparison here:
https://chrislovecnm.com/kubernetes/cni/choosing-a-cni-provider/

sidharthsurana:
In the near future we may make selection of CNI available to the users
However, we wanted to select a better default CNI at the moment

The choice of weave for CNI is based on some available comparison here:
https://chrislovecnm.com/kubernetes/cni/choosing-a-cni-provider/

Another factor for choosing weave here is that you don't need to run
an additional etcd for the CNI

Resolves #25

### DIFF
--- a/cloud/vsphere/templates.go
+++ b/cloud/vsphere/templates.go
@@ -346,457 +346,209 @@ EOF
 
 kubeadm init --config /etc/kubernetes/kubeadm_config.yaml
 
-# install calico
-cat > /tmp/calico.yaml << EOF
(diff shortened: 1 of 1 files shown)


sflxn:
Let's put this in our meeting agenda for Monday and see if there are any objections.  In the future, we should write up an issue first so the community can debate it.

sflxn:
lgtm

PR #37, opened 2018-09-04
Add the types for the provider status fields

### LINKED ISSUE #36: Add provider specific cluster controller for vsphere provider
The "API Endpoint" for the cluster ideally should be handled by the provider specific cluster controller. Without this we are dependent on the clusterctl client to drive the workflow as currently the cli updates the API Endpoint at the moment. 

We need to add a cluster controller for the vsphere implementation and have that maintain the "API Endpoint" for the cluster

This topic has been discussed in the following places:
https://github.com/kubernetes-sigs/cluster-api/issues/158
https://github.com/kubernetes-sigs/cluster-api/pull/467#issuecomment-413692314

sidharthsurana:
Add the types for the provider status fields
Adds the following status types
 * VsphereClusterProviderStatus
 * VsphereMachineProviderStatus

Single field called `LastUpdated` is added to both the types. This
will be used to track the time stamp if the last set of changes
done on the vsphere infrastructure

Partially Fixes #36


sflxn:
lgtm

PR #42, opened 2018-09-12
Add vsphere cluster controller

### LINKED ISSUE #36: Add provider specific cluster controller for vsphere provider

sidharthsurana:
The basic cluster controller implementation that reconciles the cluster object adding the `APIEndpoint` address in the status field.

This change introduces the skeleton for the workflow that in the future could handle the creation/management of the LB for the API Endpoints.

Fixes #36


cloud/vsphere/machineactuator.go
```diff
@@ -739,17 +783,57 @@ func getKubeadmToken() (string, error) {
 	return strings.TrimSpace(output), err
 }
 
+func (vc *VsphereClient) getMasterForCluster(cluster *clusterv1.Cluster) (*clusterv1.Machine, error) {
+	machines, err := vc.lister.Machines().Lister().Machines(cluster.Namespace).List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+	for _, machine := range machines {
+		//glog.Infof("%v", machine)
```
sflxn: remove commented log message if you don't want to log.

cloud/vsphere/machineactuator.go
```diff
@@ -739,17 +783,57 @@ func getKubeadmToken() (string, error) {
...
+		return "", err
+	}
+	err = ioutil.WriteFile(f.Name(), []byte(contents), 0644)
+	if err != nil {
+		return "", err
+	}
+	return f.Name(), nil
+}
+
+func ifErrRemove(pErr *error, path string) {
```
sflxn: you probably should make this an inline function in the above function since it's very specific to the temp file you're creating.

defer func() {
}()

cloud/vsphere/machineactuator.go
```diff
@@ -739,17 +783,57 @@ func getKubeadmToken() (string, error) {
...
+	return nil, fmt.Errorf("No master node found for the cluster %s", cluster.Name)
+}
+
+func createTempFile(contents string) (string, error) {
+	f, err := ioutil.TempFile("", "")
+	if err != nil {
+		return "", err
+	}
+	defer ifErrRemove(&err, f.Name())
+	if err = f.Close(); err != nil {
```
sflxn: you shouldn't close the file here.  f is already an open os.File.  Just write to it.  You should defer the close here.

cloud/vsphere/utils.go
```diff
@@ -0,0 +1,83 @@
...
+
+// GetMasterForCluster returns the master nodes for the given cluster
+func GetMasterForCluster(cluster *clusterv1.Cluster, lister v1alpha1.Interface) ([]*clusterv1.Machine, error) {
+	masters := make([]*clusterv1.Machine, 0)
+	machines, err := lister.Machines().Lister().Machines(cluster.Namespace).List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+	for _, machine := range machines {
+		//glog.Infof("%v", machine)
```
sflxn: you should remove this comment if you don't want any logging.

cloud/vsphere/utils.go
```diff
@@ -0,0 +1,83 @@
...
+		if ip, ok := machine.ObjectMeta.Annotations[VmIpAnnotationKey]; ok {
+			glog.Infof("Returning IP from machine annotation %s", ip)
+			return ip, nil
+		}
+	}
+	return "", errors.New("could not get IP")
+}
+
+func GetMachineProviderStatus(machine *clusterv1.Machine) (*vsphereconfig.VsphereMachineProviderStatus, error) {
+	if machine.Status.ProviderStatus != nil {
```
sflxn: hey, there is a Go idiom that basically says to return as soon as possible so that you don't form complex if-thens.  By checking for the valid case and creating a big if-then, you're going against the idiom.  I'd keep everything inside the if here, but structure this if-then as follows,

```
if machine.Status.ProviderStatus == nil {
  return nil, nil
}
```
...

cloud/vsphere/utils.go
```diff
@@ -0,0 +1,83 @@
...
+		if !ok {
+			return nil, fmt.Errorf("machine providerstatus failure to cast to vsphere; type: %v", gvk)
+		}
+		return status, nil
+	}
+	return nil, nil
+}
+
+func GetClusterProviderStatus(cluster *clusterv1.Cluster) (*vsphereconfig.VsphereClusterProviderStatus, error) {
+	if cluster.Status.ProviderStatus != nil {
```
sflxn: Same comment as the one I posted for MachineProviderStatus.

sidharthsurana:
@sflxn Thanks for the review, I have updated the PR to address your comments

cloud/vsphere/machineactuator.go
```diff
@@ -739,17 +783,53 @@ func getKubeadmToken() (string, error) {
...
+	return nil, fmt.Errorf("No master node found for the cluster %s", cluster.Name)
+}
+
+func createTempFile(contents string) (string, error) {
+	tmpFile, err := ioutil.TempFile("", "")
+	if err != nil {
+		glog.Warningf("Error creating temporary file")
+		return "", err
+	}
+	defer os.Remove(tmpFile.Name())
```
sflxn: Are you sure you want to remove the file?  Don't you want to do the following?

```
defer func(err error) {
  if err != nil {
      os.Remove(tmpFile.Name())
  }
} (err)
```

PR #43, opened 2018-09-13
Enable the dep pruning option to keep the vendor dir slim

sidharthsurana:
Add the Prune section in the Gopkg.toml
Run the "make build" that would run "dep ensure" removing the unused
parts from the vendor directory


frapposelli:
`−2,545,625` :+1:

/lgtm
/approve

PR #58, opened 2018-09-19
Refactor template.go out of terraform to be used as common code

### LINKED ISSUE #57: Refactor template.go out of terraform provisioner
Now that we are adding a govmomi based implementation as well, the script templates as they stand today should be used by both govmomi and terraform implementations. Thus refactoring the template.go is needed to be consumable with both implementations

sidharthsurana:
Now that we are adding a govmomi based implementation as well, the script templates as they stand today should be used by both govmomi and terraform implementations. Thus refactoring the template.go to be consumable with both implementations

Resolves #57

### DIFF
--- a/cloud/vsphere/provisioner/common/templates.go
+++ b/cloud/vsphere/provisioner/common/templates.go
@@ -14,17 +14,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package terraform
+package common
 
 import (
 	"bytes"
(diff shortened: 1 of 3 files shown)


sflxn:
@sidharthsurana you should rebase against top of tree now that the first commit has been merged.  Thast should resolve the conflicts.  Once that conflict is resolve @frapposelli can ok to test and approve it.

/lgtm

PR #63, opened 2018-10-02
Add the APIStatus field in the Cluster ProviderStatus field

### LINKED ISSUE #62: Make Cluster object provide the status of the realized cluster
The current `ClusterStatus` object does not contain any high level field that someone could look at and tell if the target cluster is `ready` for consumption. Given that the very definition of `ready` is subjective. As we could claim ready potentially at different stages:
1. master/s are ready and the APIs are up
2. workers are all deployed and nodes are in ready state

This is an open issue to be discussed in the cluster-api SIG at this moment.
Given that, I propose to further enrich the `ProviderStatus` field in the `Cluster` to track part of this status. Given the definition of `Ready`  for the cluster is not formalized, thus the proposal is to add `APIStatus` field in the `ProviderStatus` representing whether the kubernetes APIs are ready to be interacted with in the target cluster.

sidharthsurana:
* Adds a new field "APIStatus" in the ProviderStatus for Cluster type
* Adds logic in the cluster controller to keep the APIStatus up to date

Resolves #62

### DIFF
--- a/cloud/vsphere/clusteractuator.go
+++ b/cloud/vsphere/clusteractuator.go
@@ -70,7 +70,7 @@ func (vc *ClusterActuator) Reconcile(cluster *clusterv1.Cluster) error {
 		glog.Infof("Error setting the Load Balancer members for the cluster: %s", err)
 		return err
 	}
-	// Check if the target kuberenetes is ready or not, and update the ProviderStatus if change is detected
+	// Check if the target kubernetes is ready or not, and update the ProviderStatus if change is detected
 	err = vc.updateK8sAPIStatus(cluster)
 	if err != nil {
 		return err
@@ -90,11 +90,11 @@ func (vc *ClusterActuator) getClusterAPIStatus(cluster *clusterv1.Cluster) (vsph
 	masters, err := vsphereutils.GetMasterForCluster(cluster, vc.lister)
 	if err != nil {
 		glog.Infof("Error retrieving master nodes for the cluster: %s", err)
-		return vsphereconfig.NotReady, err
+		return vsphereconfig.ApiNotReady, err
 	}
 	if len(masters) == 0 {
 		glog.Infof("No masters for the cluster [%s] present", cluster.Name)
-		return vsphereconfig.NotReady, nil
+		return vsphereconfig.ApiNotReady, nil
 	}
 	// Currently we support only a single master thus the below assumption
 	// Once we start supporting multiple masters, the kubeconfig needs to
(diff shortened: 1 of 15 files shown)


cloud/vsphere/clusteractuator.go
```diff
@@ -61,6 +70,91 @@ func (vc *ClusterActuator) Reconcile(cluster *clusterv1.Cluster) error {
 		glog.Infof("Error setting the Load Balancer members for the cluster: %s", err)
 		return err
 	}
+	// Check if the target kuberenetes is ready or not, and update the ProviderStatus if change is detected
```
frodenas: nit: `kubernetes` not `kuberenetes` :)

cloud/vsphere/clusteractuator.go
```diff
@@ -61,6 +70,91 @@ func (vc *ClusterActuator) Reconcile(cluster *clusterv1.Cluster) error {
...
+	err = vc.updateK8sAPIStatus(cluster)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (vc *ClusterActuator) updateK8sAPIStatus(cluster *clusterv1.Cluster) error {
+	currentClusterAPIStatus, err := vc.getClusterAPIStatus(cluster)
```
frodenas: Is this call necessary here?

The below call to `updateClusterAPIStatus` will also [call](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/63/files#diff-437bea91b22d029880c4caf3e100903bR135) `getClusterAPIStatus`, and this will incur in a double ssh to get the kubeconfig. 

Also, by returning an error here, it means the status will not be updated when it's created.

cloud/vsphere/clusteractuator.go
```diff
@@ -61,6 +70,91 @@ func (vc *ClusterActuator) Reconcile(cluster *clusterv1.Cluster) error {
...
+}
+
+func (vc *ClusterActuator) updateClusterAPIStatus(cluster *clusterv1.Cluster, newStatus vsphereconfig.APIStatus) error {
+	oldProviderStatus, err := vsphereutils.GetClusterProviderStatus(cluster)
+	if err != nil {
+		return err
+	}
+	currentAPIStatus, err := vc.getClusterAPIStatus(cluster)
+	if err != nil {
+		return err
```
frodenas: Do you really want to return the error here? If the status is `NotReady`, wouldn't it be better to update the status with this condition?

cloud/vsphere/clusteractuator.go
```diff
@@ -61,6 +70,91 @@ func (vc *ClusterActuator) Reconcile(cluster *clusterv1.Cluster) error {
...
+	}
+	if oldProviderStatus != nil && oldProviderStatus.APIStatus == currentAPIStatus {
+		// Nothing to update
+		return nil
+	}
+	newProviderStatus := &vsphereconfig.VsphereClusterProviderStatus{}
+	// create a copy of the old status so that any other fields except the ones we want to change can be retained
+	if oldProviderStatus != nil {
```
frodenas: Technically this is not needed, because you have already checked the `oldProviderStatus` a few lines before.

sidharthsurana:
Thanks for the review @frodenas I have addressed your comments in the new patch. Please take a look again.

cloud/vsphere/clusteractuator.go
frodenas: Got it, my fault, I read the previous check as an OR, when it's an AND.

frodenas:
@sidharthsurana These changes have been merged as part of PR https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/64, right?

sidharthsurana:
> @sidharthsurana These changes have been merged as part of PR #64, right?

Yes, it did. PR#64 was stacked on top of this one

PR #71, opened 2018-10-11
Add a new MachineSpec for defining the instance on vsphere

### LINKED ISSUE #24: Feature: Add multiple nic support
Add support multiple nics on to the node machines.
This will help in setting the cluster in a way that the k8s API network and the data plane network could be possibly isolated

### LINKED ISSUE #28: Feature:  support static IP in machine definition
There are times when there is a desire to create machine nodes with static IPs.

sidharthsurana:
With the new structure we should be able to:
* Support defining multiple nics
* Support providing dhcp/static IP configuration for each nic
* Support customizing multiple disks if present
* Adds a optional flag called `vsphereCloudInit` to indicate if
  the template used has the vsphere datasource for cloud-init
  present and enabled (see [vsphere cloud-init datasource](https://github.com/akutz/cloud-init-vmware-guestinfo/))

Note: This change does not change the existing behavior of using
the `MachineVariables` string map. This will be handled in the
subsequent patches.

Partially Resolves #24
Partially Resolves #28

### DIFF
--- a/cmd/clusterctl/examples/vsphere/machines.yaml.template
+++ b/cmd/clusterctl/examples/vsphere/machines.yaml.template
@@ -22,6 +22,27 @@ items:
           disk_label: ""
           disk_size: ""
           virtual_machine_domain: ""
+        # machineSpec:
+        #   datacenter: ""
+        #   datastore: ""
+        #   resourcePool: ""
+        #   networks:
+        #   - networkName: ""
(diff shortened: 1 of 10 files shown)


frapposelli:
Please rebase the PR 🙏

sidharthsurana:
@frapposelli done! This PR is now good to be reviewed and merged

PR #75, opened 2018-10-12
Use release version dep tool

figo:
go get dep will install HEAD version which is not
what we want.

/assign @frapposelli

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,12 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Keep an existing GOPATH, make a private one if it is undefined
+GOPATH_DEFAULT := $(PWD)/.go
+export GOPATH ?= $(GOPATH_DEFAULT)
+GOBIN_DEFAULT := $(GOPATH)/bin
+export GOBIN ?= $(GOBIN_DEFAULT)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -12,12 +12,25 @@
...
+GOBIN_DEFAULT := $(GOPATH)/bin
+export GOBIN ?= $(GOBIN_DEFAULT)
+HAS_DEP := $(shell command -v dep;)
+
 .PHONY: gendeepcopy
 
 all: generate build images
 
-depend:
-	dep version || go get -u github.com/golang/dep/cmd/dep
```
sidharthsurana: Seems like your objective is still to test if the user has dep installed. if they have use it otherwise install dep and use that.
Given that is the same intent prior to this change, wouldn't it be easier if we just did this 
`dep version || curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh` 
What is the benefit of having a private GOPATH? 
In fact, if we do create the private gopath, and install the latest dep there, this script will work but after that if the user tries `dep ensure` or any other dep command they will not have dep causing confusion.  They might end up installing dep possibly a different version than the release one causing issues.  As the next run of this same file will yield different results.

PR #77, opened 2018-10-15
Fix govmomi actuator's Create when using clusterctl

### LINKED ISSUE #70: Bug: clusterctl create cluster fails with govmomi actuator
Using clusterctl, cluster creation always fails.  After some debugging, it appears kubeconfig cannot be pulled.  Root cause still unknown.

sflxn:
Clusterctl create cluster was not able to properly create the target cluster.  It would sit
in an infinite loop, attempting to clone the VM template.  The fix was to not use annotations
for task ref and vm ref during create.  Instead, use provider status.  Once this fix was in,
the Target cluster would then sit in an infinite loop, attempting to clone the VM template.
The fix was to check for nodeRef in the actuator's Exists() function.

In addition, moved utils from provisioner/common to provisioner/govmomi as it's only ever
used by the govmomi actuator.

Fixes #70

### DIFF
--- a/cloud/vsphere/provisioner/govmomi/create.go
+++ b/cloud/vsphere/provisioner/govmomi/create.go
@@ -26,7 +26,7 @@ import (
 )
 
 func (vc *Provisioner) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
-	glog.Info("Inside Govmomi provisioner create method")
+	glog.Infof("govmomi.Actuator.Create %s", machine.Spec.Name)
 	s, err := vc.sessionFromProviderConfig(cluster, machine)
 	if err != nil {
 		return err
@@ -60,7 +60,7 @@ func (vc *Provisioner) verifyAndUpdateTask(s *SessionContext, machine *clusterv1
 	if err != nil {
 		//TODO: inspect the error and act appropriately.
 		// Naive assumption is that the task does not exist any more, thus clear that from the machine
-		return vc.removeTaskRef(machine)
+		return vc.setTaskRef(machine, "")
 	}
 	switch taskmo.Info.State {
 	// Queued or Running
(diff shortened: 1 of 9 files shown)


cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -264,41 +268,72 @@ func (vc *Provisioner) removeTaskRef(machine *clusterv1.Machine) error {
...
+	}
+	return nil
 }
 
-// Sets the current task reference on the Machine object
-func (vc *Provisioner) setTaskRef(machine *clusterv1.Machine, taskref types.ManagedObjectReference) error {
-	nmachine := machine.DeepCopy()
-	if nmachine.ObjectMeta.Annotations == nil {
-		nmachine.ObjectMeta.Annotations = make(map[string]string)
+func (vc *Provisioner) setTaskRef(machine *clusterv1.Machine, taskref string) error {
```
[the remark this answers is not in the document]
sflxn: not really.  most of the code is the same, but it modifies a different field.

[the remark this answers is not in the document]
figo: you are right, one is updating`MachineRef `, and another is updating `TaskRef `, i think it make sense to has a common function to update ProviderStatus on different field.

cloud/vsphere/provisioner/govmomi/exists.go
```diff
@@ -3,13 +3,19 @@ package govmomi
...
+	"github.com/golang/glog"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 	vsphereutils "sigs.k8s.io/cluster-api-provider-vsphere/cloud/vsphere/utils"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 )
 
 func (vc *Provisioner) Exists(cluster *clusterv1.Cluster, machine *clusterv1.Machine) (bool, error) {
+	glog.Infof("govmomi.Actuator.Exists %s", machine.Spec.Name)
+	if machine.Status.NodeRef != nil {
```
sidharthsurana: I know you added this to resolve the issue with `clusterctl` CLI getting hung.. 
But IMHO in our provider's `Exists` method we should be truly checking the existence of the resource on the provider infrastructure and should be reporting on that. The `NodeRef` property is neither added nor controlled by the actuator, thus the actuator exists should not be using this check. 
As for the `clusterctl` CLI issue, I am still not sure how this check is helping, since the `NodeRef` does not get set until the pivot has happened, and the actuator's `Exists` method must have to return `true` for the master node before we could have ever triggered the pivot operation.
I think we should be fixing the code making up the `clusterctl` CLI that has logic making certain improper assumptions to begin with.

cloud/vsphere/provisioner/govmomi/update.go
```diff
@@ -65,6 +68,7 @@ func (vc *Provisioner) updateIP(cluster *clusterv1.Cluster, machine *clusterv1.M
 	if nmachine.ObjectMeta.Annotations == nil {
 		nmachine.ObjectMeta.Annotations = make(map[string]string)
 	}
+	glog.Infof("updateIP - IP = %s", vmIP)
```
sidharthsurana: switch to using `glog.V(4).Infof(...` so that we can selectively emit more or less logs

cloud/vsphere/provisioner/govmomi/update.go
```diff
@@ -48,12 +50,13 @@ func (vc *Provisioner) Update(cluster *clusterv1.Cluster, machine *clusterv1.Mac
 	}
 
 	if _, err := vsphereutils.GetIP(cluster, machine); err != nil {
+		glog.Info("actuator.Update() - did not find IP, waiting on IP")
```
sidharthsurana: switch to using `glog.V(4).Infof(...` so that we can selectively emit more or less logs

cloud/vsphere/provisioner/govmomi/utils.go
```diff
@@ -1,4 +1,4 @@
-package common
```
sidharthsurana: The main reason I had it separate was if in future we have more that 1 actuator implementations (or `variants` implementations) then this would serve as the common code that could be reused across. Essentially terraform actuator could have used this as well but since we were going to remove that anyways, thus I did not make changes in that to use this common code.
Anyways, It's fine for now but later on we may still end up re-creating something like this again based on the different `variant` implementations.

PR #86, opened 2018-10-19
Move the MachineRef property into the ProviderConfig

### LINKED ISSUE #82: Make the vm-moid reference transferable during pivot
In the recent change [PR#77](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/77) we moved the `vm-moid` property from `annotations` to `status` field. This has a side effect such that now when the pivot happens, the `status` field does not get transferred, thus on the target cluster the machine controller  does not know which is the actual vsphere vm backing the machine object. This will cause failures in the delete as the controller will not be able to know which VM to delete. We need to move the `vm-moid` reference out from the status into either an annotation or possibly the spec of the machine itself to make it transferable when we do the pivot

### LINKED ISSUE #85: Use only the types from the v1alpha1 version
Currently our code uses a mix of "v1alpha1" versioned resource and the release versioned resource present under the parent directory of the "v1alpha1".  Since currently the project is in "v1alpha1" release we should only be using the types from "v1alpha1" release.

Ideally we should remove the duplicate type information currently present under the parent directory  of the "v1alpha1" folder.

sidharthsurana:
* Move the `MachineRef` property from `ProviderStatus` to `ProviderConfig` object
* Upon creation of the `VirtualMachine` the `MachineRef` will be populated inside the spec
* Update the glog logs to use higher verbosity for improved readability of logs

Resolves #82

Change-Id: I4ef1c5dec11d0c06dd777daae6500f4fe6f13241

### DIFF
--- a/pkg/apis/vsphereproviderconfig/types.go
+++ b/pkg/apis/vsphereproviderconfig/types.go
@@ -26,6 +26,7 @@ type VsphereMachineProviderConfig struct {
 
 	// Name of the machine that's registered in the NamedMachines ConfigMap.
 	VsphereMachine string `json:"vsphereMachine"`
+	MachineRef     string `json:"machineRef,omitempty"`
 	// List of variables for the chosen machine.
 	MachineVariables map[string]string `json:"machineVariables"`
 }
@@ -44,7 +45,6 @@ type VsphereMachineProviderStatus struct {
 	metav1.TypeMeta `json:",inline"`
 
 	LastUpdated string `json:"lastUpdated"`
-	MachineRef  string `json:"machineRef"`
 	TaskRef     string `json:"taskRef"`
 }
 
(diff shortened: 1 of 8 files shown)


sidharthsurana:
/hold
Seems like after the rebase has caused some issues. Investigating right now, so hold off on reviews.

sidharthsurana:
Updated the PR, now should be good to review

sidharthsurana:
Seems like the commit in this PR already got merged as part of the commits in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/71  Since I had stacked them in order to specify dependencies acoss PRs. But pull requests on github don't seem to be well structured to express dependencies across multiple PRs. For instance in a gerrit based system, the merge of #71 would have stayed as pending until this one was merged.
Thus at this point I am going to close this PR as it is doesn't add any new thing.

PR #92, opened 2018-10-26
Add goflag parsing to pflags

### LINKED ISSUE #91: Missing glog flags in the built controller binary
Currently the --v flag is missing from the built controller binary. Thus not allowing to set the glog log level and other options

sidharthsurana:
This adds the options related to glog on the command line like `--v=4`

Resolves #91

Change-Id: I0b4756b63d4581e1560e15568b03f380eea1d9bb

### DIFF
--- a/cmd/vsphere-machine-controller/main.go
+++ b/cmd/vsphere-machine-controller/main.go
@@ -17,6 +17,8 @@ limitations under the License.
 package main
 
 import (
+	"flag"
+
 	"github.com/golang/glog"
 	"github.com/kubernetes-incubator/apiserver-builder/pkg/controller"
 	"github.com/spf13/pflag"
@@ -48,6 +50,8 @@ func init() {
 }
 
 func main() {
+	// Map go flags to pflag
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
 
 	logs.InitLogs()


sidharthsurana:
The new set of flag can be seen here with the change included
```
Usage of ./vsphere-machine-controller:
      --alsologtostderr                        log to standard error as well as files
      --kubeconfig string                      Path to kubeconfig file with authorization and master location information.
      --leader-elect                           Start a leader election client and gain leadership before executing the main loop. Enable this when running replicated components for high availability.
      --leader-elect-lease-duration duration   The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire leadership of a led but unrenewed leader slot. This is effectively the maximum duration that a leader can be stopped before it is replaced by another candidate. This is only applicable if leader election is enabled. (default 15s)
      --leader-elect-renew-deadline duration   The interval between attempts by the acting master to renew a leadership slot before it stops leading. This must be less than or equal to the lease duration. This is only applicable if leader election is enabled. (default 10s)
      --leader-elect-resource-lock endpoints   The type of resource object that is used for locking during leader election. Supported options are endpoints (default) and `configmap`. (default "endpoints")
      --leader-elect-retry-period duration     The duration the clients should wait between attempting acquisition and renewal of a leadership. This is only applicable if leader election is enabled. (default 2s)
      --log-flush-frequency duration           Maximum number of seconds between log flushes (default 5s)
      --log_backtrace_at traceLocation         when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                         If non-empty, write log files in this directory
      --logtostderr                            log to standard error instead of files (default true)
      --namedmachines string                   path to named machines yaml file
      --stderrthreshold severity               logs at or above this threshold go to stderr (default 2)
  -v, --v Level                                log level for V logs
      --vmodule moduleSpec                     comma-separated list of pattern=N settings for file-filtered logging
      --workers int                            The number of workers for controller. (default 5)
pflag: help requested
```

PR #96, opened 2018-10-26
Misc fixes

### LINKED ISSUE #94: Waiting condition for the worker nodes should be the APIStatus "Ready"
Recently we added a `APIStatus` field that tracks the status of the target k8s cluster's API status. The status turns into `Ready` once we are able to interact with the target API server. This status is what we should be waiting on when creating worker nodes as they need to generate the kubeadm token prior to cloning.
Prior to the `APIStatus` addition we used to wait on the `APIEndpoint` to show up. However, the presence of `APIEndpoint` merely does not mean the APIs are ready on the target cluster.
Thus the wait condition should be updated to depend on `APIStatus` instead of `APIEndpoint`

### LINKED ISSUE #95: Investigate failure during caching the kubeadm token
Currently when creating multiple workers are the same time, in the logs we can see the following errors consistently
`Operation cannot be fulfilled on clusters.cluster.k8s.io "test-cluster": the object has been modified; please apply your changes to the latest version and try again`

Needs investigation and possible fix

sidharthsurana:
* Updates the wait condition for the worker node to `APIStatus`
* Fix the kubeadm token caching issue by adding a jitter in the re-queue interval
* Fix few minor typos

   Resolves #94
   Resolves #95

   Change-Id: Ic3aaee93f807ebf4fd8ca3964d46510be627c2b2


pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -506,14 +511,19 @@ func (pv *Provisioner) getStartupScript(cluster *clusterv1.Cluster, machine *clu
...
-			glog.Infof("Waiting for Kubernetes API Endpoint to be populated..Retrying in %s", constants.RequeueAfterSeconds)
-			return "", &clustererror.RequeueAfterError{RequeueAfter: constants.RequeueAfterSeconds}
+		clusterstatus, err := vsphereutils.GetClusterProviderStatus(cluster)
+		if err != nil {
+			glog.Infof("Error fetching cluster ProviderStatus field: %s", err)
+			return "", err
+		}
+		if clusterstatus.APIStatus != vsphereconfigv1.ApiReady {
+			glog.Infof("Waiting for Kubernetes API Status to be \"Ready\". Retrying in %s", constants.RequeueAfterSeconds)
+			return "", &clustererror.RequeueAfterError{RequeueAfter: time.Duration(rand.Float64()*10)*time.Second + constants.RequeueAfterSeconds}
```
sflxn: Why not use exponential backoff?

sidharthsurana: This is because, we do want to query approx every 10 sec for the API ready state. As this retry starts from the time when the VM is provisioned to the time when kubeadm is completed. With exponential backoff the last retry timeout might be quite large, causing that much more time the controller will have to wait long before it starts processing the rest of the machines (worker nodes) . We want to minimize this time to a reasonable amount. The reason this small "jitter" is added so that the multiple controller loops waiting for this state could be randomly sequenced as opposed to getting called at the same time causing kubeadm token caching issue.

neolit123: SGTM, but wouldn't it be possible to map predefined timeouts to controller loops for a more deterministic approach?
loop1 -> timeout1
loop2 -> timeout2
...
new loop created -> new timeout3

e.g.
timeoutX = timeout[X-1] + increment.

just an idea.

neolit123: > with these timeouts starting from say 10 sec and increment being as small as say 1 sec will start growing 10,11,12,13 .. 34 secs by the 10th minute mark.

this was my initial proposal, yes.

> Or were you suggesting to have a fixed timeout for a given machine, and just select different timeouts for each machine that is waiting in the controller loop?

something in these lines could work too, like having a map of available unique timeouts and once a machine is up it can free the timeout it's currently using.

> I believe that the current solution with small random jitter addresses the core issue reasonably for most cases without any additional bookkeeping

the deterministic approach could be a nice-to-have for the future.
something about a product of rand() * 10 though, is that having the same timeout of say 3 loops in row would not be that uncommon.

something that can be done here is test the code with a low-entropy simulation in the lines of (e.g. [3, 3, 3, 5, 5, 5] as output from rand() * 10) to verify if the caching issue is present under such conditions.

sflxn:
lgtm

pkg/cloud/vsphere/provisioner/govmomi/create.go
frapposelli: I would definitely favor exponential backoff in this instance, there are several strategy and I can see `LinearJitterBackoff` from [pester](https://github.com/sethgrid/pester) to work just fine for this use case.

frapposelli:
/lgtm

I'll approve once the comment is addressed

pkg/cloud/vsphere/provisioner/govmomi/create.go
frapposelli: Sorry, my mistake, pester is not applicable here as we're not making the http call ourselves, still my recommendation still stands, we could be using a backoff package like https://github.com/jpillora/backoff that can be reused in the codebase and tweaked if we see fit.

PR #110, opened 2018-11-07
Overload the VMTemplate parameter for search purposes

### LINKED ISSUE #109: VMTemplate property may not be able to accurately point to the template VM
The `MachineSpec.VMTemplate` currently is treated as name of the VM to be searched. In certain cases the name of the VM can be duplicated inside a VC, thus we need a more precise mechanism to point to a VM to be used as a template. 

As discussed in the bi-weekly cluster-api-provider-vsphere-sigs meeting, for this purpose we need to allow the `VMTemplate` to also be a valid UUID that would then be matched with the VMs `InstanceUUID` property of the vsphere VM which is guaranteed to be unique with a fall back option of searching by Name if no VM is found using InstanceUUID

sidharthsurana:
The MachineSpec.VMTemplate currently is treated as name of the VM to be searched. In certain cases the name of the VM can be duplicated inside a VC, thus we need a more precise mechanism to point to a VM to be used as a template. For this purpose we are now allowing the VMTemplate to also be a valid UUID that would then be matched with the VMs InstanceUUID property which is guaranteed to be unique.

The modified logic detects if the passed string looks like a valid UUID then it tried to find the VM by InstaceUUID if not found, then it fall back on treating the string as the VM name.

Resolves #109

Change-Id: Ia5e6ca600a71ca784b54336299818b6ec892b949

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -958,7 +958,6 @@
   revision = "e3762e86a74c878ffed47484592986685639c2cd"
 
 [[projects]]
-  branch = "master"
   digest = "1:797d85ec2636b83dafa570b602919912108b913b34e213c9205214361bab722a"
   name = "sigs.k8s.io/cluster-api"
   packages = [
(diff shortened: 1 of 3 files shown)


sidharthsurana:
Verified on local environment both having the VMTemplate passed as Instance UUID as well as Name both seems to work fine.

sflxn:
lgtm

frapposelli:
/lgtm
/approve

nice job 👍

frapposelli:
merging manually to override travis.

PR #112, opened 2018-11-07
Add labels and taints to the nodes created via Machines

### LINKED ISSUE #111: Enable Node Labels and Taints on the nodes
This is a child Issue of [Issue#493](https://github.com/kubernetes-sigs/cluster-api/issues/493) on the main cluster-api repo. 

In this Issue we want to handle the labeling and taints on the node created using the vsphere provider implementation.
Based on the discussion on the cluster-api sig call, we would implement this using kubelet args for now. Once the broader sig has consensus over a better approach, we will switch to using that mechanism.

So, the behavior this Issue is trying to address is that when a user specifies a set of taints and labels inside the `Machine` object, these should be applied on the node created using the spec. These labels and taints however are only to be applied initially. Changing theses on the `Machine` object on a day-2 basis will not be reconciled back with the actual `Node` object for now.

sidharthsurana:
In the `Machine` one can specify the labels, taints. In this patch we use the kubelet args to bootstrap the node with the provided taints and labels

Resolves #111 
Change-Id: Id184e1148c8f6149c9faa40cb382bf740210bba5

### DIFF
--- a/pkg/cloud/vsphere/provisioner/common/templates.go
+++ b/pkg/cloud/vsphere/provisioner/common/templates.go
@@ -20,8 +20,10 @@ import (
 	"bytes"
 	"encoding/base64"
 	"fmt"
+	"strings"
 	"text/template"
 
+	corev1 "k8s.io/api/core/v1"
 	vsphereconfig "sigs.k8s.io/cluster-api-provider-vsphere/pkg/apis/vsphereproviderconfig"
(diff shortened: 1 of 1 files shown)


sflxn:
lgtm

sidharthsurana:
closing this so that I can create one from the main repo

sidharthsurana:
FYI https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/116 is the replacement PR

PR #124, opened 2018-11-15
Make the install and configure part of the scripts match the intent

### LINKED ISSUE #97: Cannot deploy k8s 1.12.x with current code
The kubeadm template config schema has changed in a non backwards compatible way from 1.11.x to 1.12.x  The current kubeadm template inside our code is only compatible uptill 1.11.x and breaks for k8s 1.12.x

### LINKED ISSUE #121: init script trying to install packages even in case of using preloaded template
Currently, the init scripts are broken into 2 main parts, `install` and `configure`. The idea is if a user brings in a template that is already pre-loaded with all the required packages and marks `preloaded: true` in the spec , only the `configure` part of the init script is executed that **should only configure** and ***not install*** new packages.

sidharthsurana:
Currently the configure part of the init script is also installing
packages, which should not be the case. The install part of the script
should be installing anything needed during the configure part.
This split is further used when we want to use template that is
pre-loaded with all the binaries. In that case, we only want to run
the configure part which would bring up the services

This change also updates the kubeadm config to move to v1alpha2
so that now k8s 1.12.x can be supported as well

Resolves #121 
Resolves #97 

Change-Id: I720eaf54829dd3f756dbfedb7691c004c54fcd01

### DIFF
--- a/pkg/cloud/vsphere/provisioner/common/templates.go
+++ b/pkg/cloud/vsphere/provisioner/common/templates.go
@@ -337,9 +337,6 @@ echo done.
 
 const nodeStartupScript = `
 {{ define "install" -}}
-# Disable swap otherwise kubelet won't run
-swapoff -a
-sed -i '/ swap / s/^/#/' /etc/fstab
 
 apt-get update
 apt-get install -y apt-transport-https prips
@@ -359,18 +356,7 @@ deb http://apt.kubernetes.io/ kubernetes-xenial main
 EOF
 apt-get update
 
-{{- end }} {{/* end install */}}
-
-{{ define "configure" -}}
 KUBELET_VERSION={{ .Machine.Spec.Versions.Kubelet }}
-TOKEN={{ .Token }}
-MASTER={{ index .Cluster.Status.APIEndpoints 0 | endpoint }}
-MACHINE={{ .Machine.ObjectMeta.Namespace }}/{{ .Machine.ObjectMeta.Name }}
-CLUSTER_DNS_DOMAIN={{ .Cluster.Spec.ClusterNetwork.ServiceDomain }}
-SERVICE_CIDR={{ getSubnet .Cluster.Spec.ClusterNetwork.Services }}
-NODE_LABEL_OPTION={{ if .Machine.Spec.Labels }}--node-labels={{ labelMap .Machine.Spec.Labels }}{{ end }}
-NODE_TAINTS_OPTION={{ if .Machine.Spec.Taints }}--register-with-taints={{ taintMap .Machine.Spec.Taints }}{{ end }}
-
 # Our Debian packages have versions like "1.8.0-00" or "1.8.0-01". Do a prefix
 # search based on our SemVer to find the right (newest) package version.
 function getversion() {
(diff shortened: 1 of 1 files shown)


pkg/cloud/vsphere/provisioner/common/templates.go
```diff
@@ -390,23 +376,34 @@ KUBECTL=$(getversion kubectl ${KUBELET_VERSION}-)
...
+swapoff -a
+sed -i '/ swap / s/^/#/' /etc/fstab
+
 systemctl enable docker || true
 systemctl start docker || true
 
 sysctl net.bridge.bridge-nf-call-iptables=1
 
-# kubeadm uses 10th IP as DNS server
-CLUSTER_DNS_SERVER=$(prips ${SERVICE_CIDR} | head -n 11 | tail -n 1)
```
sflxn: What are you doing in place of this?  You're removing this, but I don't see the replacement.  Actually, we already discussed it, but can you reply for historical tracking purposes?

pkg/cloud/vsphere/provisioner/common/templates.go
```diff
@@ -484,34 +461,51 @@ apt-get install -y \
...
-` +
-	"PRIVATEIP=`ip route get 8.8.8.8 | awk '{printf \"%s\", $NF; exit}'`" + `
-echo $PRIVATEIP > /tmp/.ip
-` +
-	"PUBLICIP=`ip route get 8.8.8.8 | awk '{printf \"%s\", $NF; exit}'`" + `
 
 # Set up kubeadm config file to pass parameters to kubeadm init.
 cat > /etc/kubernetes/kubeadm_config.yaml <<EOF
-apiVersion: kubeadm.k8s.io/v1alpha1
+apiVersion: kubeadm.k8s.io/v1alpha2
```
sflxn: Will this prevent us from deploying kubelet 1.10?

sidharthsurana: I tried with k8s 1.10.x and the v1alpha2 does not work.. so with this we are going to support k8s 1.11.x and k8s 1.12.x

I suppose we will need to capture the support matrix for the version of k8s we can deploy using cluster-api

pkg/cloud/vsphere/provisioner/common/templates.go
sidharthsurana: The reason this is not needed is that the kubeadm generates that value in the kubelet configuration already. Thus this is not needed anymore. I am not sure from which version of kubeadm this behavior changed, but with v 1.11.x and newer it seems to be the case.

pkg/cloud/vsphere/provisioner/common/templates.go
sflxn: We need to document this and then bring this up for discussion in both our bi-weekly provider meeting and the main cluster-api wg meeting when we come back from the holidays.

sflxn:
lgtm.  Gotta get CI fixed before merging.

pkg/cloud/vsphere/provisioner/common/templates.go
figo: is there a way of not dropping support v1.10?

figo:
@sidharthsurana @sflxn e2e ci passed now after switched ci container to pre-crd version and deploy with k8s v1.11.2,   but the question is: can we still support v1.10.x? thanks

sflxn:
We can discuss 1.10.x later.  @sidharthsurana needs this merged for their product.  I'm working on a design proposal that might pull in some ideas for supporting other versions.

PR #128, opened 2018-11-22
Update quickstart guide for post-crd

sflxn:
The quickstart guide needs to be updated with information on how to
use the repo after the CRD migration.


frapposelli:
merging manually as e2e failure has nothing to do with the current PR

Issue #35, opened 2018-09-04
Add vcsim tests for the machine actuator's create workflow


sflxn:
We can use vc sim (simulator) to run unit tests for the machine actuator.  We need unit tests for the existing machine actuator.  There may be some problems running Terraform against vc sim.  We need to identify those issues and raise them in the vcsim repo.

dougm:
@zpratt thanks, I'll have a PR for that issue soon.

sflxn:
I'm revising this issue to focus only on the create workflow.  The issue was too broad.  It was more like an epic.  I will open up new issues for further work around vcsim.

PR #133, opened 2018-11-27
Add test for govmomi actuator Create method

dougm:
First pass (towards #35) for testing Create() - adds about 38% coverage to the govmomi actuator package.


frapposelli:
/lgtm

🐑 🇮🇹

PR #135, opened 2018-11-28
Add OVA build scripts and utilities

frapposelli:
This is a very big PR, I separated the two vendor commits for ease of review.

There are a few things missing like the version is hardcoded and there are a few rough edges but it's in working conditions and ready to be iterated on.

/cc @akutz @sflxn @figo


installer/build/bootable/root/etc/inputrc
```diff
@@ -0,0 +1,42 @@
...
+set convert-meta off
+set output-meta on
+
+set bind-tty-special-chars off
+"\C-w": backward-kill-word
+"\C-u": kill-whole-line
+"\e[A":history-search-backward
+"\e[B":history-search-forward
+
+$if mode=emacs
```
dougm: :+1:

installer/build/build-ova.sh
```diff
@@ -0,0 +1,70 @@
...
+
+
+ENV_FILE="${CACHE}/installer.env"
+touch $ENV_FILE
+cat > $ENV_FILE <<EOF
+export KUBERNETES_VERSION=${KUBERNETES_VERSION:-}
+EOF
+
+echo -e "--------------------------------------------------
+buidling ova with env...\n
```
dougm: typo

installer/build/scripts/provisioners/provision_govc.sh
```diff
@@ -0,0 +1,28 @@
...
+set -euf -o pipefail
+
+DIR="$(mktemp -d)"
+cd $DIR
+GOVC_SHA256="316013c42b77d9b191a766bce53f55aeab751170704e8d5000bf3ee0d39a98f6  govc_linux_amd64.gz"
+cat > "SHA256SUMS" << EOF
+${GOVC_SHA256}
+EOF
+
+curl -L"#" -o govc_linux_amd64.gz https://github.com/vmware/govmomi/releases/download/v0.17.1/govc_linux_amd64.gz
```
dougm: not 0.19?

installer/build/container/push.sh
```diff
@@ -0,0 +1,31 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -eu
+
+OVA_REV=$(git rev-parse --verify --short=8 HEAD)
+IMAGE="cluster-api-ova-build"
+# TODO(frapposelli): find a better home for the build container
```
figo: 👍  agree that a official account is needed

akutz:
Re: signed binaries -- we can always place them in our VMC AWS S3 bucket. The thing is, if you commit them you can *never* remove them without rewriting them out of the repo's history. That will change every commit ID after the binaries were committed. I favor storing them in S3.

sidharthsurana:
> Re: signed binaries -- we can always place them in our VMC AWS S3 bucket. The thing is, if you commit them you can _never_ remove them without rewriting them out of the repo's history. That will change every commit ID after the binaries were committed. I favor storing them in S3.

+2 this idea of not storing binaries directly into the git repo. One additional tool that we can explore is https://github.com/jedbrown/git-fat This is an add-on tool for git. We use this tool internally to store binaries (reference to) in our git repo and it works pretty much as expected. The benefit of this tool is you still get to maintain the entire history of changes in the git repository without burdening the git repo with actual binary files. However, I am not sure if it supports S3 directly.

akutz:
@sidharthsurana 

I don't even think a tool is necessary. As @frapposelli said, the binaries have their own versions. We just place them in s3 with a versioned file name reflecting the correct version, and we're good to go.

frapposelli:
Ok, let’s move those off to a GS or S3 bucket :+1:

@akutz can you follow up with me offline with access to a bucket? I’ll amend the PR after the upload is done.

sidharthsurana:
@akutz  Using the tool just gives you more native feeling that these binaries are checked in to your git repository and are versioned by git.
But for this case either way works.

akutz:
FWIW @frapposelli, I have no major objection other than I know from very recent experience that it's not fun to remove them. As this is a LTS project, I imagine we don't want to rewrite the git history at a later date.

installer/build/bootable/config/builder.ovf
```diff
@@ -0,0 +1,147 @@
...
+            Version is the actual product version in the
+            form X.X.X.X where X is an unsigned 16-bit integer.
+
+            This is set by the $BUILD_OVA_REVISION
+            environment variable.
+
+            FullVersion is a descriptive version string
+            including, for example, alpha or beta designations
+            and other release criteria.
+        -->
```
zjs: Is this comment accurate?

installer/ovatools/ova-ui/main.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	return fmt.Sprintf("%sx %s", cpus, model)
+}
+
+func getMemory() string {
+	si := &syscall.Sysinfo_t{}
+	err := syscall.Sysinfo(si)
+	if err != nil {
+		panic("Austin, we have a problem... syscall.Sysinfo:" + err.Error())
```
zjs: This may be a confusing panic message to some.

installer/build/build.sh
```diff
@@ -0,0 +1,71 @@
...
+DEBUG=${DEBUG:-}
+set -e -o pipefail +h && [ -n "$DEBUG" ] && set -x
+ROOT_DIR="$GOPATH/src/sigs.k8s.io/cluster-api-provider-vsphere"
+ROOT_WORK_DIR="/go/src/sigs.k8s.io/cluster-api-provider-vsphere"
+
+ROOT_INSTALLER_DIR="${ROOT_DIR}/installer"
+ROOT_INSTALLER_WORK_DIR="${ROOT_WORK_DIR}/installer"
+
+# TAG=${DRONE_TAG:-$(git describe --abbrev=0 --tags)} # e.g. `v0.9.0`
+TAG="v0.0.1"
```
zjs: Is hard-coding the version intentional? (The above reference to `DRONE_TAG` should be removed, at least.)

installer/build/build.sh
```diff
@@ -0,0 +1,71 @@
...
+    docker.io/frapposelli/cluster-api-ova-build:latest ./build/build-ova.sh $*
+elif [ "$step" == "ova-ci" ]; then
+  echo "starting ci build..."
+  export DEBUG=${DEBUG}
+  export BUILD_OVA_REVISION=${BUILD_OVA_REVISION}
+  export TAG=${TAG}
+  export BUILD_NUMBER=${BUILD_NUMBER}
+  export DRONE_BUILD_NUMBER=${DRONE_BUILD_NUMBER}
+  export DRONE_BUILD_EVENT=${DRONE_BUILD_EVENT}
+  export DRONE_DEPLOY_TO=${DRONE_DEPLOY_TO}
```
zjs: These references to `DRONE` should be removed.

installer/build/build-ova.sh
```diff
@@ -0,0 +1,70 @@
...
+DIR=$(pwd)
+CACHE=${DIR}/bin/.cache/
+mkdir -p ${CACHE}
+KUBERNETES_DEFAULT_VERSION=$(curl -sSL https://dl.k8s.io/release/stable-1.txt)
+
+# Check if it's a file in `scripts`, URL, or REVISION
+function setenv() {
+  tmpvar=$1
+  fallback=$2
+  # if a cli argument was specified, it's a local file, gcloud bucket upload, or revision number.
```
zjs: Is this accurate?

installer/build/build-ova.sh
```diff
@@ -0,0 +1,70 @@
...
+      shift 2 # past argument
+      ;;
+    *)
+      # unknown
+      break; break;
+  esac
+done
+
+# set Kubernetes Version
+setenv KUBERNETES "$KUBERNETES_DEFAULT_VERSION"
```
zjs: Is this now equivalent to `KUBERNETES=${KUBERNETES:-$KUBERNETES_DEFAULT_VERSION}`?

installer/build/build.sh
frapposelli: Yes, it is intentional, I will be adding a comment to better explain it in code

installer/build/bootable/config/builder.ovf
frapposelli: Yes, it describes how `--version--` is substituted with the proper version during the build process

installer/build/bootable/build-app.sh
```diff
@@ -0,0 +1,110 @@
...
+# Importing the pubkey
+log2 "configuring os"
+log3 "importing local gpg key"
+rpm --import /etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY
+
+log3 "setting umask to 022"
+sed -i 's/umask 027/umask 022/' /etc/profile
+
+log3 "setting root password"
+echo 'root:Vmw@re!23' | chpasswd
```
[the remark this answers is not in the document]
frapposelli: this password is overridden during first boot as specifying a root password is required to deploy the OVA. SSH is set to never accept passwords and only work with keys.

installer/Makefile
```diff
@@ -0,0 +1,119 @@
...
+
+$(ovfenv): $(call godeps,ovatools/ovfenv/*.go)
+	@echo building ovfenv linux...
+	@GOARCH=amd64 GOOS=linux $(TIME) $(GO) build $(RACE) -ldflags "$(LDFLAGS)" -o ./$@ ./$(dir $<)
+
+$(ova-ui): $(call godeps,ovatools/ova-ui/*.go)
+	@echo building ova-ui
+	@GOARCH=amd64 GOOS=linux $(TIME) $(GO) build $(RACE) -ldflags "$(LDFLAGS)" -o ./$@ ./$(dir $<)
+
+$(toolbox): $(call godeps,toolbox/*.go)
```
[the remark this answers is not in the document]
frapposelli: I'm a fan of `toolbox` for simple use cases like this one, it also gives us the ability to extend its functionality (like [this](
https://github.com/vmware/govmomi/tree/master/toolbox#httproundtripper)) to accommodate more complex automation with Cluster API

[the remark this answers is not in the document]
akutz: Hi @frapposelli / @zjs,

installer/build/bootable/config/builder.ovf
zjs: On the subsequent lines, it seems like `FullVersion` is blank?

installer/Makefile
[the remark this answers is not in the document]
akutz: Hi @frapposelli / @zjs,

FWIW, yakity uses `open-vm-tools`, and I often have issues with them taking a while to come online in terms of visibility to the vCenter. That said, I use `open-vm-tools` because it's one less thing...

installer/Gopkg.toml
```diff
@@ -0,0 +1,36 @@
...
+# [[constraint]]
+#   name = "github.com/user/project2"
+#   branch = "dev"
+#   source = "github.com/myfork/project2"
+#
+# [[override]]
+#  name = "github.com/x/y"
+#  version = "2.4.0"
+
+required = ["github.com/golang/dep", "github.com/GoASTScanner/gas", "github.com/go-yaml/yaml", "github.com/golang/lint/golint"]
```
sflxn: We're pulling in kubernetes-sigs/yaml in the provider project.  It's based on go-yaml.  Kubernetes projects seem to like using forked versions (into the k8s.io or sigs.k8s.io repos) of outside projects.

frapposelli:
@zjs thanks a lot for reviewing 👍 I addressed most of your comments in a separate commit

The plan is to squash all the review commits into one once it's lgtm'd

installer/Gopkg.toml
frapposelli: I'm fine with pulling the same as long as they're compatible

installer/Makefile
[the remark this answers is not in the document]
dougm: Note that toolbox does support all of the guest operations.  It is possible to only expose a subset (as VIC does) or disable all (as this PR does).  The main issue is that toolbox only has a hook for in-guest authentication, but leaves it to the consumer to implement. 
https://github.com/vmware/govmomi/tree/master/toolbox#guestfilemanager-object
https://github.com/vmware/govmomi/tree/master/toolbox#guestprocessmanager

Regarding the visibility in vCenter, open-vm-tools takes ~30 sec (last I checked) to initially report IP addresses, toolbox is around 1 sec or less.

installer/build/container/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ENV GOVERSION=1.10.5
+ENV PATH=$PATH:/root/gsutil:/usr/local/go/bin:/usr/local/google-cloud-sdk/bin/
+
+RUN set -eux; \
+    tdnf remove -y toybox; \
+    tdnf install -y make tar gzip python2 python-pip sed git diffutils findutils rpm \
+    gawk util-linux which docker gptfdisk e2fsprogs grub2-efi efibootmgr parted dosfstools xz docker; \
+    curl -L'#' -k https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-200.0.0-linux-x86_64.tar.gz  | tar xzf - -C /usr/local; \
+    mkdir -p /root/.gsutil/; \
+    /usr/local/google-cloud-sdk/install.sh --quiet; \
```
sflxn: Why do we need google-cloud-sdk?  It's quite large and takes awhile to install.

installer/build/container/push.sh
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -eu
+
+OVA_REV=$(git rev-parse --verify --short=8 HEAD)
+IMAGE="cluster-api-ova-build"
+REPO="docker.io/frapposelli/"
```
sflxn: We now have a repo for the provider container if you want to use the same repo.  gcr.io/cnx-cluster-api (part of VMW's gcr account).  You just need the GCR key file, which I can provide.

installer/Makefile
```diff
@@ -0,0 +1,119 @@
+# Copyright 2018 VMware, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
frapposelli: updated the headers, double checking with legal but it should be good to go

installer/build/container/push.sh
frapposelli: let's follow up offline with that and I'll amend once it's pushed

installer/build/container/Dockerfile
frapposelli: removed

installer/build/container/push.sh
frapposelli: pushed and amended

installer/build/container/push.sh
frapposelli: moved to `gcr.io/cnx-cluster-api/cluster-api-ova-build`

frapposelli:
@akutz @figo @dougm @zjs @sflxn 

Thanks for the thorough review guys, I should have adressed all the major concerns here, last two were moving the build container to a more permanent location (`gcr.io` under the `cnx-cluster-api` project) and removing the boot/EFI binaries from the repo to a google storage bucket under the same project.

Please take a second look at it, I'll squash everything to the second commit once it LGTY.

dougm:
@frapposelli sorry I should have added a `/hold`.  Wasn't expecting it to merge since there wasn't an `/approve`, but I guess that's implicit since you are an approver.

frapposelli:
No worries, *I* should’ve put a /hold on it :sad:

PR #140, opened 2018-12-03
Dependencies & API updates

sidharthsurana:
Adjustments to the code for the following API changes:
1. ProviderConfig renamed to ProviderSpec
2. Actuator interface updated to take in a context object

Change-Id: I464c1b67a8e37b271c643c6fe9b7ea717cd22aea


sflxn:
/hold

This build will fail till we start generating the CI's template yaml.  @figo will open a PR for that.  In the mean time, we can review this PR's content.

sidharthsurana:
For all the reviewers, the actual code files that have changed are as follows, other changes are dep related:
```
M       pkg/cloud/vsphere/machineactuator.go
M       pkg/cloud/vsphere/provisioner/govmomi/create.go
M       pkg/cloud/vsphere/provisioner/govmomi/create_test.go
M       pkg/cloud/vsphere/provisioner/govmomi/delete.go
M       pkg/cloud/vsphere/provisioner/govmomi/exists.go
M       pkg/cloud/vsphere/provisioner/govmomi/session.go
M       pkg/cloud/vsphere/provisioner/govmomi/update.go
M       pkg/cloud/vsphere/utils/utils.go
```

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -25,7 +25,7 @@ import (
 	"sigs.k8s.io/cluster-api/pkg/util"
 )
 
-func (pv *Provisioner) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
+func (pv *Provisioner) Create(cctx context.Context, cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
```
sflxn: can you shorten cctx to just ctx?

sflxn:
Can you rename all context and cctx arg names to just ctx?  If it's not a go standard, it seems to be a community standard.

sflxn:
@sidharthsurana Do we need to update any template files?

sidharthsurana:
> @sidharthsurana Don't we need to update the template files?

yes, good catch. I forgot about the non-go code files.
I have updated the changeset with these changes as well now.

figo:
the e2e job is failed but reported as success: https://gubernator.k8s.io/build/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/140/pull-cluster-api-provider-vsphere-e2e/20
this problem already fixed by https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/146

i am testing a new run now

figo:
@sflxn @sidharthsurana the CI signal is green, it passed after ci container updated.
/lgtm

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -25,15 +25,15 @@ import (
 	"sigs.k8s.io/cluster-api/pkg/util"
 )
 
-func (pv *Provisioner) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
+func (pv *Provisioner) Create(ctx context.Context, cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
```
figo: @sidharthsurana just for my knowledge, i don't see this additional parameter: ctx not been used at this function, what is the purpose of it?

sflxn: this is defined in the main repo's actuator interface.  It must be here since we're implementing the interface.

sidharthsurana: This is actually one of the new parameter that was added in the actuator interface. Right now in this PR I did not make use of it. However, later on we can actually start using this context in the logging, especially when we move to klog

sflxn: klog doesn't use context from what little I've seen of it.  I may open up a few PR to add some of the VIC logging features into klog.

PR #143, opened 2018-12-04
compile clusterctl for every e2e ci job

### LINKED ISSUE #6: Create a dockerfile to build clusterctl
Create a dockerfile to build clusterctl.

This is a mirror of https://github.com/kubernetes-sigs/cluster-api/issues/289.  Since it's now up to the vendor specific providers to build clusterctl, we should look into completing this issue.

figo:
With this patch, clusterctl will be compiled, packaged into container and uploaded to gcr during the build phase.  During the test phase, the container will be used to execute with the latest version of the code.

/cc @sflxn @sidharthsurana @dougm @frapposelli 

resolve #6

### DIFF
--- a/scripts/e2e/bootstrap_job.template
+++ b/scripts/e2e/bootstrap_job.template
@@ -9,6 +9,13 @@ spec:
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
+      initContainers:
+      - name: clusterctl-bin
+      image: gcr.io/cnx-cluster-api/clusterctl:${VERSION}
+      command: ['sh', '-c', 'cp /root/clusterctl /tmp/clusterctl/clusterctl && chmod +x /tmp/clusterctl/clusterctl']
+      volumeMounts:
+        - name: shared-data
+          mountPath: /tmp/clusterctl
       containers:
       - name: cluster-api-provider-vsphere-ci
         image: gcr.io/cnx-cluster-api/cluster-api-provider-vsphere-ci:latest
@@ -53,6 +60,8 @@ spec:
             mountPath: /root/.kube
           - name: sshkeys
             mountPath: /root/.ssh
+          - name: shared-data
+            mountPath: /tmp/clusterctl/
         command:
         - "./clusterctl.sh"
       volumes:
@@ -62,4 +71,6 @@ spec:
       - name: sshkeys
         hostPath:
           path: /home/vmware/.ssh
+      - name: shared-data
+        emptyDir: {}
       restartPolicy: Never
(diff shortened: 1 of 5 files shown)


scripts/e2e/bootstrap_job.template
```diff
@@ -9,6 +9,13 @@ spec:
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
+      initContainers:
+      - name: clusterctl-bin
```
sflxn: Can you explain what you're doing here?  What is this /tmp/clusterctl/clusterctl?

scripts/e2e/bootstrap_job/clusterctl.sh
```diff
@@ -46,7 +46,7 @@ chmod +x /usr/local/bin/kubectl
 
 # run clusterctl
 echo "test ${PROVIDER_COMPONENT_SPEC}"
-./bin/clusterctl create cluster --existing-bootstrap-cluster-kubeconfig ~/.kube/config -c ./spec/cluster.yml \
+/tmp/clusterctl/clusterctl create cluster --existing-bootstrap-cluster-kubeconfig ~/.kube/config -c ./spec/cluster.yml \
```
sflxn: I still don't know what this /tmp/clusterctl/clusterctl is.  If you're building a container for clusterctl, why not run the container here?

figo: the bootstrap_job is launched by bootstrap_job.yml (generated based on bootstrap_job.template), the job spec will run a initContainer for this job, clusterctl is coming from initContainer (a shared volume mounted at /tmp/clusterctl).

sflxn:
Also, make sure to add this issue as resolved in the commit message, https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/6.

scripts/e2e/bootstrap_job/clusterctl.sh
sflxn: I now understand what you're doing.  You're using a container as a packaging system instead of using curl or wget to pull down a binary.  Instead, you're using the container to copy it's binary out so you can execute it.  It doesn't fully feel right, but there aren't any rules against it so I guess it's ok.

scripts/e2e/bootstrap_job.template
```diff
@@ -9,6 +9,13 @@ spec:
...
+      image: gcr.io/cnx-cluster-api/clusterctl:${VERSION}
+      command: ['sh', '-c', 'cp /root/clusterctl /tmp/clusterctl/clusterctl && chmod +x /tmp/clusterctl/clusterctl']
```
sflxn: If you're going to do this, is it possible to copy it to tmp/clusterctl/clusterctl-${VERSION} just to be clear what version we're putting into tmp.

PR #147, opened 2018-12-05
Add Build and Deploy docs for OVA

frapposelli:
/cc @sflxn @figo

### DIFF
--- a/docs/ova/README.md
+++ b/docs/ova/README.md
@@ -1,0 +1,87 @@
+# Cluster API Provider vSphere, OVA Installer Build process
+
+The build process uses a collection of bash scripts to launch a docker container on your local machine
+where we provision a linux OS, install dependencies, and extract the filesystem to make the OVA.
+
+- [Cluster API Provider vSphere, OVA Installer Build process](#cluster-api-provider-vsphere-ova-installer-build-process)
+  - [Usage](#usage)
+    - [Prerequisites](#prerequisites)
+    - [Build bundle and OVA](#build-bundle-and-ova)
+      - [Build script](#build-script)
+  - [Deploy](#deploy)
+  - [Using the bootstrap cluster](#using-the-bootstrap-cluster)
+
+## Usage
+
+The build process is controlled from a central script, `build.sh`. This script
+launches the build docker container and controls our provisioning and ova
+extraction through `make`.
+
+### Prerequisites
+
+The build machine must have `docker`.
+
+- `docker` for Mac: https://www.docker.com/docker-mac
+- `docker` for Windows: https://www.docker.com/docker-windows
+
+### Build bundle and OVA
+
+#### Build script
+
+The build script pulls the desired versions of each included component into the build container.
+
+> *Note:* You must specify build step `ova-dev` when calling `build.sh` from a development machine.
+
+If called without any values, `build.sh` will get Kubernetes version: `stable-1`
+```
+./build/build.sh ova-dev
+```
+
+If called with the values below, `build.sh` will include Kubernetes version v1.12.1
+```
+./build/build.sh ova-dev --kubernetes-version v1.12.1
+```
+
+## Deploy
+
+The OVA must be deployed to a vCenter.
+Deploying to ESX host is not tested at the moment.
+
+The recommended method for deploying the OVA:
+- Access the vCenter Web UI, click `vCenter Web Client` (HTML5 or Flash)
+- Right click on the desired cluster or resource pool
+- Click `Deploy OVF Template`
+- Select the URL or file and follow the prompts
+- Power on the deployed VM
+- Open the VM Console and wait for the Kubernetes cluster status to be: RUNNING
+
+Alternative, deploying with `govc` (this also requires [`jq`](https://stedolan.github.io/jq/)):
+
+First create a spec to populate the OVA fields during deployment:
+```console
+$ govc import.spec bin/cluster-api-xxxxxx.ova | jq -r . > cluster-api-spec.json
+```
+
+Now edit the file and populate the properties (Root password and SSH key at minimum)
+```console
+$ vi cluster-api-spec.json
+```
+
+Now you can import the OVA on your vSphere infrastructure
+```console
+$ export GOVC_URL='<username>:<password>@<vcenter.fqdn>'
+$ export GOVC_INSECURE=1 # This is needed if your vSphere certificate is not valid or self signed
+$ govc import.ova --options=cluster-api-spec.json cluster-api-xxxxxx.ova
+```
+
+And power it on and open the VM Console.
+```console
+$ govc vm.power -on cluster-api
+$ govc vm.console cluster-api
+```
+
+## Using the bootstrap cluster
+
+Once the cluster is in RUNNING state, you can use the `scp` command to retrieve the kubeconfig file from the Cluster API VM, this kubeconfig will be passed to clusterctl to launch the Cluster API provisioner.
+
+The Kubeconfid file inside the deployed VM lives at `/etc/kubernetes/admin.conf`.


docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+If called with the values below, `build.sh` will include Kubernetes version v1.12.1
+```
+./build/build.sh ova-dev --kubernetes-version v1.12.1
+```
+
+## Deploy
+
+The OVA must be deployed to a vCenter.
+Deploying to ESX host is not supported at the moment, it might work though.
```
sflxn: I would remove "it might work though."

docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+The build process is controlled from a central script, `build.sh`. This script
+launches the build docker container and controls our provisioning and ova
+extraction through `make`.
+
+### Prerequisites
+
+The build machine must have `docker`.
+
+- `docker for Mac`: https://www.docker.com/docker-mac
```
zjs: It seems weird to put "for Mac" in monospace.

```suggestion
- `docker` for Mac: https://www.docker.com/docker-mac
```

docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+./build/build.sh ova-dev
+```
+
+If called with the values below, `build.sh` will include Kubernetes version v1.12.1
+```
+./build/build.sh ova-dev --kubernetes-version v1.12.1
+```
+
+## Deploy
+
```
sflxn: I think we should mention something about generating the key pairs here before they start.  The top level make file will generate a set of key files for them.  When they run `make dev-yaml`, it will generate vsphere_tmp and vsphere_tmp.pub in their .ssh folder.

docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+The build process is controlled from a central script, `build.sh`. This script
+launches the build docker container and controls our provisioning and ova
+extraction through `make`.
+
+### Prerequisites
+
+The build machine must have `docker`.
+
+- `docker for Mac`: https://www.docker.com/docker-mac
+- `docker for Windows`: https://www.docker.com/docker-windows
```
zjs: Likewise, with "for Windows".

```suggestion
- `docker` for Windows: https://www.docker.com/docker-windows
```

docs/ova/README.md
sflxn: i like that much better.

docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Build bundle and OVA
+
+#### Build script
+
+The build script pulls the desired versions of each included component into the build container.
+
+> You must specify build step `ova-dev` when calling build.sh
```
zjs: But `build.sh` _does_ seem like it should be monospaced.

It may be useful to indicate that this is a note (versus, say, a quote) like other documents do.

Lastly, it might be useful to note _when_ to use `ova-dev`.

```suggestion
> *Note:* You must specify build step `ova-dev` when calling `build.sh` from a development machine.
```

docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+The recommended method for deploying the OVA:
+- Access the vCenter Web UI, click `vCenter Web Client (Flash)`
+- Right click on the desired cluster or resource pool
+- Click `Deploy OVF Template`
+- Select the URL or file and follow the prompts
+- Power on the deployed VM
+- Open the VM Console and wait for the Kubernetes cluster status to be: RUNNING
+
+Alternative, deploying with `govc` (this also requires [`jq`](https://stedolan.github.io/jq/)):
```
zjs: Any reason not to mention `ovftool` as an option as well? It works some places `govc` does not (e.g., VMC) and avoids the somewhat-ambiguous step of editing the JSON file.

docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+First create a spec to populate the OVA fields during deployment:
+```console
+$ govc import.spec bin/cluster-api-xxxxxx.ova | jq -r . > cluster-api-spec.json
```
zjs: Nit-pick: Stylistically, I dislike including the `$` prefix. It's useful in instructions that involve root and non-root steps, but in documents like this I find that it just makes commands harder to copy and paste.

```suggestion
govc import.spec bin/cluster-api-xxxxxx.ova | jq -r . > cluster-api-spec.json
```

(If this change is made, the subsequent code blocks should be updated as well.)

docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+$ govc import.ova --options=cluster-api-spec.json cluster-api-xxxxxx.ova
+```
+
+And power it on and open the VM Console.
+```console
+$ govc vm.power -on cluster-api
+$ govc vm.console cluster-api
+```
+
+### Using the bootstrap cluster
```
zjs: Should this be a second level heading? It seems like a step which follows "Deploy", not a sub-step.

```suggestion
## Using the bootstrap cluster
```

docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+Once the cluster is in RUNNING state, you can use the `scp` command to retrieve the kubeconfig file to be passed to clusterctl to launch the Cluster API provisioner.
```
figo: should you mention where the kubeconfig located inside the VM?

docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+./build/build.sh ova-dev --kubernetes-version v1.12.1
+```
+
+## Deploy
+
+The OVA must be deployed to a vCenter.
+Deploying to ESX host is not supported at the moment, it might work though.
+
+The recommended method for deploying the OVA:
+- Access the vCenter Web UI, click `vCenter Web Client (Flash)`
```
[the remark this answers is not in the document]
figo: still Flash?  html5 client is referred, correct?

docs/ova/README.md
frapposelli: I thought this would be more appropriate in the general docs for the vSphere provider, this is strictly related to the OVA / Management cluster

docs/ova/README.md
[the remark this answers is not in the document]
frapposelli: OVA deployment works on 6.5 from the Flash UI, I tested it, the issue was most likely from your environment (probably due to browser/OS combination), I tested from Windows and was working fine.

docs/ova/README.md
figo: is there other option supported other than `ova-dev`

docs/ova/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+Now edit the file and populate the properties (Root password and SSH key at minimum)
+```console
+$ vi cluster-api-spec.json
+```
+
+Now you can import the OVA on your vSphere infrastructure
+```console
+$ export GOVC_URL='<username>:<password>@<vcenter.fqdn>'
+$ govc import.ova --options=cluster-api-spec.json cluster-api-xxxxxx.ova
```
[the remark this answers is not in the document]
frapposelli: Good point, will amend once the review is done

docs/ova/README.md
frapposelli: There is `ova-ci`, which triggers the build directly (without using the docker container)

docs/ova/README.md
frapposelli: Like this?

```suggestion
Once the cluster is in RUNNING state, you can use the `scp` command to retrieve the kubeconfig file from the Cluster API VM, this kubeconfig will be passed to clusterctl to launch the Cluster API provisioner.
```

figo: i basically means the exact location of kubeconfig, i guess it is somewhere at ~/.config/kubeconfig?

docs/ova/README.md
figo: Thank you, is it possible to include this option in Doc?

docs/ova/README.md
figo: i got the VM running now, it is at /etc/kubernetes/admin.conf, it is more intuitive to mention the location in Doc, but i will leave the decision to you, thanks

docs/ova/README.md
frapposelli: We'll add it to the docs once we're using it in CI

docs/ova/README.md
frapposelli: Interestingly enough, making it harder to copy/paste is exactly why I like prepending the `$` prompt 😄. 

In general blindly copy/pasting commands from a webpage is a bad practice, this forces the user to at least be aware of what they're doing.

docs/ova/README.md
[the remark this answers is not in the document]
frapposelli: This should cover both cases, the user flow is the same

```suggestion
- Access the vCenter Web UI, click `vCenter Web Client`
```

docs/ova/README.md
frapposelli: Generally I think promoting `govc` is a good thing, but I agree with you on covering use cases that are not supported by `govc` today.

@figo is testing deployment to VMC via ovftool and Content Library, once the testing is done I will update the document

docs/ova/README.md
akutz: Hi @frapposelli,

FWIW, I also include the `$` for this very reason. It makes people consider what they are doing, and prevents blind copy/paste/execute errors.

PR #148, opened 2018-12-05
Fix kubeadm token

### LINKED ISSUE #145: bug: generated kubadm token includes log output
I’m observing a problem with the workers being unable to join a cluster because they haven an invalid token.

In fact, the /tmp/boot.sh script includes
```
TOKEN=I1204 04:46:06.528320      24 version.go:236] remote version is much newer: v1.13.0; falling back to: stable-1.12
ewe864.2i8qadn9rz541i84
```

Log output is prepended to the actual token value

The create-token utility is here:
https://github.com/kubernetes-sigs/cluster-api/blob/e1903be683739379be57f78a4095cd51726495fd/pkg/kubeadm/kubeadm.go#L56

and used:
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/fd4d9e1385779fae41393eeae59614ef12490059/pkg/cloud/vsphere/provisioner/govmomi/utils.go#L56

controller log output:
```
I1204 04:46:10.310625       1 controller.go:124] reconciling cluster a3824d30-8bf2-4a34-9645-655273c05fb5 triggers idempotent reconcile.
I1204 04:46:10.311929       1 utils.go:73] Generated kubeadm token I1204 04:46:06.528320      24 version.go:236] remote version is much newer: v1.13.0; falling back to: stable-1.12
ewe864.2i8qadn9rz541i84 from master
```

It's not clear why the log text is being included now.  It would be safer to generate the token in a programmatic way, but if we want to parse stdout, then a regexp can be used
```
KubeadmTokenPattern     = regexp.MustCompile(`(?m)[a-z0-9]{6}\.[a-z0-9]{16}`)
token = KubeadmTokenPattern.FindString(output)
```

sflxn:
Replaced the method of getting the kubeadm token.  Prior to this, we were
shelling out to kubeadm to create a token.  That proved unreliable in the
latest update to kubeadm (1.13).  Now, we generate the token ourselves and
write the token out as a secret.

Resolves #145


pkg/cloud/vsphere/provisioner/govmomi/utils.go
```diff
@@ -67,11 +65,57 @@ func (pv *Provisioner) GetKubeadmToken(cluster *clusterv1.Cluster) (string, erro
...
+		glog.Infof("Could not cache the kubeadm token on cluster object: %s", err.Error())
 	}
 	return token, err
 }
 
+// The logic for this function comes from various places in kubeadm itself
+func (pv *Provisioner) createKubeadmToken() (string, error) {
+	// Create the actual token string
+	token, err := tokenutil.GenerateBootstrapToken()
+	if err != nil || token == "" ||
```
dougm: Looked at the source for `tokenutil.GenerateBootstrapToken`, doesn't look like the extra validation of token length is needed.  But if that is the case, the error message should indicate the token length is invalid, rather than unable to create token.

pkg/cloud/vsphere/provisioner/govmomi/utils.go
```diff
@@ -67,11 +65,57 @@ func (pv *Provisioner) GetKubeadmToken(cluster *clusterv1.Cluster) (string, erro
 	ncluster.ObjectMeta.Annotations[constants.KubeadmTokenExpiryTime] = time.Now().Add(constants.KubeadmTokenTtl).Format(time.RFC3339)
 	_, err = pv.clusterV1alpha1.Clusters(cluster.Namespace).Update(ncluster)
 	if err != nil {
-		glog.Infof("Could not cache the kubeadm token on cluster object: %s", err)
+		glog.Infof("Could not cache the kubeadm token on cluster object: %s", err.Error())
```
dougm: original was fine, no need to call `Error()`

pkg/cloud/vsphere/provisioner/govmomi/utils.go
```diff
@@ -67,11 +65,57 @@ func (pv *Provisioner) GetKubeadmToken(cluster *clusterv1.Cluster) (string, erro
...
 	return token, err
 }
 
+// The logic for this function comes from various places in kubeadm itself
+func (pv *Provisioner) createKubeadmToken() (string, error) {
+	// Create the actual token string
+	token, err := tokenutil.GenerateBootstrapToken()
+	if err != nil || token == "" ||
+		len(token) < (tokenapi.BootstrapTokenIDBytes + tokenapi.BootstrapTokenSecretBytes) {
+		return "", fmt.Errorf("unable to create kubeadm token: %s", err.Error())
```
dougm: cosmetic, but no need to call `Error()`, just `err` is fine (more cases of this below).

pkg/cloud/vsphere/provisioner/govmomi/utils.go
```diff
@@ -67,11 +65,57 @@ func (pv *Provisioner) GetKubeadmToken(cluster *clusterv1.Cluster) (string, erro
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      tokenutil.BootstrapTokenSecretName(tokenID),
+			Namespace: metav1.NamespaceSystem,
+		},
+		Type: corev1.SecretType(tokenapi.SecretTypeBootstrapToken),
+		Data: data,
+	}
+
+	// Create or update the secret.  Just repeating what kubeadm is doing.
+	if _, err := pv.k8sClient.CoreV1().Secrets(metav1.NamespaceSystem).Create(secret); err != nil {
```
sidharthsurana: this needs to be the target cluster client

sflxn:
I have an update, but having trouble testing it because the main cluster-api controller is crashing and doesn't the main repo appears to be broken :(.  The main repo needs better CI and e2e testing.  Our e2e testing did not catch this problem.

```
cluster-api-system        cluster-api-controller-manager-0        0/1       CrashLoopBackOff   6          7m
```

I asked @roberthbailey if he's built and push an updated version lately, and he said not since pre-crd.  I'm going to have to spend some time today and try to get the main repo building again and get a new container.  I will likely upload one to our registry.

figo:
@sflxn i am not fully sure whether this is a test problem.
based on your description, `cluster-api-controller-manager-0` is crashing, but e2e logs shows 'vsphere-controller' is running fine, and clusterctl returns
`Done provisioning cluster. You can now access your cluster with kubectl --kubeconfig kubeconfig`

here is the log file:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/148/pull-cluster-api-provider-vsphere-e2e/32/build-log.txt

should the `vsphere-controller` return errors in the scenario?

sflxn:
PR is updated, but I'm going to add a hold.  I need to push a version of the main controller to our registry and make another PR so that our generated provider-components.yaml uses that container instead of the one from google's registry, which has a 2 month old version.

/hold

figo:
just add a little clarification here:
the current e2e test only deploy a single master cluster, which does not requires `cluster-api-manager` from main repo, so it will not catch the `cluster-api-manager` backoff.
`vsphere-provider-controller-manager` will run fine.

we are expanding e2e tests to machineSet, also adding `cluster-api-manager` pod verification and integration tests.

pkg/cloud/vsphere/provisioner/govmomi/utils.go
```diff
@@ -72,6 +67,62 @@ func (pv *Provisioner) GetKubeadmToken(cluster *clusterv1.Cluster) (string, erro
...
+
+	// Create a client to the target cluster
+	rc, err := clientcmd.RESTConfigFromKubeConfig([]byte(kubeconfig))
+	if err != nil {
+		return "", err
+	}
+
+	clusterclient, err := kubernetes.NewForConfig(rest.AddUserAgent(rc, "cluster-api-provider-vsphere"))
+	if err != nil {
+		glog.Fatalf("Invalid API configuration for kubeconfig: %v", err)
```
sidharthsurana: Why glog.Fatalf here?
This will kill the pod and then keep the actuator in a redeploy loop.
Wouldn't just a `return "", err` suffice here?

PR #152, opened 2018-12-07
e2e ci improvement

### LINKED ISSUE #153: e2e test enhancement - test CI code change from a PULL request
when developer open a PULL request, mostly the code change to scripts/e2e/* will
be tested in the same PULL, but the subfolder scripts/e2e/bootstrap_job is 
packaged into container, the container been updated separately instead of building at runtime.

we need to change this, and build this container at runtime.

figo:
1) build the bootstrap_job container in runtime.
so any changes to the ci code will be run in job

2) support Prow and Debug testrun at the same time.

/cc @sflxn 

resolve #153 #154


scripts/e2e/hack/job.yml
```diff
@@ -11,7 +11,8 @@ spec:
         key: node-role.kubernetes.io/master
       containers:
       - name: cluster-api-provider-vsphere-ci
-        image: luoh/cluster-api-provider-vsphere-prow-ci:latest
+        image: gcr.io/cnx-cluster-api/cluster-api-provider-vsphere-ci-debug:debug
+        imagePullPolicy: Never
```
sflxn: Why push a debug version (above) if you never pull it?

sflxn:
With this change, it helps a little, but if multiple people are debugging tests, they could potentially still collide, right?

scripts/e2e/hack/job.yml
figo: make push is optional only when the minikube cluster is on same machine as docker image been build, in other cases, developer can still push and pull ci debug image.

figo:
> With this change, it helps a little, but if multiple people are debugging tests, they could potentially still collide, right?

you are right, this is an incremental change

sflxn:
@figo Let's create a new issue to allow multiple debugging tests in the future.

Issue #118, opened 2018-11-09
docs: update quickstart docs based on the CRD migration


sflxn:
There is a PR https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/105 to migrate this provider to CRDs.  We need the quickstart documentation to get updated.  Perhaps we should go ahead and update it to gitbook.

PR #167, opened 2018-12-28
Fix broken link to quickstart pending resolution of #118.

davidewatson:


davidewatson:
Since KubeCon we've had an uptick of interested users and IMHO no link is better than a broken or incorrect link. The documentation in `docs/` appears to apply to the pre-CRD implementation. Am I incorrect about this? Would it be better to point to the docs directory instead of removing the quickstart link entirely?

davidewatson:
FWIW, I was able to deploy the CRD implementation roughly following the self service documentation:

```
minikube start
make prod-yaml
kubectl apply -f out/provider-components.yaml
```

frapposelli:
thanks @davidewatson sorry but the activity was slow during the winter break.

Looks like #170 fixes the link to the right position (`README.md` at the root of `docs/`), I'm going to close this PR.

/close

PR #171, opened 2019-01-07
Use `go env` to get GOPATH

benmoss:
At least on my development machine I usually leave `$GOPATH` undefined and let Go use the default. This unfortunately then breaks a lot of scripts that assume `$GOPATH` is always set. I think using `go env GOPATH` is supposed to be the right way to get the current value, since it will give the default path if the environment variable is unset.

### DIFF
--- a/installer/build/build.sh
+++ b/installer/build/build.sh
@@ -18,7 +18,7 @@
 
 DEBUG=${DEBUG:-}
 set -e -o pipefail +h && [ -n "$DEBUG" ] && set -x
-ROOT_DIR="$GOPATH/src/sigs.k8s.io/cluster-api-provider-vsphere"
+ROOT_DIR="$(go env GOPATH)/src/sigs.k8s.io/cluster-api-provider-vsphere"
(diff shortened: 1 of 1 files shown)


installer/build/build.sh
```diff
@@ -18,7 +18,7 @@
 
 DEBUG=${DEBUG:-}
 set -e -o pipefail +h && [ -n "$DEBUG" ] && set -x
-ROOT_DIR="$GOPATH/src/sigs.k8s.io/cluster-api-provider-vsphere"
+ROOT_DIR="$(go env GOPATH)/src/sigs.k8s.io/cluster-api-provider-vsphere"
```
neolit123: something to note here is that $GOPATH can be a list of paths:
https://golang.org/cmd/go/#hdr-GOPATH_environment_variable

> The GOPATH environment variable lists places to look for Go code. On Unix, the value is a colon-separated string. On Windows, the value is a semicolon-separated string. On Plan 9, the value is a list.

PR #173, opened 2019-01-08
Add support to inject custom trusted certs into the machines

### LINKED ISSUE #165: Feature: support to inject custom trusted certs into the nodes
Often, we need the ability to inject custom CA certs into the machines so that the non-standard CA signed certs can be trusted on the created machines for various reasons.

The proposal is as follows:
- Add a new optional field named `TrustedCerts` to the `MachineSpec`. This field would be a list of base64 encoded custom certs that would be added to the instance's trusted CA certificates for SSL/TLS trust establishment.

This value if present in the spec is then passed on the cloud-init to distribute the certs on the instance

sidharthsurana:
Resolves #165

Change-Id: I2f399143419553b120e28b46719e7514990fdab0


sidharthsurana:
abandoning this PR, will create a new clean PR

PR #180, opened 2019-01-24
Add support to inject custom trusted certs into the machines

### LINKED ISSUE #165: Feature: support to inject custom trusted certs into the nodes

sidharthsurana:
Resolves #165

Change-Id: I2f399143419553b120e28b46719e7514990fdab0

### DIFF
--- a/docs/design/trustedCerts.md
+++ b/docs/design/trustedCerts.md
@@ -1,5 +1,5 @@
 ## Use Case
-Often there is a need for trusting non-standard CA certs in the system. For e.g. these could be custom CA certs internal to an enterprise. In order to support secure connectivity to any secure endpoint with certs signed by the custom CA, these CA certs needs to added to the machine's trusted certs store
+Often there is a need for trusting non-standard CA certs in the system. For e.g. these could be custom CA certs internal to an enterprise. In order to support secure connectivity to any secure endpoint with certs signed by the custom CA, these CA certs need to be added to the machine's trusted certs store
 
 ## How to use
 Base64 encode the certificate and add that encoded string to the list under `trustedCerts` property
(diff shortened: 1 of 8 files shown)


pkg/cloud/vsphere/provisioner/common/templates.go
```diff
@@ -250,6 +266,14 @@ users:
   sudo: ALL=(ALL) NOPASSWD:ALL
   groups: sudo
   shell: /bin/bash
+{{- if .TrustedCerts }}
```
figo: will this affect kubeadm?  looks like not, as i don't see certs been passed to kubeadm(e.x --cert-dir).
so this trusted cert is not mean for Kubernetes cluster, correct? thanks

sidharthsurana: Well, for one cluster-api is the one responsible to deploy and setup the node. Also, I think the certs could potentially be used by kubernetes services as well (not saying it will be, but could be)

sflxn: @figo We did something similar in VIC to pull in certificates for private container registries.

docs/design/trustedCerts.md
```diff
@@ -0,0 +1,23 @@
+## Use Case
+Often there is a need for trusting non-standard CA certs in the system. For e.g. these could be custom CA certs internal to an enterprise. In order to support secure connectivity to any secure endpoint with certs signed by the custom CA, these CA certs needs to added to the machine's trusted certs store
```
sflxn: ```these CA certs need to be added...```

sidharthsurana: done!

PR #183, opened 2019-01-29
Update upstream to k8s 1.13 / latest capi master

vincepri:
Signed-off-by: Vince Prignano <vincepri@vmware.com>


sflxn:
@figo Can you look at the build job?  If I am interpreting the build logs correctly, we're failing because we have the word `error` in the logs and that's because we're pulling in libpgp-error?

vincepri:
Thanks! Figured out that the klog init in manager wasn't correct, pushing an update soon

vincepri:
/assign @sflxn
for approval

Issue #191, opened 2019-02-03
create a second target cluster with the management cluster failed


figo:
/kind bug

steps to reproduce:
1) create a bootstrap cluster
2) create a cluster with kubectl create -f cluster.yaml, machine.yaml, machineset.yaml
target cluster created fine
3) deploy another set of cluster.yaml, machine.yaml, machineset.yaml
expecting the second target cluster been created, but it is not happening.

vSphere manager log:
```
I0203 03:25:59.822033       1 round_trippers.go:405] GET https://10.96.0.1:443/apis/cluster.k8s.io/v1alpha1/machines?resourceVersion=2705697&timeoutSeconds=559&watch=true 200 OK in 3 milliseconds
I0203 03:26:03.318305       1 round_trippers.go:405] PUT https://10.96.0.1:443/apis/cluster.k8s.io/v1alpha1/namespaces/default/machines/clusterapidebug-brm87 200 OK in 4 milliseconds
I0203 03:26:11.328803       1 round_trippers.go:405] PUT https://10.96.0.1:443/apis/cluster.k8s.io/v1alpha1/namespaces/default/machines/clusterapi-machineset-2-rljh9 200 OK in 7 milliseconds
I0203 03:26:12.340076       1 round_trippers.go:405] PUT https://10.96.0.1:443/apis/cluster.k8s.io/v1alpha1/namespaces/default/machines/clusterapi-machineset-2-zwv8l 200 OK in 10 milliseconds
I0203 03:28:24.843379       1 reflector.go:428] sigs.k8s.io/cluster-api-provider-vsphere/vendor/sigs.k8s.io/cluster-api/pkg/client/informers_generated/externalversions/factory.go:117: Watch close - *v1alpha1.Machine total 6 items received
I0203 03:28:24.850594       1 round_trippers.go:405] GET https://10.96.0.1:443/apis/cluster.k8s.io/v1alpha1/machines?resourceVersion=2706806&timeoutSeconds=374&watch=true 200 OK in 6 milliseconds
```

figo:
@stobias123 i can confirm that if creating two clusters with different namespaces, it works.
specify all three types of resources: Cluster, Machine, MachineSet with unique namespace per cluster, for example:
```
apiVersion: "cluster.k8s.io/v1alpha1"
kind: Cluster
metadata:
  name: cluster1
  namespace: test1

apiVersion: "cluster.k8s.io/v1alpha1"
kind: Machine
[...]

PR #201, opened 2019-02-11
Update cluster-api, add cluster-name label to machines

vincepri:


vincepri:
/assign @sflxn @frapposelli @figo 

Need some 👀on the changes, especially in Machine* labels to make sure they are correct.

figo:
@vincepri thanks for the work, please update yaml for test as well at https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/scripts/e2e/bootstrap_job/spec/machines.template

we have yet to let test using example yaml.

figo:
@vincepri  shouldn't cluster.yaml adding label as well? both the example yaml and test yaml template.

figo:
> @vincepri shouldn't cluster.yaml adding label as well? both the example yaml and test yaml template.

found out, only the label need to match the cluster name in cluster.yaml

sidharthsurana:
@vincepri Can you provide the motivation/need for this label? Also, the current set of changes are only going to result in these labels exist on the `Machine` objects. Was your intention to have these label show up in the underlying `Node` that these `Machine` create?

figo:
> @vincepri Can you provide the motivation/need for this label? Also, the current set of changes are only going to result in these labels exist on the `Machine` objects. Was your intention to have these label show up in the underlying `Node` that these `Machine` create?

not trying to steal the conversation, just a quick answer, this is to support multi-cluster in one namespace, as discussed at issue: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/191

sidharthsurana:
Thanks @figo I see the main reasoning being https://github.com/kubernetes-sigs/cluster-api/pull/728

vendor/sigs.k8s.io/cluster-api/pkg/controller/machine/controller.go
```diff
@@ -194,18 +197,17 @@ func (r *ReconcileMachine) Reconcile(request reconcile.Request) (reconcile.Resul
...
-		Namespace: machine.Namespace,
-		// This is set so the fake client can be used for unit test. See:
-		// https://github.com/kubernetes-sigs/controller-runtime/issues/168
-		Raw: &metav1.ListOptions{
-			TypeMeta: metav1.TypeMeta{
-				APIVersion: clusterv1.SchemeGroupVersion.String(),
-				Kind:       "Cluster",
-			},
-		},
+		FieldSelector: fields.OneTermEqualSelector("metadata.name", machine.Labels[MachineClusterLabelName]),
```
figo: @vincepri, got the error: Index with name field:metadata.name does not exist.
it should be `fields.OneTermEqualSelector("metadata.name", machine.Labels[MachineClusterLabelName]).String()`?

vincepri:
Waiting for https://github.com/kubernetes-sigs/cluster-api/pull/740

PR #204, opened 2019-02-22
handle nil cluster at machine actuator

figo:
now cluster-api main has made it possible to let cluster object be NULL,
providers can choose to support or not support, but for CAPV, it panic because it assuming
cluster is not NULL, this change is to handle the situation.

a few bonus changes:
1) fix e2e.sh with right ${PROW_JOB_ID} name
2) fix sed usage at Makefile, so it works across Mac/Linux

cc @sflxn @frapposelli @vincepri

### DIFF
--- a/pkg/cloud/vsphere/provisioner/govmomi/utils.go
+++ b/pkg/cloud/vsphere/provisioner/govmomi/utils.go
@@ -25,6 +25,10 @@ var (
 	DefaultSSHPublicKeyFile = "/root/.ssh/vsphere_tmp.pub"
 )
 
+const (
+	ClusterIsNullErr = "cluster is nil"
+)
+
 func (pv *Provisioner) GetKubeadmToken(cluster *clusterv1.Cluster) (string, error) {
 	var token string
 	if cluster.ObjectMeta.Annotations != nil {
--- a/pkg/cloud/vsphere/provisioner/govmomi/utils.go
+++ b/pkg/cloud/vsphere/provisioner/govmomi/utils.go
@@ -25,10 +25,6 @@ var (
 	DefaultSSHPublicKeyFile = "/root/.ssh/vsphere_tmp.pub"
 )
 
-const (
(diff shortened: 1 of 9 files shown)


figo:
> /lgtm
> 
> One question: in which situation could a cluster be nil? User error?

nil cluster could happen when user forgot to set cluster label or set wrong label, i saw panic in my local testing (forgot to set label)

vincepri:
Should we enrich the error message making sure the user knows how to fix the error?

figo:
> Should we enrich the error message making sure the user knows how to fix the error?

good suggestion, to something like: `cluster is nil, please check cluster label setting at machine spec`?

pkg/cloud/vsphere/provisioner/govmomi/utils.go
```diff
@@ -25,6 +25,10 @@ var (
 	DefaultSSHPublicKeyFile = "/root/.ssh/vsphere_tmp.pub"
 )
 
+const (
+	ClusterIsNullErr = "cluster is nil"
```
vincepri:
```suggestion
	ClusterIsNullErr = "cluster is nil, make sure machines have `clusters.k8s.io/cluster-name` label set and the name references a valid cluster name in the same namespace"
```

vincepri:
Left a suggestion ! :)

pkg/cloud/vsphere/provisioner/govmomi/utils.go
```diff
@@ -25,6 +25,10 @@ var (
 	DefaultSSHPublicKeyFile = "/root/.ssh/vsphere_tmp.pub"
 )
 
+const (
```
sidharthsurana: Nit: This could go into `cloud/vsphere/constants/constant.go` as well

figo: looks like that is the proper place, done, thx

Issue #205, opened 2019-02-27
Generating yaml files for self-service steps is wrong using "make prod-yaml" with kustomize master branch


yuyangbj:
The project kustomize has already changed the kustomization.yaml format based on patch below.

https://github.com/kubernetes-sigs/kustomize/pull/700  == all files must be set in sub folder same as kustomization.yaml
https://github.com/kubernetes-sigs/kustomize/pull/752 == secretGenerator has removed commands field

sflxn:
The issue is more related to kustomize v2 and not kustomize.  The repo currently does work with kustomize v1.

yuyangbj:
@sflxn so should we support kustomize v2 now? Or leave it for v1 and create a new branch for v2?

sflxn:
@yuyangbj I believe the other providers are moving to v2 so we might as well move to v2 also.  I assigned myself to this issue, but if you would like to work on this yourself, I can give it to you.  We welcome contributions.

sflxn:
Oh, hi Yang.  Are you Yang from the VIC team?  If so, I didn't realize that at first.  In that case, you're welcome to contribute.

yuyangbj:
@sflxn yes, it is me :) how are you! I tried the cluster-api on my side, but you know everything I build by myself, so something is broken. I will continue to follow up the PR.

PR #206, opened 2019-02-27
Putting resources used by kustomization.yaml into its sub-folder

yuyangbj:
With kustomize PR 700 merged, in order to avoid attacking, all
resources are needed to saved in its sub-folder.

Fixes: #205

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@
 *.out
 
 minikube.kubeconfig
+config/default/vsphere_tmp.pub
+config/default/vsphere_tmp
 
 # tmp files
 *.tmp
--- a/cmd/clusterctl/examples/vsphere/generate-yaml.sh
+++ b/cmd/clusterctl/examples/vsphere/generate-yaml.sh
@@ -109,6 +109,11 @@ if [ ! -f $MACHINE_CONTROLLER_SSH_HOME$MACHINE_CONTROLLER_SSH_PRIVATE_FILE ]; th
   ssh-keygen -t rsa -f $MACHINE_CONTROLLER_SSH_HOME$MACHINE_CONTROLLER_SSH_PRIVATE_FILE  -N ""
 fi
 
+# With kustomize PR 700 merged, the resources in kustomization.yaml could only be scanned in the sub-folder
+# So putting vsphere_tmp and vsphere_tmp.pub in ../config/default folder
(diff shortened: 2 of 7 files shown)


yuyangbj:
I have  already sign the CLA, why it is still failed?

frapposelli:
@yuyangbj CLA looks fine now

/ok-to-test

figo:
@yuyangbj  thanks for the PR, kubebuilder requires `rbac` and `crds` under config, 
also we don't have to move them, kustomize suggest to use `bases` for none-local files, you do need to add kustomize.yaml in each folder, please refer to what been done at cluster-api main: https://github.com/kubernetes-sigs/cluster-api/blob/master/config/default/kustomization.yaml

figo:
feel free to cancel the hold when you updated the PR, the current e2e test does not generate yaml from scratch, so can not catch the problem, we should address this soon.

yuyangbj:
@frapposelli @figo thanks for the comments! I did not notice the kubebuilder problem in my local env, so I choose to keep resources there! Ok, I will use the bases to replace the resources and add kustomization.yaml in every folder.

yuyangbj:
@figo I updated the PR, thanks!

figo:
/lgtm 
we need to catch up on test part.

frapposelli:
/lgtm

@sflxn PTAL and approve

sidharthsurana:
> @yuyangbj Can you add the two key files that you copied to the .gitignore file so we don't accidentally check them into github in the future?

LGTM, once the .gitignore file is updated we can merge

sflxn:
@yuyangbj Do you know whether Kustomize v1 compatible with these changes?  If not, can you create a new issue to update the makefile so it can detect kustomize version and warn users to update it?

yuyangbj:
@sidharthsurana thanks for the review.

@sflxn I will update  the .gitignore and test the v1 again.

yuyangbj:
@sflxn i tested, the makefile cannot work with kustomize v1 now because the secreteGenerator:files cannot be supported in V1. I want to use the 'kustomize version' to show version, but I got an error:
Version: {KustomizeVersion:unknown GitCommit:$Format:%H$ BuildDate:1970-01-01T00:00:00Z GoOs:linux GoArch:amd64}
I see it is an old issue in kustomize... so besides waiting the fix, any recommends for the workaround?

sflxn:
@yuyangbj you don't have to deal with the makefile for this PR.  I was suggesting you open up another issue for the makefile changes.  For this PR, update this doc page with a note on Kustomize v2, https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/tree/master/docs/self-service.

yuyangbj:
@sidharthsurana @sflxn @frapposelli @figo done for a request changes, please review again.

PR #211, opened 2019-03-01
Use the machine's ID as the instance UUID of the VM

sidharthsurana:
Updates the create workflow to also look for the created
VM using the instance UUID before triggering the clone
operation. This would avoid the system to go into a recursive
clone operation in case the update for the first clone
task is missed somehow.

The updated create workflow would look like:
https://tinyurl.com/y5u8lpjl

Change-Id: I91810168193d1f2ca6002233e84f42a994917077


sidharthsurana:
> @sidharthsurana Can you add an md file in the design folder that includes your diagram? Draw.io will export a png.

Done!

PR #213, opened 2019-03-02
Fix cloning from template with snapshot and simultaneous disk expansion

### LINKED ISSUE #212: Cloning from template with snapshot and simultaneous disk expansion fails
Currently, if the source template used has snapshots and the in the target machine spec the disk is requested to be expanded the clone operation fails.

sidharthsurana:
See https://code.vmware.com/doc/preview?id=4206#/doc/vim.vm.RelocateSpec.DiskMoveOptions.html
for more details

Resolves #212
Change-Id: Ic62701aabc134a3e5038d440eccf902e52bd3351

### DIFF
--- a/pkg/cloud/vsphere/provisioner/govmomi/create.go
+++ b/pkg/cloud/vsphere/provisioner/govmomi/create.go
@@ -162,7 +162,7 @@ func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1
 		spec.Config.MemoryMB = machineConfig.MachineSpec.MemoryMB
 	}
 	spec.Config.Annotation = fmt.Sprintf("Virtual Machine is part of the cluster %s managed by cluster-api", cluster.Name)
-	spec.Location.DiskMoveType = string(types.VirtualMachineRelocateDiskMoveOptionsMoveAllDiskBackingsAndAllowSharing)
+	spec.Location.DiskMoveType = string(types.VirtualMachineRelocateDiskMoveOptionsMoveAllDiskBackingsAndConsolidate)
 	var src *object.VirtualMachine
 	if vsphereutils.IsValidUUID(machineConfig.MachineSpec.VMTemplate) {
 		// If the passed VMTemplate is a valid UUID, then first try to find it treating that as InstanceUUID


figo:
not sure i fully understand the "moveAllDiskBackingsAndAllowSharing" policy, could we include any vCenter engineer?

figo:
disk consolidation for every clone can be expensive, we should try to avoid it,
could we discuss snapshots use case itself?
   1) if user trying to use snapshot to create multiple versions of template, could user use multiple templates instead?
   2) we are not restricting snapshot, but we can detect snapshot, potentially give a warning.

sidharthsurana:
Copying the message I posted on the slack channel for the record.

To provider some background on the different disk moves types:

`moveAllDiskBackingsAndAllowSharing`: (Currently used) From the vsphere API docs:
```
All of the virtual disk's backings should be moved to the new datastore. If a disk backing is not the
child-most backing of this virtual machine, and there exists a read-only disk backing with the same
content ID on the target datastore, then this disk backing may not be copied. Instead it is acceptable
to attach to the read-only disk backing at the target datastore. A read-only disk backing is defined
as a virtual disk backing which no virtual machine is currently writing to.
```
 What this helps is if there is a snapshot present then the read-only point referred to the snapshot will be used to create the new disk chain in the target VM. (This is a bit faster since now it only copied the redo logs post that snapshot to the target VM. However the downside is since there is read-only snapshot thus the disk size cannot be modified anymore. This is what causes the issue).

`moveAllDiskBackingsAndConsolidate`: (Proposed to be used) From the vsphere API docs:
```
All of the virtual disk's backings should be moved to the new datastore. During a clone operation
or a MigrateVM_Task, any delta disk backings will be consolidated.
```
So in this case, even if there are snapshots associated with the source VM, there will be a disk consolidation that would happen as part of the clone operation now (given the source vm has snapshots).

sidharthsurana:
> disk consolidation for every clone can be expensive, we should try to avoid it,
> could we discuss snapshots use case itself?
> 
> 1. if user trying to use snapshot to create multiple versions of template, could user use multiple templates instead?
> 2. we are not restricting snapshot, but we can detect snapshot, potentially give a warning.

Yes you are absolutely correct. Ideally we would want to use the snapshots to our advantage. 
I have just opened a new Issue #216 that we can use to further improve the way we utilize the disks and their snapshots more specifically use linked clones as an option as well.

sidharthsurana:
**Note:** We would probably want to also use the snapshots directly for its various benefits. I have opened another Issue #217 for future enhancement in the same area

Issue #214, opened 2019-03-04
Update makefile to check kustomize version


sflxn:
We have updated the repo to be compatible with kustomize v2.  We should look into warning users using kustomize v1 and ask them to update.

sflxn:
cc @yuyangbj

yuyangbj:
@sflxn yes, I can take it but I cannot update the assignees by myself, do you know why?

PR #227, opened 2019-03-11
Enable the latest Kustomize version check

yuyangbj:
With the kustomize code changes, some resources, for example, 'commands'
is not supported, and new resources are introduced into the latest
code, so we add the kustomize version check.

Fixes: #214

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,8 @@ DEV_IMG ?= # <== NOTE:  outside dev, change this!!!
 VERSION ?= $(shell git describe --exact-match 2> /dev/null || \
 	   git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
 
+KUSTOMIZE_VERSION := $(shell kustomize version | awk '{ print $$2 }' | awk -F':' '{ print $$2} ')
+
 all: test manager clusterctl
 
 # Run tests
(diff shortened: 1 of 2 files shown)


yuyangbj:
I did not change the ci integration code, hope someone who own the CI could help to change it.

Makefile
```diff
@@ -22,6 +22,12 @@ DEV_IMG ?= # <== NOTE:  outside dev, change this!!!
 VERSION ?= $(shell git describe --exact-match 2> /dev/null || \
 	   git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
 
+KUSTOMIZE_VERSION = $(kustomize version | awk '{ print $2 }' | awk -F ':' '{ print $2} ')
```
sflxn: Which shell are you using?  On my Mac, I'm using bash and when I tried this code in the makefile, it doesn't work.  The variable is empty upon executing the above.  If I execute the command at a terminal, it works fine.

sflxn: It appears you want
```
KUSTOMIZE_VERSION ?= $(shell kustomize version | awk '{ print $$2 }' | awk -F ':' '{ print $$2} ')
```

Makefile
```diff
@@ -22,6 +22,12 @@ DEV_IMG ?= # <== NOTE:  outside dev, change this!!!
...
+ifneq (,$(filter $(KUSTOMIZE_VERSION), "v1", "unknown"))
```
sflxn: Does filter check if it's v1 or unknown?  What will this line do if someone builds kustomize from source and the version doesn't have a correct version?

Makefile
yuyangbj: Yes, I found the error, but I use the ':=' because KUSTOMIZATION_VERSION is not a environment variable.

Makefile
sflxn: True.  Now that I think about it, this is the behavior we want.  This code is fine.

sflxn:
This build failure may not be caused by this PR.  Investigating.

sflxn:
@yuyangbj This PR should pass CI once #231 gets merged.

PR #231, opened 2019-03-15
Pin docker to version 18.06 to ensure kubeadm's compatibility

### LINKED ISSUE #230: Ubuntu cloud init image is now failing for CAPV
Something has changed in the last 2 days that's causing CAPV to fail creating cluster.  It started happening in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/227.  I've verified that it is happening locally as well.  I tried doing code archaeology to find the root cause, but it's failing on commits that merged successfully.

This does appear to be something new around the cloud init image we are using for testing.

sflxn:
Ubuntu updated the docker version on their server that causes kubeadm init to fail.
It is preventing clusterctl from creating target clusters.  Currently, we pull the
latest version of docker off of Ubuntu's servers, which at this time is 18.09.2.
This change will pin the docker version to 18.06.x.

Resolves #230

### DIFF
--- a/pkg/cloud/vsphere/provisioner/common/templates.go
+++ b/pkg/cloud/vsphere/provisioner/common/templates.go
@@ -379,7 +379,6 @@ deb [arch=amd64] https://apt.dockerproject.org/repo ubuntu-xenial main
 EOF
 
 apt-get update
-apt-get install -y docker.io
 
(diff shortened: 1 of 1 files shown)


figo:
thanks for the fix, we can re-visit this version pinning a little later.

Issue #239, opened 2019-03-27
Latest kubeadm unable to decode our kubeadm_config.yaml during node bootstrap


sflxn:
With the latest kubeadm update, our node bootstrapping is failing.

```
+ kubeadm init --config /etc/kubernetes/kubeadm_config.yaml
unable to decode config: v1alpha2.MasterConfiguration.APIServerCertSANs: []string: API: v1alpha2.API.BindPort: AdvertiseAddress: ReadString: expects " or n, but found 0, error found in #10 byte of ...|Address":0
,"bindPort|..., bigger context ...|{"api":{"advertiseAddress":0,"bindPort":443},"apiServerCertSANs":[0,0],"apiSer|...
```

sflxn:
It appears to be an ubuntu bionic issue.

PR #240, opened 2019-03-27
Fix bootstrap script to fix kubernetes-cni issue in k8s 1.14

### LINKED ISSUE #238: After k8s 1.14 was release, installation of kubeadm and kubelet failing because it wants kubernetes-cni 0.6.0
With the latest k8s 1.14 update,  CAPV is failing on node bootstrap.  This is a known issue.  There is a work around by pinning kubernetes-cni=0.6.0-00.  This issue is here to track our resolution for v1alpha1.

Issue 75683 on the kubernetes repo.

sflxn:
With the latest k8s 1.14, there was a fix to resolve a CVE.  This required
kubernetes-cni 0.7.5 for k8s 1.14 and up.  This fix will first attempt to
install kubernetes-cni 0.7.5.  If the subsequent install of kubelet and
kubeadm fails, it will fallback kubernetes-cni to 0.6.0 and retry installing
the kubelet and kubeadm.

Resolves #238

### DIFF
--- a/pkg/cloud/vsphere/provisioner/common/templates.go
+++ b/pkg/cloud/vsphere/provisioner/common/templates.go
@@ -406,7 +406,20 @@ KUBELET=$(getversion kubelet ${KUBELET_VERSION}-)
 KUBEADM=$(getversion kubeadm ${KUBELET_VERSION}-)
 KUBECTL=$(getversion kubectl ${KUBELET_VERSION}-)
 apt-get install -y docker.io=${DOCKER_VER}
-# Explicit cni version is a temporary workaround till the right version can be automatically detected correctly
+
+### TEMPORARY solution
+# https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/238
+# Currently, ubuntu packaging includes kubernetes-cni 0.7.5 when performing
+# apt-get install kubernetes-cni.  Older versions of k8s complains and this
+# script fails.  This will install 0.7.5 for k8s 1.14 and higher.  It will
+# fall back to 0.6.0 for older versions.
+
+if [[ "${KUBELET_VERSION}" -ge "1.14" ]]; then
+	apt-get install -y kubernetes-cni=0.7.*
+else
+	apt-get install -y kubernetes-cni=0.6.*
+fi
+
 apt-get install -y kubelet=${KUBELET} kubeadm=${KUBEADM} kubectl=${KUBECTL}
 
 {{- end }}{{/* end install */}}
@@ -489,10 +502,20 @@ KUBEADM=$(getversion kubeadm ${KUBELET_VERSION}-)
 
 apt-get install -y docker.io=${DOCKER_VER}
 
-# Explicit cni version is a temporary workaround till the right version can be automatically detected correctly
-apt-get install -y \
-    kubelet=${KUBELET} \
-    kubeadm=${KUBEADM}
+### TEMPORARY solution
+# https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/238
+# Currently, ubuntu packaging includes kubernetes-cni 0.7.5 when performing
+# apt-get install kubernetes-cni.  Older versions of k8s complains and this
+# script fails.  This will install 0.7.5 for k8s 1.14 and higher.  It will
+# fall back to 0.6.0 for older versions.
+
+if [[ "${KUBELET_VERSION}" -ge "1.14" ]]; then
+	apt-get install -y kubernetes-cni=0.7.*
+else
+	apt-get install -y kubernetes-cni=0.6.*
+fi
+
+apt-get install -y kubelet=${KUBELET} kubeadm=${KUBEADM}
 
 mv /usr/bin/kubeadm.dl /usr/bin/kubeadm
 chmod a+rx /usr/bin/kubeadm
@@ -577,7 +600,7 @@ EOF
 
 {{- end }}{{/* end MajorMinorVersion 1.11 or 1.12 */}}
 
-{{ if eq .MajorMinorVersion "1.13" }}
+{{ if ge .MajorMinorVersion "1.13" }}
 cat > /etc/kubernetes/kubeadm_config.yaml <<EOF
 apiVersion: kubeadm.k8s.io/v1beta1
 kind: InitConfiguration


sflxn:
I am not sure this will pass CI as there is another issue introduced in k8s 1.14.  That issue is #239.  If this does not pass, I will add the fix for that issue into this PR when I figure out that fix.

pkg/cloud/vsphere/provisioner/common/templates.go
```diff
@@ -406,8 +406,27 @@ KUBELET=$(getversion kubelet ${KUBELET_VERSION}-)
 KUBEADM=$(getversion kubeadm ${KUBELET_VERSION}-)
 KUBECTL=$(getversion kubectl ${KUBELET_VERSION}-)
 apt-get install -y docker.io=${DOCKER_VER}
-# Explicit cni version is a temporary workaround till the right version can be automatically detected correctly
-apt-get install -y kubelet=${KUBELET} kubeadm=${KUBEADM} kubectl=${KUBECTL}
+
+### TEMPORARY solution
```
neolit123: LGTM on the temporary side.
these solutions are not ideal, but we are waiting on SIG Release to decide on how to proceed with fixing the packages (e.g. revisions).
cc @tpepper

^ you can also ask Tim on slack.

pkg/cloud/vsphere/provisioner/common/templates.go
```diff
@@ -406,8 +406,27 @@ KUBELET=$(getversion kubelet ${KUBELET_VERSION}-)
 KUBEADM=$(getversion kubeadm ${KUBELET_VERSION}-)
 KUBECTL=$(getversion kubectl ${KUBELET_VERSION}-)
 apt-get install -y docker.io=${DOCKER_VER}
-# Explicit cni version is a temporary workaround till the right version can be automatically detected correctly
-apt-get install -y kubelet=${KUBELET} kubeadm=${KUBEADM} kubectl=${KUBECTL}
```
sidharthsurana: While debugging this issue with @sflxn  seems like the upstream repo has for now the 0.7.5 version of the kubernetes-cni available for apt-get install.

Thus we would fall back to explicitly choosing the kubernetes-cni version based on the kubernetes version we are trying to install.

neolit123: ideally CNI 0.7.5 should be installed in any case for anything > 1.11, because it includes a security fix.

also i would like to point out that the packages are a "best effort" and in a flux, thus if you are pinning your base distro to Ubuntu already, you might as well just download the k8s binaries, cni binaries and systemd unit files and not rely on the packages until they are in a better shape.

sflxn: @neolit123 I noticed when I install kubernetes-cni 0.7.5 manually, it also installs kubelet and kubeadm 1.14.

neolit123: > I noticed when I install kubernetes-cni 0.7.5 manually, it also installs kubelet and kubeadm 1.14.

hm, i cannot confirm this on Debian Stretch:

```
$ apt-get install kubernetes-cni=0.7.5-00
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following NEW packages will be installed:
  kubernetes-cni
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 6473 kB of archives.
After this operation, 50.2 MB of additional disk space will be used.
Get:1 https://packages.cloud.google.com/apt kubernetes-xenial/main amd64 kubernetes-cni amd64 0.7.5-00 [6473 kB]
Fetched 6473 kB in 3s (1918 kB/s)         
debconf: delaying package configuration, since apt-utils is not installed
Selecting previously unselected package kubernetes-cni.
(Reading database ... 7341 files and directories currently installed.)
Preparing to unpack .../kubernetes-cni_0.7.5-00_amd64.deb ...
Unpacking kubernetes-cni (0.7.5-00) ...
Setting up kubernetes-cni (0.7.5-00) ...
```

again, i do encourage the vsphere provider to do this:

> ... just download the k8s binaries, cni binaries and systemd unit files and not rely on the packages until they are in a better shape.

the process i'm talking about is similar to what we are doing here for container linux:
https://kubernetes.io/docs/setup/independent/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl

except that the versions need a bump and also these complimentary non-go dependencies are nice-to-have:
https://github.com/kubernetes/release/blob/master/debian/xenial/kubelet/debian/control#L13

sidharthsurana: I agree with the approach to use the binaries directly.
However, the only question is do we want to do that right way in`v1alpha1`  or maybe do that after the milestone.
My opinion is handle this change post `v1apha1` but I will let @sflxn also chime in here.

sflxn: I think this is a temporary fix to allow us to get on with other issues.  I think we will need to revisit how we install packages post v1alpha1.  This may get changed within CAPI so I don't want to make such a radical changes that might get ripped out in a few months.

sflxn:
@neolit123 @sidharthsurana After discussing this issue with Siddharth on my logic for the backporting, I don't think it will work.  I just made pushed a much simpler logic.

btw, I also fixed another issue in the file for which I don't want to bother writing up a new issue.

pkg/cloud/vsphere/provisioner/common/templates.go
```diff
@@ -489,10 +502,20 @@ KUBEADM=$(getversion kubeadm ${KUBELET_VERSION}-)
...
+# https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/238
+# Currently, ubuntu packaging includes kubernetes-cni 0.7.5 when performing
+# apt-get install kubernetes-cni.  Older versions of k8s complains and this
+# script fails.  This sill install 0.7.5 for k8s 1.14 and higher.  It will
+# fall back to 0.6.0 for older versions.
+
+if [[ "${KUBELET_VERSION}" -ge "1.14" ]]; then
+	apt-get install -y kubernetes-cni=0.7.5-00*
+else
+	apt-get install -y kubernetes-cni=0.6.0-00*
```
sidharthsurana: lets do `apt-get install -y kubernetes-cni=0.6.*` instead so that future backports can be included

sflxn:
awaiting #241 to get merged and the CI infrastructure issue fixed.  after, will retest.

PR #250, opened 2019-04-18
Add support for specifying vsphere credentials via secrets

### LINKED ISSUE #9: [vSphere] Store provider config variables in secrets
_From @karan on May 29, 2018 17:13_

This is for security and safety, since the provider config blob can contain sensitive credentials.

/kind feature

/assign @karan

_Copied from original issue: kubernetes-sigs/cluster-api#226_

sidharthsurana:
- Adds new field vsphereCredentialSecret that can be specified to
  point to the secret that holds the credentials

Resolves #9

Change-Id: I8e43864302a51386f7f11725b0a966cc9e436c59

### DIFF
--- a/docs/design/vsphereCredentials.md
+++ b/docs/design/vsphereCredentials.md
@@ -1,0 +1,49 @@
+## Passing vsphere credentials
+For the cluster-api vsphere provider to work, the users need to provide the vsphere credentials to access the infrastructure. There are 2 ways how the users can provide these credentials.
+
+* Using kubernetes `secrets`
+   * Create a secret that contains 2 keys namely `username` and `password` in the same namespace as the desired `Cluster` object.
+      ```
+      apiVersion: v1
+      kind: Secret
+      metadata:
+        name: my-vc-credentials
+      type: Opaque
+      data:
+        # base64 encoded fields
+        username: YWRtaW5pc3RyYXRvckB2c3BoZXJlLmxvY2Fs
+        password: c2FtcGxl
+      ```
+
+   * Set the `vsphereCredentialSecret` property in the `ProviderSpec` part of the `Cluster` definition
+      ```
+      apiVersion: "cluster.k8s.io/v1alpha1"
+      kind: Cluster
+      metadata:
+        name: sample-cluster
+      spec:
+          ...
+          providerSpec:
+            value:
+              ...
+              # Credentials provided via secrets
+              vsphereCredentialSecret: "my-vc-credentials"
+      ```
+
+* Using plain text credential in the `ProviderSpec` part of the `Cluster` definition
+    ```
+    apiVersion: "cluster.k8s.io/v1alpha1"
+    kind: Cluster
+    metadata:
+      name: sample-cluster
+    spec:
+        ...
+        providerSpec:
+          value:
+            ...
+            # Credentials provided as plain text
+            vsphereUser: "administrator@vsphere.local"
+            vspherePassword: "sample"
+    ```
+
+__Note:__ If `vsphereCredentialSecret` field is set to a non empty string then the controller will ignore the `vsphereUser` and `vspherePassword` fields even if they are set.
--- a/pkg/cloud/vsphere/provisioner/govmomi/utils.go
+++ b/pkg/cloud/vsphere/provisioner/govmomi/utils.go
@@ -190,3 +190,27 @@ func (pv *Provisioner) GetKubeConfig(cluster *clusterv1.Cluster) (string, error)
 	}
 	return string(secret.Data[constants.KubeConfigSecretData]), nil
 }
+
+func (pv *Provisioner) GetVsphereCredentials(cluster *clusterv1.Cluster) (string, string, error) {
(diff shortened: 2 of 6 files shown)


pkg/cloud/vsphere/provisioner/govmomi/utils.go
```diff
@@ -190,3 +190,28 @@ func (pv *Provisioner) GetKubeConfig(cluster *clusterv1.Cluster) (string, error)
...
+		if err != nil {
+			klog.Warningf("Error reading secret %s", vsphereConfig.VsphereCredentialSecret)
+			return "", "", err
+		}
+		if username, ok := secret.Data[constants.VsphereUserKey]; ok {
+			if password, ok := secret.Data[constants.VspherePasswordKey]; ok {
+				return string(username), string(password), nil
+			}
+		}
+		klog.Errorf("Secret %s should have the keys `%s` and `%s` defined in it", vsphereConfig.VsphereCredentialSecret, constants.VsphereUserKey, constants.VspherePasswordKey)
```
figo: could 212-213 re-use an error variable?

sidharthsurana: Done! I removed the extra log print line, since that error message is going to show up in the logs regardless.

docs/design/vsphereCredentials.md
```diff
@@ -0,0 +1,61 @@
+## Passing vsphere credentials
+For the cluster-api vsphere provider to work, the users need to provide the vsphere credentials to access the infrastructure. There are 2 ways how the users can provide these credentials.
+
+1. Using kubernetes `secrets`
+   1. Create a secret that contains 2 keys namely `username` and `password` in the same namespace as the desired `Cluster` object.
```
sflxn: Can you label this a or - instead of 1?  Markdown should allow you to do -.  It's easier to follow if we don't have two 1 and two 2.

docs/design/vsphereCredentials.md
```diff
@@ -0,0 +1,61 @@
...
+      metadata:
+        name: my-vc-credentials
+      type: Opaque
+      data:
+        # base64 encoded fields
+        username: YWRtaW5pc3RyYXRvckB2c3BoZXJlLmxvY2Fs
+        password: c2FtcGxl
+      ```
+
+   2. Set the `vsphereCredentialSecret` property in the `ProviderSpec` part of the `Cluster` definition
```
sflxn: same as above

docs/design/vsphereCredentials.md
```diff
@@ -0,0 +1,61 @@
...
+                  cidrBlocks: ["10.96.0.0/12"]
+              pods:
+                  cidrBlocks: ["192.168.0.0/16"]
+              serviceDomain: "cluster.local"
+          providerSpec:
+            value:
+              apiVersion: "vsphereproviderconfig/v1alpha1"
+              kind: "VsphereClusterProviderConfig"
+              vsphereServer: "10.20.30.40"
+              vsphereCredentialSecret: "my-vc-credentials"
```
sflxn: Can you add some marker so readers know what to look for?

```
             vsphereCredentialSecret: "my-vc-credentials"         <=== HERE
```

docs/design/vsphereCredentials.md
```diff
@@ -0,0 +1,61 @@
...
+            pods:
+                cidrBlocks: ["192.168.0.0/16"]
+            serviceDomain: "cluster.local"
+        providerSpec:
+          value:
+            apiVersion: "vsphereproviderconfig/v1alpha1"
+            kind: "VsphereClusterProviderConfig"
+            vsphereServer: "10.20.30.40"
+            vsphereUser: "administrator@vsphere.local"
+            vspherePassword: "sample"
```
sflxn: Same comment as above.  add a marker `<=== HERE`.

pkg/cloud/vsphere/provisioner/govmomi/utils.go
```diff
@@ -190,3 +190,27 @@ func (pv *Provisioner) GetKubeConfig(cluster *clusterv1.Cluster) (string, error)
...
+		if err != nil {
+			klog.Warningf("Error reading secret %s", vsphereConfig.VsphereCredentialSecret)
+			return "", "", err
+		}
+		if username, ok := secret.Data[constants.VsphereUserKey]; ok {
+			if password, ok := secret.Data[constants.VspherePasswordKey]; ok {
+				return string(username), string(password), nil
+			}
+		}
+		return "", "", fmt.Errorf("Improper secret: Secret %s should have the keys `%s` and `%s` defined in it", vsphereConfig.VsphereCredentialSecret, constants.VsphereUserKey, constants.VspherePasswordKey)
```
sflxn: I don't think we should log the secrets here.

sidharthsurana: We are not logging the `Secret` rather we are printing how the secret should be.
Example output would be:
```Improper secret: Secret my-secret should have the keys `username` and `password` defined in it```

Issue #247, opened 2019-04-08
Feature: create the resource pool programmatically


sflxn:
For cluster api users on vSphere today, they must have a resource pool set up prior to creating a target cluster.  This can be handled programmatically.

dvonthenen:
Implemented and now currently testing

PR #251, opened 2019-04-19
Implements ResourcePool creation and Deploy Direct to Cluster/Host

dvonthenen:
This PR implements the following:
- If the ResourcePool does not exist in vsphere, create it just prior to cloning the master/worker nodes.
- If a ResoucePool is not specified, deploy the target cluster directly onto the cluster/host default resource pool.

Testing performed on vSphere 6.7u2:
- using an existing Resource Pool
- creating a new Resource Pool
- deploying directly to a cluster

Assumptions:
- the template was created/converted on a host within the compute cluster (ie template is owned by or it's parent is a host within the compute cluster).

Other notes:
- This was only tested on cluster, but this should also work with the effort to support standalone ESX hosts.

Addresses issue: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/247

### DIFF
--- a/pkg/cloud/vsphere/provisioner/govmomi/create.go
+++ b/pkg/cloud/vsphere/provisioner/govmomi/create.go
@@ -10,6 +10,7 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
@@ -173,13 +174,6 @@ func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1
 	}
 	spec.Location.Datastore = types.NewReference(ds.Reference())
 
-	pool, err := s.finder.ResourcePoolOrDefault(ctx, machineConfig.MachineSpec.ResourcePool)
-	if err != nil {
-		return err
-	}
-	spec.Location.Pool = types.NewReference(pool.Reference())
-	spec.PowerOn = true
-
 	spec.Config = &types.VirtualMachineConfigSpec{}
 	// Use the object UID as the instanceUUID for the VM
 	spec.Config.InstanceUuid = string(machine.UID)
@@ -223,6 +217,46 @@ func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1
 	if err != nil {
 		return fmt.Errorf("error fetching virtual machine or template properties: %s", err)
 	}
+
+	host, err := src.HostSystem(ctx)
+	if err != nil {
+		klog.Errorf("HostSystem failed. err=%s", err)
+		return err
+	}
+
+	if len(machineConfig.MachineSpec.ResourcePool) > 0 {
+		pool, err := s.finder.ResourcePoolOrDefault(ctx, machineConfig.MachineSpec.ResourcePool)
+
+		if _, ok := err.(*find.NotFoundError); ok {
+			klog.Warningf("Failed to find ResourcePool=%s err=%s. Attempting to create it.", machineConfig.MachineSpec.ResourcePool, err)
+
+			poolRoot, errRoot := host.ResourcePool(ctx)
+			if errRoot != nil {
+				klog.Errorf("Failed to find root ResourcePool. err=%s", errRoot)
+				return errRoot
+			}
+
+			klog.Info("Creating ResourcePool using default values. These values can be modified after ResourcePool creation.")
+			pool, err = poolRoot.Create(ctx, machineConfig.MachineSpec.ResourcePool, types.DefaultResourceConfigSpec())
+			if err != nil {
+				klog.Errorf("Create ResourcePool failed. err=%s", err)
+				return err
+			}
+		}
+
+		spec.Location.Pool = types.NewReference(pool.Reference())
+	} else {
+		klog.Infof("Attempting to use Host ResourcePool")
+		pool, err := host.ResourcePool(ctx)
+
+		if err != nil {
+			klog.Errorf("Host ResourcePool failed. err=%s", err)
+			return err
+		}
+
+		spec.Location.Pool = types.NewReference(pool.Reference())
+	}
+	spec.PowerOn = true
 
 	if machineConfig.MachineSpec.VsphereCloudInit {
 		// In case of vsphere cloud-init datasource present, set the appropriate extraconfig options
--- a/pkg/cloud/vsphere/provisioner/govmomi/create.go
+++ b/pkg/cloud/vsphere/provisioner/govmomi/create.go
@@ -135,8 +135,71 @@ func (pv *Provisioner) verifyAndUpdateTask(s *SessionContext, machine *clusterv1
 
 // CloneVirtualMachine clones the template to a virtual machine.
 func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
+	ctx, cancel := context.WithCancel(*s.context)
+	defer cancel()
+
+	machineConfig, err := vsphereutils.GetMachineProviderSpec(machine.Spec.ProviderSpec)
+	if err != nil {
+		return err
+	}
+
+	dc, err := s.finder.DatacenterOrDefault(ctx, machineConfig.MachineSpec.Datacenter)
+	if err != nil {
+		return err
+	}
+	s.finder.SetDatacenter(dc)
+
+	// Let's check to make sure we can find the template earlier on... Plus, we need
+	// the cluster/host info with we do happen to want to deploy direct to the cluster/host.
+	var src *object.VirtualMachine
+	if vsphereutils.IsValidUUID(machineConfig.MachineSpec.VMTemplate) {
+		// If the passed VMTemplate is a valid UUID, then first try to find it treating that as InstanceUUID
+		// In case if are not able to locate a matching VM then fall back to searching using the VMTemplate
+		// as a name
+		klog.V(4).Infof("Trying to resolve the VMTemplate as InstanceUUID %s", machineConfig.MachineSpec.VMTemplate)
+		si := object.NewSearchIndex(s.session.Client)
+		instanceUUID := true
+		templateref, err := si.FindByUuid(ctx, dc, machineConfig.MachineSpec.VMTemplate, true, &instanceUUID)
+		if err != nil {
+			return fmt.Errorf("error quering virtual machine or template using FindByUuid: %s", err)
+		}
+		if templateref != nil {
+			src = object.NewVirtualMachine(s.session.Client, templateref.Reference())
+		}
+	}
+	if src == nil {
+		klog.V(4).Infof("Trying to resolve the VMTemplate as Name %s", machineConfig.MachineSpec.VMTemplate)
+		src, err = s.finder.VirtualMachine(ctx, machineConfig.MachineSpec.VMTemplate)
+		if err != nil {
+			klog.Errorf("VirtualMachine finder failed. err=%s", err)
+			return err
+		}
+	}
+
+	host, err := src.HostSystem(ctx)
+	if err != nil {
+		klog.Errorf("HostSystem failed. err=%s", err)
+		return err
+	}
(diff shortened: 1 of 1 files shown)


pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -224,6 +217,30 @@ func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1
...
+	host, err := src.HostSystem(ctx)
+	if err != nil {
+		return fmt.Errorf("HostSystem failed. err=%s", err)
+	}
+
+	pool, err := s.finder.ResourcePoolOrDefault(ctx, machineConfig.MachineSpec.ResourcePool)
+	if err != nil {
+		klog.Warningf("Failed to find ResourcePool=%s err=%s. Attempting to create it.", machineConfig.MachineSpec.ResourcePool, err)
+
+		poolRoot, errRoot := host.ResourcePool(ctx)
```
sflxn: I think the finder has a function to get the resource pool so you don't even need to get the host, but this works too.

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -224,6 +217,30 @@ func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1
...
+		if errRoot != nil {
+			klog.Errorf("Failed to find root ResourcePool. err=%s", errRoot)
+			return errRoot
+		}
+
+		pool, err = poolRoot.Create(ctx, machineConfig.MachineSpec.ResourcePool, types.DefaultResourceConfigSpec())
```
sflxn: I would also log the fact that we're using default values for the resource pool.  It feels like we should expand the spec to allow users to specify the cpu and memory allocation value of the pool in the yaml files.  I'll add a new issue for this and link it to this issue.

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -292,8 +309,13 @@ func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1
 	// all.
 	diskMap := func(diskSpecs []vsphereconfigv1.DiskSpec) map[string]int64 {
 		diskMap := make(map[string]int64)
-		for _, s := range diskSpecs {
-			diskMap[s.DiskLabel] = s.DiskSizeGB
+		for i, s := range diskSpecs {
+			diskLabel := s.DiskLabel
+			if len(diskLabel) == 0 {
+				klog.Warningf("Missing label for VMDK. Using \"Hard disk %d\"", (i + 1))
+				diskLabel = fmt.Sprintf("Hard disk %d", (i + 1))
```
sidharthsurana: I don't think we should do this. The original intent is for the user to be sure of what they want. Thus we want them the be precise about which disk out of their template they want to resize. Having this logic may lead to unintended behavior that the user may or may not have anticipated. Since this piece touches the data I prefer that the user to be more explicit in their choice rather than us making some assumptions.

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -292,8 +309,13 @@ func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1
 	// all.
 	diskMap := func(diskSpecs []vsphereconfigv1.DiskSpec) map[string]int64 {
 		diskMap := make(map[string]int64)
-		for _, s := range diskSpecs {
-			diskMap[s.DiskLabel] = s.DiskSizeGB
+		for i, s := range diskSpecs {
+			diskLabel := s.DiskLabel
+			if len(diskLabel) == 0 {
```
sflxn: great!

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -330,6 +352,10 @@ func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1
 	// Add new nics based on the user info
 	nicid := int32(-100)
 	for _, network := range machineConfig.MachineSpec.Networks {
+		if len(network.NetworkName) == 0 {
```
sflxn: 👍

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: Yup sounds good!

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: Was one of the things initially tripping me up when first deploying. 👍

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: Actually, I should be able to also call the root RP from the template itself. Testing it out now.

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: Makes sense... how about if there is only a single VMDK, we attempt to use hard disk label. If the label does not match even using the default value, then the deployment is going to fail anyways.

sidharthsurana: The point in the `disks` section of the spec is optional.. the user if they don't want to make any disk changes they can simply leave that section out and that would work just fine.

For the Network part, yes it will fail. But the proper way is we need to report it back on the machineStatus to let the user know what went wrong, or what was the missing info. We haven't implemented that yet, but it would be great if you can add that as the default mechanism of giving the user feedback on what went wrong!

figo: the providerSpec required and optional items are not been documented, it is in api code, but not obvious to users, we need to improve on this. (obvious not for this PR)

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: Yea, I just looked at doing it that way, but this seems to work without a bunch of additional code and handles the standalone ESX host case using the same code path.

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: yea @figo that's what I was kinda working around.

By default, when you generate the `./out` files, it creates the yaml for both the disks and network fields. I ended up getting stuck for a bit because this was not immediately obvious without a lot of trial and error.

This was an attempt to  use the default generated `./out` files and also give the user experience a "it just works" type feel. Either way, I can remove both code blocks. Just let me know what the consensus is.

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: updated

pkg/cloud/vsphere/provisioner/govmomi/create.go
sflxn: Sidharth has a good point.  When I saw this, I just thought of it as more validation logic.  We should remove this.

dvonthenen: no problem. will remove it.

dvonthenen:
Have updated based on feedback and removed those sections

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -224,6 +217,31 @@ func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1
...
 	}
 
+	host, err := src.HostSystem(ctx)
+	if err != nil {
+		return fmt.Errorf("HostSystem failed. err=%s", err)
+	}
+
+	pool, err := s.finder.ResourcePoolOrDefault(ctx, machineConfig.MachineSpec.ResourcePool)
+	if err != nil {
+		klog.Warningf("Failed to find ResourcePool=%s err=%s. Attempting to create it.", machineConfig.MachineSpec.ResourcePool, err)
```
dougm: You can limit the error case where pool doesn't exist like so:
```go
if _, ok := err.(*find.NotFoundError); !ok {
   return err
}
// ... attempt to create the pool
```

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -224,6 +217,31 @@ func (pv *Provisioner) cloneVirtualMachine(s *SessionContext, cluster *clusterv1
 		return fmt.Errorf("error fetching virtual machine or template properties: %s", err)
 	}
 
+	host, err := src.HostSystem(ctx)
+	if err != nil {
+		return fmt.Errorf("HostSystem failed. err=%s", err)
+	}
+
+	pool, err := s.finder.ResourcePoolOrDefault(ctx, machineConfig.MachineSpec.ResourcePool)
```
sidharthsurana: Can you see if we can make the rp optional? if the user does not specify one then things should work too.

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: Good suggestion... will update.

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: Looking into this...

dvonthenen: Implemented and testing now...

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: Implemented and testing now

dvonthenen:
This is ready for review... initial PR comment has been updated.

/hold cancel

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -134,33 +135,83 @@ func (pv *Provisioner) verifyAndUpdateTask(s *SessionContext, machine *clusterv1
...
+	if vsphereutils.IsValidUUID(machineConfig.MachineSpec.VMTemplate) {
+		// If the passed VMTemplate is a valid UUID, then first try to find it treating that as InstanceUUID
+		// In case if are not able to locate a matching VM then fall back to searching using the VMTemplate
+		// as a name
+		klog.V(4).Infof("Trying to resolve the VMTemplate as InstanceUUID %s", machineConfig.MachineSpec.VMTemplate)
+		si := object.NewSearchIndex(s.session.Client)
+		instanceUUID := true
+		templateref, err := si.FindByUuid(ctx, dc, machineConfig.MachineSpec.VMTemplate, true, &instanceUUID)
+		if err != nil {
+			return fmt.Errorf("error quering virtual machine or template using FindByUuid: %s", err)
```
sflxn: querying.  Missing y.

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -134,33 +135,83 @@ func (pv *Provisioner) verifyAndUpdateTask(s *SessionContext, machine *clusterv1
...
 		return err
 	}
-	ctx, cancel := context.WithCancel(*s.context)
-	defer cancel()
+	s.finder.SetDatacenter(dc)
 
-	var spec types.VirtualMachineCloneSpec
-	machineConfig, err := vsphereutils.GetMachineProviderSpec(machine.Spec.ProviderSpec)
+	// Let's check to make sure we can find the template earlier on... Plus, we need
+	// the cluster/host info with we do happen to want to deploy direct to the cluster/host.
```
sflxn: Second sentence doesn't read right.

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: fixed

pkg/cloud/vsphere/provisioner/govmomi/create.go
dvonthenen: fixed

dvonthenen:
Updated with changes...

PR #253, opened 2019-04-25
Add support for specifying ntp servers in the MachineSpec

### LINKED ISSUE #244: Feature: Support ntp
At times the time on the hypervisors can be off, thus causing various issues when running the k8s cluster. For .e.g. Certificated that are generated may become invalid due to time sync discrepancies.

Thus we should be able to define ntp servers in the machine specs, so that they all should come up with the same time.

sidharthsurana:
Resolves #244

Change-Id: I0cb436d3e778686ab70ad01e119c94eb4407e28d

### DIFF
--- a/config/crds/vsphereproviderconfig_v1alpha1_vsphereclusterproviderconfig.yaml
+++ b/config/crds/vsphereproviderconfig_v1alpha1_vsphereclusterproviderconfig.yaml
@@ -26,6 +26,8 @@ spec:
           type: string
         metadata:
           type: object
+        vsphereCredentialSecret:
+          type: string
         vspherePassword:
           type: string
         vsphereServer:
@@ -33,9 +35,8 @@ spec:
         vsphereUser:
           type: string
       required:
-      - vsphereUser
-      - vspherePassword
       - vsphereServer
+      type: object
   version: v1alpha1
 status:
   acceptedNames:
--- a/docs/design/ntpServers.md
+++ b/docs/design/ntpServers.md
@@ -1,0 +1,16 @@
+## Use Case
+For a distributed system to operate properly, you need the time to be synchronized across the machines. For instance, when a master node is created kubeadm generates a bunch of certificates for the system to use for various purposes. Now in case, the time of the master node and the worker nodes that come up later is not the same there is a possibility that the certificates generated would not work. This is just a simple example of why time synchronization across the nodes is important. A simple way to ensure that is by synchronizing the time on all the machines with NTP server/s.
+
+## How to use
+Provide the list of ntp servers accessible in your environment via the `ntpServers` property as shown below
+```
+providerSpec:
+  value:
+    apiVersion: "vsphereproviderconfig/v1alpha1"
+    kind: "VsphereMachineProviderConfig"
+    machineSpec:
+      ...
+      ntpServers:
+      - 0.pool.ntp.org
+      - 1.pool.ntp.org
+```
--- a/pkg/apis/vsphereproviderconfig/v1alpha1/vsphereclusterproviderconfig_types.go
+++ b/pkg/apis/vsphereproviderconfig/v1alpha1/vsphereclusterproviderconfig_types.go
@@ -47,10 +47,10 @@ type VsphereClusterProviderConfig struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	VsphereUser             string `json:"vsphereUser"`
-	VspherePassword         string `json:"vspherePassword"`
+	VsphereUser             string `json:"vsphereUser,omitempty"`
+	VspherePassword         string `json:"vspherePassword,omitempty"`
 	VsphereServer           string `json:"vsphereServer"`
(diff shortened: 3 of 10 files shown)


config/crds/vsphereproviderconfig_v1alpha1_vsphereclusterproviderconfig.yaml
```diff
@@ -26,16 +26,17 @@ spec:
...
+        vsphereCredentialSecret:
+          type: string
         vspherePassword:
           type: string
         vsphereServer:
           type: string
         vsphereUser:
           type: string
       required:
-      - vsphereUser
```
figo: is there a way to express that: vsphereUser, vspherePassword, vsphereCredentialSecret can not empty at the same time?

sidharthsurana: Not that I know of. I looked at using the `oneof` but seems like this file is auto-generated and I don't know a good way to express that.
If you have some ideas please feel free to let me know.

pkg/apis/vsphereproviderconfig/v1alpha1/vsphereclusterproviderconfig_types.go
```diff
@@ -47,10 +47,10 @@ type VsphereClusterProviderConfig struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	VsphereUser             string `json:"vsphereUser"`
-	VspherePassword         string `json:"vspherePassword"`
+	VsphereUser             string `json:"vsphereUser,omitempty"`
```
sflxn: This looks like a patch for a previous commit and has nothing to do with ntp servers at all.  It's fine, but you need to document this in the commit.  Also, I noticed your commit has no description body.  Please add that, and comment that you also fixed some issues with a previous commit.

https://chris.beams.io/posts/git-commit/

config/crds/vsphereproviderconfig_v1alpha1_vsphereclusterproviderconfig.yaml
figo: after a little bit research, i don't aware of anything like this in openAPI schema level.

with that, could you handle the situation at `GetVsphereCredentials()`, return error if all three are empty from usr input. currently? currently it returns with username, password both = nil, then fail implicitly later. which is not idea!

sflxn: Anything that we do to perform validation is good but also futile.  Even if we error out, the current machine controller doesn't know how to "kick out" of an error condition.  It will keep trying to recreate the machine on any error.  Ideally, we would have something like a validating webhook to check the validity of the spec and decide whether to allow it or not.  That way, the request wouldn't even get to the machine controller.  We could also define a standard set of errors that all providers follow and have the machine controller understand those errors.

pkg/apis/vsphereproviderconfig/v1alpha1/vsphereclusterproviderconfig_types.go
sidharthsurana: Done

docs/design/ntpServers.md
```diff
@@ -0,0 +1,16 @@
+## Use Case
+Often for the kuberentes cluster to operate properly, you need the time to be synchronized across the machines. For instance, when a master node is created kubeadm generates a bunch of certificates for the system to use for various purposes. Now in case, the time of the master node and the worker nodes that come up later is not the same there is a possibility that the certificates generated would not work. This is just a simple example of why time synchronization across the nodes is important. A simple way to ensure that is by synchronizing the time on all the machines with NTP server/s.
+
+## How to use
+Provide the list of ntp servers accessible in your environment via the `ntpServers` property like below
```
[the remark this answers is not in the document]
sidharthsurana: done

docs/design/ntpServers.md
```diff
@@ -0,0 +1,16 @@
+## Use Case
+Often for the kuberentes cluster to operate properly, you need the time to be synchronized across the machines. For instance, when a master node is created kubeadm generates a bunch of certificates for the system to use for various purposes. Now in case, the time of the master node and the worker nodes that come up later is not the same there is a possibility that the certificates generated would not work. This is just a simple example of why time synchronization across the nodes is important. A simple way to ensure that is by synchronizing the time on all the machines with NTP server/s.
```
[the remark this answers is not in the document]
sidharthsurana: done

config/crds/vsphereproviderconfig_v1alpha1_vsphereclusterproviderconfig.yaml
sidharthsurana: I agree a proper webhook implementation is what might be a good solution. Also, we need to have a proper status field similar to say how pods have. Where we could tell the user that the machine/cluster deployment has errored out with the appropriate error details. And when in such state the controller should stop retrying.

figo: i will write an issue for this, looks like we just have to live with the current situation.

PR #254, opened 2019-04-26
[NKS-1351] Support Debian templates

sverrirsig:
This change to the injection scripts for user cluster masters/nodes allows Debian based templates with cloud-init to be used instead of Ubuntu.


sverrirsig:
This PR went further across the wire than intended - mi scusi.

PR #257, opened 2019-05-01
Explicitly delete the node object bound to the Machine on deletion

### LINKED ISSUE #256: Node not removed when Machine is deleted
The default behavior observed in k8s 1.11.x was that when one deletes the Machine object, the node associated with the machine would automatically get removed from the node list in a couple of minutes. This behavior got changed soon after 1.11.6 or so. Thus with k8s 1.12.x or newer, the node object is left behind even after the Machine object and the underlying VM is deleted.

sidharthsurana:
In the latest version of the kubernetes the node object is not automatically removed once the underlying VM is deleted. Thus this leaves the system with additional nodes that are in `NotReady`
state for which the VMs have already been removed. Ultimately, this should be handled in the k/k controllers. However, for the time being, the cluster-api controller can handle the node
removal upon machine deletion

Resolves #256

### DIFF
--- a/pkg/cloud/vsphere/provisioner/govmomi/delete.go
+++ b/pkg/cloud/vsphere/provisioner/govmomi/delete.go
@@ -9,6 +9,7 @@ import (
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


sidharthsurana:
I accidentally messed up my fork of the repo and had to re-create it. Thus I will recreate a new pull request for this again

PR #260, opened 2019-05-08
Replace relative path with absolute for ssh call

corrieb:
It would seem that some shells don't expand relative paths correctly in exec.Command(). The ssh call in clusterctl on the Mac was failing with:

```
I0507 14:04:00.305127    2608 getkubeconfig.go:60] Waiting for kubeconfig on vs-master-rk7lw to become ready...
I0507 14:04:00.305545    2608 utils.go:170] pulling kubeconfig (using ssh) from 10.118.69.111
Warning: Identity file ~/.ssh/vsphere_tmp not accessible: No such file or directory.
I0507 14:04:00.780473    2608 utils.go:183] ssh failed with error = exit status 255
I0507 14:04:00.780534    2608 getkubeconfig.go:63] error getting kubeconfig: exit status 255
```
Simple fix was to get the home dir for the current user and set it as an absolute path

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.10.3 as builder
+FROM golang:1.12 as builder
 
 # Copy in the go src
 WORKDIR /go/src/sigs.k8s.io/cluster-api-provider-vsphere
--- a/pkg/cloud/vsphere/utils/utils.go
+++ b/pkg/cloud/vsphere/utils/utils.go
@@ -9,6 +9,7 @@ import (
 	"math"
 	"os"
 	"os/exec"
+	"os/user"
 	"strings"
 	"time"
 
@@ -168,14 +169,24 @@ func GetKubeConfig(cluster *clusterv1.Cluster, master *clusterv1.Machine) (strin
 		return "", err
 	}
 	klog.Infof("pulling kubeconfig (using ssh) from %s", ip)
+
+	usr, err := user.Current()
+	if err != nil {
+		klog.Infof("failed to get current user: %s", err.Error())
+		return "", err
+	}
+	sshDir := fmt.Sprintf("%s/.ssh/vsphere_tmp", usr.HomeDir)
+	klog.Infof("using ssh key in %s", sshDir)
+
 	var out bytes.Buffer
 	cmd := exec.Command(
-		"ssh", "-i", "~/.ssh/vsphere_tmp",
+		"ssh", "-i", sshDir,
 		"-q",
 		"-o", "StrictHostKeyChecking no",
 		"-o", "UserKnownHostsFile /dev/null",
 		fmt.Sprintf("ubuntu@%s", ip),
 		"sudo cat /etc/kubernetes/admin.conf")
+
 	cmd.Stdout = &out
 	cmd.Stderr = os.Stderr
 	err = cmd.Run()


corrieb:
Well pppffff. See https://github.com/golang/go/issues/14626. Apparently this is an issue if the binaries are cross-compiled. This seems to have been fixed 6 months ago, but maybe isn't in the compiler in use here. Seems like a less reliable alternative to rely on $USER? I'm open to suggestions.

figo:
> Well pppffff. See [golang/go#14626](https://github.com/golang/go/issues/14626). Apparently this is an issue if the binaries are cross-compiled. This seems to have been fixed 6 months ago, but maybe isn't in the compiler in use here. Seems like a less reliable alternative to rely on $USER? I'm open to suggestions.

https://github.com/mitchellh/go-homedir this one looks compelling

frapposelli:
Looks like this was fixed in go 1.12, we should rev up the compiler version.

sflxn:
If this is fixed in 1.12, we should be fine.  I believe k8s require 1.12 now.  We need to update the document to bump the go requirements.

sflxn:
We need to also update the Dockerfile to go 1.12.  I'll create an issue to update the repo to go 1.12.

sflxn:
Sorry, did not see that e2e failed.  If Fabio is right about 1.12, then this should work if we update the Dockerfile to 1.12.  @corrieb Can you try that?

sflxn:
I looked at the golang sources at https://github.com/golang/go/blob/master/src/os/user/lookup_stubs.go#L22

If os.Getuid() doesn't return a string representing a number >= 0, then it'll be "".  Then the call to lookupUserId() will fail.

https://github.com/golang/go/blob/master/src/os/user/lookup_stubs.go#L25

I suspect this is what's happening in the container env.  That might be why the update to golang 1.12.5 still fails in e2e.  As a test, we can make a call to `id := os.Getuid()` and print out id.  If it's empty, that would match up to my suspicion.

frapposelli:
I would suggest to take in https://github.com/mitchellh/go-homedir as @figo suggested, it's a popular library and then we can revisit this problem at a later date once it's figured out upstream in golang.

sflxn:
Yeah, I agree.  Using Mitchell's project (if it works in a container) is the best path for now.

Btw, I have figured out the bug in the golang sources.  It turns out it's a caching problem.  I'm going to submit a PR to fix the problem to them when I figure out the best way to do it.

sflxn:
Ignore my last comment.  I was still using the old Dockerfile.  After updating golang in the Dockerfile, it works for me in my environment so I no longer think it's a golang problem.

I looked at the build output and see it's using the latest golang 1.12 container to build.  Something is wrong in our CI environment, and it's not apparent to me yet.

figo:
> Yeah, I agree. Using Mitchell's project (if it works in a container) is the best path for now.
> 
> Btw, I have figured out the bug in the golang sources. It turns out it's a caching problem. I'm going to submit a PR to fix the problem to them when I figure out the best way to do it.
> 
> [golang/go#31949](https://github.com/golang/go/issues/31949)

@sflxn  i don't feel this is a CI issue, but i could be wrong, CI build clusterctl fresh.(that is this PR about)
i need to find more time to come back and investigate, but feel free to take lead.

figo:
> Ignore my last comment. I was still using the old Dockerfile. After updating golang in the Dockerfile, it works for me in my environment so I no longer think it's a golang problem.
> 
> I looked at the build output and see it's using the latest golang 1.12 container to build. Something is wrong in our CI environment, and it's not apparent to me yet. @figo

@sflxn i don't feel this is a CI issue, but i could be wrong, CI build clusterctl fresh.(that is this PR about)
i need to find more time to come back and investigate, but feel free to take lead

figo:
found the problem, please update the Dockerfile for clusterctl, https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/cmd/clusterctl/Dockerfile

cc @corrieb

sflxn:
@corrieb @figo Yup, that's it.  I saw that in the build too, but I didn't think it should affect the controller.  However, Hui pointed out that it's the clusterctl calling that code that is producing the message, and it's built with the older Golang container.  Good job, Hui!

Ben, when you update that dockerfile, would you mind updating the golang version in this markdown page also?  :)

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/docs/self-service/README.md

PR #262, opened 2019-05-10
Add namespace flag to limit to a single namespace

ncdc:
Add optional --namespace flag so the controllers only watch for cluster
and machine objects in a single namespace.

Signed-off-by: Andy Goldstein <goldsteina@vmware.com>


ncdc:
@detiber fyi

PR #269, opened 2019-05-28
Update build.sh with build-ova.sh

jimangel:
It appears the build script has moved.


figo:
@jimangel thank you for the pull, just want to clarify one thing, the OVA build entry point is still `build.sh`, `build.sh` is a wrapper on top of `build-ova.sh` to support run build either with container or without.

jimangel:
Ah, sorry, my mistake! Thanks

/close

PR #271, opened 2019-05-29
Explicitly delete the node object bound to the Machine on deletion

### LINKED ISSUE #256: Node not removed when Machine is deleted

sidharthsurana:
In the latest version of the kubernetes, the node object is not automatically removed once the underlying VM is deleted. Thus this leaves the system with additional nodes that are in `NotReady`
state for which the VMs have already been removed. Ultimately, this should be handled in the k/k controllers. However, for the time being, the cluster-api controller can handle the node
removal upon machine deletion

Resolves #256

### DIFF
--- a/pkg/cloud/vsphere/provisioner/govmomi/delete.go
+++ b/pkg/cloud/vsphere/provisioner/govmomi/delete.go
@@ -9,6 +9,7 @@ import (
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/constants"
 	vsphereutils "sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/utils"
@@ -20,7 +21,27 @@ func (pv *Provisioner) Delete(ctx context.Context, cluster *clusterv1.Cluster, m
 	if cluster == nil {
 		return errors.New(constants.ClusterIsNullErr)
 	}
+	if exists, _ := pv.Exists(ctx, cluster, machine); exists {
+		err := pv.powerOffAndDestroy(ctx, cluster, machine)
+		if err != nil {
+			return err
+		}
+	}
 
+	// Delete the node object bound to the Machine via the Status.NodeRef reference
+	if machine.Status.NodeRef != nil {
+		// Do a force delete (grace period set to 0) as the underlying VM is already deleted
+		err := pv.k8sClient.Core().Nodes().Delete(machine.Status.NodeRef.Name, metav1.NewDeleteOptions(0))
+		if err != nil {
+			// Log the warning for unable to delete and that the user can delete the node manually. This error
+			// should not hold the Machine delete operation since the underlying VM is already deleted at this point
+			klog.Warningf("Could not remove the node %s bound to the machine automatically. Please manually remove the node if it is not already removed. Error encountered: %s", machine.Status.NodeRef.Name, err.Error())
+		}
+	}
+	return nil
+}
+
+func (pv *Provisioner) powerOffAndDestroy(ctx context.Context, cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
 	s, err := pv.sessionFromProviderConfig(cluster, machine)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


pkg/cloud/vsphere/provisioner/govmomi/delete.go
```diff
@@ -20,52 +21,69 @@ func (pv *Provisioner) Delete(ctx context.Context, cluster *clusterv1.Cluster, m
 	if cluster == nil {
 		return errors.New(constants.ClusterIsNullErr)
 	}
+	if exists, _ := pv.Exists(ctx, cluster, machine); exists {
+		err := pv.powerOffAndDestroy(ctx, cluster, machine)
+		if err != nil {
```
vincepri: This could be simplified with `if err := ...; err != nil`

pkg/cloud/vsphere/provisioner/govmomi/delete.go
```diff
@@ -20,52 +21,69 @@ func (pv *Provisioner) Delete(ctx context.Context, cluster *clusterv1.Cluster, m
...
+	err = s.session.RetrieveOne(deletectx, vmref, []string{"name", "runtime.powerState"}, &vm)
+	if err != nil {
+		return err
+	}
+	pv.eventRecorder.Eventf(machine, corev1.EventTypeNormal, "Killing", "Killing machine %v", machine.Name)
+	vmo := object.NewVirtualMachine(s.session.Client, vmref)
+	if vm.Runtime.PowerState == types.VirtualMachinePowerStatePoweredOn {
+		task, err := vmo.PowerOff(deletectx)
 		if err != nil {
+			klog.Infof("Error trigerring power off operation on the Virtual Machine %s", vm.Name)
```
vincepri: Should this be klog.Errorf instead of Info?

pkg/cloud/vsphere/provisioner/govmomi/delete.go
```diff
@@ -20,52 +21,69 @@ func (pv *Provisioner) Delete(ctx context.Context, cluster *clusterv1.Cluster, m
...
 		}
-		var vm mo.VirtualMachine
-		vmref := types.ManagedObjectReference{
-			Type:  "VirtualMachine",
-			Value: moref,
-		}
-		err = s.session.RetrieveOne(deletectx, vmref, []string{"name", "runtime.powerState"}, &vm)
+		err = task.Wait(deletectx)
 		if err != nil {
+			klog.Infof("Error powering off the Virtual Machine %s", vm.Name)
```
vincepri: Same as above, re:Errorf

pkg/cloud/vsphere/provisioner/govmomi/delete.go
```diff
@@ -20,52 +21,69 @@ func (pv *Provisioner) Delete(ctx context.Context, cluster *clusterv1.Cluster, m
...
-		if taskinfo.State == types.TaskInfoStateSuccess {
-			klog.Infof("Virtual Machine %v deleted successfully", vm.Name)
-			pv.eventRecorder.Eventf(machine, corev1.EventTypeNormal, "Killed", "Machine %v deletion complete", machine.Name)
-			return nil
-		}
+	}
+	task, err := vmo.Destroy(deletectx)
+	taskinfo, err := task.WaitForResult(deletectx, nil)
+	if taskinfo.State == types.TaskInfoStateSuccess {
+		klog.Infof("Virtual Machine %v deleted successfully", vm.Name)
```
vincepri: Instead of `%v` should we use `%q` so the name is put in quotes? Makes reading the logs a little easier

pkg/cloud/vsphere/provisioner/govmomi/delete.go
```diff
@@ -20,52 +21,69 @@ func (pv *Provisioner) Delete(ctx context.Context, cluster *clusterv1.Cluster, m
...
 		pv.eventRecorder.Eventf(machine, corev1.EventTypeNormal, "Killed", "Machine %v deletion complete", machine.Name)
-		klog.Errorf("VM Deletion failed on pv with following reason %v", taskinfo.Reason)
-		return errors.New("VM Deletion failed")
+		return nil
 	}
-	return nil
+	pv.eventRecorder.Eventf(machine, corev1.EventTypeNormal, "Killed", "Machine %v deletion complete", machine.Name)
+	klog.Errorf("VM Deletion failed on pv with following reason %v", taskinfo.Reason)
+	return errors.New("VM Deletion failed")
```
vincepri: Should we return the name of the Virtual Machine as part of the error? Might help with debugging

sidharthsurana: done

pkg/cloud/vsphere/provisioner/govmomi/delete.go
sidharthsurana: nice suggestion :) done

pkg/cloud/vsphere/provisioner/govmomi/delete.go
sidharthsurana: done

pkg/cloud/vsphere/provisioner/govmomi/delete.go
sidharthsurana: done

pkg/cloud/vsphere/provisioner/govmomi/delete.go
sidharthsurana: done

pkg/cloud/vsphere/provisioner/govmomi/delete.go
```diff
@@ -20,52 +21,68 @@ func (pv *Provisioner) Delete(ctx context.Context, cluster *clusterv1.Cluster, m
 	if cluster == nil {
 		return errors.New(constants.ClusterIsNullErr)
 	}
+	if exists, _ := pv.Exists(ctx, cluster, machine); exists {
+		if err := pv.powerOffAndDestroy(ctx, cluster, machine); err != nil {
+			return err
```
figo: i think we should start to return `controllerError.RequeueAfterError` in order for controller to re-queue, otherwise, there will be many cases that machine object been deleted, but those VMs or nodes never been deleted.

pkg/cloud/vsphere/provisioner/govmomi/delete.go
```diff
@@ -20,52 +21,68 @@ func (pv *Provisioner) Delete(ctx context.Context, cluster *clusterv1.Cluster, m
...
+		}
+	}
+
+	// Delete the node object bound to the Machine via the Status.NodeRef reference
+	if machine.Status.NodeRef != nil {
+		// Do a force delete (grace period set to 0) as the underlying VM is already deleted
+		err := pv.k8sClient.Core().Nodes().Delete(machine.Status.NodeRef.Name, metav1.NewDeleteOptions(0))
+		if err != nil {
```
figo: we could let controller re-try by returning `controllerError.RequeueAfterError`, my thinking is we should try to achieve consistent state, ignore the error (throw warning) means: node deletion may success or fail, it will be two possiable states.

akutz:
Hi @sidharthsurana,

Is this still a valid PR? If so, are you able to rebase it?

akutz:
Hi @sidharthsurana,

Do you know if this is still an issue?

PR #279, opened 2019-06-04
Update CRDs

akutz:
This patch updates the generated CRDs to be consistent with the sources from which the CRDs are generated.

cc @frapposelli @figo


akutz:
@figo Could you possibly approve this? @frapposelli? It is blocking #280

config/crds/vsphereproviderconfig_v1alpha1_vsphereclusterproviderconfig.yaml
```diff
@@ -42,5 +36,5 @@ status:
   acceptedNames:
     kind: ""
     plural: ""
-  conditions: []
-  storedVersions: []
+  conditions: null
```
sidharthsurana: IIRC the `null` here causes issues.
This generated CRD has this diff but was intentionally not merged because of the `null` issue.
I always thought this was because of some incorrect version of the tooling on my local machine.
Can you verify once that this change will not cause an issue when deploying the CAPV?

akutz: Hi @sidharthsurana,

I won't get into the details, but we spent some time (@frapposelli, @figo, and myself) figuring out the last time someone actually generated the CRDs ahead of committing them. It wasn't since the first commit. They're very stale. We used `kustomize` v2 exactly, and we all generated the same deltas. If this causes a failure, then the failure is in the project or its settings, because there's currently a bug in CI where the CRDs aren't actually generated properly on the server-side due to the CI jobs not including or downloading `kustomize`.

If the generated CRD is invalid, then the sources are invalid or the tooling is invalid. Either way, if there is *any* **manual** step between generating the CRD and using the CRD, then that's a bug.

akutz:
/hold

Need to update the CI process to have customize

config/crds/vsphereproviderconfig_v1alpha1_vsphereclusterproviderconfig.yaml
akutz: > Can you please verify once that this change will not cause an issue when deploying the CAPV?

Hi @sidharthsurana,

If it does, then good, because if generating the CRDs causes CAPV to fail, then then either the CRD sources, the tooling, or the process needs to be fixed. It has to be manual or generated -- it can't be both.

akutz:
/hold cancel

This is the wrong PR that needs to be held. It's #280.

akutz:
Hi @figo and @frapposelli,

Any objections to getting this merged?

config/crds/vsphereproviderconfig_v1alpha1_vsphereclusterproviderconfig.yaml
sidharthsurana: I agree, but the point is, if this CRD breaks things, then we should fix the source code so that it generates the right CRD rather than checking in CRD that may not work.

figo:
> Hi @figo and @frapposelli,
> 
> Any objections to getting this merged?
> 
> Regarding @sidharthsurana's concern, @figo, please correct me if I'm wrong, but doesn't the e2e test use the CRD during the deploy process? So if the PR passes the e2e test, the CRD is valid?

no, ci is using a static version provider-component.yaml right now:
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/scripts/e2e/bootstrap_job/spec/provider-components-v2.0.template

i finally find time to dig around kube-builder, i think capi and cap-* all hand edited config/crd/*.yaml, otherwise, description should be empty.

tested with `condition: null` got error below:
```
Waiting for kubectl apply... unknown error couldn't kubectl apply, output: namespace "vsphere-provider-system" configured
error: error validating "STDIN": error validating data: [ValidationError(CustomResourceDefinition.status): missing required field "conditions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionStatus, ValidationError(CustomResourceDefinition.status): missing required field "storedVersions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionStatus]; if you choose to ignore these errors, turn validation off with --validate=false
: exit status 1
```
i was not fully understand the implication when we discussed yesterday. 
could this be a kubebuilder issue? (i am using kubebuilder 1.0.8)

cc @sidharthsurana

akutz:
I'm glad you caught this @figo. And kudos to @sidharthsurana for the warning. If the CRDs are currently hand-edited then we definitely need to get the sources fixed.

/hold

sidharthsurana:
> no, ci is using a static version provider-component.yaml right now:
> https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/scripts/e2e/bootstrap_job/spec/provider-components-v2.0.template
> 
> i finally find time to dig around kube-builder, i think capi and cap-* all hand edited config/crd/*.yaml, otherwise, description should be empty.
> 
> tested with `condition: null` got error below:
> 
> ```
> Waiting for kubectl apply... unknown error couldn't kubectl apply, output: namespace "vsphere-provider-system" configured
> error: error validating "STDIN": error validating data: [ValidationError(CustomResourceDefinition.status): missing required field "conditions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionStatus, ValidationError(CustomResourceDefinition.status): missing required field "storedVersions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionStatus]; if you choose to ignore these errors, turn validation off with --validate=false
> : exit status 1
> ```
> 
> i was not fully understand the implication when we discussed yesterday.
> could this be a kubebuilder issue? (i am using kubebuilder 1.0.8)

That's what I thought I remembered from the time we moved to CRDs. It most likely could be a kubebuilder issue. Although I couldn't find any specific issue

akutz:
Hi @figo and @sidharthsurana,

I figured out the issue. It was a bad version of the controller-tools. In fact, @andrewsykim noted that this version was based on a commit ID that was no longer even available if resolved today. I updated the controller-tools to match CAPA's, and the CRDs have been updated as well. They now match as they did before with one small change each -- and this change matches the layout in CAPA's CRDs. I think the extra `type: object` was an artifact of manual creation, not realizing it was unnecessary.

Please review and approve. Thanks!

PR #283, opened 2019-06-05
Install kustomize v2 if missing

akutz:
This patch installs kustomize v2.0.0 if the tool is missing and provides a centralized means to install build tooling in `hack/ensure-tools.sh`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,9 @@ config/default/vsphere_tmp
 # tmp files
 *.tmp
 
+# Sed backup files
+*-e
+
 *.bzl
 *.bazel
 
--- a/Makefile
+++ b/Makefile
@@ -36,17 +36,15 @@ DEV_IMG ?= # <== NOTE:  outside dev, change this!!!
 VERSION ?= $(shell git describe --exact-match 2> /dev/null || \
 	   git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
 
-KUSTOMIZE := $(shell command -v kustomize 2>/dev/null)
-ifeq (,$(strip $(KUSTOMIZE)))
-$(KUSTOMIZE):
-	GO111MODULE=off go get sigs.k8s.io/kustomize
-endif
-
-build: clusterctl manager
-
-# Run tests
-test: generate fmt vet manifests
-	go test ./pkg/... ./cmd/... -coverprofile cover.out
+# Ensure the tooling required by the build is present in hack/.bin and that
+# the tool directory is in the PATH
+TOOLS_BIN_DIR := $(abspath hack/.bin)
+export PATH := $(TOOLS_BIN_DIR):$(PATH)
+KUSTOMIZE := $(TOOLS_BIN_DIR)/kustomize
+TOOLS += $(KUSTOMIZE)
+tools $(TOOLS):
+	hack/ensure-tools.sh
+.PHONY: tools
 
 # Build manager binary
 manager: fmt vet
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -18,5 +18,8 @@ config/default/vsphere_tmp
 # tmp files
 *.tmp
 
+# Sed backup files
+*-e
```
andrewsykim: yes please

akutz:
> I'm not sure that installing kustomize is the right approach here, but I think the version check should be saner. Or we can assume the user is using v2, log that requirement and forget about it.

Hi @andrewsykim,

So I'm not sure if having `=v2.0.0` is actually required any longer. This was when we thought it was the specific version of Kustomize as to the reason that CRDs and other YAML were resulting in invalid output. But as the discussion in #279 shows, it wasn't the necessarily the version of a tool that was the cause for the invalid, generated output. It seems the generated output was once-upon-a-time created manually, and the generation process was never actually used/verified.

However, there *is* precedent for installing tooling. Other CAPI providers do the same. That's what Bazel does. And again, we're not so much installing Kustomize as downloading the binary into a specific directory away from the standard PATH or system dirs.

andrewsykim:
> And again, we're not so much installing Kustomize as downloading the binary into a specific directory away from the standard PATH or system dirs.

Ah, missed this part. Fine with me then :)

Makefile
```diff
@@ -98,14 +101,11 @@ dev-push:
 ###################################
 
 # Create YAML file for deployment
-prod-yaml:
-ifeq ($(KUSTOMIZE_VERSION),$(filter $(KUSTOMIZE_VERSION),unknown v1))
-	@echo "please upgrade kustomize version to v2 at least."
-else
+prod-yaml: | $(KUSTOMIZE)
+	@$(KUSTOMIZE) version 2>&1 | grep -q 'KustomizeVersion:v\{0,1\}[2-9][[:digit:]]\{0,\}' || { echo "kustomize v2+ required" 1>&2; exit 1; }
```
andrewsykim: If we're dictating the version of kustmoize, do we still need these version checks?

akutz: Hi @andrewsykim,

Sí, because the only time Kustomize is installed to a sub-directory of the project is if `command -v kustomize` returns a non-zero exit code. Meaning we'll use an existing version of Kustomize if it's present.

akutz:
Hi @andrewsykim,

Can you please review for approval if you're satisfied with the answers to your earlier concerns?

Makefile
andrewsykim: On linux, `kustomize version` doesn't have the `v` prefix which breaks the Makefile:
```
$ kustomize version
Version: {KustomizeVersion:2.0.3 GitCommit:a6f65144121d1955266b0cd836ce954c04122dc8 BuildDate:2019-03-05T20:37:42Z GoOs:linux GoArch:amd64}
```

Does this account for that?

akutz:
```

Comment (akutz) [reply Makefile]: Sí, that's what `v\{0,1\}` does. It says either zero or one `v`.

**san v**
```shell
$ cat <<EOF | grep -o 'KustomizeVersion:v\{0,1\}[2-9][[:digit:]]\{0,\}'
> $ kustomize version
> Version: {KustomizeVersion:2.0.3 GitCommit:a6f65144121d1955266b0cd836ce954c04122dc8 BuildDate:2019-03-05T20:37:42Z GoOs:linux GoArch:amd64}
> EOF
KustomizeVersion:2
```

**cum v**
```shell
$ cat <<EOF | grep -o 'KustomizeVersion:v\{0,1\}[2-9][[:digit:]]\{0,\}'
> $ kustomize version
> Version: {KustomizeVersion:v2.0.3 GitCommit:a6f65144121d1955266b0cd836ce954c04122dc8 BuildDate:2019-03-05T20:37:42Z GoOs:linux GoArch:amd64}
> EOF
KustomizeVersion:v2
```

andrewsykim: thanks

akutz: FWIW, the regex says `KustomizeVersion:` followed by an optional `v` followed by a digit that is `2-9` followed one or more additional, optional digits.

Basically, it has to be at least version 2 *something*, but we don't care what.

andrewsykim: Hope this doesn't break with kustomize v10.* :trollface:

PR #287, opened 2019-06-06
restore e2e test cleanup

figo:
restore cleanup vms

cc @akutz


akutz:
Crap. Didn't mean for that to get committed...

PR #295, opened 2019-06-07
remove the default weave CNI 

andrewsykim:
CNI installation should not be part of target cluster provisioning. This PR updates cloud init bootscript to not automatically install weave CNI, however, for the initial target cluster created with `clusterctl` will still install weave by default. Discussions on-going to see if weave as a default still makes sense. 

Also updates the default pod and service CIDRs to `100.96.0.0/11` and `100.64.0.0/13` respectively


cmd/clusterctl/examples/vsphere/addons.yaml.template
```diff
@@ -6,4 +6,205 @@ metadata:
     storageclass.kubernetes.io/is-default-class: "true"
 provisioner: kubernetes.io/vsphere-volume
 parameters:
-  datastore: ""
\ No newline at end of file
+  datastore: ""
+
+# install weave CNI
```
andrewsykim: @akutz we might have discussed this earlier, but technically we don't need this either for the `kind` cause since kind installs CNI for us right?

akutz: That's correct. And even if it doesn't, I'm not sure Weave should ever be present in the addons file.

sidharthsurana:
Let's wait for the v1alpha2 to shape up at least till the time when CNI installation is handled by the other operators/controllers.

/hold

akutz:
/hold cancel

@sidharthsurana Lets chat offline.

cmd/clusterctl/examples/vsphere/addons.yaml.template
```diff
@@ -6,4 +6,205 @@ metadata:
...
+          labels:
+            name: weave-net
+        spec:
+          containers:
+            - name: weave
+              command:
+                - /home/weave/launch.sh
+              env:
+                - name: IPALLOC_RANGE
+                  value: ${POD_CIDR}
```
andrewsykim: oops, missed this, will fix this soon

PR #297, opened 2019-06-07
Add addons.yaml into ci deployment

figo:
since we moved the CNI from cloud-init spec to addons

cc @andrewsykim @akutz

### DIFF
--- a/scripts/e2e/bootstrap_job/Makefile
+++ b/scripts/e2e/bootstrap_job/Makefile
@@ -7,8 +7,12 @@ REGISTRY ?=gcr.io/cnx-cluster-api/cluster-api-provider-vsphere-ci
 all: build push clean
 .PHONY : all
 
+.PHONY : copyspec
+copyspec:
+	cp ../../../cmd/clusterctl/examples/vsphere/addons.yaml.template ./spec/addons.template
+
 .PHONY : build
-build: 
+build: copyspec
 	docker build . --tag $(REGISTRY):$(VERSION)
         
 push: build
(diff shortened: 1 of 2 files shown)


akutz:
Hi @figo,

Is there anyway we can verify that this works? The VM we're using for e2e was deployed with Weave I imagine, so how are we validating that Calico is actually working?

figo:
> Hi @figo,
> 
> Is there anyway we can verify that this works? The VM we're using for e2e was deployed with Weave I imagine, so how are we validating that Calico is actually working?

VM in e2e is vanilla ubuntu, so do not have any cni pre-builtin

akutz:
> VM in e2e is vanilla ubuntu, so do not have any cni pre-builtin

Hi @figo,

Except we're not verifying anything about what's being stood up I thought? So how do we know CNI installed successfully?

figo:
> > VM in e2e is vanilla ubuntu, so do not have any cni pre-builtin
> 
> Hi @figo,
> 
> Except we're not verifying anything about what's being stood up I thought? So how do we know CNI installed successfully?

yes, we are not verifying it right now,  we have discussed expanding e2e tests in other threads, but cni  need be part of deployment since it is core

akutz:
So my question is how do we know the new CNI applied to the deployed cluster is working?

akutz:
Also, CNI is not core except for the initial target cluster to which the management components are pivoted.

figo:
@akutz  let me ask the question in this way, what do you suggest to change this PR?

akutz:
I guess nothing. I just hoped we could verify that the new addon with the new CNI was working

PR #299, opened 2019-06-09
Shell with GOPATH

akutz:
This patch updates the Makefile to use a custom shell based on `hack/shell-with-gopath.sh` to ensure that even when building outside the GOPATH, K8s tooling will still work.

This patch also:

* Updates any go:generate code to use scripts in `hack/` since go:generate ignores the current, working directory
* Removes the Makefile from `config/crds` and puts that logic into `hack/update-generated.sh` since recursive Make does not use the custom shell
* Remove the CRD validation added in 50dc7f0264761df87288f5aeea883c52401521f1. This will be replaced with a new Prow pre-submit test that will call `hack/verify-crds.sh`

**One Major Caveat**
This caveat was fixed in da23c9547b340206e5af8941261d28dfdeb9d370
~~If you routinely build CAPV in a container on you desktop with something like `docker run -it --rm -v "$(pwd)":/build -w /build golang make`, please note that the `.gopath` directory created on your host won't work once its parent directory is mapped into a container. Nor will the `.gopath` directory created from inside the container work on the host.~~

cc @andrewsykim


akutz:
/hold

This breaks when running `make vendor` because Go tooling cannot have symlinks anywhere in the `GOPATH`.

akutz:
/hold cancel

I fixed the issue with `make vendor` by allowing Make targets to `SHELL_WITH_GOPATH := 0` as a target-specific variable, which disables the symlink in the program `hack/shell-with-gopath.sh`.

akutz:
I opened https://github.com/kubernetes/test-infra/pull/12962 to replace the previous method of verifying committed CRDs. That PR is blocked by this PR.

PR #300, opened 2019-06-10
Use envsubst for clusterctl yaml generation

andrewsykim:
Use envsubst where possible instead of `sed` for yaml templating and generation. 

Also some clean up:
* removes all kubecon demo related code/yaml files
* adds out/ to .gitignore
* removed kustomize v2 checks <- unless there's a good reason to keep this


akutz:
Hi @andrewsykim,

Instead of removing the kustomize checks, please take a look at #283. It has the checks, but just ensures it's *at least* 2.0.

Makefile
```diff
@@ -152,15 +142,11 @@ prod-push:
...
-	@echo "updating kustomize image patch file for manager resource"
-	sed -i'' -e 's@image: .*@image: '"$(CI_IMG):$(VERSION)"'@' ./config/default/vsphere_manager_image_patch.yaml
-	cmd/clusterctl/examples/vsphere/generate-yaml.sh
+	VSPHERE_MANAGER_IMG=${CI_IMG}:${VERSION} cmd/clusterctl/examples/vsphere/generate-yaml.sh
 
 ci-image: generate fmt vet manifests
 	docker build . -t "$(CI_IMG):$(VERSION)"
 	docker build . -f cmd/clusterctl/Dockerfile -t "$(CLUSTERCTL_CI_IMG):$(VERSION)"
-	@echo "updating kustomize image patch file for manager resource"
-	sed -i.tmp -e 's@image: .*@image: '"$(CI_IMG):$(VERSION)"'@' ./config/default/vsphere_manager_image_patch.yaml
```
andrewsykim: Wasn't clear to me why we override the image after the image is built, is it to make `generate-yaml.sh` easier to run afterwards?

figo: you are right, this is redundant

akutz: ^ @figo?

figo:
@andrewsykim have you tested on a linux distribution with envsubst? just FYI to you, the ci is not running against the target: ci-yaml yet, whatever change on generate-yaml.sh will not automatically tested by ci yet.

andrewsykim:
@figo yup, tested on Linux
```
$ uname
Linux
$ make prod-yaml
VSPHERE_MANAGER_IMG=gcr.io/cnx-cluster-api/vsphere-cluster-api-provider:0.2.0 cmd/clusterctl/examples/vsphere/generate-yaml.sh
Done generating ./out/provider-components.yaml

*** Finished creating initial example yamls in ./out

    You need to update the machines.yaml and cluster.yaml files with information about your
    vSphere environment and information about the cluster you want to create.

Enjoy!
```

Makefile
```diff
@@ -105,15 +105,11 @@ vendor:
 # Create YAML file for deployment
 dev-yaml: | $(KUSTOMIZE)
 	@$(KUSTOMIZE) version 2>&1 | grep -q 'KustomizeVersion:v\{0,1\}[2-9][[:digit:]]\{0,\}' || { echo "kustomize v2+ required" 1>&2; exit 1; }
-	@echo "updating kustomize image patch file for manager resource"
-	sed -i'' -e 's@image: .*@image: '"${DEV_IMG}"'@' ./config/default/vsphere_manager_image_patch.yaml
-	cmd/clusterctl/examples/vsphere/generate-yaml.sh
+	VSPHERE_MANAGER_IMG=${DEV_IMG} cmd/clusterctl/examples/vsphere/generate-yaml.sh
```

If there's nothing left in the target but this line, then why don't you move the kustomize version check to `generate-yams.sh`?

Makefile
```diff
@@ -127,15 +123,11 @@ dev-push:
 # Create YAML file for deployment
 prod-yaml: | $(KUSTOMIZE)
 	@$(KUSTOMIZE) version 2>&1 | grep -q 'KustomizeVersion:v\{0,1\}[2-9][[:digit:]]\{0,\}' || { echo "kustomize v2+ required" 1>&2; exit 1; }
-	@echo "updating kustomize image patch file for manager resource"
-	sed -i'' -e 's@image: .*@image: '"${PRODUCTION_IMG}"'@' ./config/default/vsphere_manager_image_patch.yaml
-	cmd/clusterctl/examples/vsphere/generate-yaml.sh
+	VSPHERE_MANAGER_IMG=${PRODUCTION_IMG} cmd/clusterctl/examples/vsphere/generate-yaml.sh
```

If there's nothing left in the target but this line, then why don't you move the kustomize version check to `generate-yams.sh`?

Makefile
andrewsykim: Would you prefer `generate-yaml.sh` also use `hack/.bin/kustomize`?

akutz: I don't really care. At that point there's no cause to do so. However, if you notice at the top, we're exporting the PATH to prefer `hack/.bin`, so the script will use the specific version if called via the Makefile.

andrewsykim: Would prefer to get rid of the version check entirely then

andrewsykim: removed the version check

andrewsykim: removed the version check

akutz: That makes sense. Keep the order only dependency to enforce the download. Thanks.

andrewsykim:
Comments addressed, PTAL 

cc @sidharthsurana

hack/ensure-tools.sh
```diff
@@ -53,4 +53,14 @@ ensure_kustomize() {
     chmod 0755 "${_lcl_bin}"
 }
 
+ensure_envsubst() {
+  echo "ensure-tools: envsubst"
+
+  if [ ! envsubst --help 2>/dev/null ]; then
```
akutz: Hi @andrewsykim,

A few notes:

* The brackets aren't necessary. Please enable the [shellcheck](https://www.shellcheck.net) plug-in for you IDE.
* Please change to `if ! command -v envsubst >/dev/null 2>&1` to be consistent with the standard way to check the existence of commands.

figo:
another thing we have not discussed in this PR is: envsubst isn't widely available as sed in linux containers, for example, the Prow e2e container does not have envsubst by default.

but i like the current implementation of ensubst.

akutz:
> another thing we have not discussed in this PR is: envsubst isn't widely available as sed in linux containers, for example, the Prow e2e container does not have envsubst by default.
> 
> but i like the current implementation of ensubst.

Hi @figo,

True, but it's also easy enough to add. @andrewsykim is going to add it in a subsequent commit. It's available on a mac via `brew install gettext`, and on Linux via the same package. Super simple to get there, and CAPA uses it, so we know it's possible on Prow.

akutz:
Hi @figo,

However, I also wouldn't be opposed to using `sed` for this either. Lord knows I use it for a ton. It is easy enough to do. I'll leave it up to @andrewsykim to sort this all out though as he volunteered to work on this.

andrewsykim:
Addressed comments in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/302

PR #301, opened 2019-06-10
[WIP] Refactor CRUD operations

moshloop:
First off I am sorry for the size of this PR, I went down a rabbit hole, and this is what emerged at the end.

My goal with the refactoring (like most) was to make future changes quicker and safer. 

This refactoring includes:
* Breaking down the monolithic `create.go` in something a bit more manageable. 
* Reduce the function argument length's by introducing new `GovmomiMachine` and `GovmomiCluster` cluster objects that include references to CAPI, K8s and govc object that were being copied across function calls
* Reducing duplicated code into common methods (While I agree with the go idiom of A little copy and paste is better than a little dependency, I think within a package there should be limited duplication of code)
* Removing duplicated log and return errors (I might not be aware of the context or go idiom used here, but from my experience logging and then returning an error is an anti-pattern, You do 1 or the other, not both)
* Moving most utils methods into the `GovmomiMachine` or `GovmomiCluster` objects
* Decoupling of CAPI behavior vs general infrastructure behavior (i.e. when introducing new CAPI behavior you shouldn't need to go through working provisioning code that isn't being changed)

1 new feature: 
* Thin cloning for ESXi 

While this PR is not ready for merge, I wanted to get it out there to a) ensure it isn't duplicating any existing efforts b) Get early review, feedback, and criticism. 

TODO:
- Unit tests against vcsim
- Integration testing against ESXi and vCenter
- More inline comments and comments for exported methods

### DIFF
--- a/pkg/cloud/vsphere/clusteractuator.go
+++ b/pkg/cloud/vsphere/clusteractuator.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/klog"
 	vsphereconfigv1 "sigs.k8s.io/cluster-api-provider-vsphere/pkg/apis/vsphereproviderconfig/v1alpha1"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/constants"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/provisioner/govmomi"
 	vsphereutils "sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/utils"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	clusterv1alpha1 "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1"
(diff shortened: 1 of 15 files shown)


akutz:
Hi @moshloop,

I appreciate this, and I will definitely take a closer look soon. Just FYI, as I'm sure you've noticed, we're currently in the process of refactoring large swaths of CAPV to bring it into sync with CAPA. A PR this large simply won't survive the refactor, although there are bits and pieces we may be able to use. I suggest holding off on this for now for another two to three weeks until you see where things settle.

moshloop:
I think this PR will make syncing with CAPA much easier - This is essentially creating a Machine Scope (and I could rename it as such) that also encapsulates the ec2 services.

timothysc:
@moshloop could you break up into discrete smaller changes then?  I know it's a bit of extra work, but it's also generally best practice when engaging in upstream k/k as well.

akutz:
/hold 

Awaiting an opportunity to discuss this with @moshloop to sync up on the short-term direction of CAPV and the goal and plans to get it in sync with CAPA.

moshloop:
My thinking for breaking this down:

* Introduce MachineScope and refactor Exists
* Add helper methods to MachineScope and refactor Delete
* Add helper methods to MachineScope and refactor Update
* Introduce ClusterScope and refactor Create
* Add ESXi thin provision

PR #303, opened 2019-06-11
Fix base64 bug in hack/image-post-create-config.sh

codenrhoden:
This patch fixes an issue with `hack/image-post-create-config.sh`
where it failed on macOS due to it not supporting the `-w` flag for the
`base64` program.


codenrhoden:
/assign @akutz 

Big thanks to Andrew for showing me the easy way to enhance this!

PR #305, opened 2019-06-12
clusterctl: templatize more env vars for clusterctl yaml generation

andrewsykim:

Adding the following env vars for generating yaml files for the clusterctl bootstrap cluster:
```
VSPHERE_DATACENTER
VSPHERE_DATASTORE
VSPHERE_NETWORK
VSPHERE_RESOURCE_POOL
VSPHERE_FOLDER
VSPHERE_TEMPLATE
VSPHERE_DISK
```

### DIFF
--- a/cmd/clusterctl/examples/vsphere/generate-yaml.sh
+++ b/cmd/clusterctl/examples/vsphere/generate-yaml.sh
@@ -117,57 +117,57 @@ export VSPHERE_DISK_SIZE_GB=${VSPHERE_DISK_SIZE_GB:-20}
 
 # validate all required variables before generating any files
 
-if [ -z ${VSPHERE_USER} ]; then
+if [ -z "${VSPHERE_USER}" ]; then
   echo "env var VSPHERE_USER is required" 1>&2
   exit 1
 fi
 
-if [ -z ${VSPHERE_PASSWORD} ]; then
+if [ -z "${VSPHERE_PASSWORD}" ]; then
   echo "env var VSPHERE_PASSWORD is required" 1>&2
   exit 1
 fi
 
-if [ -z ${VSPHERE_SERVER} ]; then
+if [ -z "${VSPHERE_SERVER}" ]; then
   echo "env var VSPHERE_SERVER is required" 1>&2
   exit 1
 fi
 
-if [ -z $CAPV_MANAGER_IMAGE ]; then
+if [ -z "${CAPV_MANAGER_IMAGE}" ]; then
   echo "env var CAPV_MANAGER_IMAGE is required" 1>&2
   exit 1
 fi
 
-if [ -z $VSPHERE_DATACENTER ]; then
+if [ -z "${VSPHERE_DATACENTER}" ]; then
   echo "env var VSPHERE_DATACENTER is required" 1>&2
   exit 1
 fi
 
-if [ -z $VSPHERE_DATASTORE ]; then
+if [ -z "${VSPHERE_DATASTORE}" ]; then
   echo "env var VSPHERE_DATASTORE is required" 1>&2
   exit 1
 fi
 
-if [ -z $VSPHERE_NETWORK ]; then
+if [ -z "${VSPHERE_NETWORK}" ]; then
   echo "env var VSPHERE_NETWORK is required" 1>&2
   exit 1
 fi
 
-if [ -z $VSPHERE_RESOURCE_POOL ]; then
+if [ -z "${VSPHERE_RESOURCE_POOL}" ]; then
   echo "env var VSPHERE_RESOURCE_POOL is required" 1>&2
   exit 1
 fi
 
-if [ -z $VSPHERE_FOLDER ]; then
+if [ -z "${VSPHERE_FOLDER}" ]; then
   echo "env var VSPHERE_FOLDER is required" 1>&2
   exit 1
 fi
 
-if [ -z $VSPHERE_TEMPLATE ]; then
+if [ -z "${VSPHERE_TEMPLATE}" ]; then
   echo "env var VSPHERE_TEMPLATE is required" 1>&2
   exit 1
 fi
 
-if [ -z $VSPHERE_DISK ]; then
+if [ -z "${VSPHERE_DISK}" ]; then
   echo "env var VSPHERE_DISK is required" 1>&2
   exit 1
 fi
--- a/cmd/clusterctl/examples/vsphere/generate-yaml.sh
+++ b/cmd/clusterctl/examples/vsphere/generate-yaml.sh
@@ -98,14 +98,22 @@ mkdir -p ${OUTPUT_DIR}
 
 # all variables used for yaml generation
 
-CLUSTER_NAME=${CLUSTER_NAME:-vsphere-cluster}
-SERVICE_CIDR=${SERVICE_CIDR:-100.64.0.0/13}
-CLUSTER_CIDR=${CLUSTER_CIDR:-100.96.0.0/11}
+export CLUSTER_NAME=${CLUSTER_NAME:-vsphere-cluster}
+export SERVICE_CIDR=${SERVICE_CIDR:-100.64.0.0/13}
+export CLUSTER_CIDR=${CLUSTER_CIDR:-100.96.0.0/11}
 
-VSPHERE_USER=${VSPHERE_USER:-}
-VSPHERE_PASSWORD=${VSPHERE_PASSWORD:-}
-VSPHERE_SERVER=${VSPHERE_SERVER:-}
-VSPHERE_MANAGER_IMG=${VSPHERE_MANAGER_IMG:-}
+export VSPHERE_USER=${VSPHERE_USER:-}
+export VSPHERE_PASSWORD=${VSPHERE_PASSWORD:-}
+export VSPHERE_SERVER=${VSPHERE_SERVER:-}
+export VSPHERE_MANAGER_IMG=${VSPHERE_MANAGER_IMG:-}
+export VSPHERE_DATACENTER=${VSPHERE_DATACENTER:-}
+export VSPHERE_DATASTORE=${VSPHERE_DATASTORE:-}
+export VSPHERE_NETWORK=${VSPHERE_NETWORK:-}
+export VSPHERE_RESOURCE_POOL=${VSPHERE_RESOURCE_POOL:-}
+export VSPHERE_FOLDER=${VSPHERE_FOLDER:-}
+export VSPHERE_TEMPLATE=${VSPHERE_TEMPLATE:-}
+export VSPHERE_DISK=${VSPHERE_DISK:-}
+export VSPHERE_DISK_SIZE_GB=${VSPHERE_DISK_SIZE_GB:-20}
 
 # validate all required variables before generating any files
 
@@ -126,6 +134,46 @@ fi
 
 if [ -z $VSPHERE_MANAGER_IMG ]; then
   echo "env var VSPHERE_MANAGE_IMG is required"
+  exit 1
+fi
+
+if [ -z $VSPHERE_DATACENTER ]; then
+  echo "env var VSPHERE_DATACENTER is required"
+  exit 1
+fi
+
+if [ -z $VSPHERE_DATASTORE ]; then
+  echo "env var VSPHERE_DATASTORE is required"
+  exit 1
+fi
+
+if [ -z $VSPHERE_NETWORK ]; then
+  echo "env var VSPHERE_NETWORK is required"
+  exit 1
+fi
+
+if [ -z $VSPHERE_RESOURCE_POOL ]; then
+  echo "env var VSPHERE_RESOURCE_POOL is required"
+  exit 1
+fi
+
+if [ -z $VSPHERE_FOLDER ]; then
+  echo "env var VSPHERE_FOLDER is required"
+  exit 1
+fi
+
+if [ -z $VSPHERE_TEMPLATE ]; then
+  echo "env var VSPHERE_TEMPLATE is required"
+  exit 1
+fi
+
+if [ -z $VSPHERE_DISK ]; then
+  echo "env var VSPHERE_DISK is required"
+  exit 1
+fi
+
+if [ ${VSPHERE_DISK_SIZE_GB} -lt 20 ]; then
+  echo "env var VSPHERE_DISK_SIZE_GB must be >= 20" 1>&2
   exit 1
 fi
 
--- a/cmd/clusterctl/examples/vsphere/generate-yaml.sh
+++ b/cmd/clusterctl/examples/vsphere/generate-yaml.sh
@@ -98,7 +98,7 @@ mkdir -p ${OUTPUT_DIR}
 
 # all variables used for yaml generation
 
-export CLUSTER_NAME=${CLUSTER_NAME:-vsphere-cluster}
+export CLUSTER_NAME=${CLUSTER_NAME:-capv-mgmt-example}
 export SERVICE_CIDR=${SERVICE_CIDR:-100.64.0.0/13}
 export CLUSTER_CIDR=${CLUSTER_CIDR:-100.96.0.0/11}
 
--- a/cmd/clusterctl/examples/vsphere/generate-yaml.sh
+++ b/cmd/clusterctl/examples/vsphere/generate-yaml.sh
@@ -36,8 +36,8 @@ CLUSTER_API_CRD_PATH=./vendor/sigs.k8s.io/cluster-api/config
 VSPHERE_CLUSTER_API_CRD_PATH=./config
 
 PROVIDERCOMPONENT_GENERATED_FILE=${OUTPUT_DIR}/provider-components.yaml
-VSPHERE_MANAGER_TEMPLATE_FILE=${TEMPLATE_DIR}/vsphere_manager_image_patch.yaml.template
-VSPHERE_MANAGER_GENERATED_FILE=$VSPHERE_CLUSTER_API_CRD_PATH/default/vsphere_manager_image_patch.yaml
+CAPV_MANAGER_TEMPLATE_FILE=${TEMPLATE_DIR}/capv_manager_image_patch.yaml.template
+CAPV_MANAGER_GENERATED_FILE=$VSPHERE_CLUSTER_API_CRD_PATH/default/capv_manager_image_patch.yaml
 
 MACHINE_CONTROLLER_SSH_PUBLIC_FILE=vsphere_tmp.pub
 MACHINE_CONTROLLER_SSH_PUBLIC=
@@ -105,7 +105,7 @@ export CLUSTER_CIDR=${CLUSTER_CIDR:-100.96.0.0/11}
 export VSPHERE_USER=${VSPHERE_USER:-}
 export VSPHERE_PASSWORD=${VSPHERE_PASSWORD:-}
 export VSPHERE_SERVER=${VSPHERE_SERVER:-}
-export VSPHERE_MANAGER_IMG=${VSPHERE_MANAGER_IMG:-}
+export CAPV_MANAGER_IMAGE=${CAPV_MANAGER_IMAGE:-}
 export VSPHERE_DATACENTER=${VSPHERE_DATACENTER:-}
 export VSPHERE_DATASTORE=${VSPHERE_DATASTORE:-}
 export VSPHERE_NETWORK=${VSPHERE_NETWORK:-}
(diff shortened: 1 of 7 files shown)


cmd/clusterctl/examples/vsphere/generate-yaml.sh
```diff
@@ -98,14 +98,21 @@ mkdir -p ${OUTPUT_DIR}
...
 
-CLUSTER_NAME=${CLUSTER_NAME:-vsphere-cluster}
-SERVICE_CIDR=${SERVICE_CIDR:-100.64.0.0/13}
-CLUSTER_CIDR=${CLUSTER_CIDR:-100.96.0.0/11}
-
-VSPHERE_USER=${VSPHERE_USER:-}
-VSPHERE_PASSWORD=${VSPHERE_PASSWORD:-}
-VSPHERE_SERVER=${VSPHERE_SERVER:-}
-VSPHERE_MANAGER_IMG=${VSPHERE_MANAGER_IMG:-}
+export CLUSTER_NAME=${CLUSTER_NAME:-vsphere-cluster}
```
akutz: Hi @andrewsykim,

Maybe CAPA did this, and it looks like we used to, but if we default to a cluster name, will this cause problems?

cmd/clusterctl/examples/vsphere/generate-yaml.sh
```diff
@@ -98,14 +98,21 @@ mkdir -p ${OUTPUT_DIR}
...
+export SERVICE_CIDR=${SERVICE_CIDR:-100.64.0.0/13}
```
akutz: Hi @andrewsykim,

Perhaps make a note that these CIDRs must match those used by whatever CNI add-on is installed?

cmd/clusterctl/examples/vsphere/generate-yaml.sh
```diff
@@ -129,6 +136,41 @@ if [ -z $VSPHERE_MANAGER_IMG ]; then
   exit 1
 fi
 
+if [ -z $VSPHERE_DATACENTER ]; then
+  echo "env var VSPHERE_DATACENTER is required"
```
akutz: Hi @andrewsykim,

Anywhere `echo` is used to indicate an error, please use `echo "your message" 1>&2` to ensure the message is printed to `stderr`. 

Please note this applies to several other lines, but I won't repeat this comment there.

cmd/clusterctl/examples/vsphere/machines.yaml.template
```diff
@@ -4,7 +4,7 @@ items:
   - apiVersion: "cluster.k8s.io/v1alpha1"
     kind: Machine
     metadata:
-      generateName: vs-master-
+      name: ${CLUSTER_NAME}-mgmt
```
akutz: Hi @andrewsykim,

Is `generateName` not something CAPV uses? Or is it some old CAPI label no longer used?

cmd/clusterctl/examples/vsphere/machines.yaml.template
```diff
@@ -14,29 +14,21 @@ items:
...
-                - xxxx
-                - yyyy
             numCPUs: 2
             memoryMB: 2048
-            template: ""
+            template: "${VSPHERE_TEMPLATE}"
             disks:
-            - diskLabel: ""
+            - diskLabel: "${VSPHERE_DISK}"
               diskSizeGB: 20
```
akutz: Hi @andrewsykim,

Please templatize the disk size and do a quick check on the value. It should never be *smaller* than 20GiB as that is the file system size used when creating the images. Cloud-init will automatically *grow* the file system if the disk is larger than 20 GiB, but I'm not sure what happens if the disk is smaller than 20 GiB. Probably nothing good.

cmd/clusterctl/examples/vsphere/machineset.yaml.template
```diff
@@ -15,30 +15,25 @@ spec:
...
+	  kind: "VsphereMachineProviderConfig"
+	  machineSpec:
+	    datacenter: "${VSPHERE_DATACENTER}"
+	    datastore: "${VSPHERE_DATASTORE}"
+	    resourcePool: "${VSPHERE_RESOURCE_POOL}"
+	    vmFolder: "${VSPHERE_FOLDER}"
+	    networks:
+	    - networkName: "${VSPHERE_NETWORK}"
+	      ipConfig:
+		networkType: static
```
akutz: Hi @andrewsykim,

Please see my above comment on the `networkType`.

cmd/clusterctl/examples/vsphere/machineset.yaml.template
```diff
@@ -15,30 +15,25 @@ spec:
...
+	    numCPUs: 2
+	    memoryMB: 2048
+	    template: "${VSPHERE_TEMPLATE}"
+	    disks:
+	    - diskLabel: "${VSPHERE_DISK}"
+	      diskSizeGB: 20
```
akutz: Hi @andrewsykim,

Please see my above comment on disk size.

cmd/clusterctl/examples/vsphere/machines.yaml.template
```diff
@@ -14,29 +14,21 @@ items:
...
             preloaded: false
-            ntpServers:
-            - 0.pool.ntp.org
-            - 1.pool.ntp.org
       versions:
         kubelet: 1.11.3
```
akutz: Hi @andrewsykim,

Note to selves -- we need to update the default kubelet and controlPlane versions to 1.13.6.

cmd/clusterctl/examples/vsphere/machines.yaml.template
```diff
@@ -14,29 +14,21 @@ items:
...
-            resourcePool: ""
+            datacenter: "${VSPHERE_DATACENTER}"
+            datastore: "${VSPHERE_DATASTORE}"
+            resourcePool: "${VSPHERE_RESOURCE_POOL}"
+            vmFolder: "${VSPHERE_FOLDER}"
             networks:
-            - networkName: ""
+            - networkName: "${VSPHERE_NETWORK}"
               ipConfig:
                 networkType: static
```
akutz: Hi @andrewsykim,

I think the networkType should default to dhcp, right? Otherwise we should force the selection of IP address, subnet, gateway information, and possibly DNS servers.

cmd/clusterctl/examples/vsphere/generate-yaml.sh
```diff
@@ -98,14 +98,21 @@ mkdir -p ${OUTPUT_DIR}
...
-VSPHERE_SERVER=${VSPHERE_SERVER:-}
-VSPHERE_MANAGER_IMG=${VSPHERE_MANAGER_IMG:-}
+export CLUSTER_NAME=${CLUSTER_NAME:-vsphere-cluster}
+export SERVICE_CIDR=${SERVICE_CIDR:-100.64.0.0/13}
+export CLUSTER_CIDR=${CLUSTER_CIDR:-100.96.0.0/11}
+
+export VSPHERE_USER=${VSPHERE_USER:-}
+export VSPHERE_PASSWORD=${VSPHERE_PASSWORD:-}
+export VSPHERE_SERVER=${VSPHERE_SERVER:-}
+export VSPHERE_MANAGER_IMG=${VSPHERE_MANAGER_IMG:-}
```
akutz: Hi @andrewsykim,

Another quick question -- what's the manager image? Is this the manager OVA we are deprecating?

cmd/clusterctl/examples/vsphere/generate-yaml.sh
andrewsykim: This is only the bootstrap cluster and we automatically install calico there with the same CIDRs injected

cmd/clusterctl/examples/vsphere/machines.yaml.template
andrewsykim: The `generateName` field is confusing IMO because it only works from the context of `clusterctl`. If you try to apply a machine resource with `generateName` it just errors on validation

cmd/clusterctl/examples/vsphere/machines.yaml.template
andrewsykim: template the k8s version too?

cmd/clusterctl/examples/vsphere/generate-yaml.sh
andrewsykim: No this is the CAPV controller, it's called a "manager" in the statefulset spec

cmd/clusterctl/examples/vsphere/machines.yaml.template
akutz: That makes sense. With the default being whatever version we support in the images.

cmd/clusterctl/examples/vsphere/machines.yaml.template
akutz: Huh. That *is* odd. Do we expect the templates to be used for anything *but* `clusterctl` though?

cmd/clusterctl/examples/vsphere/generate-yaml.sh
akutz: Hi @andrewsykim,

Only if some other CNI plug-in were to come along that is better than Calico in every way that we would want to use. Do you feel there's a chance that could happen?

cmd/clusterctl/examples/vsphere/generate-yaml.sh
akutz: Got it. Thank you.

cmd/clusterctl/examples/vsphere/generate-yaml.sh
andrewsykim: I'm assuming this is OK because it's an "example" run through but I have no problems changing this

cmd/clusterctl/examples/vsphere/machines.yaml.template
andrewsykim: We don't, BUT I imagine the generated machine objects will be referenced for future machines created against the management cluster

cmd/clusterctl/examples/vsphere/generate-yaml.sh
akutz: Stupid question -- this isn't the same template used to create the management cluster? I thought these templates were used for both?

cmd/clusterctl/examples/vsphere/machines.yaml.template
akutz: Hi @andrewsykim,

Maybe it's the time of day, but I've read this:

> I imagine the generated machine objects will be referenced for future machines created against the management cluster

six times, and I still don't understand what you mean. :)

cmd/clusterctl/examples/vsphere/machines.yaml.template
andrewsykim: Should we default the value to 20 though?

akutz: I'm just saying it shouldn't be allowed to be less than 20. So:

```shell
DISK_SIZE_GIB="${DISK_SIZE_GIB:-20}"
[[ ${DISK_SIZE_GIB} =~ ^[0-9]+$ ]] || DISK_SIZE_GIB=20
[ "${DISK_SIZE_GIB}" -ge "20" ] || DISK_SIZE_GIB=20
```

cmd/clusterctl/examples/vsphere/generate-yaml.sh
andrewsykim: Yup, but I'm assuming someone that wants to stand up a production management cluster would set `CLUSTER_NAME` to something that is more reasonable. `vsphere-cluster` as a default for `cmd/clusterctl/examples` seems okay to me

andrewsykim: fwiw https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cmd/clusterctl/examples/aws/generate-yaml.sh#L27

andrewsykim: Reverting to something like `vsphere-mgmt-test1` might be better so users know to not use the example cluster for production

akutz: All fair points. I guess I just look at this with fresh eyes instead of improving what we have. Looks good, thanks.

cmd/clusterctl/examples/vsphere/machines.yaml.template
andrewsykim: So `generate-yaml.sh` generates a cluster, machine and machinset resource. This is used by clusterctl for the bootstrap cluster that is eventually pivoted into the mgmt cluster. I imagine that the generated YAML will be the first _working_ cluster, machine and machinset spec that most users of CAPV will have and will likely use it (as a reference) for target clusters created against the mgmt cluster created by clusterctl. 

If you try to apply a machine resource with the `generateName` field directly against the mgmt cluster (i.e. not with clusterctl), it will not work s

cmd/clusterctl/examples/vsphere/machines.yaml.template
andrewsykim: okay fixed this

andrewsykim:
The commits are split into meaningful parts (like the CAPV_MANAGER_IMAGE rename) and the stderr redirect changes so prefer not to squash them. Don't feel strongly about it and can squash them if needed

cmd/clusterctl/examples/vsphere/machines.yaml.template
figo: the name `-mgmt` is not clear to me, how about `-controlplane` or `-master`?

akutz: Not `-master` as that convention is one we actively want to avoid.

figo: agree, we should use `-controlplane`

andrewsykim: From my perspective, appending `-controlplane` to the management cluster nodes is redundant because I don't think we would advise adding non control plane nodes there.

figo: could you clarify what do you mean regards management cluster? i could be out of date in term of naming, there are `bootstrap cluster` and `target cluster`, this `machine.yaml` is for controlplace node of the `target cluster`, whether it been used as management cluster node or not all falls under user's preferences, but many of previous discussions are actually around bootstrap cluster functioning as management cluster.

akutz: Hi @figo,

We're not longer supporting a non-pivot deployment strategy, so the management cluster is the initial target cluster deployed by the bootstrap cluster, to which management resources are pivoted.

figo: is there doc or issue about the latest design? i definitely need to understand more about rationale behind it,  @sidharthsurana

andrewsykim: Remember this is also just an example walk through. The example generated yaml DOES use bootstrap, if you want your own management cluster you decide the name of the node anyways

akutz: > Remember this is also just an example walk through. The example generated yaml DOES use bootstrap, if you want your own management cluster you decide the name of the node anyways

I agree this is confusing though. As I've mentioned elsewhere, it's super unclear if the purpose of the templates is for the initial target cluster only, or to be used to create additional clusters later.

andrewsykim: I will try to make this clear in the docs coming up, naming the node to something like `vpshere-mgmt-test1` like I mentioned before might make this super clear. Thoughts?

akutz: I think a basic PlantUML diagram we can put on our README with names like "bootstrap cluster", "initial target/management cluster", "additional cluster n-n-1" will be useful.

andrewsykim: Updated the default cluster name to `capv-mgmt-example` and the machine to `capv-mgmt-example-controlplane`

cmd/clusterctl/examples/vsphere/generate-yaml.sh
```diff
@@ -98,34 +98,88 @@ mkdir -p ${OUTPUT_DIR}
...
+export CAPV_MANAGER_IMAGE=${CAPV_MANAGER_IMAGE:-}
+export VSPHERE_DATACENTER=${VSPHERE_DATACENTER:-}
+export VSPHERE_DATASTORE=${VSPHERE_DATASTORE:-}
+export VSPHERE_NETWORK=${VSPHERE_NETWORK:-}
+export VSPHERE_RESOURCE_POOL=${VSPHERE_RESOURCE_POOL:-}
+export VSPHERE_FOLDER=${VSPHERE_FOLDER:-}
+export VSPHERE_TEMPLATE=${VSPHERE_TEMPLATE:-}
+export VSPHERE_DISK=${VSPHERE_DISK:-}
+export VSPHERE_DISK_SIZE_GB=${VSPHERE_DISK_SIZE_GB:-20}
+export VSPHERE_NETWORK_TYPE=${VSPHERE_NETWORK_TYPE:-dhcp}
```
sidharthsurana: Can this not be a var driven from user input? since it they choose staticIP then the rendered yaml is not sufficient?

cmd/clusterctl/examples/vsphere/machineset.yaml.template
```diff
@@ -15,30 +15,25 @@ spec:
...
+	    numCPUs: 2
+	    memoryMB: 2048
+	    template: "${VSPHERE_TEMPLATE}"
+	    disks:
+	    - diskLabel: "${VSPHERE_DISK}"
+	      diskSizeGB: ${VSPHERE_DISK_SIZE_GB}
+	    preloaded: false
       versions:
         kubelet: 1.11.3
       roles:
```
sidharthsurana: Can we remove the rolese section. It is not used is just a leftover from older design.
```
roles:
.....
```

andrewsykim: That seems like a significant change because some of the CAPV logic still depends on the roles being there (at least last time I checked). Either way sounds like something I can address in a follow-up PR :)

cmd/clusterctl/examples/vsphere/generate-yaml.sh
andrewsykim: I might be missing something, I've been able to create VMs with `static` network type even without having the IP in the network spec. Is there something in CAPV allocating that automatically anyways?

akutz: Hi @andrewsykim,

The image they're using will default to a fallback network if the provided config is incomplete or invalid. But that doesn't mean that we want the YAML to be/look invalid.

Unless we pivot on the network type and then change the YAML based on `static` or `dhcp`, I recommend we stick with just DHCP and let people add the rest if they want. Although doing an `if/else` shouldn't be too hard either.

The issue, then, is that for static IPs you'd need to have indexed env vars to use with each machine in the spec.

cmd/clusterctl/examples/vsphere/machineset.yaml.template
akutz: Hi @sidharthsurana,

I personally would like more information on the RBAC anyway as CAPA seems to generate their RBAC manifests through the controller gen tool, and it looks like we intended to do that at one point but didn't get that far.

sidharthsurana: This has nothing to do with RBAC. This is a just yet another API field in the spec, that IIRC pre-dates late 2018
I am surprised CAPA uses that since the method in CAPI that checks whether a machine is a control plane machine does depend on this

sidharthsurana: > That seems like a significant change because some of the CAPV logic still depends on the roles being there (at least last time I checked). Either way sounds like something I can address in a follow-up PR :)
@andrewsykim
I don't recall any CAPV logic depending on that. Can you point me to one? maybe I am mistaken.

cmd/clusterctl/examples/vsphere/generate-yaml.sh
andrewsykim: Okay so hard code on dhcp and document how to configure with static?

cmd/clusterctl/examples/vsphere/machineset.yaml.template
andrewsykim: Ah, I was mistakening this for versions :). Either way, don't want to mix up those changes in this PR, will update in follow- up

PR #307, opened 2019-06-13
Support for kubeadm bootstrapping & using new imgs

akutz:
This patch adds support for using kubeadm for bootstrapping onto the new CAPV images. This is to align CAPV's features with those in CAPA. It is completely untested and this PR is only so people can start providing feedback.

cc @sidharthsurana @figo @frapposelli

I would also appreciate some CAPA folks like @detiber, @ncdc, and @vincepri giving this a looksee as many of the changes are based on CAPA code. Thank you!


akutz:
HI @andrewsykim,

Please note that this PR requires the `set` label on machines to indicate whether a machine is a member of the `controlplane` or a `node`. This is similar to how CAPA works.

Currently the `set` value is `master` in the example manifests, but we should update it to `controlplane`.

detiber:
lgtm at a quick glance

akutz:
> HI @andrewsykim,
> 
> Please note that this PR requires the `set` label on machines to indicate whether a machine is a member of the `controlplane` or a `node`. This is similar to how CAPA works.
> 
> Currently the `set` value is `master` in the example manifests, but we should update it to `controlplane`.

I've rebased this PR so that the old values of `set` and `node-type` are both supported as well via a utils call for getting a machine's role.

figo:
could not find change related to cloud-init template: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/pkg/cloud/vsphere/provisioner/common/vc/scripts.go#L65

is it still pulling image from upstream?

akutz:
> could not find change related to cloud-init template: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/pkg/cloud/vsphere/provisioner/common/vc/scripts.go#L65
> 
> is it still pulling image from upstream?

Hi @figo,

I need to just remove those packages. We're no longer using anything related to the old cloud-init user data method. It's all based on the CAPA cloud-init user data now.

pkg/cloud/vsphere/machineactuator.go
```diff
@@ -54,22 +64,116 @@ func NewGovmomiMachineActuator(m manager.Manager, clusterV1alpha1 clusterv1alpha
...
-		if err != nil {
-			klog.Error(err)
-			return err
+func (vc *VsphereClient) Create(
+	ctx context.Context,
+	cluster *clusterv1.Cluster,
+	machine *clusterv1.Machine) error {
+
+	if vc.provisioner == nil {
+		return errors.New("No provisioner available")
```
andrewsykim: nit: error should start with lower case

pkg/cloud/vsphere/machineactuator.go
```diff
@@ -54,22 +64,116 @@ func NewGovmomiMachineActuator(m manager.Manager, clusterV1alpha1 clusterv1alpha
...
+	if cluster == nil {
+		return errors.Errorf(
+			"missing cluster for machine %s/%s",
+			machine.Namespace, machine.Name)
+	}
+
+	machineRole := vsphereutils.GetMachineRole(machine)
+	if machineRole == "" {
+		return errors.Errorf(
+			"Unable to get machine role while creating machine "+
```
andrewsykim: nit: lowercase error

pkg/cloud/vsphere/machineactuator.go
```diff
@@ -54,22 +64,116 @@ func NewGovmomiMachineActuator(m manager.Manager, clusterV1alpha1 clusterv1alpha
...
+			"unable to get control plane machines while creating machine "+
+				"%s=%s %s=%s %s=%s %s=%s",
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Namespace,
+			"machine-namespace", machine.Namespace,
+			"machine-name", machine.Name)
+	}
+
+	// Init the control plane by creating this machine.
+	if machineRole == "controlplane" && len(controlPlaneMachines) <= 1 {
```
andrewsykim: Should `controlplane` be a const?

andrewsykim: Also `len(controlPlaneMachines) == 0`?

pkg/cloud/vsphere/machineactuator.go
```diff
@@ -54,22 +64,116 @@ func NewGovmomiMachineActuator(m manager.Manager, clusterV1alpha1 clusterv1alpha
...
+		if err != nil {
+			klog.V(2).Info(errors.Wrap(err, msg))
+			return errors.Cause(err)
+		}
+		klog.V(2).Info(msg)
+		return &capierr.RequeueAfterError{RequeueAfter: time.Minute * 1}
+	}
+
+	// Get a Kubernetes client for the cluster.
+	kubeClient, _ := vsphereutils.GetKubeClientForCluster(cluster, vc.lister)
```
andrewsykim: why ignore err?

pkg/cloud/vsphere/machineactuator.go
```diff
@@ -54,22 +64,116 @@ func NewGovmomiMachineActuator(m manager.Manager, clusterV1alpha1 clusterv1alpha
...
+	online, _, err := vsphereutils.GetControlPlaneStatus(cluster, vc.lister)
+	if !online {
+		msg := fmt.Sprintf("unable to join machine to control plane until it is online "+
+			"%s=%s %s=%s %s=%s %s=%s",
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Namespace,
+			"machine-namespace", machine.Namespace,
+			"machine-name", machine.Name)
+		if err != nil {
+			klog.V(2).Info(errors.Wrap(err, msg))
```
andrewsykim: klog.Error?

pkg/cloud/vsphere/machineactuator.go
```diff
@@ -54,22 +64,116 @@ func NewGovmomiMachineActuator(m manager.Manager, clusterV1alpha1 clusterv1alpha
...
+			return errors.Cause(err)
+		}
+		klog.V(2).Info(msg)
+		return &capierr.RequeueAfterError{RequeueAfter: time.Minute * 1}
```
andrewsykim: Put the default RequeueAfter value in a const?

pkg/cloud/vsphere/machineactuator.go
```diff
@@ -95,3 +199,19 @@ func (vc *VsphereClient) Exists(ctx context.Context, cluster *clusterv1.Cluster,
 
 	return false, fmt.Errorf("No provisioner available")
 }
+
+// GetControlPlaneMachines retrieves all control plane nodes from a MachineList
+func GetControlPlaneMachines(machineList *clusterv1.MachineList) []*clusterv1.Machine {
```
andrewsykim: Might be missing something, but don't see this called anywhere?

pkg/cloud/vsphere/machineactuator.go
```diff
@@ -95,3 +199,19 @@ func (vc *VsphereClient) Exists(ctx context.Context, cluster *clusterv1.Cluster,
...
+	for _, m := range machineList.Items {
+		if m.Spec.Versions.ControlPlane != "" {
+			cpm = append(cpm, m.DeepCopy())
+		}
+	}
+	return cpm
+}
+
+// defining equality as name and namespace are equivalent and not checking any other fields.
+func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
```
andrewsykim: same with this one, where is this called?

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -4,49 +4,115 @@ import (
...
+				"%s=%s %s=%s %s=%s %s=%s",
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Namespace,
+			"machine-namespace", machine.Namespace,
+			"machine-name", machine.Name)
+	}
+
+	klog.V(4).Infof("Creating machine with GoVmomi %s=%s %s=%s %s=%s %s=%s %s=%s",
+		"cluster-namespace", cluster.Namespace,
+		"cluster-name", cluster.Namespace,
```
andrewsykim: cluster.Name

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -4,49 +4,115 @@ import (
...
 	}
 
-	klog.V(4).Infof("govmomi.Actuator.Create %s", machine.Name)
+	machineRole := vsphereutils.GetMachineRole(machine)
+	if machineRole == "" {
+		return errors.Errorf(
+			"Unable to get machine role while creating machine with GoVmomi "+
+				"%s=%s %s=%s %s=%s %s=%s",
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Namespace,
```
andrewsykim: cluster.Name

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -4,49 +4,115 @@ import (
...
+		"machine-namespace", machine.Namespace,
+		"machine-name", machine.Name,
+		"machine-role", machineRole)
+
 	s, err := pv.sessionFromProviderConfig(cluster, machine)
 	if err != nil {
-		return err
+		return errors.Wrapf(
+			err,
+			"Unable to get session while creating machine with GoVmomi "+
```
andrewsykim: nit: start with lowercase

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -4,49 +4,115 @@ import (
...
+				"%s=%s %s=%s %s=%s %s=%s",
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Namespace,
```
andrewsykim: cluster.Name

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -4,49 +4,115 @@ import (
...
 	if err != nil {
-		return err
-	}
-	klog.V(4).Infof("Using session %v", usersession)
+		return errors.Wrapf(
+			err,
+			"Unable to get user session while creating machine with GoVmomi "+
+				"%s=%s %s=%s %s=%s %s=%s",
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Namespace,
```
andrewsykim: cluster.Name

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -4,49 +4,115 @@ import (
...
-	usersession, err := s.session.SessionManager.UserSession(createctx)
+
+	userSession, err := s.session.SessionManager.UserSession(createctx)
 	if err != nil {
-		return err
-	}
-	klog.V(4).Infof("Using session %v", usersession)
+		return errors.Wrapf(
+			err,
+			"Unable to get user session while creating machine with GoVmomi "+
```
andrewsykim: nit: start witih lower

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -4,49 +4,115 @@ import (
...
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Namespace,
+			"machine-namespace", machine.Namespace,
+			"machine-name", machine.Name)
+	}
+
+	klog.V(4).Infof("Got user session while creating machine with GoVmomi "+
+		"%s=%s %s=%s %s=%s %s=%s %s=%s %s=%v",
+		"cluster-namespace", cluster.Namespace,
+		"cluster-name", cluster.Namespace,
```
andrewsykim: cluster.Name

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -4,49 +4,115 @@ import (
...
+		"machine-namespace", machine.Namespace,
+		"machine-name", machine.Name,
+		"machine-role", machineRole,
+		"user-session", userSession)
```
andrewsykim: Haven't looked at the entire userSession struct, but is it safe to log this? Also how big is this struct? Might make logs unreadable if you dump the entire struct like this.

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -56,24 +122,245 @@ func (pv *Provisioner) Create(ctx context.Context, cluster *clusterv1.Cluster, m
 	// as this Machine. If found, that VM is the right match for this machine
 	vmRef, err := pv.findVMByInstanceUUID(ctx, s, machine)
 	if err != nil {
-		return err
+		return errors.Wrapf(
+			err,
+			"Unable to get find VM by instance UUID while creating machine with GoVmomi "+
+				"%s=%s %s=%s %s=%s %s=%s",
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Namespace,
```
andrewsykim: cluster.Name - there's a few other places won't comment on all of them

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -56,24 +122,245 @@ func (pv *Provisioner) Create(ctx context.Context, cluster *clusterv1.Cluster, m
...
+	}
+
+	caCertHash, err := certificates.GenerateCertificateHash(clusterConfig.CAKeyPair.Cert)
+	if err != nil {
+		return err
+	}
+
+	var controlPlaneEndpoint string
+	if bootstrapToken != "" {
+		controlPlaneEndpoint, _ = vsphereutils.GetControlPlaneEndpoint(cluster, nil)
```
andrewsykim: why ignore error?

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -56,24 +122,245 @@ func (pv *Provisioner) Create(ctx context.Context, cluster *clusterv1.Cluster, m
...
+			JoinConfiguration: joinConfigurationYAML,
+		})
+		if err != nil {
+			return nil
+		}
+
+		userDataYAML = userData
+
+	default:
+		return errors.Errorf("Unknown node role %q", machineRole)
```
andrewsykim: nit: `unknown`

pkg/cloud/vsphere/provisioner/govmomi/create.go
```diff
@@ -322,58 +603,12 @@ func (pv *Provisioner) cloneVirtualMachineOnVCenter(s *SessionContext, cluster *
 	}
 	spec.PowerOn = true
 
-	if machineConfig.MachineSpec.VsphereCloudInit {
```
andrewsykim: Remove this field from machine spec?

pkg/cloud/vsphere/services/certificates/certificates.go
```diff
@@ -0,0 +1,343 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
andrewsykim: 2019

pkg/cloud/vsphere/services/userdata/controlplane.go
```diff
@@ -0,0 +1,303 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
andrewsykim: 2019

pkg/cloud/vsphere/services/userdata/controlplane_test.go
```diff
@@ -0,0 +1,51 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
andrewsykim: 2019

pkg/cloud/vsphere/services/userdata/node.go
```diff
@@ -0,0 +1,52 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
andrewsykim: 2019

pkg/cloud/vsphere/services/userdata/userdata.go
```diff
@@ -0,0 +1,92 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
andrewsykim: 2019

pkg/cloud/vsphere/utils/utils.go
```diff
@@ -1,31 +1,304 @@
...
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/services/certificates"
 )
 
+const (
+	defaultBindPort = 6443
+)
+
+// GetMachineRole returns either "controlplane", "node", or an empty
+// string if the role could be ascertained.
+func GetMachineRole(machine *clusterv1.Machine) string {
```
andrewsykim: Possible to unit test this?

pkg/cloud/vsphere/utils/utils.go
```diff
@@ -1,31 +1,304 @@
...
+	}
+	if controlPlaneEndpoint := clusterProviderConfig.ClusterConfiguration.ControlPlaneEndpoint; controlPlaneEndpoint != "" {
+		klog.V(2).Infof("got control plane endpoint from cluster config %s=%s %s=%s %s=%s",
+			"control-plane-endpoint", controlPlaneEndpoint,
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Name)
+		return controlPlaneEndpoint, nil
+	}
+
+	if client == nil {
```
andrewsykim: nit: move this to the top of the function?

pkg/cloud/vsphere/utils/utils.go
```diff
@@ -1,31 +1,304 @@
...
+		}
+	}
+	if bindPort == 0 {
+		bindPort = defaultBindPort
+	}
+	return bindPort
+}
+
+// GetKubeConfig returns the kubeconfig for the given cluster.
+func GetKubeConfig(
```
andrewsykim: nice 👍

pkg/cloud/vsphere/utils/utils.go
```diff
@@ -1,31 +1,304 @@
...
+
+	if err := getControlPlaneStatus(cluster, client); err != nil {
+		return false, "", errors.Wrapf(
+			&capierr.RequeueAfterError{RequeueAfter: constants.RequeueAfterSeconds},
+			"unable to get control plane status %s=%s %s=%s: %v",
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Name,
+			err)
+	}
+	controlPlaneEndpoint, _ := GetControlPlaneEndpoint(cluster, client)
```
andrewsykim: why ignore error?

pkg/cloud/vsphere/clusteractuator.go
```diff
@@ -213,47 +210,76 @@ func (ca *ClusterActuator) provisionLoadBalancer(cluster *clusterv1.Cluster) err
...
-				ca.eventRecorder.Eventf(ncluster, corev1.EventTypeWarning, "Failed Update", "Error in updating API Endpoint: %s", err)
-				klog.Infof("Error in updating the status: %s", err)
-				return err
-			}
-			ca.eventRecorder.Eventf(ncluster, corev1.EventTypeNormal, "Updated", "Updated API Endpoint to %v", ip)
-		}
+		return err
 	}
+
+	host, szPort, err := net.SplitHostPort(controlPlaneEndpoint)
```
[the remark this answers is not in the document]
vincepri: It might be something related to the value being a string

pkg/cloud/vsphere/utils/utils.go
akutz: The first three functions don't require `client` and can return the result without it. That's why it's at this location. Here and beyond `client` is required, but not before. And there are callers that send in `nil` clients by design.

pkg/cloud/vsphere/utils/utils.go
akutz: Done

pkg/cloud/vsphere/machineactuator.go
akutz: Fixed

pkg/cloud/vsphere/clusteractuator.go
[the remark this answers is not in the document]
akutz: It is. It's an old C-ism to mean a null-terminated string. See https://en.wikipedia.org/wiki/Hungarian_notation and search for `sz`

pkg/cloud/vsphere/clusteractuator.go
```diff
@@ -213,47 +210,76 @@ func (ca *ClusterActuator) provisionLoadBalancer(cluster *clusterv1.Cluster) err
...
+	cluster = cluster.DeepCopy()
+
+	if len(cluster.Status.APIEndpoints) > 0 {
+		return nil
+	}
+
+	controlPlaneEndpoint, err := vsphereutils.GetControlPlaneEndpoint(cluster, ca.lister)
+	if err != nil {
+		klog.Error(err)
+		if err, ok := errors.Cause(err).(*clustererror.RequeueAfterError); ok {
```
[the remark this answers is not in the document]
akutz: No longer relevant.

pkg/cloud/vsphere/clusteractuator.go
```diff
@@ -213,47 +210,76 @@ func (ca *ClusterActuator) provisionLoadBalancer(cluster *clusterv1.Cluster) err
...
+	if _, err := ca.clusterV1alpha1.Clusters(cluster.Namespace).UpdateStatus(cluster); err != nil {
+		err = errors.Wrapf(err,
+			"unable to update cluster with API endpoint "+
+				"%s=%s %s=%s %s=%s",
+			"api-endpoint", controlPlaneEndpoint,
+			"cluster-namespace", cluster.Namespace,
+			"cluster-name", cluster.Name)
+
+		ca.eventRecorder.Eventf(cluster, corev1.EventTypeWarning, "Failed Update", err.Error())
+		klog.Error(err)
```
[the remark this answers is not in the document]
akutz: Nope, and we don't any more.

pkg/cloud/vsphere/machineactuator.go
akutz: Re: constant, probably, but it's not used in too many places. I'll update it regardless.

pkg/cloud/vsphere/machineactuator.go
akutz: This isn't the default. This is a special case.

pkg/cloud/vsphere/provisioner/govmomi/create.go
akutz: It was logged before, so I'm still logging it. We can double-check it though.

pkg/cloud/vsphere/provisioner/govmomi/create.go
akutz: I wasn't going to remove any fields until we're in sync on templates and such.

pkg/cloud/vsphere/utils/utils.go
andrewsykim: But if we return error anyways when client is nil isn't it redundant to even call the first 3 functions?

akutz: The first three functions can return the ControlPlaneEndpoint without the client. Beyond that point the client is needed.

pkg/cloud/vsphere/clusteractuator.go
```diff
@@ -56,210 +57,175 @@ func NewClusterActuator(clusterV1alpha1 clusterv1alpha1.ClusterV1alpha1Interface
...
 
-// ensureLoadBalancerMembers would be responsible for keeping the master API endpoints
-// synced with the lb members at all times.
-func (ca *ClusterActuator) ensureLoadBalancerMembers(cluster *clusterv1.Cluster) error {
-	// This is the temporary implementation until we do the proper LB implementation
-	err := ca.setMasterNodeIPAsEndpoint(cluster)
-	if err != nil {
-		klog.Infof("Error registering master node's IP as API Endpoint for the cluster: %s", err)
-		return err
+	// Do not update Machine if nothing has changed
```
figo: Machine -> Cluster?

pkg/apis/vsphereproviderconfig/v1alpha1/vsphereclusterproviderstatus_types.go
```diff
@@ -0,0 +1,37 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
moshloop: nit: s/2018/2019/

pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+// Reconcile will create or update the cluster
+func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) (result error) {
+	ctx, err := context.NewClusterContext(&context.ClusterContextParams{
+		Cluster:    cluster,
+		Client:     a.client,
+		CoreClient: a.coreClient,
+		Logger:     klogr.New().WithName("[cluster-actuator]"),
```
moshloop: why not use the actual cluster name instead of [cluster-actuator] ?

pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -0,0 +1,132 @@
...
+	coreClient corev1.CoreV1Interface) *Actuator {
+
+	return &Actuator{
+		client:     client,
+		coreClient: coreClient,
+	}
+}
+
+// Reconcile will create or update the cluster
+func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) (result error) {
```
moshloop: nit: `result error` is a little confusing for me, I would expect `result` to be an object and error to be named `e` or `err` ?

pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -0,0 +1,132 @@
...
+	defer func() {
+		if result == nil {
+			record.Eventf(ctx.Cluster, "ReconcileSuccess", "reconciled cluster %q", ctx)
+		} else {
+			record.Warnf(ctx.Cluster, "ReconcileFailure", "failed to reconcile cluster %q: %v", ctx, result)
+		}
+	}()
+
+	ctx.Logger.V(2).Info("reconciling cluster")
+	defer ctx.Patch(utils.GetControlPlaneStatus)
```
moshloop: nit: why the need for defer? Why not just `return ctx.Patch(utils.GetControlPlaneStatus)` or similar after PKI.reconsole ? perhaps it just needs a comment as to why?

akutz: Hi @moshloop,

Because PKI reconcile may not always be the last thing called. This ensure resources are always patched.

pkg/cloud/vsphere/actuators/cluster/actuator.go
akutz: Hi @moshloop,

It's because if you use `err`, you can't use `err` in the function where `err` may be returned from a tuple, and I don't like not being able to use that. The only reason I use a named return param is to be able to set or check it during `defer` logic.

pkg/cloud/vsphere/actuators/cluster/actuator.go
akutz: Hi @moshloop,

Please see the cluster and machine contexts. The names *are* used. This is the initial name.

akutz:
> Besides for a couple of nits I like it
> 
> In terms of project backlog: At what point do todo's become issues?

Hi @moshloop,

I think we can discuss that this week. I just posted this to our `#yelling` channel in VMware Slack:

> IT WORKS!!!!!!!!!
> IT REALLY WORKS!!!!
> CAPV REFACTOR FINALLY WORKS!!! 
> https://gist.github.com/akutz/765a2e4220fa1a9f86c11248607e6fdc
> https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/307/pull-cluster-api-provider-vsphere-e2e/1140313958880645120

There's a bunch of cleanup that still needs to occur. The top three items, for me, are:

1. Add VMC ELB support
2. Create a concurrency model for selecting the initial control plane member by implemented a distributed lock with the cluster metadata and resource versions
3. The VM network address issue. I will open an issue on this. IMO it's a **massive** problem we need to solve before CAPV can scale.

akutz:
/test pull-cluster-api-provider-vsphere-e2e

There's a possibility e0d2de470f63836e3180713939b72d11f7e6af10 may have actually caused e2e to fail again by cleaning up resources before the controller issues the final patch against the machine. I'm going to re-run e2e a few more times to see if this is an intermittent error. If it fails consistently I will disable cleanup again. If that causes things to work, I'll work with @figo to figure out what to do.

akutz:
Ugh, the issue I mentioned in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/307#issuecomment-502475291 appears to be a flaky bit to do with ordering. We probably want to resolve this before merging this PR.

cc @figo @moshloop @frapposelli @andrewsykim

akutz:
I’m removing WiP, but going to place a hold until we figure out the flaky e2e that fails occasionally. 

/hold

moshloop:
Unless I am missing something I think this removes support for ESXi ?

figo:
@akutz the one failure you mentioned has this log:
```
I0616 18:20:06.651685       1 round_trippers.go:438] PATCH https://10.96.0.1:443/apis/cluster.k8s.io/v1alpha1/namespaces/default/machines/clusterapi-ag4zoyiq-w6bv7 500 Internal Server Error in 1 milliseconds
```
the ci code should not contributing to this error, the potential cause is from ci infra: VMC, could it be?

akutz:
> @akutz the one failure you mentioned has this log:
> 
> ```
> I0616 18:20:06.651685       1 round_trippers.go:438] PATCH https://10.96.0.1:443/apis/cluster.k8s.io/v1alpha1/namespaces/default/machines/clusterapi-ag4zoyiq-w6bv7 500 Internal Server Error in 1 milliseconds
> ```
> 
> the ci code should not contributing to this error, the potential cause is from ci infra: VMC, could it be?

Hi @figo,

I don't think this has anything to do with VMC unless it's the API server no longer running because the deployed VM was destroyed. The error appears to be originating from the bootstrap or `clusterctl` indicating it can no longer send a patch to the API server regarding an updated object (in this case the machine).

akutz:
> Unless I am missing something I think this removes support for ESXi ?

Hi @moshloop,

As Mr. Tyler crooned, you didn't [miss a thing](https://youtu.be/JkK8g6FMEXE). Yes, ESXi support was removed, albeit temporarily. I made a note [here](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/307/files#diff-0edf72958a0e8f8a6b9198a2bdbd8d0cR270) that I temporarily disabled ESXi cloning to simplify the refactor of the `create.go` source with the new logging, etc.

The decision to support ESXi directly is in the air at the moment, but if you would like to go ahead and create a branch based on this PR and add ESXi support back into the mix, I'll happily reintroduce it for now, until we have a more thorough discussion on the matter.

pkg/cloud/vsphere/actuators/machine/actuator.go
```diff
@@ -0,0 +1,231 @@
...
+	// Init the control plane by creating this machine.
+	if machineRole == context.ControlPlaneRole && len(controlPlaneMachines) == 1 {
+		if err := govmomi.Create(ctx, ""); err != nil {
+			return errors.Wrapf(err, "failed to create machine as initial member of the control plane %q", ctx)
+		}
+		return nil
+	}
+
+	// Join the existing cluster.
+	online, _, _ := kubeclient.GetControlPlaneStatus(ctx.ClusterContext)
```
andrewsykim: check err

pkg/cloud/vsphere/context/cluster_context.go
```diff
@@ -0,0 +1,368 @@
...
+	if secretName := clusterConfig.VsphereCredentialSecret; secretName != "" {
+		if params.CoreClient == nil {
+			return nil, errors.Errorf("credential secret %q specified without core client", secretName)
+		}
+		logr.V(4).Info("fetching vsphere credentials", "secret-name", secretName)
+		secret, err := params.CoreClient.Secrets(params.Cluster.Namespace).Get(secretName, metav1.GetOptions{})
+		if err != nil {
+			return nil, errors.Wrapf(err, "error reading secret %q for cluster %s/%s", secretName, params.Cluster.Namespace, params.Cluster.Name)
+		}
+		user2, userOk := secret.Data[constants.VsphereUserKey]
```
andrewsykim: nit: `userByte` or `userData`

pkg/cloud/vsphere/context/cluster_context.go
```diff
@@ -0,0 +1,368 @@
...
+			return nil, errors.Wrapf(err, "improperly formatted secret %q for cluster %s/%s", secretName, params.Cluster.Namespace, params.Cluster.Name)
+		}
+		user, pass = string(user2), string(pass2)
+		logr.V(2).Info("found vSphere credentials")
+	}
+
+	return &ClusterContext{
+		Context:       parentContext,
+		Cluster:       params.Cluster,
+		ClusterCopy:   params.Cluster.DeepCopy(),
```
andrewsykim: Passing along a deep copy of a resource in a context seems odd, is this what CAPA does?

pkg/cloud/vsphere/services/govmomi/create.go
```diff
@@ -56,6 +56,13 @@ const (
 
 	// nodeRole is the label assigned to every node in the cluster.
 	nodeRole = "node-role.kubernetes.io/node="
+
+	// macAddrOUI is the OUI prefix to use with generated MAC addresses
+	// Use "00:50:56" as it is VMware's OUI.
+	// The preferred OUI is "00:00:0f", the old NeXt OUI, but the in-tree
+	// cloud provider has a list of valid OUIs, and this OUI is not in the
+	// list.
+	macAddrOUI = "00:50:56"
```
andrewsykim: macAddrOUIPrefix

pkg/cloud/vsphere/services/govmomi/create.go
```diff
@@ -638,3 +646,12 @@ func byteToGiB(n int64) int64 {
 func giBToByte(n int64) int64 {
 	return int64(n * int64(math.Pow(1024, 3)))
 }
+
+// newMACAddr generates a new MAC address using VMware's OUI.
+func newMACAddr() (string, error) {
```
andrewsykim: Maybe for another PR, but we can unit test this by running it through `net.ParseMac`

pkg/cloud/vsphere/services/metadata/metadata_test.go
```diff
@@ -0,0 +1,216 @@
...
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tc.ctx.Machine.Name = tc.name
+			actVal, err := metadata.New(tc.ctx)
+			if err != nil {
+				t.Fatal(err)
+			}
+			buf, err := base64.StdEncoding.DecodeString(actVal)
```
andrewsykim: nit: validate against actual string values?

pkg/cloud/vsphere/context/cluster_context.go
andrewsykim: Ah, I'm recalling our previous conversations now. Guessing it's needed to make the cluster patch request work?

go.mod
```diff
@@ -2,13 +2,19 @@ module sigs.k8s.io/cluster-api-provider-vsphere
 
 go 1.12
 
+replace (
+	github.com/go-logr/logr => github.com/go-logr/logr v0.1.0
+	github.com/go-logr/zapr => github.com/go-logr/zapr v0.1.0
+	k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20181127025237-2b1284ed4c93
```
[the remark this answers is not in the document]
akutz: We're using the necessary version.

pkg/cloud/vsphere/context/machine_context.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	return GetMachineRole(c.Machine)
+}
+
+// IPAddr returns the machine's IP address.
+func (c *MachineContext) IPAddr() string {
+	if c.Machine == nil {
+		return ""
+	}
+	return c.Machine.Annotations[constants.VmIpAnnotationKey]
```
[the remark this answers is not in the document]
akutz: Go supports checking nil maps for key queries - https://play.golang.org/p/kOy6dWFYrZI

pkg/cloud/vsphere/actuators/cluster/actuator.go
andrewsykim: hmm starting to see the benefits for consistency though

pkg/cloud/vsphere/context/cluster_context.go
```diff
@@ -0,0 +1,368 @@
...
+
+// GetControlPlaneMachines returns the control plane machines for the cluster.
+func (c *ClusterContext) GetControlPlaneMachines() ([]*clusterv1.Machine, error) {
+	machines, err := c.GetMachines()
+	if err != nil {
+		return nil, err
+	}
+	controlPlaneMachines := []*clusterv1.Machine{}
+	for _, machine := range machines {
+		if role := GetMachineRole(machine); role == ControlPlaneRole {
```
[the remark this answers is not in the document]
akutz: Hi @andrewsykim,

Previously, and in CAPV, the node's role is checked both ways. I standardized on `set` as it makes more sense.

pkg/cloud/vsphere/context/cluster_context.go
```diff
@@ -0,0 +1,368 @@
...
+
+// Pass returns the password used to access the vSphere endpoint.
+func (c *ClusterContext) Pass() string {
+	return c.pass
+}
+
+// CanLogin returns a flag indicating whether the cluster config has
+// enough information to login to the vSphere endpoint.
+func (c *ClusterContext) CanLogin() bool {
+	return c.ClusterConfig.VsphereServer != "" && c.user != ""
```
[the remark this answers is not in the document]
akutz: I'd hash it if it was required, and it isn't. The server and user name is all that matters. If a person changes their password it doesn't change the account used to access the system.

pkg/cloud/vsphere/context/machine_context.go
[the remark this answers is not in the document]
andrewsykim: ah good point

pkg/cloud/vsphere/context/cluster_context.go
akutz: Sí. You compare to the original copy to generate the JSON patch.

pkg/cloud/vsphere/services/govmomi/create.go
akutz: The OUI *is* the prefix.

pkg/cloud/vsphere/services/metadata/metadata_test.go
akutz: Unless I have authoritative, expected data, any generated data is going to be based on previous generated data. For now I just wanted to make sure *looked* okay.

cmd/manager/main.go
```diff
@@ -20,36 +20,34 @@ import (
...
-	// Get a config to talk to the apiserver
-	cfg, err := config.GetConfig()
-	if err != nil {
-		klog.Fatalf("Failed to get config: %s", err.Error())
-	}
+	cfg := config.GetConfigOrDie()
 
-	syncPeriod := 120 * time.Second
+	// Setup a Manager
+	syncPeriod := 10 * time.Minute
```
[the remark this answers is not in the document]
akutz: Hi @andrewsykim,

Because I literally copied over our existing manager from the one from CAPA and just replaced the actuator init calls.

pkg/cloud/vsphere/actuators/cluster/actuator.go
akutz: Hi @andrewsykim and @moshloop,

There's only one error case *for now*. That will change as soon as we introduce downloading the OVA into the content library, VMC LB support, etc. I'm leaving this as is as it's consistent with the other functions and will be needed in not even weeks.

go.mod
```diff
@@ -2,13 +2,19 @@ module sigs.k8s.io/cluster-api-provider-vsphere
...
+	k8s.io/klog => k8s.io/klog v0.2.0
```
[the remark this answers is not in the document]
akutz: Yes, we do. Please commit a8ea97726a202bbdeb7c5c0bb64becf8254ee317.

pkg/cloud/vsphere/services/govmomi/create.go
akutz: Unit test that we correctly formatted six bytes as a colon-separated hexadecimal string?

pkg/cloud/vsphere/constants/constants.go
```diff
@@ -5,6 +5,7 @@ import (
 )
 
 const (
+	DefaultBindPort                  = 6443
```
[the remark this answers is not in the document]
akutz: Hi @andrewsykim,

I wanted to introduce this as a default, not as the authoritative value. I plan to remove the other and see if it breaks anything as the new constant is used in many more places.

pkg/cloud/vsphere/services/govmomi/create.go
andrewsykim: No harm in that IMO, a unit test using `ParseMac` should be super simple. If you leave it out here let me know and happy to add it as a follow-up :)

akutz: Fair enough. I'll add it in the commit used to address this round of feedback.

pkg/cloud/vsphere/actuators/machine/actuator.go
akutz: Hi @andrewsykim,

The error is ignored because the reaction is to requeue. Since the reaction doesn't depend on the error, only whether `online` is `true` or `false` (and will be `false` if there *is* an error), the error can be safely ignored in this case.

pkg/cloud/vsphere/services/govmomi/create.go
akutz: This commit was removed from this PR

andrewsykim:
/lgtm
/hold

Leaving hold cancel to @akutz

cmd/manager/main.go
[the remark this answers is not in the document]
sidharthsurana: I think the value we have is actually more appropriate for the CAPV than the one coming from CAPA

pkg/cloud/vsphere/context/machine_role.go
```diff
@@ -0,0 +1,53 @@
...
+const (
+	// ControlPlaneRole indicates a machine is a member of the control plane.
+	ControlPlaneRole MachineRole = "controlplane"
+
+	// NodeRole indicates a machine is a member of the cluster.
+	NodeRole MachineRole = "node"
+)
+
+// GetMachineRole returns the Machine's role.
+func GetMachineRole(machine *clusterv1.Machine) MachineRole {
```
sidharthsurana: Why deviate from the way CAPI determines which machine is a control plane machine or not?
There is a utils method `IsControlPlaneMachine` in CAPI that gives that answer.

[CAPI method: IsControlPlaneMachine](https://github.com/kubernetes-sigs/cluster-api/blob/91179deda81cc4b96740a1d182f64a293844dae5/pkg/util/util.go#L134-L137)
```// IsControlPlaneMachine checks machine is a control plane node.
func IsControlPlaneMachine(machine *clusterv1.Machine) bool {
	return machine.Spec.Versions.ControlPlane != ""
}
```

akutz: Hi @sidharthsurana,

Please see this [Slack conversation](https://kubernetes.slack.com/archives/C8TSNPY4T/p1560807049390100) where I dug into this very same issue the other day :)

pkg/apis/vsphereproviderconfig/v1alpha1/vspheremachineproviderstatus_types.go
```diff
@@ -0,0 +1,37 @@
...
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// VsphereMachineProviderStatus defines the observed state of VsphereMachineProviderConfig
+// +k8s:openapi-gen=true
+type VsphereMachineProviderStatus struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	TaskRef string `json:"taskRef"`
```
moshloop: I don't think this TaskRef works, or at least not in the way it is currently being used. 

When trying to patch in the TaskRef you get a `422 Unprocessable Entity` error, previously TaskRef was being set by Machine.UpdateStatus()
Not sure if this is due to how the API is defined (I tried making it optional with no difference) trying to patch a `runtime.RawExtension` in a status

figo:
the CI has completed the switching to kind as bootstrap cluster, 
there is one issue i need to follow up after this merge: ci trying to read controller manager log from bootstrap cluster but it will be deleted, although ci has all test job log and able to determine test success or failure based on test job status.
  
feel free to cancel hold.

pkg/apis/vsphereproviderconfig/v1alpha1/vspheremachineproviderstatus_types.go
akutz: Thank you @moshloop for helping fix the broken task logic!

moshloop: I am testing against ESXi, so maybe it is a new code path I have introduced - will submit a PR and see if that passes on VMC

pkg/cloud/vsphere/context/machine_role.go
akutz: Resolved in 1009d0afb6e811276a645302f719c85e8ad67acc

PR #309, opened 2019-06-15
Use envsubst for template file

figo:
also keep a single provider-components.yaml file


figo:
it missing cncf-cla label, not sure why, but i will try re-test or force push again, hopefully the label can be added.

PR #313, opened 2019-06-17
Add License Headers To Code And Script Files

### LINKED ISSUE #292: Add license headers
There are many files that don't have license headers, we should add them.

james65535:
Some code and script files lacked license headers.  This commit adds headers only to files which lacked any header at all.  It does not modify existing license headers.

Fixes #292

### DIFF
--- a/installer/build/bootable/root/etc/profile.d/kubernetes.sh
+++ b/installer/build/bootable/root/etc/profile.d/kubernetes.sh
@@ -1,3 +1,17 @@
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


akutz:
/hold

Hi @james65535,

This looks great! However, I'm concerned that this will be a difficult merge due to #307. Since there aren't *that* many files in the mix, I think we should hold this until 307 is merged, and then you can see what still lacks the headers. Thanks!

james65535:
Rebased on top of latest changes.  Will do another scan for missing headers before this is ready to go.

james65535:
/hold cancel

tidied up commit and there are now new files with missing license headers.

akutz:
Hi @james65535,

You're going to hate me, but...have you seen #334?

PR #322, opened 2019-06-19
refactor: reduce logging verbosity + add additional context

moshloop:


pkg/cloud/vsphere/context/cluster_context.go
```diff
@@ -102,7 +102,7 @@ func NewClusterContext(params *ClusterContextParams) (*ClusterContext, error) {
 	if logr == nil {
 		logr = klogr.New().WithName("default-logger")
 	}
-	logr = logr.WithName(params.Cluster.APIVersion).WithName(params.Cluster.Namespace).WithName(params.Cluster.Name)
+	logr = logr.WithName(params.Cluster.Name)
```
akutz: The current model is in sync with CAPA and provides necessary information. What is the reason for changing it?

moshloop: Adding the full qualified API version extends the line length significantly, leaving less room real estate for meaningful log data

akutz: I understand your concern, but because logs could be interspersed from multiple incoming calls, having the context of the object being logged is important as log data is flushed in real-time. If you still wish to make this change, please bring it up in the next upstream call. For now let's leave it as is so we can merge the rest of the PR : )

moshloop: 👍

akutz: Hi @moshloop,

FWIW, we literally just got done demoing 0.3.0 where the logs verbosity helped. A prior machine set was still being deleted during the creation of a new one, and log entries were being interlaced with each other.

pkg/cloud/vsphere/services/govmomi/create.go
```diff
@@ -267,23 +267,23 @@ func Create(ctx *context.MachineContext, bootstrapToken string) error {
...
+	ctx.Logger.V(6).Info("finding vm by instance UUID", "instance-uuid", ctx.Machine.UID)
 
 	vmRef, err := ctx.Session.FindByInstanceUUID(ctx, string(ctx.Machine.UID))
 	if err != nil {
 		return "", errors.Wrapf(err, ctx.String())
 	}
 
 	if vmRef != nil {
-		ctx.Logger.V(2).Info("found machine by instance UUID", "vmRef", vmRef.Reference().Value, "instance-uuid", ctx.Machine.UID)
+		ctx.Logger.V(5).Info("found machine by instance UUID", "vmRef", vmRef.Reference().Value, "instance-uuid", ctx.Machine.UID)
```
figo: nit, i don't see a real difference between v5 and v6

akutz: Hi @figo,

It was changed from 2 to 5.

figo: i know, but my question is why not 6? as several other logs in this change, or do we really differentiate level 5 and 6?

akutz: Good question. Honestly I want to see someone write a guide on what to log and at what level. I may create an issue on the topic.

PR #324, opened 2019-06-19
clusterctl: VM folder should be optional

andrewsykim:

/assign @akutz

### DIFF
--- a/cmd/clusterctl/examples/vsphere/generate-yaml.sh
+++ b/cmd/clusterctl/examples/vsphere/generate-yaml.sh
@@ -158,8 +158,7 @@ if [ ! "${CAPV_YAML_VALIDATION:-1}" = "0" ]; then
   fi
 
   if [ -z "${VSPHERE_FOLDER}" ]; then
-    echo "env var VSPHERE_FOLDER is required" 1>&2
-    exit 1
+    echo "WARNING: VSPHERE_FOLDER is not set, defaulting to root folder"
   fi
 
   if [ -z "${VSPHERE_TEMPLATE}" ]; then


figo:
good catch, although vm folder is required for VMC env

/lgtm

akutz:
> good catch, although vm folder is required for VMC env
> 
> /lgtm

Technically it's required for any env where you can't deploy at the root. But for envs sans folders, this blocks them. We basically have an inflexible generator at the moment that we need to improve to tolerate different environment configs.

andrewsykim:
/hold

Should we log warning if it's not set?

akutz:
Hi @andrewsykim,

I recommend the following:

```shell
[ ! "${VM_FOLDER+x}" ] || VM_FOLDER="${VM_FOLDER:?empty}"
echo "VM_FOLDER=${VM_FOLDER}"
```

The above script says either `VM_FOLDER` must be not set, or, if set, must not be the empty value. Otherwise an error is printed. For example:

**`VM_FOLDER` is defined, but empty**

```shell
$ cat <<EOF | VM_FOLDER= bash -s --
[ ! "\${VM_FOLDER+x}" ] || VM_FOLDER="\${VM_FOLDER:?empty}"
echo "\${VM_FOLDER}"
EOF
bash: line 1: VM_FOLDER: empty
```

**`VM_FOLDER` is not defined**

```shell
$ cat <<EOF | bash -s --
[ ! "\${VM_FOLDER+x}" ] || VM_FOLDER="\${VM_FOLDER:?empty}"
echo "\${VM_FOLDER}"
EOF

```

**`VM_FOLDER` is defined and not empty**

```shell
$ cat <<EOF | VM_FOLDER="workloads" bash -s --
[ ! "\${VM_FOLDER+x}" ] || VM_FOLDER="\${VM_FOLDER:?empty}"
echo "\${VM_FOLDER}"
EOF
workloads
```

/hold

andrewsykim:
In this case we don't want to log if VM_FOLDER is set so it might be an easier check than that 🤔

andrewsykim:
@akutz updated, let me know what you think

PR #330, opened 2019-06-20
Add clusterctl tools container for generating example yaml

andrewsykim:
Signed-off-by: Andrew Sy Kim kiman@vmware.com

This adds a clusterctl-tools container which has kustomize and envsubst already installed. This should make the clusterctl example workflow much easier since the only requirement is now docker which is already required for Kind.


Makefile
```diff
@@ -54,6 +44,12 @@ manager: fmt vet
 clusterctl: fmt vet
 	go build -o bin/clusterctl ./cmd/clusterctl
 
+# Build the clusterctl-tools container used for generating example clusterctl yaml
+clusterctl-tools:
+ifeq "$(shell docker images -q clusterctl-tools)" ""
+	docker build -t clusterctl-tools -f cmd/clusterctl/examples/tools/Dockerfile .
```
andrewsykim: Building locally for now but pushing to a registry might be a good next step

cmd/clusterctl/examples/tools/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM golang:1.12
```
andrewsykim: golang:1.12 is a beefy image, ~700MB. I couldn't find anything smaller than this that has go installed. If we think this is too big for a tools container we could install go ourselves on a slimmer base image, thoughts?

Makefile
```diff
@@ -54,6 +44,12 @@ manager: fmt vet
 clusterctl: fmt vet
 	go build -o bin/clusterctl ./cmd/clusterctl
 
+# Build the clusterctl-tools container used for generating example clusterctl yaml
+clusterctl-tools:
+ifeq "$(shell docker images -q clusterctl-tools)" ""
```

This should be:

```makefile
ifeq (,$(strip $(shell docker images -q clusterctl-tools))
```

Also, I'm not sure it should be a Make `ifeq` statement since this will fail if anyone runs any target and Docker isn't present or running. Perhaps make it part of the recipe instead of part of the target. Please consider rewriting this as:

```makefile
clusterctl-tools:
	[ -n "$(docker images -q clusterctl-tools)" ] || docker build -t clusterctl-tools -f cmd/clusterctl/examples/tools/Dockerfile .
```

Makefile
```diff
@@ -98,13 +94,25 @@ vendor:
...
+
 ####################################
 # DEVELOPMENT Build and Push targets
 ####################################
 
 # Create YAML file for deployment
-dev-yaml: | $(KUSTOMIZE)
-	CAPV_MANAGER_IMAGE=${DEV_IMG} cmd/clusterctl/examples/vsphere/generate-yaml.sh
+dev-yaml: clusterctl-tools
+	docker run -v $(PWD):/go/src/sigs.k8s.io/cluster-api-provider-vsphere \
```

I don't know that we should assume `PWD` is always set. At the top of the Makefile please consider adding:

```makefile
PWD := $(abspath .)
```

Makefile
```diff
@@ -98,13 +94,25 @@ vendor:
...
+	  -w /go/src/sigs.k8s.io/cluster-api-provider-vsphere \
+	  $(GENERATE_YAML_ENV_VARS) clusterctl-tools \
+	  sh -c "CAPV_MANAGER_IMAGE=${DEV_IMG} cmd/clusterctl/examples/vsphere/generate-yaml.sh"
```

Since the k/k scripts use `bash`, please use `bash -c` since `sh` isn't always linked to a bash shell.

Makefile
```diff
@@ -98,13 +94,25 @@ vendor:
 	cp -rf --no-preserve=mode "$${_src}" "$${_dst}"
 .PHONY: vendor
 
+# GENERATE_YAML_ENV_VARS is the list of all environment variables expected by generate-yaml.sh
+# to build example manifests for clusterctl. The list of environment variables here should contain
+# the environment variables expected in generate-yaml.sh
+GENERATE_YAML_ENV_VARS = -e CLUSTER_NAME -e SERVICE_CIDR -e CLUSTER_CIDR -e CAPV_YAML_VALIDATION \
```

Please use `:=` when assigning variables in Make that do not ever need to be [recursively evaluated](https://www.gnu.org/software/make/manual/html_node/Setting.html).

Makefile
```diff
@@ -54,6 +44,12 @@ manager: fmt vet
 clusterctl: fmt vet
 	go build -o bin/clusterctl ./cmd/clusterctl
 
+# Build the clusterctl-tools container used for generating example clusterctl yaml
+clusterctl-tools:
```

This target should be marked as `PHONY`

.gitignore
```diff
@@ -27,9 +27,6 @@ config/default/vsphere_tmp
 # Ignore sed backup files
 *-e
 
-# Tooling
-.bin/
```
akutz: Hi @andrewsykim,

Please leave this as is. It isn't hurting anything, and I doubt anything ever put there should be committed anyway.

Makefile
```diff
@@ -98,13 +94,25 @@ vendor:
...
+		      -e KUBERNETES_VERSION
+
 ####################################
 # DEVELOPMENT Build and Push targets
 ####################################
 
 # Create YAML file for deployment
-dev-yaml: | $(KUSTOMIZE)
-	CAPV_MANAGER_IMAGE=${DEV_IMG} cmd/clusterctl/examples/vsphere/generate-yaml.sh
+dev-yaml: clusterctl-tools
```

This should be an order-only dependency. Please change it to:

```make
dev-yaml: | clusterctl-tools
```

Makefile
```diff
@@ -120,8 +128,11 @@ dev-push:
 ###################################
 
 # Create YAML file for deployment
-prod-yaml: | $(KUSTOMIZE)
-	CAPV_MANAGER_IMAGE=${PRODUCTION_IMG} cmd/clusterctl/examples/vsphere/generate-yaml.sh
+prod-yaml: clusterctl-tools
```

This should be an order-only dependency. Please change it to:

```make
prod-yaml: | clusterctl-tools
```

Makefile
```diff
@@ -139,8 +150,11 @@ prod-push:
 ###################################
 
 # Create YAML file for deployment into CI
-ci-yaml: | $(KUSTOMIZE)
-	CAPV_MANAGER_IMAGE=${CI_IMG}:${VERSION} cmd/clusterctl/examples/vsphere/generate-yaml.sh
+ci-yaml: clusterctl-tools
```

This should be an order-only dependency. Please change it to:

```make
ci-yaml: | clusterctl-tools
```

andrewsykim:
> It looks like the only thing being tied to Docker is the generation of the YAML, correct? In which case then perhaps that isn't the biggest problem. As long as we're able to update the CRDs sans Docker, I think it may be fine.

Correct, this is only for generating example yaml for clusterctl which previously required kustomize and envsubst to be installed. I'd like to take this a step further and build `clusterctl` in docker which also removes the need for having a go environment to use clusterctl.

cmd/clusterctl/examples/tools/Dockerfile
akutz: Hi @andrewsykim,

Unless we need Go you could use `debian:stretch` as a base and then just build the tools you need in a build phase and copy them into the final phase.

akutz:
> > It looks like the only thing being tied to Docker is the generation of the YAML, correct? In which case then perhaps that isn't the biggest problem. As long as we're able to update the CRDs sans Docker, I think it may be fine.
> 
> Correct, this is only for generating example yaml for clusterctl which previously required kustomize and envsubst to be installed. I'd like to take this a step further and build `clusterctl` in docker which also removes the need for having a go environment to use clusterctl.

Hi @andrewsykim,

I don't want you to spend **too** much time on this as the plan is still to move to Bazel sooner than later, at which point this entire PR becomes moot.

andrewsykim:
> But I don't want that to be the default means to build clusterctl as it's an unnecessary requirement for building locally.

Good point, I'll add a separate target for this one.

figo:
i like the idea, this will help simplifying e2e.

cmd/clusterctl/examples/tools/Dockerfile
andrewsykim: Going to use `debian:stretch-slim` unless there are objections

akutz: That's what I meant anyway, so go for it. FYI, you may wish to set `GOOS`, `GOHOSTOS`, etc. in the container just so anything that depends on those will find them already set.

andrewsykim:
Comments addressed, PTAL

Makefile
```diff
@@ -98,13 +100,25 @@ vendor:
 	cp -rf --no-preserve=mode "$${_src}" "$${_dst}"
 .PHONY: vendor
 
+# GENERATE_YAML_ENV_VARS is the list of all environment variables expected by generate-yaml.sh
+# to build example manifests for clusterctl. The list of environment variables here should contain
+# the environment variables expected in generate-yaml.sh
+GENERATE_YAML_ENV_VARS := -e CLUSTER_NAME -e SERVICE_CIDR -e CLUSTER_CIDR -e CAPV_YAML_VALIDATION \
```

By the way, I forgot to say, thank you for teaching me about this feature of Docker. I didn't realize you could use `-e` to set an env var without also the value. This is nice.

cmd/clusterctl/examples/tools/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM debian:stretch-slim
```
akutz: Hi @andrewsykim,

Please remind me to give you creds to the CAPV GCS project so you can get this image pushed there.

PR #332, opened 2019-06-20
Do not log requeue errors as failures

akutz:
This patch updates the actuator error handling so that requeue errors are no longer logged as failures.

### DIFF
--- a/pkg/cloud/vsphere/actuators/actuators.go
+++ b/pkg/cloud/vsphere/actuators/actuators.go
@@ -17,6 +17,8 @@ limitations under the License.
 package actuators
 
 import (
+	"fmt"
+
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 
@@ -28,6 +30,7 @@ import (
 )
 
 type patchContext interface {
+	fmt.Stringer
 	GetObject() runtime.Object
 	GetLogger() logr.Logger
 	Patch() error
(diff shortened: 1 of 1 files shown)


pkg/cloud/vsphere/actuators/actuators.go
```diff
@@ -31,6 +31,7 @@ type patchContext interface {
 	GetObject() runtime.Object
 	GetLogger() logr.Logger
 	Patch() error
+	String() string
```
andrewsykim: Can use `fmt.Stringer` here

pkg/cloud/vsphere/actuators/actuators.go
```diff
@@ -50,7 +51,11 @@ func PatchAndHandleError(ctx patchContext, opName string, opErr error) error {
 	// and log the op error if one is replaced with the requeue error.
 	requeueErr, isRequeueErr := errors.Cause(err).(*clustererr.RequeueAfterError)
 	if isRequeueErr {
-		ctx.GetLogger().Error(opErr, "op failed", "object", ctx.GetObject())
+		// If the requeue error is not the outer-most error then log the outer-most
+		// error since it will be dropped in favor of the requeue.
+		if requeueErr != err {
+			ctx.GetLogger().V(6).Info("requeue after error", "object", ctx.String(), "error", err)
```
andrewsykim: I might be misunderstanding here, so `err` here is any opErr that is _not_ a requeue error? I feel the log level for that should be lower?

akutz: Hi @andrewsykim,

This is when an error **contained a requeue error** (the requeue error was wrapped), but the error is itself **not** a requeue error. For example:

1. A VM cannot be cloned because there's not enough disk space -- a requeue error is returned
2. The VM clone function wraps the requeue error with "vm clone failed"

The PR makes it so "vm clone failed" is still logged at `V(6)` while the inner requeue error is returned to initiate a requeue.

andrewsykim: Gotcha, so a "vm clone failed" is log level 6? Isn't that a real error we should log always?

akutz: There's certainly nothing to prevent someone from logging that elsewhere, but by the time an error reaches this point we can't make any guesses as to the severity of the error. Additionally, somewhere on the call stack, someone decided that the error wasn't severe enough to prevent a retry. Therefore if a requeue was part of the error, it's my opinion that any other, associated error, shouldn't be ignored, but also shouldn't clog the logs.

andrewsykim: Makes sense, thanks for clarifying!

PR #334, opened 2019-06-20
Remove the installer

akutz:
This patch removes the installer OVA as it is no longer used.


akutz:
Hi @figo,

I imagine we'll need to update Prow to no longer build this.

sidharthsurana:
/lgtm
/hold
hold just so that @figo can make any required changes to CI

codenrhoden:
If @figo doesn't have the bandwidth for this, I may be able to take this part on. There will be more changes to this repo needed as well, since the `installer` directory is referenced in several places in `scripts/e2e/e2e.sh`.

akutz:
/assign @figo
/assign @codenrhoden 

I'll let the two of you fight it out to see who takes this one :)

figo:
> If @figo doesn't have the bandwidth for this, I may be able to take this part on. There will be more changes to this repo needed as well, since the `installer` directory is referenced in several places in `scripts/e2e/e2e.sh`.

feel free to take on it.

/lgtm

codenrhoden:
@akutz (or @figo if you know the details), can you give a short description of how the e2e tests are executed, now that the `installer` image is no longer used? I see *a lot* of images in [gcr.io](https://console.cloud.google.com/gcr/images/cnx-cluster-api/GLOBAL) and frankly don't understand why there are so many.

From what I see, Prow is calling `scripts/ci-test.sh` and `scripts/e2e/e2e.sh` on presubmits, and `scripts/e2e/e2e.sh ova` as a periodic.

It looks to me that when run with the `ova` argument, the script builds a new copy of the `installer` image and pushes that, then *runs* that new container to build a new ovf, and uploads that to VMC for testing. As part of building that new ovf, it builds the ovf, uploads it, and starts that ovf as the "bootstrap VM", waiting until the IP of that VM becomes available.

Before I rip that part out, I would like to know what happens in the new world, just so that I don't remove too much! In other words, do we still need the `e2e.sh` script to start a bootstrap VM at all? If so, where does it come from? It looks like if `ova` is not passed as an arg to `e2e.sh`, rather than building and uploading a new OVF, it is using an existing one named "prow", and always reverting it to a snapshot named `cluster-api-provider-vsphere-ci-0.0.2` before running. Does that logic stick around too?

codenrhoden:
Even though https://github.com/kubernetes/test-infra/pull/13120 has not merged, since #335 has, it should be safe to merge this now. When the prow job runs `scripts/e2e/e2e.sh ova`, which only happens every 24 hours, the "ova" part will just be ignored anyways.

akutz:
Hi @andrewsykim,

Could you lgtm this? Travis indicated it's safe to merge, and it should speed things up when gone. Thanks!

PR #336, opened 2019-06-21
clusterctl: generate yaml in docker as current user

andrewsykim:

Run generate-yaml.sh in docker as the current user, otherwise generated yaml files belong to root which end up being unread-able to other users.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,6 +117,7 @@ GENERATE_YAML_ENV_VARS := -e CLUSTER_NAME -e SERVICE_CIDR -e CLUSTER_CIDR -e CAP
 dev-yaml: | clusterctl-tools
 	docker run -v $(CWD):/go/src/sigs.k8s.io/cluster-api-provider-vsphere \
 	  -w /go/src/sigs.k8s.io/cluster-api-provider-vsphere \
+	  -u $$(id -u):$$(id -g) \
 	  $(GENERATE_YAML_ENV_VARS) clusterctl-tools \
 	  bash -c "CAPV_MANAGER_IMAGE=${DEV_IMG} cmd/clusterctl/examples/vsphere/generate-yaml.sh"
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -117,6 +117,7 @@ GENERATE_YAML_ENV_VARS := -e CLUSTER_NAME -e SERVICE_CIDR -e CLUSTER_CIDR -e CAP
 dev-yaml: | clusterctl-tools
 	docker run -v $(CWD):/go/src/sigs.k8s.io/cluster-api-provider-vsphere \
 	  -w /go/src/sigs.k8s.io/cluster-api-provider-vsphere \
+	  -u $(shell id -u):$(shell id -g) \
```

Every use of `shell` starts a new sub-process. Please either:

1. Define both once as Makefile vars, or
2. Move the logic to the shell: `$$(id -u):$$(id -g)` so running `make` doesn't need to call these unless calling these targets

Makefile
```diff
@@ -138,6 +139,7 @@ dev-push:
 prod-yaml: | clusterctl-tools
 	docker run -v $(CWD):/go/src/sigs.k8s.io/cluster-api-provider-vsphere \
 	  -w /go/src/sigs.k8s.io/cluster-api-provider-vsphere \
+	  -u $(shell id -u):$(shell id -g) \
```

Please see my comment above.

Makefile
```diff
@@ -161,6 +163,7 @@ prod-push:
 ci-yaml: | clusterctl-tools
 	docker run -v $(CWD):/go/src/sigs.k8s.io/cluster-api-provider-vsphere \
 	  -w /go/src/sigs.k8s.io/cluster-api-provider-vsphere \
+	  -u $(shell id -u):$(shell id -g) \
```

Please see my comment above.

akutz:
Hi @andrewsykim,

Please update with either suggestion and then feel free to remove the hold.

/hold
/lgtm

andrewsykim:
@akutz updated, PTAL

sidharthsurana:
nice trick 👍 
/lgtm

PR #337, opened 2019-06-21
cluster actuator: sync kubeconfig secret

andrewsykim:

Syncs the kubeconfig for clusters as Secrets. The naming format is Secrets is `<cluster-name>-kubeconfig`. 

Fixes #

**Release note**:

```release-note
Sync the kubeconfig for a cluster as a Kubernetes Secret. 
```

/assign @akutz

### DIFF
--- a/pkg/cloud/vsphere/actuators/cluster/actuator.go
+++ b/pkg/cloud/vsphere/actuators/cluster/actuator.go
@@ -17,10 +17,15 @@ limitations under the License.
 package cluster
 
 import (
+	"fmt"
 	"net"
 	"strconv"
 
 	"github.com/pkg/errors"
+
+	apiv1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog/klogr"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
@@ -77,6 +82,13 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) (opErr error) {
 		return err
 	}
 
+	// reconcileKubeConfig can fail because control plane endpoints
+	// aren't ready so return requeue error
+	if err := a.reconcileKubeConfig(ctx); err != nil {
+		return errors.Wrapf(&clusterErr.RequeueAfterError{RequeueAfter: constants.DefaultRequeue},
+			"error reconciling kubeconfig secret for cluster %q", ctx)
+	}
+
 	return nil
 }
 
@@ -96,6 +108,13 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster) (opErr error) {
 	}()
 
 	ctx.Logger.V(2).Info("deleting cluster")
+
+	// if deleteKubeconfig fails, return requeue error so kubeconfig
+	// secret is properly cleaned up
+	if err := a.deleteKubeConfig(ctx); err != nil {
+		return errors.Wrapf(&clusterErr.RequeueAfterError{RequeueAfter: constants.DefaultRequeue},
+			"error deleting kubeconfig secret for cluster %q", ctx)
+	}
 
 	return nil
 }
@@ -128,10 +147,16 @@ func (a *Actuator) GetKubeConfig(cluster *clusterv1.Cluster, machine *clusterv1.
 	if err != nil {
 		return "", err
 	}
+
+	if machine == nil {
+		return kubeclient.GetKubeConfig(clusterContext)
+	}
+
 	machineContext, err := context.NewMachineContextFromClusterContext(clusterContext, machine)
 	if err != nil {
 		return "", err
 	}
+
 	return kubeclient.GetKubeConfig(machineContext)
 }
 
@@ -180,4 +205,36 @@ func (a *Actuator) reconcileReadyState(ctx *context.ClusterContext) error {
 	ctx.Logger.V(6).Info("cluster is ready")
 
 	return nil
+}
+
+// reconcileKubeConfig creates a Kubernetes Secret holding the kubeconfig
+// for a cluster if the secret does not already exist
+func (a *Actuator) reconcileKubeConfig(ctx *context.ClusterContext) error {
+	cluster := ctx.Cluster
+
+	kubeConfig, err := a.GetKubeConfig(cluster, nil)
+	if err != nil {
+		return errors.Wrap(err, "error generating kubeconfig")
+	}
+
+	secret := &apiv1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%s-kubeconfig", cluster.Name),
+		},
+		StringData: map[string]string{
+			"value": kubeConfig,
+		},
+	}
+
+	if _, err := a.coreClient.Secrets(cluster.Namespace).Create(secret); err != nil &&
+		!apierrors.IsAlreadyExists(err) {
+		return errors.Wrap(err, "error creating kubeconfig secret")
+	}
+
+	return nil
+}
+
+func (a *Actuator) deleteKubeConfig(ctx *context.ClusterContext) error {
+	secretName := fmt.Sprintf("%s-kubeconfig", ctx.Cluster.Name)
+	return a.coreClient.Secrets(ctx.Cluster.Namespace).Delete(secretName, &metav1.DeleteOptions{})
 }


pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -115,7 +142,7 @@ func (a *Actuator) GetIP(cluster *clusterv1.Cluster, machine *clusterv1.Machine)
 
 // GetKubeConfig returns the contents of a Kubernetes configuration file that
 // may be used to access the cluster.
-func (a *Actuator) GetKubeConfig(cluster *clusterv1.Cluster, machine *clusterv1.Machine) (string, error) {
+func (a *Actuator) GetKubeConfig(cluster *clusterv1.Cluster, _ *clusterv1.Machine) (string, error) {
```
andrewsykim: Assumed it's safe to ignore the machine resource here since cluster context provides enough information to generate the kubeconfig.

akutz: Hi @andrewsykim,

That's not accurate. Please see this squashed commit comment from bd46a0bf5fc6c2c88e5d8b1f6bf254bb3d4309ab:

> This patch allows a machine to return a control plane endpoint to satisfy the ProviderDeployer interface. Please see the Slack conversation https://kubernetes.slack.com/archives/C8TSNPY4T/p1560696064368600 for more information.

pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -72,6 +77,28 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) (opErr error) {
...
+	secret := &apiv1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%s-kubeconfig", cluster.Name),
+		},
+		StringData: map[string]string{
+			"value": kubeConfig,
+		},
+	}
+
+	_, err = a.coreClient.Secrets(cluster.Namespace).Create(secret)
```
akutz: Hi @andrewsykim,

Please collapse into a single line since the non-error return value is ignored. In fact, since [`apierrors.IsAlreadyExist`](https://github.com/kubernetes/apimachinery/blob/d16981aedf33f88b7434bd3bdbcc2e1f940aecd0/pkg/api/errors/errors.go#L465-L468) calls [`apierrors.ReasonForError`](https://github.com/kubernetes/apimachinery/blob/d16981aedf33f88b7434bd3bdbcc2e1f940aecd0/pkg/api/errors/errors.go#L616-L623) and they can both accept a `nil` error, please rewrite like so:

```golang
	if _, err := a.coreClient.Secrets(cluster.Namespace).Create(secret); !apierrors.IsAlreadyExists(err) {
		return errors.Wrapf(err, "error creating kubeconfig secret for cluster %q", ctx)
	}
```

pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -72,6 +77,28 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) (opErr error) {
 		return errors.Wrapf(err, "unable to reconcile certs while reconciling cluster %q", ctx)
 	}
 
+	// sync kubeconfig secrets for cluster
+	kubeConfig, err := a.GetKubeConfig(cluster, nil)
```
akutz: Hi @andrewsykim,

Please consider renaming `kubeConfig` to `kubeConfigForTargetCluster` to make it super descriptive.

pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -72,6 +77,28 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) (opErr error) {
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to get kubeconfig for cluster %q", ctx)
+	}
+
+	secret := &apiv1.Secret{
```
akutz: Hi @andrewsykim,

I was about to say the same thing regarding `secret` and descriptive, but maybe consider refactoring this entire block into a function `reconcileTargetClusterKubeConfigSecret` or some such thing. That way everything inside is implicitly about the target cluster, and it keeps the outer `Reconcile` function neat and tidy.

akutz:
/hold

(since editing the PR review and adding hold didn't work)

pkg/cloud/vsphere/actuators/cluster/actuator.go
andrewsykim: Not sure if `kubeConfigForTargetCluster` is accurate here since this applies for any cluster represented by the cluster resource which also includes the management cluster

pkg/cloud/vsphere/actuators/cluster/actuator.go
andrewsykim: The control plane endpoints are fetched in the cluster status though? I tested this change as well and it works as expected.

andrewsykim: The params that GetKubeConfig checks in the context only checks fields set by the initial cluster context

pkg/cloud/vsphere/actuators/cluster/actuator.go
akutz: Fair. I still think it's worth pushing this new code into a separate function (like ReconcileCertificates).

andrewsykim: Added a `reconcileKubeConfig` method

andrewsykim:
Comments addressed, PTAL

andrewsykim:
Sorry, don't review yet, forgot to add Delete code

pkg/cloud/vsphere/actuators/cluster/actuator.go
akutz: So `GetKubeConfig` [invokes](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/f5893540c99c40f46f9520070d6964b463705ea1/pkg/cloud/vsphere/services/kubeclient/kubeclient.go#L49) `ctx.ControlPlaneEndpoint()`.

When `ctx KubeContext`, as passed into `GetKubeConfig`, is a `MachineContext`, [this](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/f5893540c99c40f46f9520070d6964b463705ea1/pkg/cloud/vsphere/context/machine_context.go#L163-L182) is what gets called:

```golang
// ControlPlaneEndpoint returns the control plane endpoint for the cluster.
// This function first attempts to retrieve the control plane endpoint with
// ClusterContext.ControlPlaneEndpoint.
// If no endpoint is returned then this machine's IP address is used as the
// control plane endpoint if the machine is a control plane node.
// Otherwise an error is returned.
func (c *MachineContext) ControlPlaneEndpoint() (string, error) {


	if controlPlaneEndpoint, _ := c.ClusterContext.ControlPlaneEndpoint(); controlPlaneEndpoint != "" {
		return controlPlaneEndpoint, nil
	}


	ipAddr := c.IPAddr()
	if ipAddr == "" || !c.HasControlPlaneRole() {
		return "", errors.New("unable to get control plane endpoint")
	}
	controlPlaneEndpoint := net.JoinHostPort(ipAddr, strconv.Itoa(int(c.BindPort())))
	c.Logger.V(2).Info("got control plane endpoint from machine", "control-plane-endpoint", controlPlaneEndpoint)
	return controlPlaneEndpoint, nil
}
```

So what actually happens is the Cluster's fields are checked first, and if nothing is returned, the IP from the machine is used to build the control plane endpoint.

andrewsykim:
Delete added, PTAL

pkg/cloud/vsphere/actuators/cluster/actuator.go
akutz: I submitted the previous comment too soon.

> So what actually happens is the Cluster's fields are checked first, and if nothing is returned, the IP from the machine is used to build the control plane endpoint.

Why is this important? Well, because the Cluster's fields aren't even updated until the cluster is patched. Calling the new `reconcileKubeConfig` function before the first time a cluster is patched won't actually work.

Here's what I recommend:

* Move the logic that assigns a cluster's APIEndpoints out of the `ClusterContext` patch call and into `reconcileControlPlaneEndpoint` in the cluster actuator. Put that above the `reconcileKubeConfig` call. If the `reconcileControlPlaneEndpoint` fails, then requeue the cluster. Once `reconcileControlPlaneEndpoint` works you can then call `reconcileKubeConfig`.

Also, please leave the `GetKubeConfig` function the way it was as it is consistent with the deployer, and the two context structs are designed to work with it.

andrewsykim: Ah I see.. and if we use cluster context then a random machine in the control plane will be used. Any ideas on how to pass a machine context without any references to machines from `Reconcile`? We can list machines and pick one but that wouldn't be much different from what cluster context does with it's `ControlPlaneEndpoint`

pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -93,6 +102,10 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster) (opErr error) {
 
 	ctx.Logger.V(2).Info("deleting cluster")
 
+	if err := a.deleteKubeConfig(ctx); err != nil {
+		return errors.Wrapf(err, "error deleting kubeconfig for cluster %q", ctx)
```
akutz: Hi @andrewsykim,

Should this be a requeue error? I'm not sure, which is why I'm asking.

pkg/cloud/vsphere/actuators/cluster/actuator.go
akutz: > then a random machine in the control plane will be used

Not random -- the oldest control plane node is always used.

akutz: I know we resolved this via Zoom, but I wanted to clarify the above in case someone else reads this :)

pkg/cloud/vsphere/actuators/cluster/actuator.go
andrewsykim: Realizing though that `!apierrors.IsAlreadyExists(err)` returns true for `err == nil` in which case we don't want to return err.

andrewsykim: Ah, can still collapse as `err != nil && !apierrors.IsAlreadyExists(err)` :)

PR #342, opened 2019-06-24
Fixes failed JSON patches

akutz:
This patch fixes an issue with the way patching an object with a JSON patch would fail when the status was modified first.

Since cluster and machine objects have their status values changed during any given call stack, the current patch model sometimes fails because JSONPatching either object type fails when status is included. This is also why the status is patched separately later.

This PR fixes things by ensuring the status of either object is the same as it was when the call was enqueued ahead of generating the JSON patch. After the patch the status is put back ahead of the status patch.

Please see [this Slack thread](https://kubernetes.slack.com/archives/C8TSNPY4T/p1561400433201000) for more information.

**Which issue(s) this PR fixes**: NA

### DIFF
--- a/pkg/cloud/vsphere/context/cluster_context.go
+++ b/pkg/cloud/vsphere/context/cluster_context.go
@@ -301,21 +301,19 @@ func (c *ClusterContext) ControlPlaneEndpoint() (string, error) {
 
 // Patch updates the cluster on the API server.
 func (c *ClusterContext) Patch() error {
-	ext, err := v1alpha1.EncodeClusterSpec(c.ClusterConfig)
+	// Ensure the provider spec is encoded.
+	newProviderSpec, err := v1alpha1.EncodeClusterSpec(c.ClusterConfig)
 	if err != nil {
 		return errors.Wrapf(err, "failed encoding cluster spec for cluster %q", c)
 	}
-	newStatus, err := v1alpha1.EncodeClusterStatus(c.ClusterStatus)
-	if err != nil {
-		return errors.Wrapf(err, "failed encoding cluster status for cluster %q", c)
-	}
-	ext.Object = nil
-	newStatus.Object = nil
+	c.Cluster.Spec.ProviderSpec.Value = newProviderSpec
 
-	c.Cluster.Spec.ProviderSpec.Value = ext
+	// Make sure the status isn't part of the JSON patch.
+	newStatus := c.Cluster.Status.DeepCopy()
+	c.Cluster.Status = clusterv1.ClusterStatus{}
+	c.ClusterCopy.Status.DeepCopyInto(&c.Cluster.Status)
 
-	// Build a patch and marshal that patch to something the client will
-	// understand.
+	// Build and marshal a patch for the cluster object, minus the status.
 	p, err := patch.NewJSONPatch(c.ClusterCopy, c.Cluster)
 	if err != nil {
 		return errors.Wrapf(err, "failed to create new JSONPatch for cluster %q", c)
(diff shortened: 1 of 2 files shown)


akutz:
Hi @andrewsykim,

Just waiting on this for the alpha. Please take a look when you get a chance.

pkg/cloud/vsphere/context/cluster_context.go
```diff
@@ -344,7 +341,10 @@ func (c *ClusterContext) Patch() error {
 		c.Cluster.ResourceVersion = result.ResourceVersion
 	}
 
-	c.Cluster.Status.ProviderStatus = newStatus
+	// Put the status back.
+	c.Cluster.Status = clusterv1.ClusterStatus{}
```
chuckha: I was expecting to see `result` on line 332 get its status updated (and if there was no patch, then you could use the existing object) and have that saved.

But if it works like this then that's 💯

pkg/cloud/vsphere/context/cluster_context.go
```diff
@@ -344,7 +341,10 @@ func (c *ClusterContext) Patch() error {
 		c.Cluster.ResourceVersion = result.ResourceVersion
```
sidharthsurana: shouldn't we simply do
`c.Cluster = result`
That way `c.Cluster` always points to the latest persisted version?

pkg/cloud/vsphere/context/machine_context.go
```diff
@@ -266,7 +263,9 @@ func (c *MachineContext) Patch() error {
 		c.Machine.ResourceVersion = result.ResourceVersion
```
sidharthsurana: Same comment as above.

pkg/cloud/vsphere/context/cluster_context.go
akutz: Hi @sidharthsurana,

Patching is the last thing that happens, so the RV only matters for the status update really. And besides, the RV *is* updated on the the existing cluster. It's also what CAPA does FWIW.

pkg/cloud/vsphere/context/cluster_context.go
akutz: Hi @chuckha,

Hmm, we just followed the CAPA model at https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/ace9cb7b1ab5b3d5ced5663e44656ac94ad78b2c/pkg/cloud/aws/actuators/scope.go#L204. Maybe there was a reason for it?

pkg/cloud/vsphere/context/cluster_context.go
sidharthsurana: True, but my only concern is that are we making an assumption that only this code is going to be the one that will ever mutate the Cluster or Machine object?
In the future, there may be more controller looking and possibly mutating the same object. Thus I feel in a probably very slim chance but we may overwrite someone else's changes done via outside this controller.

akutz: No, this will always be the last code per the defer. If it's not we can revisit. Either way, the resource version *is* updated. :) I just don't want to make that change here.

akutz: That's right, if the patch fails, then we return an error. Otherwise, if the patch succeeds, then the cluster object *is* the same as the one on the server.

sidharthsurana: I guess for now it is okay since there are no other additional controllers mutating these objects.
If I think about it again, the risk here is probably not that this method doesn't successfully update the server object. But rather it is that after this update the `c.Cluster` object may not capture the most current persisted object.

PR #343, opened 2019-06-24
re-enable in-tree vsphere cloud provider

andrewsykim:

In https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/307 we temporarily removed the vSphere cloud provider, this PR adds it back using the updated custom kubeadm config mechanism. 

Fixes #

**Release note**:

```release-note
Re-enable the vSphere cloud provider. 
```

/assign @akutz

### DIFF
--- a/pkg/cloud/vsphere/services/govmomi/create.go
+++ b/pkg/cloud/vsphere/services/govmomi/create.go
@@ -44,6 +44,12 @@ const (
 
 	// nodeRole is the label assigned to every node in the cluster.
 	nodeRole = "node-role.kubernetes.io/node="
+
+	// the Kubernetes cloud provider to use
+	cloudProvider = "vsphere"
+
+	// the cloud config path read by the cloud provider
+	cloudConfigPath = "/etc/kubernetes/vsphere.conf"
 )
 
 // Create creates a new machine.
@@ -91,6 +97,19 @@ func generateUserData(ctx *context.MachineContext, bootstrapToken string) ([]byt
 
 	// apply values based on the role of the machine
 	if ctx.HasControlPlaneRole() {
+		cloudConfig, err := userdata.NewCloudConfig(&userdata.CloudConfigInput{
+			User:         ctx.ClusterConfig.VsphereUser,
+			Password:     ctx.ClusterConfig.VspherePassword,
+			Server:       ctx.ClusterConfig.VsphereServer,
+			Datacenter:   ctx.MachineConfig.MachineSpec.Datacenter,
+			ResourcePool: ctx.MachineConfig.MachineSpec.ResourcePool,
+			Datastore:    ctx.MachineConfig.MachineSpec.Datastore,
+			// assume the first VM network found for the vSphere cloud provider
+			Network: ctx.MachineConfig.MachineSpec.Network.Devices[0].NetworkName,
+		})
+		if err != nil {
+			return nil, err
+		}
 
 		if bootstrapToken != "" {
 			ctx.Logger.V(2).Info("allowing a machine to join the control plane")
@@ -111,7 +130,7 @@ func generateUserData(ctx *context.MachineContext, bootstrapToken string) ([]byt
 						kubeadm.WithTaints(ctx.Machine.Spec.Taints),
 						kubeadm.WithNodeRegistrationName(hostnameLookup),
 						kubeadm.WithCRISocket(containerdSocket),
-						//kubeadm.WithKubeletExtraArgs(map[string]string{"cloud-provider": cloudProvider}),
+						kubeadm.WithKubeletExtraArgs(map[string]string{"cloud-provider": cloudProvider}),
 					),
 				),
 				kubeadm.WithLocalAPIEndpointAndPort(localIPV4Lookup, int(bindPort)),
@@ -131,6 +150,7 @@ func generateUserData(ctx *context.MachineContext, bootstrapToken string) ([]byt
 				FrontProxyCAKey:   string(ctx.ClusterConfig.FrontProxyCAKeyPair.Key),
 				SaCert:            string(ctx.ClusterConfig.SAKeyPair.Cert),
 				SaKey:             string(ctx.ClusterConfig.SAKeyPair.Key),
+				CloudConfig:       cloudConfig,
 				JoinConfiguration: joinConfigurationYAML,
 			})
 			if err != nil {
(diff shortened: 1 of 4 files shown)


akutz:
Hi @andrewsykim,

When I was going to do this I was going to add a new field to the CRD called "ExtraFiles" with some version of the cloud-init file schema. I was going to use that to add the CCM config. The reason is because it could also be used to add support for writing custom CAs to the machines that need them to access the vSphere endpoint.

pkg/cloud/vsphere/services/userdata/controlplane.go
```diff
@@ -24,6 +24,28 @@ import (
 )
 
 const (
+	cloudConfig = `[Global]
+insecure-flag = "1" # set to 1 if the vCenter uses a self-signed cert
+datacenters = "{{ .Datacenter }}"
+
+[VirtualCenter "{{ .Server }}"]
+user = "{{ .User }}"
```
akutz: Hi @andrewsykim,

I really don't want to send the user and pass via cloud-init in user data. Please set the Kubernetes secret name for the cloud provider credentials and then add that secret to the target cluster.

akutz: This [doc](https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html#supported-parameters) discusses the params for pointing to the secret and the expected format, etc.

andrewsykim:
Verification:

## Nodes
$ kubectl get no
NAME                                  STATUS   ROLES    AGE   VERSION
workload-example-controlplane-1       Ready    master   36m   v1.13.6
workload-example-machineset-1-8576d   Ready    node     13m   v1.13.6
workload-example-machineset-1-m8vpd   Ready    node     12m   v1.13.6

## Nodes has `spec.ProviderID` and `NodeAddresses` set
```
$ kubectl get nodes workload-example-machineset-1-8576d -o yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    kubeadm.alpha.kubernetes.io/cri-socket: /var/run/containerd/containerd.sock
    node.alpha.kubernetes.io/ttl: "0"
    projectcalico.org/IPv4Address: 192.168.3.79/24
    projectcalico.org/IPv4IPIPTunnelAddr: 100.101.96.192
    volumes.kubernetes.io/controller-managed-attach-detach: "true"
  creationTimestamp: "2019-06-25T17:41:07Z"
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/os: linux
    kubernetes.io/hostname: workload-example-machineset-1-8576d
    node-role.kubernetes.io/node: ""
  name: workload-example-machineset-1-8576d
  resourceVersion: "3356"
  selfLink: /api/v1/nodes/workload-example-machineset-1-8576d
  uid: 69ba186e-9770-11e9-836c-005056b0c6e3
spec:
  podCIDR: 100.96.2.0/24
  providerID: vsphere://4230fc1e-c619-c04d-b4d5-df19d147734d
status:
  addresses:
  - address: 192.168.3.79
    type: ExternalIP
  - address: 192.168.3.79
    type: InternalIP
  - address: workload-example-machineset-1-8576d
    type: Hostname
  allocatable:
```

### Delete a VM 
```
$ govc vm.destroy workload-example-machineset-1-8576d
```

### Node is deleted in Kubenetse
```
$ kubectl get no
NAME                                  STATUS   ROLES    AGE   VERSION
workload-example-controlplane-1       Ready    master   37m   v1.13.6
workload-example-machineset-1-m8vpd   Ready    node     14m   v1.13.6
```

pkg/cloud/vsphere/services/govmomi/create.go
```diff
@@ -91,6 +97,23 @@ func generateUserData(ctx *context.MachineContext, bootstrapToken string) ([]byt
 
 	// apply values based on the role of the machine
 	if ctx.HasControlPlaneRole() {
+		if len(ctx.MachineConfig.MachineSpec.Network.Devices) == 0 {
+			return nil, errors.New("machine spec has no network devices")
```
akutz: Hi @andrewsykim,

Let's move this to the machine actuator and make it a dependency before govmomi create is ever called.

andrewsykim: sgtm

andrewsykim: We don't do validation on any field, should we just address this in a follow-up PR?

andrewsykim: I only added it initially since it's the only place that could cause a panic here

pkg/cloud/vsphere/services/govmomi/create.go
```diff
@@ -91,6 +97,23 @@ func generateUserData(ctx *context.MachineContext, bootstrapToken string) ([]byt
...
+
+		cloudConfig, err := userdata.NewCloudConfig(&userdata.CloudConfigInput{
+			User:         ctx.ClusterConfig.VsphereUser,
+			Password:     ctx.ClusterConfig.VspherePassword,
+			Server:       ctx.ClusterConfig.VsphereServer,
+			Datacenter:   ctx.MachineConfig.MachineSpec.Datacenter,
+			ResourcePool: ctx.MachineConfig.MachineSpec.ResourcePool,
+			Datastore:    ctx.MachineConfig.MachineSpec.Datastore,
+			// assume the first VM network found for the vSphere cloud provider
+			Network: ctx.MachineConfig.MachineSpec.Network.Devices[0].NetworkName,
```
akutz: This is fine for now. We need to clearly document this though.

pkg/cloud/vsphere/services/govmomi/create.go
akutz: > We don't do validation on any field, should we just address this in a follow-up PR?

sgtm

pkg/cloud/vsphere/services/userdata/controlplane.go
```diff
@@ -209,9 +246,20 @@ type ContolPlaneJoinInput struct {
 	SaKey             string
 	BootstrapToken    string
 	ELBAddress        string
+	CloudConfig       string
 	JoinConfiguration string
 }
 
+type CloudConfigInput struct {
```
akutz: We need to add a comment to this exported type. I don't want to make reconciling things to where we can enable listing worse. However, since this is passing CI for now, let's get back to it. I don't want to chance Prow today.

pkg/cloud/vsphere/services/userdata/controlplane.go
```diff
@@ -292,6 +342,17 @@ func JoinControlPlane(input *ContolPlaneJoinInput) (string, error) {
 	return userData, err
 }
 
+func NewCloudConfig(input *CloudConfigInput) (string, error) {
```
akutz: I'm curious why this is its own function as opposed to part of the cloud init user data func. Just to make testing easier?

andrewsykim: eyah

akutz: I saw that 'eyah'. I thought you were pulling a Howard Dean and excitedly answering :)

andrewsykim: haha, btw I forgot to add comments for this exported func, just pushed a change

pkg/cloud/vsphere/services/userdata/controlplane.go
```diff
@@ -24,6 +24,28 @@ import (
...
+datacenters = "{{ .Datacenter }}"
+
+[VirtualCenter "{{ .Server }}"]
+user = "{{ .User }}"
+password = "{{ .Password }}"
+
+[Workspace]
+server = "{{ .Server }}"
+datacenter = "{{ .Datacenter }}"
+folder = "{{ .ResourcePool }}"
```
mylesagray: This actually needs to be a "VM Folder" - not the same as the Resource Pool: https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html#supported-parameters

So for example, one might place running VMs in the RP `cluster01/Resources` but the "VM Folder" they live in is purely organisation and could be something like `Testing/K8s/cluster-name`.

andrewsykim: The cloud config here was taken from what was set on CAPV previously https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/0ec7a864653df44316d6747601d8d2927cb0f3ea/pkg/cloud/vsphere/provisioner/common/templates.go#L311-L331 which has the resource pool as the folder, I'm just going to assume that was also a bug, will open a bug shortly, thanks Myles!

PR #352, opened 2019-06-25
Create simple make target for container images

codenrhoden:
The previous make targets prod-images, prod-push, ci-images, and ci-push
essentially did the same thing for prod vs CI. The prod targets were
never meant to be invoked by CI and used a hard-coded image version that
needed to be updated whenever a new tag was made. The CI targets are
invoked on every PR by CI, and pushed a new CAPV and clusterctl image
tagged with the git SHA.

Instead, we now combine them into a single set of targets, build-images
and push-images, that set the image version to a simpler version of `git
describe` that will always show the changes since the last annotated
tag. This conveys more meaning than a simple SHA, and the git hash is
still included. With that done, we can then just always build and push
images in the same fashion, with no distinction between CI and
Production. This will allow the CI system to build images for us both as
part of CI/testing, and post merge to master.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,15 +26,17 @@ CWD := $(abspath .)
 #      nested GOPATH.
 SHELL := hack/shell-with-gopath.sh
 
-# Image URL to use all building/pushing image targets
-PRODUCTION_IMG ?= gcr.io/cnx-cluster-api/vsphere-cluster-api-provider:0.3.0-alpha.1
-CI_IMG ?= gcr.io/cnx-cluster-api/vsphere-cluster-api-provider
-CLUSTERCTL_CI_IMG ?= gcr.io/cnx-cluster-api/clusterctl
+# Image URLs to use for building/pushing image targets
+CAPV_IMG ?= gcr.io/cnx-cluster-api/vsphere-cluster-api-provider
+CLUSTERCTL_IMG ?= gcr.io/cnx-cluster-api/clusterctl
 DEV_IMG ?= # <== NOTE:  outside dev, change this!!!
 
 # Retrieves the git hash
-VERSION ?= $(shell git describe --exact-match 2> /dev/null || \
-	   git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
+VERSION ?= $(shell git describe --always --dirty --abbrev=8)
+
+CAPV_IMG_VERSION := $(CAPV_IMG):$(VERSION)
+CLUSTERCTL_IMG_VERSION := $(CLUSTERCTL_IMG):$(VERSION)
+CAPV_PROD_IMG ?= gcr.io/cnx-cluster-api/vsphere-cluster-api-provider:0.3.0-alpha.1
 
 # Build manager binary
 manager: fmt vet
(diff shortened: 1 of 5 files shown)


codenrhoden:
/assign @akutz 
/assign @andrewsykim 

This PR would allow us to add a Prow postsubmit job that builds and publishes new CAPV controller and clusterctl image. It also makes it such that the container image version is going to be along the lines of `v0.3.0-alpha.0-11-g891b92bc` instead of a simple git hash each time, even for CI builds.

A couple things to note -- this does not make any changes to the `cluster-api-provider-vsphere-ci` image that is **also** built as part of the e2e tests. This also doesn't do anything about deleting the CI images that are build after a PR test is run. That should be something we tackle in the future.  There isn't much point in keeping those around, and the registry where we upload those to has 700+ images in it that will never be used again.

codenrhoden:
Also, I kept the `prod-yaml` target that uses a hard-coded reference to the current "release" of CAPV so that it can be used to generate valid YAML against our current release. I'm not sure of a better way to handle that just yet, but @akutz may have some idea. I don't like the idea of having to change a hardcoded version in the Makefile every time a release is made.

akutz:
Hi @codenrhoden,

So one thing we discussed is changing the image to just be `:latest` since we're not tagging the CI images as latest. That's what #367 does in the script.

codenrhoden:
@akutz, can you clarify a little on what you want to see re: `latest`?

Right now, we do tag the CI-built images with the SHA of the PR. This SHA is not going to match what ends up being present on master after the PR merges. This PR would change that tag from being a SHA to being something akin to `v0.3.0-alpha.0-11-g891b92bc`, but that 8-char SHA at the end still would not match the SHA on master.

This is in prep to make a postsubmit job in Prow that would build and push images *again* after the PR merges, tagged with the same method as above. Are you saying that you would want the tag of `latest` only applied to what results from the postsubmit? That would make sense to me -- I wouldn't want to tag the PR/CI images tagged `latest`. Additionally, for the `postsubmit` built images, I think it would be good to tag the image with both `latest` and the version, so that when `latest` is moved later there is still a tag on the image.

I guess, ideally, the workflow would be like this:

CI-presubmit:

- build and push image tagged with `git describe --always --dirty --abbrev=8`
- run E2E tests
- delete CI image

postsubmit:

 - build and push image tagged with `latest` **and** `git describe --always --dirty --abbrev=8`

If you agree with that approach, I should make a few more changes...
/hold

akutz:
Hi @codenrhoden,

So what I keep forgetting to reconcile is that the Makefile value for the image is used both as the thing to build, push, *and* as the thing used to generate the yaml. That's why I moved the image as `:latest` to the `hack/generate-yaml.sh` file. I think we should get rid of the `*-yaml` targets in the Makefile entirely. We can leave the image prefix, and set the version to be `$(VERSION)` when building and pushing post-submits as well as a push to `latest` from the same build (during the post-submit).

The problem is that the image in the Makefile is also used by people generating yaml.

Let's change that. Let's move all the YAML gen out of the Makefile and make the docs say to call `hack/generate-yaml.sh` directly, and if you want a version of the image other than `latest` then you must use the `-m` flag.

Thoughts?

codenrhoden:
Hi @akutz,

I definitely agree.

For anyone following, I had an offline conversation with @akutz where we agreed on an approach around all this. I'm going to write up the broader release scope for how we deal with CI images from PRs, builds from master with `postsubmits`, tagging, image tagging for releases and `latest`, etc.

In the meantime, though, a couple immediate first steps are to build off of #367 and update the docs and Makefile as Andrew suggested to just always use the script directly for generation of YAML. Then, with a simplified Makefile, add a script that can be called by CI that itself calls the simplified Makefile to do exactly what is happening with CI today.

codenrhoden:
This work is heading a different direction and will look differently. No need to keep this around.

PR #355, opened 2019-06-25
Modify -ci image makefile to default to build only

codenrhoden:
When running `make` with no arguments in scripts/e2e/bootstrap_job, the
default behavior was to build the image, push the image to gcr.io, and
delete the local copy of the image. That's not a useful default.
Instead, change the "all" target to just have "build" as a pre-req,
making "make" the equivalent of "make build". If someone wants to push
images, they can run "make push" instead. Deleting the local image is an
explicit step left to "make clean".

Fixes #

**Release note**:

```release-note

```

/hold

### DIFF
--- a/scripts/e2e/bootstrap_job/Makefile
+++ b/scripts/e2e/bootstrap_job/Makefile
@@ -4,7 +4,7 @@ VERSION ?= $(shell git describe --exact-match 2> /dev/null || \
                  git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
 REGISTRY ?=gcr.io/cnx-cluster-api/cluster-api-provider-vsphere-ci
 
-all: build push clean
+all: build
 .PHONY : all
 
 .PHONY : copyspec
@@ -14,7 +14,7 @@ copyspec:
 .PHONY : build
 build: copyspec
 	docker build . --tag $(REGISTRY):$(VERSION)
-        
+
 push: build
 	@echo "logging into gcr.io registry with key file"
 	@echo $$GCR_KEY_FILE | docker login -u _json_key --password-stdin gcr.io
--- a/scripts/e2e/e2e.sh
+++ b/scripts/e2e/e2e.sh
@@ -156,7 +156,7 @@ fi
 
 export VERSION="${vsphere_controller_version}"
 make ci-push
-cd ./scripts/e2e/bootstrap_job && make && cd .. || exit 1
+cd ./scripts/e2e/bootstrap_job && make push && cd .. || exit 1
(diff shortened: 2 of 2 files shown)


codenrhoden:
So this is really interesting. Even though the image is "delete", it still shows up in gcr.io and I can pull it directly with `docker pull`. Yet the logs show: 
>docker image rm -f gcr.io/cnx-cluster-api/cluster-api-provider-vsphere-ci:bae68e6d1987441d243fac288cad062a15054b06
Untagged: gcr.io/cnx-cluster-api/cluster-api-provider-vsphere-ci:bae68e6d1987441d243fac288cad062a15054b06

I wonder if something with the GCR_KEY file associated perms doesn't allow delete?

Regardless, I think this change makes more sense. The only reason these tests are working, I believe, is because deletes aren't working *at all*.

akutz:
> So this is really interesting. Even though the image is "delete", it still shows up in gcr.io and I can pull it directly with `docker pull`. Yet the logs show:
> 
> > docker image rm -f gcr.io/cnx-cluster-api/cluster-api-provider-vsphere-ci:bae68e6d1987441d243fac288cad062a15054b06
> > Untagged: gcr.io/cnx-cluster-api/cluster-api-provider-vsphere-ci:bae68e6d1987441d243fac288cad062a15054b06
> 
> I wonder if something with the GCR_KEY file associated perms doesn't allow delete?
> 
> Regardless, I think this change makes more sense. The only reason these tests are working, I believe, is because deletes aren't working _at all_.

Wow. Feel free to increase the perms on the key being used.

codenrhoden:
Still not quite sure what is going on here... The roles seem to be setup directly, and the docs indicate this should work. I'll have to try this out manually with my own user/role and see if a `docker image rm` command works again gcr.io. I couldn't find any docs about how well `docker` commands work against gcr.io. All the docs reference `gcloud` commands.

codenrhoden:
Okay, just leaving some notes for myself as I was being silly. I had it in my head that job was trying to delete the image from the registry, so with that assumption in my mind I thought it wasn't working...

`docker rm` doesn't remove images from a remote registry, only local. I knew that, but completely forgot. :) So the `make clean` target here was really just a convenience method to help local devs cleanup, and not meant to remove the the image from the gcr.io registry.

That being said, I don't think we should have `make` default to delete something that was just made, so there are still changes to make...

Furthermore, we really don't need the CI image to hang around longer than the PR. I still think we should delete the image, and that requires `gcloud`. I'll open an issue later to cover that.

codenrhoden:
@akutz this should be good to `/approve`. Though if you want to wait until after the release of of v0.3.0, that makes sense too.

PR #356, opened 2019-06-25
Net devices by order, remove invalid IP addrs

akutz:
This PR removes the UUID used to track the network device specs to network device mappings. Now the order of the devices added is assumed  to be the order of the devices retrieved later. This was confirmed via the test cases in the following gists:

* https://gist.github.com/dougm/6061dcf7bf44cc21fae185143923f1fa
* https://gist.github.com/dougm/6061dcf7bf44cc21fae185143923f1fa/revisions

This PR also removes any IP addresses from the list returned from a VM if the addresses are loopback, link local unicast, link local multicast, invalid, or unspecified.

**Which issue(s) this PR fixes**: NA

**Release note**:
```release-note
NONE
```

cc @dougm 
/assign @andrewsykim

### DIFF
--- a/pkg/cloud/vsphere/services/govmomi/net/net.go
+++ b/pkg/cloud/vsphere/services/govmomi/net/net.go
@@ -33,11 +33,6 @@ type NetworkStatus struct {
 	// connected to the VM.
 	Connected bool `json:"connected,omitempty"`
 
-	// UUID is stored as the ExternalID field on a network device and uniquely
-	// identifies the device as one that was created from a known network
-	// spec.
-	UUID string `json:"uuid"`
-
 	// IPAddrs is one or more IP addresses reported by vm-tools.
 	// +optional
 	IPAddrs []string `json:"ipAddrs,omitempty"`
@@ -80,7 +75,6 @@ func GetNetworkStatus(
 			nic := dev.GetVirtualEthernetCard()
 			netStatus := NetworkStatus{
 				MACAddr: nic.MacAddress,
-				UUID:    nic.ExternalId,
 			}
 			if obj.Guest != nil {
 				for _, i := range obj.Guest.Net {
--- a/pkg/cloud/vsphere/services/govmomi/net/net.go
+++ b/pkg/cloud/vsphere/services/govmomi/net/net.go
@@ -18,6 +18,7 @@ package net
 
 import (
 	"context"
+	"net"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -90,4 +91,26 @@ func GetNetworkStatus(
 	}
 
 	return allNetStatus, nil
+}
+
+// IsExternalIPAddr returns a nil error if the provided IP address is valid and
+// can be used externally to the VM.
+func IsExternalIPAddr(addr string) error {
+	var reason string
+	a := net.ParseIP(addr)
+	if a == nil {
+		reason = "invalid"
+	} else if a.IsUnspecified() {
+		reason = "unspecified"
+	} else if a.IsLinkLocalMulticast() {
+		reason = "link-local-mutlicast"
+	} else if a.IsLinkLocalUnicast() {
+		reason = "link-local-unicast"
+	} else if a.IsLoopback() {
+		reason = "loopback"
+	}
+	if reason != "" {
+		return errors.Errorf("failed to validate ip addr=%v: %s", addr, reason)
+	}
+	return nil
 }
--- a/pkg/cloud/vsphere/services/govmomi/net/net.go
+++ b/pkg/cloud/vsphere/services/govmomi/net/net.go
@@ -93,9 +93,9 @@ func GetNetworkStatus(
 	return allNetStatus, nil
 }
 
-// IsExternalIPAddr returns a nil error if the provided IP address is valid and
-// can be used externally to the VM.
-func IsExternalIPAddr(addr string) error {
+// ErrOnLocalOnlyIPAddr returns an error if the provided IP address is
+// accessible only on the VM's guest OS.
+func ErrOnLocalOnlyIPAddr(addr string) error {
 	var reason string
 	a := net.ParseIP(addr)
 	if a == nil {
--- a/pkg/cloud/vsphere/services/govmomi/net/net_test.go
+++ b/pkg/cloud/vsphere/services/govmomi/net/net_test.go
@@ -22,7 +22,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/services/govmomi/net"
 )
 
-func TestIsExternalIPAddr(t *testing.T) {
+func TestErrOnLocalOnlyIPAddr(t *testing.T) {
 	testCases := []struct {
 		name      string
 		ipAddr    string
@@ -66,7 +66,7 @@ func TestIsExternalIPAddr(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			if err := net.IsExternalIPAddr(tc.ipAddr); err != nil {
+			if err := net.ErrOnLocalOnlyIPAddr(tc.ipAddr); err != nil {
 				t.Log(err)
 				if !tc.expectErr {
 					t.Fatal(err)
--- a/pkg/cloud/vsphere/services/govmomi/net/net_test.go
+++ b/pkg/cloud/vsphere/services/govmomi/net/net_test.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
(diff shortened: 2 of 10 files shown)


pkg/cloud/vsphere/services/govmomi/net/net.go
```diff
@@ -97,3 +92,25 @@ func GetNetworkStatus(
 
 	return allNetStatus, nil
 }
+
+// IsExternalIPAddr returns a nil error if the provided IP address is valid and
+// can be used externally to the VM.
+func IsExternalIPAddr(addr string) error {
```
andrewsykim: `IsExternalIPAddr` can be confused with publically external IP, prefer something like "IsNonLocalIPAddr", thoughts?

pkg/cloud/vsphere/services/govmomi/net/net_test.go
```diff
@@ -0,0 +1,79 @@
...
+		{
+			name:      "link-local-multicast-ipv6",
+			ipAddr:    "FF02:0:0:0:0:0:1:3",
+			expectErr: true,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if err := net.IsExternalIPAddr(tc.ipAddr); err != nil {
+				t.Log(err)
```
andrewsykim: can remove this since there's `t.Fatal` below

pkg/cloud/vsphere/services/govmomi/update.go
```diff
@@ -108,19 +108,14 @@ func reconcileNetworkStatus(ctx *context.MachineContext, vm *object.VirtualMachi
...
 	// Update the MAC addresses in the machine's network config as well. This
 	// is required in order to generate the metadata.
-	for _, netStatus := range allNetStatus {
-		for i := range ctx.MachineConfig.MachineSpec.Network.Devices {
-			dev := &ctx.MachineConfig.MachineSpec.Network.Devices[i]
-			if netStatus.UUID == dev.UUID && netStatus.MACAddr != dev.MACAddr {
-				ctx.Logger.V(4).Info("updating MAC address for device", "device-uuid", dev.UUID, "network-name", dev.NetworkName, "old-mac-addr", dev.MACAddr, "new-mac-addr", netStatus.MACAddr)
-				dev.MACAddr = netStatus.MACAddr
-			}
+	for i := range ctx.MachineConfig.MachineSpec.Network.Devices {
```
andrewsykim: Wondering if at the very least we should log warning if the two lengths don't match?

pkg/cloud/vsphere/services/govmomi/util.go
```diff
@@ -95,22 +95,32 @@ func getNetworkStatus(ctx *context.MachineContext) ([]v1alpha1.NetworkStatus, er
...
 	}
 	return apiNetStatus, nil
 }
 
+func sanitizeIPAddrs(ctx *context.MachineContext, ipAddrs []string) []string {
+	if len(ipAddrs) == 0 {
+		return nil
+	}
+	newIPAddrs := []string{}
+	for _, i := range ipAddrs {
```
andrewsykim: s/i/ipAddr

pkg/cloud/vsphere/services/govmomi/vcenter/clone.go
```diff
@@ -187,16 +186,16 @@ func getNetworkSpecs(
 		// "types.BaseVirtualEthernetCard" as a "types.BaseVirtualDevice".
 		nic := dev.(types.BaseVirtualEthernetCard).GetVirtualEthernetCard()
 
+		if netSpec.MACAddr != "" {
+			nic.MacAddress = netSpec.MACAddr
+			nic.AddressType = "Manual"
```
andrewsykim: Is this a recognized address type by vSphere or arbitrary?

pkg/cloud/vsphere/services/govmomi/net/net.go
sidharthsurana: Nit:
If possible can we name this method differently?
like `IsNonLocalIPAddr(addr string)`

`External IP` seems to have a different meaning in my mind and may cause confusion to some.

akutz: Hi @andrewsykim,

Maybe, but the docs are pretty clear I think. If I have to push a code change, I'll consider a name change.

pkg/cloud/vsphere/services/govmomi/net/net_test.go
akutz: Hi @andrewsykim,

Not when I tried it. It showed both.

pkg/cloud/vsphere/services/govmomi/update.go
akutz: Hi @andrewsykim,

We log that above. In fact, we return an error.

pkg/cloud/vsphere/services/govmomi/util.go
akutz: Hi @andrewsykim,

I used `i` instead of `ipAddr` because of the name of the parameter. It's a short function, and I don't think there's much room for confusion. Again, if I update code for some reason I'll change it to `addr`.

pkg/cloud/vsphere/services/govmomi/vcenter/clone.go
akutz: Hi @andrewsykim,

It's one of [three, valid values](https://www.vmware.com/support/developer/converter-sdk/conv60_apireference/vim.vm.device.VirtualEthernetCard.html#addressType).

pkg/cloud/vsphere/services/govmomi/net/net_test.go
akutz: image

akutz: FWIW, I thought so too, but it's different in the test case pattern.

pkg/cloud/vsphere/services/govmomi/update.go
andrewsykim: ah, missed it thank you

pkg/cloud/vsphere/services/govmomi/net/net.go
andrewsykim: sure

akutz: Hi @sidharthsurana,

Please see above. The meaning is documented in the function's GoDoc.

akutz: Hi @andrewsykim and @sidharthsurana,

FWIW, I originally called this `IsNonLinkLocal`, but I test for other things too. Hence the flip on the name.

akutz: Hi @andrewsykim and @sidharthsurana,

Please check now and see what you think.

pkg/cloud/vsphere/services/govmomi/vcenter/clone.go
akutz: Hi @andrewsykim,

Please check now to see if this helps.

dougm: Note that there is a generated constant you can use: `types.VirtualEthernetCardMacTypeManual`

pkg/cloud/vsphere/services/govmomi/util.go
akutz: Hi @andrewsykim,

Updated.

pkg/cloud/vsphere/services/govmomi/net/net_test.go
akutz: Hi @andrewsykim,

I'd like to resolve this. Want to verify I'm not crazy here?

pkg/cloud/vsphere/services/govmomi/net/net.go
akutz: Hi @andrewsykim and @sidharthsurana,

It's been renamed.

pkg/cloud/vsphere/services/govmomi/net/net_test.go
dougm: Yeah it'll print both.  Looks like @akutz just wants to always Log() the error and only Fatal() when the test expects an error.

pkg/cloud/vsphere/services/govmomi/update.go
andrewsykim: ha, I reported this bug so I should have known this 😆

pkg/cloud/vsphere/services/govmomi/vcenter/clone.go
andrewsykim: ^ a good TODO for @james65535

PR #367, opened 2019-06-26
Add new flags to generate-yaml

### LINKED ISSUE #359: QoL: Have generate-yaml.sh generate folders based on cluster name and type
/kind feature

I would like generate-yaml.sh (any maybe, by extension - it's invocation method) to take a cluster name and type argument.

Valid types would be:

`prod, dev, ci`

Name would simply be the name applied to that K8s cluster when deployed, whether it's management or workload.

As an example: 

`hack/generate-yaml.sh -t ci -n workload-cluster-1`

This would output to the dir `./out/ci/workload-cluster-1`

Another example:

`hack/generate-yaml.sh -t prod -n mgmt-cluster-1`

This would output to the dir `./out/prod/mgmt-cluster-1`

The first step would be to introduce these flags into `generate-yaml.sh` but ideally, whatever method is used to call it (currently `make prod-yaml`) would take this argument, I don't think the `Makefile` is ideal here, maybe it `generate-yaml.sh` just has another abstraction script or binary in front of it that handles image build and pushing along with handing off `yaml` generation.

**Environment:**

- Cluster-api-provider-vsphere version: 0.3.0-alpha.1
- Kubernetes version: (use `kubectl version`): 1.14.3
- OS (e.g. from `/etc/os-release`): macOS

akutz:
This patch adds new flags to the generate-yaml script.

* -c can be used to specify the name of the cluster
* -m is used to specify the name of the CAPV manager image

If -c is specified then the output directory becomes "./out/CLUSTER_NAME". If -o OUT_DIR is specified then the output directory is "OUT_DIR". Use -o to specify explicit output directories, otherwise the directories are unique based on the name of the cluster.

Fixes #359 

**Release note**:

```release-note
* Adds -c CLUSTER_NAME and -m CAPV_MANAGER_IMAGE flags to generate-yaml.sh
```

### DIFF
--- a/hack/generate-yaml.sh
+++ b/hack/generate-yaml.sh
@@ -22,11 +22,13 @@ set -o pipefail
 # script is located.
 cd "$(dirname "${BASH_SOURCE[0]}")/.."
 
-OUT_DIR=./out
+OUT_DIR=
 TPL_DIR=./cmd/clusterctl/examples/vsphere
 
 OVERWRITE=
+CLUSTER_NAME='capv-mgmt-example'
 ENV_VAR_REQ=':?required'
+CAPV_MANAGER_IMAGE="${CAPV_MANAGER_IMAGE:-gcr.io/cnx-cluster-api/vsphere-cluster-api-provider:latest}"
 
 # Initialize the paths for Kustomize and Python
 KUSTOMIZE="$(command -v kustomize 2>/dev/null || echo "docker")"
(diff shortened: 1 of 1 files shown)


akutz:
Hi @mylesagray,

Please take a look.

/assign @andrewsykim,

This is ready for an lgtm IMO.

akutz:
/assign @codenrhoden 

Maybe relevant to what you're doing as well?

mylesagray:
LGTM

akutz:
You’re okay with the lack of “type”? I figured since it really served no purpose, letting “out” be an override works the same way.

codenrhoden:
/lgtm
/hold
hold is there just so @mylesagray can chime in on `type`. I think the way this is makes sense.

mylesagray:
I agree, type is unnecessary- ship it.

PR #368, opened 2019-06-26
Fix exists logic since Exist doesn't requeue

akutz:
This patch fixes how the existence checks for VMs are handled since the MachineController doesn't reenqueue Exists calls that return a requeue error. Therefore all requeue logic had to be shifted into the CUD operations.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: NA

I haven't tested this locally at all yet. My morning has been nuts. I just wanted to push a WiP PR so that reviews could start. I plan to test and handle bugs after lunch.

**Release note**:

```release-note
NONE
```

/assign @sidharthsurana

### DIFF
--- a/pkg/cloud/vsphere/services/govmomi/util.go
+++ b/pkg/cloud/vsphere/services/govmomi/util.go
@@ -26,8 +26,10 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/apis/vsphereproviderconfig/v1alpha1"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/constants"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/context"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/services/govmomi/net"
+	clustererror "sigs.k8s.io/cluster-api/pkg/controller/error"
 )
 
 func findVM(ctx *context.MachineContext) (*object.VirtualMachine, error) {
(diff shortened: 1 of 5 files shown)


pkg/cloud/vsphere/services/govmomi/util.go
```diff
@@ -165,3 +167,76 @@ func getExistingMetadata(ctx *context.MachineContext) (string, error) {
...
+
+	// Check to see if there is an in-flight task.
+	if task := getTask(ctx); task == nil {
+		ctx.MachineStatus.TaskRef = ""
+	} else {
+		ctx := context.NewMachineLoggerContext(ctx, task.Reference().Value)
+
+		// Since a task was discovered, let's find out if it indicates a VM is
+		// being, or has been, created/cloned.
+		ctx.Logger.V(4).Info("task found", "state", task.Info.State)
```
akutz: By the way @sidharthsurana, I had the realization last night (not yet having time to change the code to reflect this realization), that we actually don't care at all about the task. Whether it's queued or running or anything. I suppose we should check to see if it exists, and if not then delete the ref. But regardless, we only care *if* there **is** a task. If so, we requeue. And we requeue until there is no task and then process the resultant state from there.

Thoughts?

sidharthsurana: One case that comes to mind is race between the user issues a delete (while the clone task was running) and the time the actuator got into action to delete. Now, if the task is running then we want to cancel it. if it succeeded in the small duration, then we got to find and delete the VM directly.

_thinking about any other situations as well..._

akutz: Hi @sidharthsurana,

So task cancellation is the one case, but we don't currently support that. But given this case, at worst the create must complete before the delete then occurs. It's not the worst thing. 

If there's nothing other than that, this would greatly simplify understanding the code. And we could always add the exceptions in as we encounter them.

akutz: By the way, task cancellation is something CAPI's model would have to support as well. Otherwise I'm pretty sure the controller won't process the next call before the previous one is complete.

To that end, what you're discussing is handling the case of the task being cancelled out from under Kubernetes. It's a fair point, but I don't think we care about that right.

However one issue with my proposal it that I describe removing the task ref only when the task can't be found. That's problem since the task ref may live for a while in the collector. We still need to look at the task state and remove the task ref if the task is completed, success or failure.

akutz: Hi @sidharthsurana,

I updated the commit. I didn't change much. I just simplified the result of success. We don't need to set the MoRef here. That will happen (or won't) when the FindVM call runs below.

pkg/cloud/vsphere/services/govmomi/util.go
```diff
@@ -165,3 +167,67 @@ func getExistingMetadata(ctx *context.MachineContext) (string, error) {
 
 	return string(metadataBuf), nil
 }
+
+// lookupVM returns the VM, an RequeueError because the VM has an in-flight,
+// or an unexpected error.
+func lookupVM(ctx *context.MachineContext) (*object.VirtualMachine, error) {
+
+	// Check to see if there is an in-flight task.
```
akutz: Check if task ref is empty

pkg/cloud/vsphere/services/govmomi/util.go
```diff
@@ -165,3 +167,67 @@ func getExistingMetadata(ctx *context.MachineContext) (string, error) {
...
+	}
+
+	// Otherwise verify that the MoRef may be used to return the name of the VM.
+	moRef := types.ManagedObjectReference{
+		Type:  "VirtualMachine",
+		Value: ctx.MachineConfig.MachineRef,
+	}
+	var obj mo.VirtualMachine
+	if err := ctx.Session.RetrieveOne(ctx, moRef, []string{"name"}, &obj); err != nil {
+		// The name lookup fails, therefore the VM does not exist.
```
akutz: // TODO(akutz) Don't verify the name, but add comment as to why we don't verify the name.

@sidharthsurana

akutz:
Thanks for the review @sidharthsurana, please feel free to lgtm when you're ready.

PR #370, opened 2019-06-26
Add support for Machine maintenance window

### LINKED ISSUE #369: Support taking a `Machine` temporarily offline
/kind feature

In vsphere env, there are cases when an admin might want to power off some VMs in order to do routine software/hardware maintenance . The current behavior for CAPV actuator is to power ON the VM if it is not already powered on. This means even if the admin intentionally powered OFF a VM it would automatically be powered on by the actuator.
This we need a way for the user to express their intent on the `Machine` object so that the controller would not automatically change the state of the VM for the time being. 

I suggest we could define a special annotation for the `Machine` object that could express that intent.
e.g. `capv.sigs.k8s.io/offline` 
The intent would be for the actuator to sort of skip reconciling the `Machine` object temporarily.

sidharthsurana:
This patch introduces a new annotation "capv.sigs.k8s.io/maintenance"
that currenly anyone can put on the Machine object to instruct the
actuator to skip any kind of reconciliation on the machine object.

This can be used for various situations like(only sample use cases):
- admin wants to perform maintenance on the underlying infra for which
  they might need to power down VMs
- create bulk machine objects without the actuator creating them all at once,
  but only have them realized as needed by removing this annotation

Change-Id: I5a22c7abbe819f34a05c482c19d8625a661eaeaf

Fixes #369

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/vsphere/actuators/machine/actuator.go
+++ b/pkg/cloud/vsphere/actuators/machine/actuator.go
@@ -78,6 +78,11 @@ func (a *Actuator) Create(
 		return err
 	}
 
+	if _, ok := machine.Annotations[constants.MaintenanceAnnotationLabel]; ok {
+		return errors.Wrapf(&clustererr.RequeueAfterError{RequeueAfter: constants.DefaultRequeue},
+			"detected `capv.sigs.k8s.io/maintenance` annotation on the machine, skipping all operations for machine %q", ctx)
+	}
+
 	defer func() {
 		opErr = actuators.PatchAndHandleError(ctx, "Create", opErr)
 	}()
@@ -111,6 +116,11 @@ func (a *Actuator) Delete(
 		})
 	if err != nil {
 		return err
+	}
+
+	if _, ok := machine.Annotations[constants.MaintenanceAnnotationLabel]; ok {
+		return errors.Wrapf(&clustererr.RequeueAfterError{RequeueAfter: constants.DefaultRequeue},
+			"detected `capv.sigs.k8s.io/maintenance` annotation on the machine, skipping all operations for machine %q", ctx)
 	}
 
 	defer func() {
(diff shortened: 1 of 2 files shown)


pkg/cloud/vsphere/actuators/machine/actuator.go
```diff
@@ -78,6 +78,11 @@ func (a *Actuator) Create(
 		return err
 	}
 
+	if _, ok := machine.Annotations[constants.MaintenanceAnnotationLabel]; ok {
```
akutz: Hi @sidharthsurana,

FWIW, all the `if` statements could be simplified as `if machine.Annotations[constants.MaintenanceAnnotationLabel] != ""` unless you envision using an empty value?

sidharthsurana:
> /hold
> 
> Hi @sidharthsurana,
> 
> Thank you for doing this! Two things:
> 
> 1. Please use the annotation constant in any message. That way things are consistent, even if labels change.
> 2. You cannot wrap the Requeue error if this check is above `PatchAndHandleError`. That function is what checks the error stack for a wrapped Requeue error and returns it if one exists. As it is you should probably change your returns to:
> 
> ```go
> // Should this be a 4 or 6? Thoughts?
> ctx.Logger.V(4).Info("skipping operations on machine", "reason", "annotation", "annotation-key", constants.MaintenanceAnnotationLabel)
> return &clustererr.RequeueAfterError{RequeueAfter: constants.DefaultRequeue}
> ```

Thanks for the suggestion. Updated the code!

pkg/cloud/vsphere/actuators/machine/actuator.go
akutz: I guess I don't ever see someone put an annotation with an empty value, but fair enough :)

PR #381, opened 2019-06-27
Updates to build static clusterctl 

akutz:
This patch updates clusterctl targets to build statically linked binaries.

Fixes #

I was building `clusterctl` for releases manually, and this way we can rely on the Docker build for those too.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,13 +42,19 @@ manager: fmt vet
 
 # Build the clusterctl binary
 clusterctl: fmt vet
-	go build -o bin/clusterctl ./cmd/clusterctl
+	CGO_ENABLED=0 go build -ldflags '-extldflags "-static" -w -s' \
+	-o bin/clusterctl."$${GOOS:-$$(go env GOHOSTOS)}"_"$${GOARCH:-$$(go env GOHOSTARCH)}" \
+	./cmd/clusterctl
+	@cp -f bin/clusterctl."$${GOOS:-$$(go env GOHOSTOS)}"_"$${GOARCH:-$$(go env GOHOSTARCH)}" bin/clusterctl
 
 clusterctl-in-docker:
 	docker run --rm -v $(CWD):/go/src/sigs.k8s.io/cluster-api-provider-vsphere \
 	  -w /go/src/sigs.k8s.io/cluster-api-provider-vsphere \
-	  -e GOOS -e GOHOSTOS golang:1.12 \
-	  go build -o bin/clusterctl ./cmd/clusterctl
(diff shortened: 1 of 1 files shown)


akutz:
@andrewsykim I updated this to build the normal clusterctl as statically linked as well.

Makefile
```diff
@@ -42,13 +42,18 @@ manager: fmt vet
...
 
 clusterctl-in-docker:
 	docker run --rm -v $(CWD):/go/src/sigs.k8s.io/cluster-api-provider-vsphere \
 	  -w /go/src/sigs.k8s.io/cluster-api-provider-vsphere \
-	  -e GOOS -e GOHOSTOS golang:1.12 \
-	  go build -o bin/clusterctl ./cmd/clusterctl
+	  -e CGO_ENABLED=0 -e GOOS="$${GOOS:-linux}" -e GOARCH="$${GOARCH:-amd64}" \
+	  golang:1.12 \
+	  go build -ldflags '-extldflags "-static" -w -s' \
+	  -o bin/clusterctl."$${GOOS:-linux}"_"$${GOARCH:-amd64}" ./cmd/clusterctl
```
andrewsykim: The bin name changing is going to break some docs and blogs

akutz: Then I'll add the same hack I have in the normal target. We should always include the arch in a Docker output. This is really just fixing that. But I'll add that hack for now.

akutz: Actually, the clusterctl binaries are on the releases pages now. It should be grab them from there. If a person wants to build it locally then they should be fine handling this IMO.

akutz: Fixed for now. We should change this later. This target shouldn't be in the Makefile IMO.

Makefile
```diff
@@ -42,13 +42,19 @@ manager: fmt vet
 
 # Build the clusterctl binary
 clusterctl: fmt vet
-	go build -o bin/clusterctl ./cmd/clusterctl
+	CGO_ENABLED=0 go build -ldflags '-extldflags "-static" -w -s' \
+	-o bin/clusterctl."$${GOOS:-$$(go env GOHOSTOS)}"_"$${GOARCH:-$$(go env GOHOSTARCH)}" \
```
andrewsykim: The GOOS and GOHOSTARCH overrides doesn't seem necessary for this case?

andrewsykim: oh nvm this is just the bin name

akutz: /hold

akutz: Wait, never mind. Thought I saw a bug. There's not.

akutz: /hold cancel

PR #385, opened 2019-06-27
wip: PROPOSAL: release process

codenrhoden:
This doc is a proposal for how to approach releases in the future. It's really a combination of what's happening already and where (I'm proposing) we should go. 

Fixes #

The main things this proposal would change versus today are:

- changes the version used on CAPV manager images from always being the git SHA, to being something like `v0.3.0-alpha.0-11-g891b92bc`, which conveys a lot more meaning
- mandates that the CI images built as part of a PR are not pushed to the same place as images built from `master`
- we start publishing `clusterctl` along with the CAPV manager binary as part of the automated process

I made the doc be slightly generic in places to adhere to "don't mandate an implementation". For example, saying "a public registry" rather than "gcr.io". We would of course keep using gcr.io for container images and GCS buckets for files.

I also didn't mention Prow specifically, and I ignore some of it's current limitations. I imply that it's possible to just build a new image when just a tag is pushed. I think that *is* possible with Prow from the code I've looked at so far, but I also know it's not working currently. I'm pretty sure we can get that to work, though, and I've called out an exception just in case we can't for a while.

I put in a few **TODO** sections where I have open questions. For example, i'm not sure if we need to publish a new `clusterctl` binary for every master version or not.

**Release note**:

```release-note

```

/assign @akutz 
/assign @andrewsykim 
/assign @frapposelli

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -1,0 +1,72 @@
+# CAPV Release Process
+
+This document outlines the strategy and process for building and releasing CAPV images.
+
+## Goals
+
+* Define what images/artifacts are built and where they are published
+* Define when and how the git repo is tagged
+* Define how and where container images are built
+* Define what tags are applied to container images and under what scenarios those tags may change
+
+## Non-Goals
+
+* To mandate any specific implementation to achieve Goals
+* To place requirements on intermediate/temporary images that are built solely for purposes of CI and E2E testing
+
+## Artifacts
+
+The CAPV project has three main artifacts that are for public consumption
+
+* CAPV manager container image
+  * This container image contains the main CAPV manager binary, and is what is deployed to a K8s cluster to make it a CAPI management cluster capable of deploying workload clusters.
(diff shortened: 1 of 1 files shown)


codenrhoden:
Okay, I updated this doc to more closely align to what I've discussed with @akutz, to pass the new markdown linter (I hope), and to align with work in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/426.

I definitely want some more eyes on this.

/assign @yastij 
/assign @dims

codenrhoden:
Some other good feedback that is relevant here: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/426#issuecomment-512291343

frapposelli:
@codenrhoden is this PR still current?

codenrhoden:
> @codenrhoden is this PR still current?

Not really. The majority of it was implemented though, and the release process is mostly on auto-pilot.

/close

PR #389, opened 2019-06-28
Fix Kubeadm Init and Join NodeRegistration Options

james65535:

Kubeadm init and join Node Registration options currently overwrites values passed in through the provider config.  This fix replaces the NewNodeRegistration function so that it does not start with empty Node Registration Options

### DIFF
--- a/pkg/cloud/vsphere/services/kubeadm/noderegistration.go
+++ b/pkg/cloud/vsphere/services/kubeadm/noderegistration.go
@@ -26,13 +26,12 @@ import (
 // NodeRegistrationOption is a function that sets a value on a Kubeadm NodeRegistrationOptions.
 type NodeRegistrationOption func(*kubeadmv1beta1.NodeRegistrationOptions)
 
-// NewNodeRegistration will create a kubeadmNodeRegistrationOptions with some defaults and the ability to provide customizations.
-func NewNodeRegistration(opts ...NodeRegistrationOption) kubeadmv1beta1.NodeRegistrationOptions {
-	nro := &kubeadmv1beta1.NodeRegistrationOptions{}
+// SetNodeRegistration will create a kubeadmNodeRegistrationOptions with some defaults and the ability to provide customizations.
+func SetNodeRegistration(base *kubeadmv1beta1.NodeRegistrationOptions, opts ...NodeRegistrationOption) kubeadmv1beta1.NodeRegistrationOptions {
 	for _, opt := range opts {
-		opt(nro)
+		opt(base)
 	}
-	return *nro
+	return *base
 }
 
 // WithTaints will set the taints on the NodeRegistration.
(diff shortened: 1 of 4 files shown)


pkg/cloud/vsphere/services/kubeadm/noderegistration.go
```diff
@@ -27,12 +27,11 @@ import (
 type NodeRegistrationOption func(*kubeadmv1beta1.NodeRegistrationOptions)
 
 // NewNodeRegistration will create a kubeadmNodeRegistrationOptions with some defaults and the ability to provide customizations.
-func NewNodeRegistration(opts ...NodeRegistrationOption) kubeadmv1beta1.NodeRegistrationOptions {
-	nro := &kubeadmv1beta1.NodeRegistrationOptions{}
+func SetNodeRegistration(base *kubeadmv1beta1.NodeRegistrationOptions, opts ...NodeRegistrationOption) kubeadmv1beta1.NodeRegistrationOptions {
```
akutz: Please lint. It would tell you that the function’s name doesn’t match the comment.

Also, please indicate why the function was renamed?

andrewsykim: We updated this locally, this is stale branch, git issues

james65535: Will lint.
I renamed it to match the naming in CAPA.

james65535:
still need to run e2e test to verify it fixes the original problem of supplying a custom kubeadm init and join configuration.  /hold

pkg/cloud/vsphere/services/kubeadm/noderegistration.go
akutz: Cool.

akutz:
Feel free to cancel the hold once comment is fixed.

pkg/cloud/vsphere/services/kubeadm/noderegistration_test.go
```diff
@@ -68,7 +69,8 @@ func TestNewNodeRegistration(t *testing.T) {
 					"node-labels": "test value one,test value two",
 				},
 			},
-			actual: kubeadm.NewNodeRegistration(
+			actual: kubeadm.SetNodeRegistration(
+				&kubeadmv1beta1.NodeRegistrationOptions{},
```
andrewsykim: As discussed, in a follow-up PR, add a test case where the base config has some values.

andrewsykim:
/lgtm

Will let you hold cancel @james65535 after you've had a chance to test

james65535:
Manual testing came back successful, was able to verify kubeadm custom init/join settings are applied to workload machines where previously they were not.

pkg/cloud/vsphere/services/kubeadm/noderegistration_test.go
james65535: ok will do.

PR #391, opened 2019-06-28
Bump etcd start timeout to 30s

### LINKED ISSUE #390: pull-cluster-api-provider-vsphere-test  is flaky
/kind bug

We've seen the test `pull-cluster-api-provider-vsphere-test` be very flaky, and it always fails at the same spot:

```
2019/06/27 23:38:15 failed to start the controlplane. retried 5 times: timeout waiting for process etcd to start
FAIL	sigs.k8s.io/cluster-api-provider-vsphere/pkg/apis/vsphereproviderconfig/v1alpha1	106.627s
```

Digging into this, the error is actually coming out of [controller-runtime/pkg/envtest](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/envtest/server.go), which in turn is using [testing-frameworks/integration](https://github.com/kubernetes-sigs/testing_frameworks/tree/master/integration) under the hood.

In short, `etcd` fails to start up, and starting this defaults to 5 retries with a 20 second timeout. Hence, when it flakes, the test is always just over 100 seconds.

When the test works, I've seen it take anywhere from 12 to 30 seconds, which to me indicates that the first attempt timed out at 20 seconds, then then the second attempt worked fine and the test *usually* takes about 12-13 seconds.  On my local machine, the tests takes 4-5 seconds. I have to drop the (adjustable) timeout for the control plane to under 4 seconds before the timeout is reliably hit.

The number of retries is hardcoded at 5.

It's not clear to me if adjusting the timeout up from the default of 20 seconds would solve the issue, or if the container running the tests needs more resources. It's probably worth bumping the timeout to 30 seconds and see if that helps. if it doesn't, take the different approach of requesting more resources for this particular job.

not flake

**Environment:**

- Cluster-api-provider-vsphere version:  v0.3.0

codenrhoden:
Add KUBEBUILDER_CONTROLPLANE_START_TIMEOUT=30s to the ci-test.sh script,
which is invoked by CI. Hoping to address test flakiness.

attempts to fix #390 

**Release note**:

```release-note

```

### DIFF
--- a/scripts/ci-test.sh
+++ b/scripts/ci-test.sh
@@ -24,4 +24,4 @@ cd $REPO_ROOT && \
 	source ./scripts/fetch_ext_bins.sh && \
 	fetch_tools && \
 	setup_envs && \
-	make test
+	KUBEBUILDER_CONTROLPLANE_START_TIMEOUT=30s make test


codenrhoden:
/hold

apologies for the coming spam, as I am going to re-run the flaky test a bunch of times to see if this patch actually improves things.

/test pull-cluster-api-provider-vsphere-test

codenrhoden:
/test pull-cluster-api-provider-vsphere-test

3/3 success so far, with 2/3 having the problematic test taking ~25s, which previously would have failed when hitting the 20s timeout. I think we are on to something here.

codenrhoden:
Blargh, it failed on the 5th run with a runtime of `150.733s`, which of course is 5*30s. Since two of the tests took 25s, those are tests that would have failed before, however we don't actually know that since we don't know what would have happened on the retry.

I was hoping to see a test that took > 30s, meaning that it worked on a subsequent retry. Right now I don't have much indication that that it's not always the case that if things dont' work the first time, they fail on all the subsequent retries. 🤔 

I don't want to just keep bumping up the timeout, because if it is a resource issue and once in a while we just get a pod that's going to perform so slowly it always fails, the failure is just going to take that much longer...

Anyone have thoughts here? We could try a 60s timeout (potentially making this one test take up to 5 minutes) and then also try to request more resources for this test?

akutz:
Hi @yastij,

You've been working with etcd weirdness lately. Any advice?

yastij:
@codenrhoden - etcd startup can very flaky when dedicating a low amount of resources. We might also want to make [this](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/envtest/server.go#L214) tunable

codenrhoden:
@yastij  Yeah I was surprised to see that was hardcoded. I think my main concern is that I have seen so far that this is pretty much "all or nothing". Either it works the first time, or it fails after 5 times. 

I take that to mean that however long it takes to start etcd the first time (whether is 4, 15, 25, or 60 seconds), it's going to take that same amount of time on each retry. So I'm not sure that more tries helps. But again, this is more a feeling than actual data I have on this. That's why I'm going to keep running the test a few more times and see if there are any obvious ones where the test time is greater than the timeout -- meaning that it had to retry at least once.

/test pull-cluster-api-provider-vsphere-test

akutz:
Hi @codenrhoden,

Don't know if you saw [this in Slack](https://kubernetes.slack.com/archives/CKFGK3SSD/p1561736627086200), but I mentioned checking the Prow job's resources earlier, re: @yastij's comment about resource exhaustion on start.

yastij:
@codenrhoden - do we have any logs on etcd startup ? we should be able to get some if [these](https://github.com/kubernetes-sigs/testing_frameworks/blob/57f07443c2d47d9c7f932a9a4fc87fbf3408215d/integration/etcd.go#L55-L56) are populated

codenrhoden:
@yastij not at the moment, but looks like we will get it if I set `KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT=true`. I'll give that a shot.

@akutz I definitely saw it. That's why I've mentioned a few times whether we should explore that in parallel with this. We are in agreement, for sure.

akutz:
@codenrhoden Sorry, I missed your responses. Mea culpa.

codenrhoden:
Okay, I added a minimum resource request of 0.5CPU in https://github.com/kubernetes/test-infra/pull/13264#issuecomment-506873012

Will run this a few more times to see what happens, and if there is a general trending up or down of test time.

As for etcd logs, I  haven't pushed a change to get those yet because I haven't been able to make that work locally. Will pick that up again next week.

/test pull-cluster-api-provider-vsphere-test

codenrhoden:
/test pull-cluster-api-provider-vsphere-test

Latest tests are completing in 12-14s. Maybe we've finally cracked the code. :)

yastij:
I think we can merge this

codenrhoden:
> I think we can merge this

Yep, things are looking good. Went through all the latest tests for this commit on test grid and the times are now consistent. Flakiness be gone!

PR #392, opened 2019-06-29
reduce requeue period for expected wait operations

moshloop:
This PR reduces the re-queue period for standard operations like power on/off, destroy etc.. 
Most of the time these operations return quickly and waiting 20 seconds to retry wastes time.

This should reduce testing time without adding too much logging verbosity


akutz:
Hi @moshloop,

I disagree that 20 seconds is too long. We commonly see multiple requeues, even at 20s. I've thought of *increasing* the wait time. Perhaps this is specific to ESX?

As it is, I see no compelling reason to reduce wait times. At best the default should stay as is and then have some provider config value that is used if defined. That way your tests can switch to 3s, but the default can remain at 20s.

moshloop:
Maybe making the default timeout configurable via a CLI argument is a better alternative ?

moshloop:
I think implementing this as a provider config would be unnecessarily complex  
* what happens if you have multiple clusters/provider configs ?
* would the value start as a default, and then get updated once the cluster is reconciled?

akutz:
/close 

Replaced by #420

PR #387, opened 2019-06-28
Thread-safe cluster init/join logic

akutz:
**What this PR does / why we need it**:
This PR implements thread-safe cluster init/join logic, preventing two control plane machines from trying to initialize the cluster at the same time.

There are two diagrams describing the workflow of this PR:

1. [Cluster controller](https://github.com/akutz/cluster-api-provider-vsphere/blob/feature/cluster-admission/docs/design/cluster-admission-cluster-controller.svg)
2. [Machine controller](https://github.com/akutz/cluster-api-provider-vsphere/blob/feature/cluster-admission/docs/design/cluster-admission-machine-controller.svg)

Related to kubernetes-sigs/cluster-api-provider-aws#779

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:

**Special notes for your reviewer**:
Please note that this PR is scoped to **just** the happy-path. Machine deletions will be handled at a later date.

This PR is dependent upon #395 and will be rebased once #395 is merged.

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
* Fix a race condition when creating a new cluster with multiple control plane machines at the same time.
```

PR #393, opened 2019-06-30
Update object patch logic

akutz:
This patch fixes a bug in the patch logic for clusters and machines. If the provider spec and status are not reencoded back into their original locations, no patch is generated for either. Please note this does not affect the cluster/machine spec/status -- only the provider analogues.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/vsphere/actuators/actuators.go
+++ b/pkg/cloud/vsphere/actuators/actuators.go
@@ -25,7 +25,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	clustererr "sigs.k8s.io/cluster-api/pkg/controller/error"
 
-	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/constants"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/record"
 )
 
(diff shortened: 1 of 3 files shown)


akutz:
This bug is the cause for the failures in #387 when the updated machine ready logic patch was applied. Once this PR is merged, #387 will be rebased so the revert is no longer part of that PR.

pkg/cloud/vsphere/actuators/actuators.go
```diff
@@ -25,29 +25,24 @@ import (
...
-	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/constants"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/record"
 )
 
 type patchContext interface {
 	fmt.Stringer
 	GetObject() runtime.Object
 	GetLogger() logr.Logger
-	Patch() error
+	Patch()
```
andrewsykim: Wondering if we can keep the err return value if we just separate the patch update for spec and status into separate methods

akutz: Hi @andrewsykim,

I considered it, but it really breaks down into how you consider patching. If patching the object fails, you don't want to update the status, because the status may not reflect the object data known to Kubernetes.

And I've decided that anything you do in the patch should be idempotent anyway, and any subsequent patch should reconcile previous failures.

Additionally, if a patch fails due to a conflict, you pretty much **have** to wait until the next entry. You simply don't have the knowledge to take the updated server view of an object or its status and merged it with your data. That's why the conflict occurred in the first place. 

Which all leads back to this -- patching must be reentrant. If an object or status patch failing causes a provider to enter a corrupt state then the provider's design is flawed.

akutz:
There have been no objections to this PR, so I'm removing the hold. We can always update/revert any bugs discovered. As it stands, `master` is not correctly updating provider status on clusters or machines.

/hold cancel

PR #394, opened 2019-06-30
CAPI 1.4

akutz:
**What this PR does / why we need it**:
This patch updates the CAPI dependency to v0.1.4. The big change is support for the NodeRef controller, meaning Machines will now automatically have their NodeRef set once the cloud provider reconciles the cluster node.

~~This patch also updates the Cluster/Machine ready state logic to be dependent upon the presence of a NodeRef.~~

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:
This PR depends on #393 and will be rebased once that PR is merged.

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
* CAPI 1.4
```

PR #395, opened 2019-07-01
Ready-state logic changes

akutz:
This patch updates the ready-state logic for the Cluster and Machine objects. A machine is not marked as ready until it is possible to connect to the target cluster and verify the machine is represented
there as a node.
    
Additionally, this patch relocates the code to create the KubeConfig secret from the Cluster's Reconcile function to the Machine's Update function. This is because the KubeConfig secret must be created before a Cluster can verify its ready state. If the Cluster didn't create the KubeConfig until after it was ready, then a circular dependency would be created, and a Cluster would never appear as ready.
    
Finally, several existing calls have been changed to rely on the CAPI NodeRef/Remote packages in order to get a client for a target cluster or retrieve a KubeConfig for the target cluster.

Fixes #

This PR depends on #394 and will be rebased once #394 merges.

**Release note**:

```release-note

```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -85,10 +85,10 @@ func main() {
 	record.InitFromRecorder(mgr.GetRecorder("vsphere-controller"))
 
 	// Initialize cluster actuator.
-	clusterActuator := cluster.NewActuator(cs.ClusterV1alpha1(), coreClient)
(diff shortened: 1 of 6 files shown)


akutz:
Hi @ncdc,

If you have time, this is ready for review. It passes e2e tests, so I'm confident that the changes work, but maybe there's a more efficient means of using the new packages. Thanks!

PR #396, opened 2019-07-01
Use pkg/deployer in the Cluster actuator

akutz:
This package removes the functions GetIP and GetKubeConfig from the Cluster actuator and has it defer to pkg/deployer for the analogous logic. This removes any confusion about the two functions as they existed in the Cluster actuator. They were not to get any special values from secrets, etc. They were just to perform the exact same behavior as from the deployer package.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/vsphere/actuators/cluster/actuator.go
+++ b/pkg/cloud/vsphere/actuators/cluster/actuator.go
@@ -36,7 +36,6 @@ import (
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/constants"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/context"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/services/certificates"
-	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/services/kubeclient"
 )
 
 // Actuator is responsible for maintaining the Cluster objects.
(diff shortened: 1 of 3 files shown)


pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -115,47 +114,6 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster) (opErr error) {
 	return nil
 }
 
-// GetIP returns the control plane endpoint for the cluster.
-func (a *Actuator) GetIP(cluster *clusterv1.Cluster, machine *clusterv1.Machine) (string, error) {
-	clusterContext, err := context.NewClusterContext(&context.ClusterContextParams{
-		Cluster:    cluster,
-		Client:     a.client,
```
andrewsykim: Loosing access to clients here, guessing that's okay in this case?

akutz: Yes. I realized that the deployer was never supposed to have that access in the first place. It was just the way CAPA was built. The deployer simply builds a kubeconfig or parses the control plane endpoint out of one of those objects.

pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -115,47 +114,6 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster) (opErr error) {
...
-		CoreClient: a.coreClient,
-	})
-	if err != nil {
-		return "", err
-	}
-
-	if machine == nil {
```
andrewsykim: Do we need an equivalent machine == nil check here in the Deployer?

akutz: The deployer always existed, so it shouldn’t need to be changed. But I’m happy to add it.

akutz: Fixed.

andrewsykim: Thought cluster actuator calls this but and needed the machine == nil check, looks like that changed recently though

pkg/deployer/deployer.go
```diff
@@ -23,6 +23,19 @@ import (
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/services/kubeclient"
 )
 
+var d Deployer
+
+// GetIP returns the control plane endpoint for the cluster.
+func GetIP(cluster *clusterv1.Cluster, machine *clusterv1.Machine) (string, error) {
```
andrewsykim: Where is this actually getting called, I thought the methods part of `Deployer` below are enough for the provider?

akutz: I just did this as a convenience thing.

akutz:
Hi @andrewsykim,

If there's no other feedback, can you please lgtm? Thanks!

PR #397, opened 2019-07-01
Distroless manager image

### LINKED ISSUE #258: Fix image security issues
/kind bug

Please refer to https://github.com/kubernetes-sigs/cluster-api-provider-ibmcloud/issues/115#issuecomment-489579864 for detail. There are some security issues for the controller image, we should fix it.

akutz:
This patch updates the CAPV manager image to a static, distroless image. This saves space as well as enhances security by removing possible attack vectors.

Fixes #258

**Release note**:

```release-note
* CAPV manager now based on Google's static, distroless image
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,38 @@
-# Build the manager binary
-FROM golang:1.12 as builder
+################################################################################
+##                               BUILD ARGS                                   ##
+################################################################################
+# This build arg allows the specification of a custom Golang image.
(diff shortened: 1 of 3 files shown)


akutz:
Thanks @figo. If the hold is for others to comment then I may remove it soon. This is a basic change without much room for feedback. The e2e test passed, and really the only point up for discussion is whether to make this change at all, and I don’t think anyone would argue against it as long as it doesn’t break the image, which doesn’t appear to be the case.

PR #405, opened 2019-07-03
Updates `generateUserData` to fill default VMFolder if not provided.

### LINKED ISSUE #401: kube-controller-manager does not start if "vmFolder" is not provided in config
"vmFolder" is an optional parameter to be set while generating the manifests. "vmFolder" in `VsphereMachineProviderConfig` defaults to an empty string while generating the manifests.

Using these manifests with `clusterctl` results in a management cluster in which `kube-controller-manager` does not work, resulting in a broken management cluster.

**Possible reason**:
Spinning up of `kube-controller-manager` needs to know the correct "vmFolder" which by default is an empty string. The data that is passed to the `SetCloudInitUserData` method is not modified to fill the values like "vmFolder", "resoursePool" if missing because `SetCloudInitUserData` is called before we get the correct "vmFolder", "resourcePool", etc using the `FolderOrDefault`, `ResourcePoolOrDefault`, etc functions.

**Possible fixes**:
* Make the "vmFolder" field required.
    - **Advantages**: Easy implementation
    - **Disadvantages**: Might be enforcing a field that should not be enforced.
* Make code changes to retrieve correct default values before calling `SetCloudInitUserData`. 
     - **Advantages**: The user-facing `VsphereMachineProviderConfig` can remain as is.
     - **Disadvantages**: The argument passed to `SetCloudInitUserData` is not a struct but an array of bytes. Modifying it to support default values will need changes elsewhere too (possibly the `generateUserData` function).

/cc @figo

ykakarap:
cloudconfig needs a valid `vmFolder` value. Since it is an optional field in the `VsphereMachineProviderConfig` (whose values are used to generate the cloudconfig) as of now we need to fill in a valid default value if not provided.

Fixes #401 

The `generateUserData` is updated to fill in a default value for `VMFolder` in cloudconfig if it is not provided.
This needed a change in `govmomi`. https://github.com/vmware/govmomi/pull/1515 makes the required changes. 
The `go.mod`, `go.sum` files and the `vendor` folder are updated to reflect this new govmomi dependency.

FYI, looks like the `vendor` folder also has some changes from previous updates.

**Release note**:

```release-note

```


ykakarap:
I still need to sign the Contributor License Agreement. Having some issues with that.

akutz:
Hi @ykakarap,

Please rebase this after running `make vendor`. That's how you update the project's vendored deps. It re-adds some required, non-Go assets pruned by Go modules.

PR #411, opened 2019-07-04
Recreate VM if it is deleted by infrastructure

### LINKED ISSUE #409: Externally delete VM won't automatically heal
/kind bug

If I use kubectl delete machine to delete a VM, clusterapi will automatically grow one back. However, if I manually delete one VM from vCenter, clusterapi won't remedy the cluster. 

clusterapi should resurrect the node. 

**Environment:**

- Cluster-api-provider-vsphere version: 0.3.0 
- Kubernetes version: (use `kubectl version`): 1.13.6 
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

akutz:
This patch fixes a bug where a VM is not recreated if it is deleted by the infrastructure, but still exists as a machine in CAPV.

Fixes #409

**Release note**:

```release-note
* Fixes a bug where a deleted VM is not recreated
```

### DIFF
--- a/pkg/cloud/vsphere/services/govmomi/update.go
+++ b/pkg/cloud/vsphere/services/govmomi/update.go
@@ -42,8 +42,14 @@ func Update(ctx *context.MachineContext) error {
 	if err != nil {
 		return err
 	}
+
+	// If no VM was found then it means the machine has a MachineRef that refers
+	// to a VM that no longer exists. Clear the existing MachineRef and requeue
+	// this machine so that a new VM is created.
 	if vm == nil {
-		return errors.Errorf("vm should already exist for %q", ctx)
+		ctx.Logger.V(2).Info("vm should exist but cannot be found, recreating", "machine-ref", ctx.MachineConfig.MachineRef)
+		ctx.MachineConfig.MachineRef = ""
+		return &clustererror.RequeueAfterError{RequeueAfter: constants.DefaultRequeue}
 	}
 
 	if err := reconcileNetwork(ctx, vm); err != nil {


pkg/cloud/vsphere/services/govmomi/update.go
```diff
@@ -42,8 +42,14 @@ func Update(ctx *context.MachineContext) error {
...
 		return err
 	}
+
+	// If no VM was found then it means the machine has a MachineRef that refers
+	// to a VM that no longer exists. Clear the existing MachineRef and requeue
+	// this machine so that a new VM is created.
 	if vm == nil {
-		return errors.Errorf("vm should already exist for %q", ctx)
+		ctx.MachineConfig.MachineRef = ""
+		ctx.Logger.V(2).Info("vm does not exist; requeue for creation")
```
frapposelli: Maybe make this error message more specific to the condition, like “VM was deleted by an external entity, recreating”

akutz: Hi @frapposelli,

I'm more than happy to make the error more descriptive, but truthfully we don't know *why* or *how* or by *whom* the VM was removed. We just know it's no longer there. I'm all for adding some code later to try and find a task/event that provides additional information related to the deletion event.

akutz: Hi @frapposelli,

I think I *should* rename this to something like "vm cannot be found; requeue for creation". Because truthfully the VM *could* exist, but we don't have access to it or its ID has changed. I may even want to use this PR to update the permanent ID to the BIOS UUID instead of the MoRef CAPV currently uses.

frapposelli: If I'm looking at the CAPV log, I would be fine with just knowing that the condition that triggered a recreate is that the vm *existed, but is no longer there*. More details are absolutely welcome though.

akutz: Hi @frapposelli,

Fair. I think I'll just update the log for now then, since anything more requires some thinking and quite a bit of work for what you originally requested.

akutz: Hi @frapposelli,

Please check now, thanks!

frapposelli: 👍

PR #414, opened 2019-07-05
Rename API group to "vsphere.cluster.k8s.io"

akutz:
This patch renames the CAPV API group to `vsphere.cluster.k8s.io`. This is after a conversation with Andy from CAPI, and he advised to drop the `sigs` from the API group and agrees this is a decent choice to simplify the current API group. All annotations are now also based on this root.

**Release note**:

```release-note
* The CAPV API group is now vsphere.cluster.k8s.io.
```


frapposelli:
Awesome stuff @akutz 👍 

/lgtm

holding for @sidharthsurana , PTAL

/hold

docs/getting_started.md
```diff
@@ -194,20 +194,19 @@ kind: Cluster
...
-            cidrBlocks: ["100.96.0.0/11"]
-        serviceDomain: "cluster.local"
-    providerSpec:
-      value:
-        apiVersion: "vsphereproviderconfig/v1alpha1"
-        kind: "VsphereClusterProviderConfig"
-        vsphereUser: "<REDACTED>"
-        vspherePassword: "<REDACTED>"
-        vsphereServer: "<REDACTED>"
-        vsphereCredentialSecret: ""
```
sidharthsurana: Can we keep the  `vsphereCredentialSecret` field in the doc?

akutz: Hi @sidharthsurana,

I removed it because we temporarily removed the field from the cluster provider spec. I have another branch I'm preparing that augments credentials by allowing you to initialize one or more credentials structs that then get turned into secrets on the management cluster (and eventually the workload clusters).

akutz: But I can put it back here so that it's only removed once the other stuff is ready.

docs/getting_started.md
```diff
@@ -218,20 +217,20 @@ metadata:
 spec:
   providerSpec:
     value:
-      apiVersion: vsphereproviderconfig/v1alpha1
-      kind: VsphereMachineProviderConfig
+      apiVersion: vsphere.cluster.k8s.io/v1alpha1
+      kind: VsphereMachineProviderSpec
       machineSpec:
```
sidharthsurana: need to remove the `machineSpec` from here.

docs/getting_started.md
```diff
@@ -250,30 +249,30 @@ spec:
...
+        machineset-name: prod-workload-machineset
         cluster.k8s.io/cluster-name: prod-workload
     spec:
       providerSpec:
         value:
-          apiVersion: vsphereproviderconfig/v1alpha1
-          kind: VsphereMachineProviderConfig
+          apiVersion: vsphere.cluster.k8s.io/v1alpha1
+          kind: VsphereMachineProviderSpec
           machineSpec:
```
sidharthsurana: remove `machineSpec` from here as well

docs/self-service/README.md
```diff
@@ -104,8 +104,8 @@ items:
   spec:
       providerSpec:
         value:
-          apiVersion: "vsphereproviderconfig/v1alpha1"
-          kind: "VsphereMachineProviderConfig"
+          apiVersion: "vsphere.cluster.k8s.io/v1alpha1"
+          kind: "VsphereMachineProviderSpec"
         machineSpec:
```
sidharthsurana: same

docs/self-service/README.md
```diff
@@ -84,8 +84,8 @@ spec:
         serviceDomain: "cluster.local"
     providerSpec:
       value:
-        apiVersion: "vsphereproviderconfig/v1alpha1"
-        kind: "VsphereClusterProviderConfig"
+        apiVersion: "vsphere.cluster.k8s.io/v1alpha1"
+        kind: "VsphereClusterProviderSpec"
         vsphereUser: "administrator@vsphere.local"
```
sidharthsurana: change `vsphereUser` ->  `username`
change `vspherePassword` ->  `password`

docs/getting_started.md
akutz: Actually, I'm going to leave this out. The field would be omitted when marshaled anyway as it is empty. This is not an example of all possible fields, but the example of what happens when the manifests are created.

We cannot expect every manifest to include every possible configuration. The k/k community has routinely decided that GoDocs are the means by which people can discover additional field information.

Eventually there will be better documentation for how to use credential secrets.

akutz: Hi @frapposelli,

Just getting your opinion here. There's just zero reason for the secret before the management cluster exists anyway. So I don't really think it belongs in any of our generated data since that's all generally for the pivot, and if people know how to create secrets, then they know how to find the docs on how to use them IMO.

And as I said, I'm working on a refactor to allow for a single credentials manager defined as part of the cluster spec to be used by both the cloud providers as well as the machine specs (they can select which secret to use in order to partition creation of machines in different locations)

PR #416, opened 2019-07-07
Provide default SSH access to machines

### LINKED ISSUE #413: generate ssh key pair for VM if user does not set sshAuthorizedKeys
/kind feature

generate ssh key pair of VM if user does not set sshAuthorizedKeys, so that user can SSH to vm for debugging

N/A

cc @akutz

akutz:
This patch introduces a new field to the ClusterProviderSpec: `SSHKeyPair`. This field will automatically be populated with a public and private key that may be used to access machines deployed to the cluster.

Fixes #413

**Release note**:

```release-note
* Provide default SSH access to machines
```


moshloop:
Why is this needed? Accessing a machine via SSH would be an anti-pattern in most cases?

akutz:
> Why is this needed? Accessing a machine via SSH would be an anti-pattern in most cases?

Hi @moshloop,

I disagree that this is anti-pattern. The CAPA provider uses AWS's native ability to inject SSH keys into machines to do the same. This is to enhance CAPV to support a default key pair for remote access.

In another location you asked how this differs from `SSHAuthorizedKeys`. I attempted to explain that in the GoDocs:

>  SSHKeyPair is the SSH key pair for accessing provisioned machines. This field is set at runtime and should not be assigned by an operator. Providing access to existing SSH keys is handled by SSHAuthorizedKeys. Please see SSHAuthorizedKeys for information on accessing the machines.

In other words, `SSHKeyPair` is always generated to guarantee there is a means to remotely access a provisioned machine. The field `SSHAuthorizedKeys` is for operators to provide additional public keys that should be granted access.

akutz:
Hi @frapposelli / @ncdc ,

I'd love your opinions on this. Thanks!

moshloop:
> The CAPA provider uses AWS's native ability to inject SSH keys into machines to do the same. 

Unless I am mistaken, CAPA specifies the SSH Key Name for AWS to inject, however, that Key refers to a public key only,  the private key needs to be saved externally to AWS

akutz:
Hi @moshloop,

That's correct, but the point is that it guarantees there's always SSH access if desired. This change doesn't inject a private key. It just generates a key pair to ensure there's always SSH access, even if someone neglected to add something to `SSHAuthorizedKeys`.

ncdc:
I'm torn on this one. SSH access may be nice, and if you forget to set the authorized keys, this is a nice fallback. On the other hand, we want to be moving away from storing secret data in resources other than `secrets` (although this is adding "just one more" to a set of secrets in the spec).

How would you feel about generating a `secret` instead and storing a ref to it in the spec?

I'm also curious for @detiber's take since he predates me on the project.

akutz:
Hi @ncdc,

I totally agree. I even brought this up in another [conversation](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/419#discussion_r301150738). The issue that I raised there is that `cluserctl` doesn't really support a first-class notion of a `secrets.yaml` or some manner to inject secret data into the bootstrap cluster. Else I would do this that way. That would also require secrets created pre-pivot be automatically moved to the initial target/management cluster a la https://github.com/kubernetes-sigs/cluster-api/issues/1112.

pkg/apis/vsphere/v1alpha1/vsphereclusterproviderspec_types.go
```diff
@@ -60,37 +60,68 @@ type VsphereClusterProviderSpec struct {
 	//
 	// The default user for CentOS is "centos".
 	// The default user for Ubuntu is "ubuntu".
+	// +optional
 	SSHAuthorizedKeys []string `json:"sshAuthorizedKeys,omitempty"`
 
+	// SSHKeyPair is the SSH key pair for accessing provisioned machines.
+	// This field is set at runtime and should not be assigned by an operator.
+	// Providing access to existing SSH keys is handled by SSHAuthorizedKeys.
```
figo: SSHAuthorizedKeys -> GetSSHAuthorizedKeys?

akutz: Hi @figo,

`SSHAuthorizedKeys` is the field for a list of public SSH keys that can access the remote machines.

`GetSSHAuthorizedKeys` is a function that returns the concatenation of `SSHAuthorizedKeys` and the plain-text, public key from `SSHKeyPair`, if it is set.

figo: yup, it is trivial, my question is regards to the comment:
`Providing access to existing SSH keys is handled by SSHAuthorizedKeys` 
is it means by `function GetSSHAuthorizedKeys`?  thanks

akutz: It's referring to the field `SSHAuthorizedKeys`. It's a Godoc convention to refer to other fields by name.

figo: thx

moshloop:
@randomvariable do you have any comments on this from a CIS perspective?

randomvariable:
The CIS benchmark for Kubernetes doesn't have anything to say on SSH access as it's a not related to Kubernetes itself.

FWIW, if we haven't already for CAPA, we were definitely going to make SSH access optional in case people wanted to use other things like SSH certificates, LDAP or AWS SSM Session Manager.

frapposelli:
Should the generation of the key pair (if not provided) be delegated to `clusterctl`?

ncdc:
`clusterctl` is somewhat "frozen" pending the outcome of https://github.com/kubernetes-sigs/cluster-api/issues/1065 and https://github.com/kubernetes-sigs/cluster-api/issues/1085

frapposelli:
@ncdc `clusterctl` or whatever tool triggers the creation of a cluster, in other words, not letting the provider create those keys.

PR #419, opened 2019-07-07
Cloud provider credentials as secret

### LINKED ISSUE #194: [Security] vsphere cloud provider configuration should use secrets for credentials
Currently the cloud config that we generate as `/etc/kubernetes/cloud-config/cloud-config.yaml` file on the master node contains the vsphere credentials in plain text format. vsphere cloud provider supports providing these sensitive credentials via secrets out of the box. (for e.g see the test case [here](https://github.com/kubernetes/kubernetes/blob/153bc23365c85a3b9a02a350bf0c4e9e420d78b0/pkg/cloudprovider/providers/vsphere/credentialmanager_test.go#L30)). Note that this approach is supported by both in-tree and out-of-tree version of the vsphere cloud provider.
We need to move towards setting the vcenter credentials via secrets for the consumption of the vsphere cloud provider configuration.
However, the only catch here is that today we use kubeadm to setup and configure the kubernetes service with the vsphere cloud provider. Thus, within the workflow if kubeadm we do not have a means to create the secret needed for the vsphere cloud provider prior to starting it. Thus we would need to overcome this limitation somehow.

Note: This issue is different that #9 as that is related to credentials stored as part of the cluster-api objects and this issue is about consumption of the credentials for the vsphere cloud provider that we configure for the target cluster.

akutz:
This patch changes the way the cloud provider obtains its credential information. Instead of directly via the cloud config file, a Kubernetes Secret object is created on the target cluster used by the cloud
provider. This is a much more secure design as credentials are no longer transported via the GuestInfo RPC mechanism as base64-encoded, plain-text data, over an unencrypted transport.

Fixes #194

**Release note**:

```release-note
* The cloud provider credentials are now transported securely and stored/accessed as a Kubernetes secret "kube-system/cloud-provider-vsphere-credentials"
```

### DIFF
--- a/pkg/cloud/vsphere/context/cluster_context.go
+++ b/pkg/cloud/vsphere/context/cluster_context.go
@@ -36,6 +36,7 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	client "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1"
 	clusterUtilv1 "sigs.k8s.io/cluster-api/pkg/util"
+	controllerClient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/patch"
 
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/apis/vsphere/v1alpha1"
@@ -45,27 +46,29 @@ import (
 
 // ClusterContextParams are the parameters needed to create a ClusterContext.
 type ClusterContextParams struct {
-	Context    context.Context
-	Cluster    *clusterv1.Cluster
-	Client     client.ClusterV1alpha1Interface
-	CoreClient corev1.CoreV1Interface
-	Logger     logr.Logger
+	Context          context.Context
+	Cluster          *clusterv1.Cluster
+	Client           client.ClusterV1alpha1Interface
+	CoreClient       corev1.CoreV1Interface
+	ControllerClient controllerClient.Client
+	Logger           logr.Logger
 }
 
 // ClusterContext is a Go context used with a CAPI cluster.
 type ClusterContext struct {
 	context.Context
-	Cluster       *clusterv1.Cluster
-	ClusterCopy   *clusterv1.Cluster
-	ClusterClient client.ClusterInterface
-	ClusterConfig *v1alpha1.VsphereClusterProviderSpec
-	ClusterStatus *v1alpha1.VsphereClusterProviderStatus
-	CoreClient    corev1.CoreV1Interface
-	Logger        logr.Logger
-	client        client.ClusterV1alpha1Interface
-	machineClient client.MachineInterface
-	user          string
-	pass          string
+	Cluster          *clusterv1.Cluster
+	ClusterCopy      *clusterv1.Cluster
+	ClusterClient    client.ClusterInterface
+	ClusterConfig    *v1alpha1.VsphereClusterProviderSpec
+	ClusterStatus    *v1alpha1.VsphereClusterProviderStatus
+	CoreClient       corev1.CoreV1Interface
+	ControllerClient controllerClient.Client
+	Logger           logr.Logger
+	client           client.ClusterV1alpha1Interface
+	machineClient    client.MachineInterface
+	user             string
+	pass             string
 }
 
 // NewClusterContext returns a new ClusterContext.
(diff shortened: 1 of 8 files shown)


moshloop:
+1
fixes #194

akutz:
Hi @andrewsykim / @figo / @frapposelli,

Could one of you please give this a look and lgtm? Thanks!

pkg/cloud/vsphere/services/kubeclient/kubeclient.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Context is used to get a Kubernetes client for a target cluster.
+type Context interface {
+	context.Context
+	GetCluster() *clusterv1.Cluster
+	GetControllerClient() controllerClient.Client
+}
+
+// New returns a new client for the target cluster using the KubeConfig
+// secret stored in the management cluster.
```
figo: this looks like a general library to get kubernetes core client from context, so it could be not specific to `target cluster` or `management cluster`?

pkg/cloud/vsphere/services/kubeclient/kubeclient.go
```diff
@@ -0,0 +1,48 @@
...
+func New(ctx Context) (corev1.CoreV1Interface, error) {
+	client, err := remotev1.NewClusterClient(ctx.GetControllerClient(), ctx.GetCluster())
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get client for target cluster")
```
figo: the same question for using `target cluster` in log

pkg/cloud/vsphere/context/cluster_context.go
```diff
@@ -45,27 +46,29 @@ import (
...
+	ClusterClient    client.ClusterInterface
+	ClusterConfig    *v1alpha1.VsphereClusterProviderSpec
+	ClusterStatus    *v1alpha1.VsphereClusterProviderStatus
+	CoreClient       corev1.CoreV1Interface
+	ControllerClient controllerClient.Client
+	Logger           logr.Logger
+	client           client.ClusterV1alpha1Interface
+	machineClient    client.MachineInterface
+	user             string
+	pass             string
```
figo: not your change, but the password in ClusterSpec should be encoded? i should user should pass encoded value instead of plain text password

akutz: It never was, and if we want to discuss that, let’s file a different issue. I don’t believe it has any value as we can’t encrypt it since we’d need to supply the key, and base64 encoding it just obfuscates, not protects.

pkg/cloud/vsphere/services/kubeclient/kubeclient.go
akutz: Hi @figo,

It’s specific to target cluster (per the godocs), and it does not get it from the context.

pkg/cloud/vsphere/services/kubeclient/kubeclient.go
akutz: Hi @figo,

Same answer.

pkg/cloud/vsphere/context/cluster_context.go
frapposelli: Agree, if keeping plaintext credentials in memory is a problem let’s file an issue for that. Encoding to base64 is not helping in this case.

akutz: Hi @frapposelli,

Yeah, and I only do this (@figo is incorrect, this *is* my code :) ) since the user and pass can come from a secret or the cluster spec. I can always pull them from the secret each time if I need to do so.

pkg/cloud/vsphere/services/kubeclient/kubeclient.go
akutz: Hi @figo,

Sorry, was on a walk when I responded. The client is formed by using the `remotev1` package from CAPI 0.1.4, which pulls the credentials and endpoint for the target cluster from a Secret stored on the management cluster (bootstrap or permanent) that contains the KubeConfig for a target cluster.

figo: Sorry, it is base on context not from, which godocs are you referring to? I just could not see this code that specific to Target cluster

akutz: Hi @figo,

The ones on line 36:

> // New returns a new client for the target cluster using the KubeConfig
> // secret stored in the management cluster.

pkg/cloud/vsphere/context/cluster_context.go
akutz: Let's not forget, however, that we put plain-text credentials in a cluster spec for the initial target cluster since we cannot provide a secret to `clusterctl` in any easy manner today. It's not ideal, that's for sure.

pkg/cloud/vsphere/services/kubeclient/kubeclient.go
akutz: Hi @figo,

By the way:

> so it could be not specific to target cluster or management cluster

It occurs to me that you may be making a distinction here that I do not. In this case, the initial target cluster which becomes the management cluster post-pivot is *still* a **target cluster** pre-pivot.

pkg/cloud/vsphere/context/cluster_context.go
figo: Let's discuss in a separate issue, it is not about this change, sorry for causing confsut

PR #421, opened 2019-07-08
Exit early if command fails on bootstrap

codenrhoden:
I've seen a few cases where the E2E test reports successful, even though it failed pretty early on. An example is this [log](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-cluster-api-provider-vsphere-e2e/1144662693072867328). In that example, the KIND cluster creation failed, but the test reported success.

This patch adds a wrapper that checks the return code when remote operations are
done, such that we can exit when there is a failure. Track when the bootstrap
VM is created so that we clean up properly if there is a failure.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/scripts/e2e/e2e.sh
+++ b/scripts/e2e/e2e.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Copyright 2018 The Kubernetes Authors.
 #
@@ -25,16 +25,31 @@
 # /root/ssh/.jumphost/jumphost-key
 # /root/ssh/.bootstrapper/bootstrapper-key
 
-# the first argument can be empty or string:"ovf"
+set -o errexit   # exits immediately on any unexpected error (does not bypass traps)
+set -o nounset   # will error if variables are used without first being defined
+set -o pipefail  # any non-zero exit code in a piped command causes the pipeline to fail with that code
+
+# VM_CREATED is set to "1" when the VM has been successfully created.
+VM_CREATED=
+
+# CONTEXT is set to "prow" when the env var PROW_JOB_ID is set, otherwise set
+# to "debug"
+CONTEXT=
+
+# TARGET_VM_PRE contains a string to uniquely indentify the newly created VM(s)
+TARGET_VM_PRE=
+
+# RANDOM_STR holds a random string
+RANDOM_STR=""
 
 get_random_str() {
-   if [ -z "$random_str" ]; then
-      random_str=$(tr -dc 'a-z0-9' < /dev/urandom | fold -w 8 | head -n 1)
+   if [ -z "${RANDOM_STR}" ]; then
+      RANDOM_STR=$(date | { md5sum || md5; } 2>/dev/null | cut -c 1-8)
    fi
 }
 
 fill_file_with_value() {
-  newfilename="$(echo "$1" | sed 's/template/yml/g')"
+  newfilename="${1//template/yml}"
   rm -f "$newfilename" temp.sh
   ( echo "cat <<EOF >$newfilename";
     cat "$1";
@@ -63,8 +78,8 @@ delete_vm() {
    if [ "$vm" = "" ]; then
       vm=$(govc find / -type m -name "$1")
    fi
-   govc vm.power -off "${vm}"
    govc vm.destroy "${vm}"
+   unset VM_CREATED
 }
 
 get_bootstrap_vm() {
@@ -86,6 +101,7 @@ get_bootstrap_vm() {
       echo "bootstrap vm ip is empty"
       exit 1
    fi
+   VM_CREATED=1
    echo "bootstrapper VM ip: ${bootstrap_vm_ip}"
 }
 
@@ -135,16 +151,22 @@ start_docker() {
    done
 }
 
+on_exit() {
+  [ "${VM_CREATED}" ] || return 0
+  get_bootstrap_vm "${CONTEXT}"
+  delete_vm "${TARGET_VM_PRE}"
+}
+
 # the main loop
+trap on_exit EXIT
 vsphere_controller_version=""
-context=""
 if [ -z "${PROW_JOB_ID}" ] ; then
-   context="debug"
+   CONTEXT="debug"
    start_docker
    vsphere_controller_version=$(shell git describe --exact-match 2> /dev/null || \
       git describe --match="$(git rev-parse --short=8 HEAD)" --always --dirty --abbrev=8)
 else
-   context="prow"
+   CONTEXT="prow"
    if [ -z "${PULL_PULL_SHA}" ] ; then
       # for periodic job
       vsphere_controller_version="${PROW_JOB_ID}"
@@ -155,20 +177,22 @@ else
 fi
 
 export VERSION="${vsphere_controller_version}"
-make ci-push
-cd ./scripts/e2e/bootstrap_job && make push && cd .. || exit 1
 echo "build vSphere controller version: ${vsphere_controller_version}"
 
 # set target cluster vm name prefix
 get_random_str
-target_vm_prefix="clusterapi-""$random_str"
-export_base64_value "TARGET_VM_PREFIX" "$target_vm_prefix"
+TARGET_VM_PRE="clusterapi-""${RANDOM_STR}"
+export_base64_value "TARGET_VM_PREFIX" "${TARGET_VM_PRE}"
 
 # install_govc
 go get -u github.com/vmware/govmomi/govc
 
+# Push new container images
+make ci-push
+cd ./scripts/e2e/bootstrap_job && make push && cd .. || exit 1
+
 # get bootstrap VM
-get_bootstrap_vm "$context"
+get_bootstrap_vm "${CONTEXT}"
 
 # bootstrap with kind
 export bootstrap_vm_ip="${bootstrap_vm_ip}"
@@ -191,8 +215,6 @@ run_cmd_on_bootstrap "${bootstrap_vm_ip}" "kubectl --kubeconfig ${kubeconfig_pat
 run_cmd_on_bootstrap "${bootstrap_vm_ip}" 'bash -s' < wait_for_job.sh
 ret="$?"
 
-# cleanup
-get_bootstrap_vm "$context"
-delete_vm "$target_vm_prefix"
+# cleanup done automatically by trap
 
 exit "${ret}"


scripts/e2e/e2e.sh
```diff
@@ -94,6 +93,18 @@ export_base64_value() {
    export "$1"="$base64_value"
 }
 
+exit_cleanup_on_err() {
+  "$@"
```
akutz: Hi @codenrhoden,

We really need to employ shellcheck on this script...Anyway, this should be rewritten as (please note this depends on the comment [above]()):

```shell
if ! eval "${@}"; then
  local exit_code="${?}"
  if [ "${VM_CREATED}" ]; then
    get_bootstrap_vm "${context}"
    delete_vm "${target_vm_prefix}"
  fi
  exit "${exit_code}"
fi
```

scripts/e2e/e2e.sh
```diff
@@ -86,6 +84,7 @@ get_bootstrap_vm() {
       echo "bootstrap vm ip is empty"
       exit 1
    fi
+   vm_created="true"
```
akutz: Hi @codenrhoden,

This is a global variable since it is consumed in `exit_cleanup_on_err` and so should be upper case and defined earlier with a comment:

```shell
# VM_CREATED is set to "true" when the VM has been successfully created.
VM_CREATED=
```

Then it should be assigned, well, anything when created:

```shell
   VM_CREATED=1
```

scripts/e2e/e2e.sh
```diff
@@ -138,6 +149,7 @@ start_docker() {
 # the main loop
 vsphere_controller_version=""
 context=""
+vm_created="false"
```
akutz: Hi @codenrhoden,

`/vm_created="false"/unset VM_CREATED/`

codenrhoden:
@akutz 

> Maybe not in this, but we need to implement shellcheck on these scripts

Agreed. I've actually run it before, and this script only complains about a couple of things. One that I'm not sure how to fix yet, which is why I haven't done it. =) I did run it to make sure I didn't add any new issues, and it caught a couple of things I fixed before submitting. I wonder if you are using different options than me. When I run it on the version of the script in this PR, all I get is this:

```
$ shellcheck --version
ShellCheck - shell script analysis tool
version: 0.6.0
---
$ shellcheck scripts/e2e/e2e.sh

In scripts/e2e/e2e.sh line 92:
   base64_value=$(echo -n "$2" | base64 -w 0)
                       ^-- SC2039: In POSIX sh, echo flags are undefined.


In scripts/e2e/e2e.sh line 157:
      git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
                           ^-----------------------------^ SC2046: Quote this to prevent word splitting.
```

I'd much rather take the `trap` route, as it is much cleaner. I'll give that a shot.

akutz:
Hi @codenrhoden,

> ```shell
> In scripts/e2e/e2e.sh line 92:
>    base64_value=$(echo -n "$2" | base64 -w 0)
>                        ^-- SC2039: In POSIX sh, echo flags are undefined.
> ```

The `echo` command doesn't include the `-n` flag for POSIX-compliant shells. The `-n` flag says not to emit the trailing `\n` character. Instead do `$(printf '%s' "${2}" | base64 -w0)`

> ```shell
> In scripts/e2e/e2e.sh line 157:
>      git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
>                           ^-----------------------------^ SC2046: Quote this to prevent word splitting.
> ```

Change this line to:

```shell
 git describe --match="$(git rev-parse --short=8 HEAD)" --always --dirty --abbrev=8)
```

akutz:
Hi @codenrhoden,

>  I wonder if you are using different options than me. 

As it turns out, no. I get the same warnings as you. I guess I'm just used to encountering a lot more of the program's advice due to my crappy scripts, so I've been forced to learn a lot more conventional style/guidelines related to shell scripting. Hence the above suggestions.

codenrhoden:
> Hi @codenrhoden,
> 
> > I wonder if you are using different options than me.
> 
> As it turns out, no. I get the same warnings as you. I guess I'm just used to encountering a lot more of the program's advice due to my crappy scripts, so I've been forced to learn a lot more conventional style/guidelines related to shell scripting. Hence the above suggestions.

Or maybes it's an `sh` vs `bash` thing? I'm not sure. I wasn't meaning to come off so dense that you had to show me how to fix each error, more that I was surprised when you said some new things need to be rewritten in order to satisfy shellcheck, when it seemed like it was pretty minor. 

I think I'll open a separate PR just to fix the shellcheck issues to make things cleaner.

akutz:
Hi Travis,

Sorry. I was scanning a lot of PRs and probably didn’t take long enough to read yours. You didn’t come off dense. If anything l should have paid more attention. Thanks again for handling this!

codenrhoden:
I moved the shellcheck fixes to https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/423. I'm going to add the exit trap now, and will change the global variables.

scripts/e2e/e2e.sh
```diff
@@ -135,16 +149,22 @@ start_docker() {
    done
 }
 
+on_exit() {
+  [ "${VM_CREATED}" ] || return 0
```
codenrhoden: @akutz One question for this part... Will this `return 0` effect the exit code at all? Meaning if something wrong happened early on, the VM was never created (so `VM_CREATED` is not set), but we want the script to exit with a non-zero return code, will this keep that from happening? I don't think it will since this is a `return` not an `exit`, but just wanted to make sure.

codenrhoden:
@akutz  Okay, I updated this to use a trap and incorporated your other suggestions. There are *a lot* of other global variables in use... I didn't changed them all to uppercase just yet, just the ones I was using in this PR.

scripts/e2e/e2e.sh
akutz: I've literally no idea. Let's see:

```shell
$ cat <<EOF | bash -s; echo "${?}"
on_exit() { exit 1; }
trap on_exit EXIT
EOF
1
```

```shell
cat <<EOF | bash -s; echo "${?}"
on_exit() { return 1; }
trap on_exit EXIT
EOF
0
```

Looks like the delegate can return a non-zero value and not affect the exit code.

codenrhoden:
Also, looks like `pipefail` is not an option for `sh` scripts... :/

Why aren't we just using bash? :)

akutz:
> Also, looks like `pipefail` is not an option for `sh` scripts... :/
> 
> Why aren't we just using bash? :)

Make the change :) (add the header too if you don't mind)

codenrhoden:
> Make the change :) (add the header too if you don't mind)

Sorry, what header?

akutz:
> > Make the change :) (add the header too if you don't mind)
> 
> Sorry, what header?

I meant the license header, but I realized it was already there.

codenrhoden:
Well this is getting fun. The `pipefail` option is making the command `RANDOM_STR=$(tr -dc 'a-z0-9' < /dev/urandom | fold -w 8 | head -n 1)` fail... 

```
+ get_random_str
+ '[' -z '' ']'
++ tr -dc a-z0-9
++ head -n 1
++ fold -w 8
+ RANDOM_STR=7sc7k3q6
+ on_exit
+ '[' '' ']'
+ return 0
+ EXIT_VALUE=141
```

I'm not sure why one of the pipe stages returns non-zero, but since I have to `Ctrl-C` out of the `tr` command, I always get rc 130, which is different than 141... But that's on Mac and not Linux... regardless, I'm thinking of just switching to ` RANDOM_STR=$(date | { md5sum || md5; } 2>/dev/null | cut -c 1-8)`. It should be just as effective for our needs.

To help debug this, I also moved the `make` commands that push new images be just before creating the bootstrap VM. That means the images aren't built and pushed until we actually spin up a VM that needs them. That made the latest version of the test fail in 18 seconds instead of several minutes...

codenrhoden:
Only remaining issue now is with the cleanup. It goes to delete the VM for the target cluster, but can't find it:

```
+ delete_vm Y2x1c3RlcmFwaS05NTdhMmZhMg==
++ govc find / -type m -name 'Y2x1c3RlcmFwaS05NTdhMmZhMg==-*'
+ vm=
+ '[' '' = '' ']'
++ govc find / -type m -name Y2x1c3RlcmFwaS05NTdhMmZhMg==
+ vm=
+ govc vm.power -off ''
govc: vm '' not found
+ EXIT_VALUE=1
```

As expected, turning on `errexit` is uncovering lots of issues.

I'm still trying to figure out what is happening at the end here. After the target cluster is created via `clusterctl`, we aren't doing any sort of cleanup: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/c41d7ce144227067b8a8c8732e288f8579f4e7fa/scripts/e2e/bootstrap_job/clusterctl.sh#L67-L82

That's a different issue to fix that we were already aware of. Since we aren't doing `clusterctl delete`, the VM that is created as part of `clusterctl create` is still hanging around, and it is that VM that `e2e.sh` is trying to delete at the end but can't find. However, I feel like if this was a long-standing problem, we'd see dozens of old VMs hanging around. I'm trying to check if that's the case, but can't get into VMC to look at the moment.

akutz:
> Only remaining issue now is with the cleanup. It goes to delete the VM for the target cluster, but can't find it:
> 
> ```
> + delete_vm Y2x1c3RlcmFwaS05NTdhMmZhMg==
> ++ govc find / -type m -name 'Y2x1c3RlcmFwaS05NTdhMmZhMg==-*'
> + vm=
> + '[' '' = '' ']'
> ++ govc find / -type m -name Y2x1c3RlcmFwaS05NTdhMmZhMg==
> + vm=
> + govc vm.power -off ''
> govc: vm '' not found
> + EXIT_VALUE=1
> ```
> 
> As expected, turning on `errexit` is uncovering lots of issues.
> 
> I'm still trying to figure out what is happening at the end here. After the target cluster is created via `clusterctl`, we aren't doing any sort of cleanup:
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/c41d7ce144227067b8a8c8732e288f8579f4e7fa/scripts/e2e/bootstrap_job/clusterctl.sh#L67-L82
> 
> That's a different issue to fix that we were already aware of. Since we aren't doing `clusterctl delete`, the VM that is created as part of `clusterctl create` is still hanging around, and it is that VM that `e2e.sh` is trying to delete at the end but can't find. However, I feel like if this was a long-standing problem, we'd see dozens of old VMs hanging around. I'm trying to check if that's the case, but can't get into VMC to look at the moment.

Agreed. A few notes:

* `govc vm.destroy` powers a VM off for you. You don't need to power it off first.
* The name `Y2x1c3RlcmFwaS05NTdhMmZhMg==` is base64-encoded. Are you sure the scripts decode that before trying to find the VM based on the name?

codenrhoden:
> Agreed. A few notes:
>govc vm.destroy powers a VM off for you. You don't need to power it off first.

true. This is not a change I made. But I might as well remove it.
>The name Y2x1c3RlcmFwaS05NTdhMmZhMg== is base64-encoded. Are you sure the scripts decode that before trying to find the VM based on the name?

Also not something that should have changed... but that's been my same line of thinking. That shouldn't be the actual name of the VM. I'll keep looking.

codenrhoden:
> > The name Y2x1c3RlcmFwaS05NTdhMmZhMg== is base64-encoded. Are you sure the scripts decode that before trying to find the VM based on the name?
> 
> Also not something that should have changed... but that's been my same line of thinking. That shouldn't be the actual name of the VM. I'll keep looking.

Oh wait, I know what has happened. When I made `target_vm_prefix` become `TARGET_VM_PREFIX` because it's global, that name is actually now clashing with the `TARGET_VM_PREFIX` env var that gets created with a call to `export_base64_value`. So, it is a new problem that I created. I can fix that at least. =)

scripts/e2e/e2e.sh
```diff
@@ -25,16 +25,32 @@
 # /root/ssh/.jumphost/jumphost-key
 # /root/ssh/.bootstrapper/bootstrapper-key
 
-# the first argument can be empty or string:"ovf"
+set -o errexit   # exits immediately on any unexpected error (does not bypass traps)
+set -o nounset   # will error if variables are used without first being defined
+set -o pipefail  # any non-zero exit code in a piped command causes the pipeline to fail with that code
+set -x
```
codenrhoden: I will remove this before merge. Just needed it for debugging.

codenrhoden:
@akutz okay, this should finally be in good shape. =) Thanks for all your help with it.

PR #423, opened 2019-07-09
Fix shellcheck issues for scripts/e2e/e2e.sh

codenrhoden:
Fix shellcheck issues for scripts/e2e/e2e.sh

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/scripts/e2e/e2e.sh
+++ b/scripts/e2e/e2e.sh
@@ -90,7 +90,7 @@ get_bootstrap_vm() {
 }
 
 export_base64_value() {
-   base64_value=$(echo -n "$2" | base64 -w 0)
+   base64_value=$(printf '%s' "${2}" | { base64 -w0 || base64; } 2>/dev/null)
    export "$1"="$base64_value"
 }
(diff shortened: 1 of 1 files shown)


scripts/e2e/e2e.sh
```diff
@@ -90,7 +90,7 @@ get_bootstrap_vm() {
 }
 
 export_base64_value() {
-   base64_value=$(echo -n "$2" | base64 -w 0)
+   base64_value=$(printf '%s' "${2}" | base64 -w0)
```
figo: thanks for making the change.

is it possible to make `base64` works the same on MacOS and Linux?
currently `base64 -w` does not work on MacOS, i remember you did something similar in other change.

codenrhoden: yeah I thought about doing that here too... I wasn't sure if this script was ever run on MacOS or not. but yeah, I can do that.

codenrhoden:
@figo the `base64` change is now in place.

PR #432, opened 2019-07-12
Cluster actuator needs delete permission for secret.

vijaykatam:
Cluster delete is failing because  delete secret permission is missing.

Drive-by: Setters for extra config are returning errors which are not being handled, fix by
removing error return as it never returns error.

Cluster delete is failing because secret delete permission is missing.

**Release note**:

```release-note
Add delete secret permission for cluster actuator.
```

### DIFF
--- a/pkg/cloud/vsphere/actuators/cluster/actuator.go
+++ b/pkg/cloud/vsphere/actuators/cluster/actuator.go
@@ -45,7 +45,7 @@ import (
 
 //+kubebuilder:rbac:groups=vsphere.cluster.k8s.io,resources=vsphereclusterproviderspecs;vsphereclusterproviderstatuses,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=cluster.k8s.io,resources=clusters;clusters/status,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=,resources=secrets,verbs=create;get;list;watch
+//+kubebuilder:rbac:groups=,resources=secrets,verbs=create;get;list;watch;delete
 //+kubebuilder:rbac:groups="",resources=nodes;events;configmaps,verbs=get;list;watch;create;update;patch;delete
 
 // Actuator is responsible for maintaining the Cluster objects.
--- a/pkg/cloud/vsphere/services/govmomi/extra/config.go
+++ b/pkg/cloud/vsphere/services/govmomi/extra/config.go
@@ -27,7 +27,7 @@ type Config []types.BaseOptionValue
 
 // SetCloudInitUserData sets the cloud init user data at the key
 // "guestinfo.userdata" as a base64-encoded string.
-func (e *Config) SetCloudInitUserData(data []byte) error {
+func (e *Config) SetCloudInitUserData(data []byte) {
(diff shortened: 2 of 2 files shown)


pkg/cloud/vsphere/services/govmomi/extra/config.go
```diff
@@ -27,7 +27,7 @@ type Config []types.BaseOptionValue
 
 // SetCloudInitUserData sets the cloud init user data at the key
 // "guestinfo.userdata" as a base64-encoded string.
-func (e *Config) SetCloudInitUserData(data []byte) error {
+func (e *Config) SetCloudInitUserData(data []byte) {
```
akutz: Hi @vijaykatam,

Please remove this and the other function signature change as they are not integral to this patch. Feel free to open a separate PR for the functions. Thank you!

pkg/cloud/vsphere/services/govmomi/extra/config.go
```diff
@@ -38,13 +38,11 @@ func (e *Config) SetCloudInitUserData(data []byte) error {
...
 		},
 	)
-
-	return nil
 }
 
 // SetCloudInitMetadata sets the cloud init user data at the key
 // "guestinfo.metadata" as a base64-encoded string.
-func (e *Config) SetCloudInitMetadata(data []byte) error {
+func (e *Config) SetCloudInitMetadata(data []byte) {
```
akutz: Hi @vijaykatam,

Please remove this and the other function signature change as they are not integral to this patch. Feel free to open a separate PR for the functions. Thank you!

pkg/cloud/vsphere/actuators/cluster/actuator.go
```diff
@@ -45,7 +45,7 @@ import (
 
 //+kubebuilder:rbac:groups=vsphere.cluster.k8s.io,resources=vsphereclusterproviderspecs;vsphereclusterproviderstatuses,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=cluster.k8s.io,resources=clusters;clusters/status,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=,resources=secrets,verbs=create;get;list;watch
+//+kubebuilder:rbac:groups=,resources=secrets,verbs=create;get;list;watch;delete
```
akutz: Hi @vijaykatam,

Great catch, thank you! Please make sure you run `make` ahead of submitting the PR. This will regenerate the RBAC files that need to be part of the commit. Additionally, the bootstrap spec will need to be updated as well. To update the bootstrap spec, run `make` and then do the following:

```shell
hack/generate-yaml.sh -f -d -c ci -m '$VSPHERE_CONTROLLER_VERSION' && \
cp -f out/ci/provider-components.yaml scripts/e2e/bootstrap_job/spec/provider-components.template
```

Then, with your editor of choice, please modify `scripts/e2e/bootstrap_job/spec/provider-components.template` by adding `-v=6` as an argument for the CAPV manager's pod spec. For example:

```shell
$ diff out/ci/provider-components.yaml scripts/e2e/bootstrap_job/spec/provider-components.template 
1183a1184
>         - -v=6
```

PR #386, opened 2019-06-28
properly log requeue error messages with klogr

andrewsykim:
Signed-off-by: Andrew Sy Kim <kiman@vmware.com>

**What this PR does / why we need it**:
k8s.io/klogr does not log raw error values (it calls `json.Marshal` on the value) so we have to explicitly call `err.Error()` to pass the string value. 

Before:
```
I0628 14:43:20.470520       1 actuators.go:59] [cluster-actuator]/cluster.k8s.io/v1alpha1/default/workload-log-err/workload-log-err-controlplane-1 "level"=6 "msg"="requeue after error"  "error"={} "object"="default/workload-log-err"
```

After:
```
I0628 14:46:56.782602       1 actuators.go:59] [cluster-actuator]/cluster.k8s.io/v1alpha1/default/workload-log-err/workload-log-err-controlplane-1 "level"=6 "msg"="requeue after error"  "error"="unable to get control plane status for cluster \"default/workload-log-err\": unable to get kubeconfig for cluster \"default/workload-log-err\": unable to get control plane endpoint: requeue in: 20s" "object"="default/workload-log-err"
```

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note

```

PR #445, opened 2019-07-19
Bug fixes and improvements for v0.3.1

andrewsykim:

* cherry-picks https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/386 
* uses empty string for CRI socket which allows kubeadm to determine the CRI socket
* logs when the provider manager is started 

**Release note**:

```release-note
* cherry-picks https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/386 
* uses empty string for CRI socket which allows kubeadm to determine the CRI socket
* logs when the provider manager is started 
```

### DIFF
--- a/pkg/cloud/vsphere/services/govmomi/create.go
+++ b/pkg/cloud/vsphere/services/govmomi/create.go
@@ -39,8 +39,8 @@ const (
 	// hostnameLookup resolves via cloud init and uses cloud provider's metadata service to lookup its own hostname.
 	hostnameLookup = "{{ ds.meta_data.hostname }}"
 
-	// containerdSocket is the path to containerd socket.
-	containerdSocket = "/var/run/containerd/containerd.sock"
+	// criSocket is the path to the CRI socket to connect. If empty kubeadm will try to auto-detect this value.
+	criSocket = ""
 
 	// nodeRole is the label assigned to every node in the cluster.
 	nodeRole = "node-role.kubernetes.io/node="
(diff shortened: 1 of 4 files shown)


pkg/cloud/vsphere/services/govmomi/create.go
```diff
@@ -39,8 +39,8 @@ const (
 	// hostnameLookup resolves via cloud init and uses cloud provider's metadata service to lookup its own hostname.
 	hostnameLookup = "{{ ds.meta_data.hostname }}"
 
-	// containerdSocket is the path to containerd socket.
-	containerdSocket = "/var/run/containerd/containerd.sock"
+	// criSocket is the path to the CRI socket to connect. If empty kubeadm will try to auto-detect this value.
+	criSocket = ""
```
andrewsykim: @detiber would love your thoughts on this - it would allow machine images that prefer docker as the container runtime.

akutz:
Hi @andrewsykim,

I believe e2e failed due to the socket file change. FWIW, I think I tried this once, and if the socket isn't specified, kubeadm has issues initializing.

pkg/cloud/vsphere/services/govmomi/create.go
detiber: Longer term, my thoughts here is that we could add fields like this to a centralized location in cluster-api, which would then be able to impact both the bootstrapping behavior (setting the right kubeadm configuration) and also the right image lookup behavior, so that both configurations would match with less likelihood of a user shooting themselves in the foot.
/cc @ncdc @vincepri

ncdc: Also see https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/863/files for how we adjusted CAPA to default to containerd but allow the user to override.

andrewsykim:
/close

Closing this for now since most of the fixes will be in v0.4.0 which is being released soon

PR #448, opened 2019-07-22
The manifests command

akutz:
This patch introduces a new command, `manifests`, used to produce the manifests necessary to use CAPV. This command may also be extended/reused by other CAPI providers to do the same.

The manifests command links to Kustomize directly, therefore removing the need for an external kustomize binary. Additionally, the manifest command will process any files in a kustomization root that end with `.template` and interpolate them against a data structure, injecting the result into a virtual filesystem. Therefore a kustomization can link to a patch that doesn't exist as long as the equivalent template file does.

- [x] Improve the `README`
- [x] Generate `machineset.yaml`
- [x] Generate `addons.yaml`

**Release note**:

```release-note
* The "manifests" command is used to generate YAML manifests for the CAPI provider
```


cmd/manifests/README.md
```diff
@@ -0,0 +1,88 @@
...
+```shell
+$ ls -al
+-rw-r--r--  1 akutz  staff    19K Jul 23 12:00 addons.yaml
+-rw-r--r--  1 akutz  staff   484B Jul 23 12:00 cluster.yaml
+-rw-r--r--  1 akutz  staff   640B Jul 23 12:00 machines.yaml
+-rw-r--r--  1 akutz  staff   936B Jul 23 12:00 machineset.yam
+-rw-r--r--  1 akutz  staff    99K Jul 23 12:00 provider-components.yaml
+```
+
+For a full list of the flags and options, use `./manifests -?`.
```
andrewsykim: why not `manifests --help` like most other CLIs?

cmd/manifests/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Virtual patches
+
+The `manifests` program introduces the concept of _virtual patches_ for Kustomize.
+
+_A virtual patch is any file located in a kustomization root that ends with `.template`._
```
andrewsykim: What is "kustomization root" exactly? A directory with a `kustmization.yaml` file?

cmd/manifests/README.md
```diff
@@ -0,0 +1,88 @@
...
+2. Define a type that adheres to the following interface:
+
+    ```golang
+    type Provider interface {
+        GetClusterProviderSpec() (runtime.Object, error)
+        GetMachineProviderSpec() (runtime.Object, error)
+    }
+    ```
+
+3. Call `app.Run(myProvider)`
```
andrewsykim: Worth calling out the flag definitions too?

cmd/manifests/pkg/kustomize/kustomize.go
```diff
@@ -0,0 +1,264 @@
...
+import (
+	"bytes"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"text/template"
+
+	"github.com/pkg/errors"
+	"sigs.k8s.io/kustomize/k8sdeps/kunstruct"
```
andrewsykim: Are these imports meant to be importable outside of sigs.k8s.io/kustomize?

config/default/manager_log_level_patch.yaml.template
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - args:
+        - --logtostderr
+        - --v={{ .ManagerLogLevel }}
```
andrewsykim: I feel something like this should have a default though?

andrewsykim: Ah I see, the default is specified in the flag

akutz:
I am closing this and will likely re-open it as some form of `clustervsphereadm`. For whatever reason no one in the CAPI or CAPV communities seemed interested in looking at the work :(

PR #451, opened 2019-07-24
Change e2e script/job to use new image locations

codenrhoden:
This patch updates e2e.sh and related pieces to use images located at
gcr.io/cluster-api-provider-vsphere/pr instead of the previous location.

It also eliminates the need for a temporary 'clusterctl' container image
as part of the e2e process, because that binary is now being pushed as
part of the automated builds. Therefore we just download the binary
instead of building and pushing a new image just to copy the binary out
of it.

Fixes #

@akutz You'll see in the `e2e.sh` script the use of `-k` with `release.sh`. This is why I needed that change in Prow. The e2e tests at this stage need all of the env var sets from that first secret, but that secret also sets `GCR_KEY_FILE`, which would make the `release.sh` script fail... I need all of this stuff to work first and then we can remove `GCR_KEY_FILE` from the first set of exports and then remove this.

I realize I have missed a step, though. I didn't add the preset with the key file to the e2e test. I need to do that before this test will pass. Darn. Back to Prow!

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,10 +26,6 @@ CWD := $(abspath .)
 #      nested GOPATH.
 SHELL := hack/shell-with-gopath.sh
 
-# Image URL to use all building/pushing image targets
-CI_IMG ?= gcr.io/cnx-cluster-api/vsphere-cluster-api-provider
-CLUSTERCTL_CI_IMG ?= gcr.io/cnx-cluster-api/clusterctl
-
 # Retrieves the git hash
 VERSION ?= $(shell git describe --always --dirty)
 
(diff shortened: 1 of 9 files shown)


codenrhoden:
Well that's weird, seeing this:

```
Activated service account credentials for: [capv-prow@cluster-api-provider-vsphere.iam.gserviceaccount.com]
pushing gcr.io/cluster-api-provider-vsphere/pr/manager:v0.4.0-alpha.0-8-gd063ce6c
The push refers to repository [gcr.io/cluster-api-provider-vsphere/pr/manager]
ad1c61ec633f: Preparing
01092e5921c5: Preparing
unauthorized: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: https://cloud.google.com/container-registry/docs/advanced-authentication
WARNING: [capv-prow@cluster-api-provider-vsphere.iam.gserviceaccount.com] appears to be a service account. Service account tokens cannot be revoked, but they will expire automatically. To prevent use of the service account token earlier than the expiration, revoke the parent service account or service account key.
Revoked credentials:
 - capv-prow@cluster-api-provider-vsphere.iam.gserviceaccount.com
                       Credentialed Accounts
```

But that account should work just fine. I'll have to debug a bit.

codenrhoden:
/retest
Last failure was a timeout on the image pull, and we see that somewhat regularly, so trying again.

If no progress there, I'll submit a second PR that fixes the push on it's own, and that will make the regular post-submit pushes work. No reason that couldn't be working now.

codenrhoden:
The latest E2E job appears stuck... I've never seen that. 25 minutes and no job log. hopefully there is a timeout for it or something? I have no idea.

codenrhoden:
Success!! 🎉

codenrhoden:
If this looks good and goes in first, then #460 is not necessary.

PR #456, opened 2019-07-24
Fix issue generating manifests due to Docker perms

### LINKED ISSUE #453: Generated manifests should be modifiable on the local host
/kind feature

The Getting Started guide has the following command:

```shell
docker run --rm \
  -v "$(pwd)/management-cluster":/out \
  -v "$(pwd)/envvars.txt":/out/envvars.txt:ro \
  gcr.io/cluster-api-provider-vsphere/release/manifests:latest \
  -c management-cluster
```

However, the above command produces manifests with a permissions level that prevents modification on the host.

The manifests should be modifiable on the host.

The fixed command is:

```shell
docker run --rm \
  --user "$(id -u):$(id -g)" \
  -v "$(pwd)/management-cluster":/out \
  -v "$(pwd)/envvars.txt":/out/envvars.txt:ro \
  gcr.io/cluster-api-provider-vsphere/release/manifests:latest \
  -c management-cluster
```

However, this also requires that the out directory be writeable by the provided user and group inside of the container.

akutz:
This patch fixes an issue with generating manifests using the manifests image due to Docker permissions.

Fixes #453

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/tools/generate-yaml/Dockerfile
+++ b/hack/tools/generate-yaml/Dockerfile
@@ -49,6 +49,9 @@ COPY ./vendor/sigs.k8s.io/cluster-api/config ./vendor/sigs.k8s.io/cluster-api/co
 # Copy in the examples directory.
 COPY ./cmd/clusterctl/examples/vsphere/*.template ./cmd/clusterctl/examples/vsphere/
 
+# Ensure all the directories in the /build directory are writeable.
+RUN find . -type d -exec chmod 0777 \{\} \;
+
 # This build arg specifies the value of the -m flag for generate-yaml.sh, the
 # CAPV manager image inserted into the generated provider components.
 ARG CAPV_MANAGER_IMAGE=gcr.io/cnx-cluster-api/vsphere-cluster-api-provider:latest
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -94,6 +94,7 @@ With the above environment variable file it is now possible to generate the mani
 
 ```shell
 $ docker run --rm \
+  --user "$(id -u):$(id -g)" \
   -v "$(pwd)/management-cluster":/out \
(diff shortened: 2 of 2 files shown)


hack/tools/generate-yaml/Dockerfile
```diff
@@ -49,6 +49,9 @@ COPY ./vendor/sigs.k8s.io/cluster-api/config ./vendor/sigs.k8s.io/cluster-api/co
 # Copy in the examples directory.
 COPY ./cmd/clusterctl/examples/vsphere/*.template ./cmd/clusterctl/examples/vsphere/
 
+# Ensure all the directories in the /build directory are writeable.
+RUN find . -type d -exec chmod 0777 \{\} \;
```
andrewsykim: Maybe I'm missing something, but this updates `/build` to have 0777 but the yaml files are still generated under `/out`.

akutz: Because before the user couldn't write the generated manager image patch into the directory since they were `root:root 0755`. This allows the mapped in `user:group` to write to `/out`

andrewsykim: But `/out` is still not accessible I think. If you mount the parent directory and let `generate-yamal.sh` create the directory it works:
```
docker run --rm   --user "$(id -u):$(id -g)" -e OUT_DIR=/build/capv/out/   -v "$(pwd)/":/build/capv/   -v "$(pwd)/envvars.txt":/out/envvars.txt:ro   gcr.io/cluster-api-provider-vsphere/release/manifests:latest   -c management-cluster
```

But as long as `/out` is being mounted directly it is owned by root (I think).

akutz: *sigh*

For those watching this, it turns out this behaves differently whether you're running Docker on Mac or Linux. Fun.

PR #462, opened 2019-07-25
Create base images for bootstrap and manifests

codenrhoden:
Both the bootstrap and manifests container image install OS packages, and since
these images are built every time during E2E tests, that is time
consuming. We can speed this up by creating base images to build from,
only pulling in what is needed and not doing OS updates.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/hack/tools/bootstrap-base/Dockerfile
+++ b/hack/tools/bootstrap-base/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


codenrhoden:
/assign @akutz 
When I was  debugging my issues with E2E yesterday, I noticed that a significant amount of time was spent building the manifests and bootstrap images. This was because they both do OS package installs. But those package installs don't change from run to run, so this could be done once in a base image that is re-used, speeding up E2E for us.

My plan here would be to:

 - have this PR merged
 - with this PR merged, build and push the base images
 - Open a new PR that consumes those base images

akutz:
Hi @codenrhoden,

Love it. Thanks for this!

/lgtm
/approve

codenrhoden:
@akutz The E2E test appears stuck again. Weird - second time in 24 hours. Last night I got it unstuck by doing a force-push, which killed and restarted the job. I'll wait a bit more, but if doesn't move along I may do that here. I'll fix the location of the `kustomize`binary and re-push. Then I think it would need another `/lgtm`. I've never done that after something went into the merge pool. 🤷‍♂  But hopefully none of that will be needed.

PR #463, opened 2019-07-25
Use base images for bootstrap and manifests images

codenrhoden:
Opening this PR as an example of how we would use the base images in #462 

Fixes #

**Release note**:

```release-note

```
/hold

### DIFF
--- a/hack/tools/generate-yaml/Dockerfile
+++ b/hack/tools/generate-yaml/Dockerfile
@@ -13,28 +13,11 @@
 # limitations under the License.
 
 ################################################################################
-##                            DOWNLOAD KUSTOMIZE                              ##
-################################################################################
-FROM debian:stretch-slim as kustomize
-RUN apt-get --assume-no update && apt-get -y install curl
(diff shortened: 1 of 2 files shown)


hack/tools/generate-yaml/Dockerfile
```diff
@@ -12,29 +12,15 @@
...
+FROM ${BASE_IMAGE}
 LABEL "maintainer" "Andrew Kutz <akutz@vmware.com>"
 
-# Install python
-RUN apt-get --assume-no update && apt-get -y install curl python
-
-# Copy kustomize from the kustomize build stage.
-COPY --from=kustomize /kustomize /usr/local/bin/kustomize
+# move kustomize to an executable path
+RUN mv /kustomize /usr/local/bin/kustomize
```
codenrhoden: This step could be avoided if we just do it in the base image instead.

codenrhoden: @akutz I'll leave it for now since the base images PR was approved. Not exactly a time consuming step. :)

PR #466, opened 2019-07-26
Troubleshooting guide

akutz:
This patch adds a troubleshooting guide.

Fixes #

**Release note**:

```release-note
* Added a troubleshooting guide
```

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,0 +1,331 @@
+# Troubleshooting
+
+This is a guide on how to troubleshoot issues related to the Cluster API provider for vSphere (CAPV).
+
+* [Troubleshooting](#troubleshooting)
+  * [Debugging issues](#debugging-issues)
+    * [Bootstrapping with logging](#bootstrapping-with-logging)
+      * [Adjusting log levels](#adjusting-log-levels)
+        * [Adjusting the CAPI manager log level](#adjusting-the-capi-manager-log-level)
+        * [Adjusting the CAPV manager log level](#adjusting-the-capv-manager-log-level)
+        * [Adjusting the `clusterctl` log level](#adjusting-the-clusterctl-log-level)
+      * [Accessing the logs in the bootstrap cluster](#accessing-the-logs-in-the-bootstrap-cluster)
+        * [Exporting the kubeconfig](#exporting-the-kubeconfig)
+        * [Following the CAPI manager logs](#following-the-capi-manager-logs)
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -0,0 +1,341 @@
+# Troubleshooting
+
+This is a guide on how to troubleshoot issues related to the Cluster API provider for vSphere (CAPV).
+
+* [Troubleshooting](#Troubleshooting)
```
andrewsykim: Recommend using https://github.com/ekalinin/github-markdown-toc in the future :)

docs/troubleshooting.md
```diff
@@ -0,0 +1,341 @@
...
+  Host:         1.2.3.4
+```
+
+If the above VM is no longer needed then it may be deleted using `govc` as well:
+
+```shell
+docker run --rm \
+  -e GOVC_URL -e GOVC_USERNAME -e GOVC_PASSWORD \
+  vmware/govc \
+  vm.destroy -k management-cluster-controlplane-1
```
andrewsykim: Is it safe to recommend this without setting DC, ResourcePool, etc?

docs/troubleshooting.md
```diff
@@ -0,0 +1,341 @@
...
+```
+
+#### A static IP address must include the segment length
+
+Another common error is to omit the segment length when using a static IP address. For example:
+
+```yaml
+network:
+  devices:
+  preferredAPIServerCidr: "192.168.5.0/24"
```
andrewsykim: Wrong cidr here?

andrewsykim: or maybe not but it's confusing without defining the 2nd network yet

andrewsykim:
/lgtm
/hold 

Left some non-blocking comments

docs/troubleshooting.md
akutz: I do.

ScreenCapture-2019 07 26-15 29 00

docs/troubleshooting.md
akutz: Fair.

docs/troubleshooting.md
akutz: (I just forgot to update the ToC)

akutz: Rather I use https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one.

akutz:
Hi @andrewsykim,

I addressed all of your feedback. Please take a quick look and re-add the lgtm as you see fit. Thanks!

Issue #469, opened 2019-07-29
Machine actuator errors on multiple network devices even if some are valid


andrewsykim:
/kind bug

**What steps did you take and what happened:**
See this comment https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/465#issuecomment-515565029. The 2nd network device is invalid because it has dhcp disabled but no static IP addrs is defined but the 1st network device should still be used for the control plane endpoint. 

**What did you expect to happen:**
When there are multiple network devices, the machine actuator will error if 1 of N network devices does not return a valid IP address, even though the other network devices do. It should succeed even if 1 network devices returns a valid address that can be used for the control plane endpoint. 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

andrewsykim:
cc @jpweber @akutz 
/assign

PR #470, opened 2019-07-29
Do not error on multiple network devices if at least one has a valid IP address

andrewsykim:

In the case where there are multiple network devices and one of them returns a valid control plane endpoint, the machine actuator could still error if the other network devices don't return an IP address. The actuator should succeed if at least one network device returns a valid address.

Fixes #https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/469

**Release note**:

```release-note
* Relaxes network requirements for VMs to support a multi-network device configuration when only a single network device has a valid IP address
```

### DIFF
--- a/pkg/cloud/vsphere/context/cluster_context.go
+++ b/pkg/cloud/vsphere/context/cluster_context.go
@@ -309,13 +309,14 @@ func (c *ClusterContext) ControlPlaneEndpoint() (string, error) {
 		return "", errors.Wrap(err, "error creating machine context while searching for control plane endpoint")
 	}
 
-	if ipAddr := machineCtx.IPAddr(); ipAddr != "" {
-		controlPlaneEndpoint := net.JoinHostPort(ipAddr, strconv.Itoa(int(machineCtx.BindPort())))
-		machineCtx.Logger.V(2).Info("got control plane endpoint from machine", "control-plane-endpoint", controlPlaneEndpoint)
-		return controlPlaneEndpoint, nil
+	ipAddr, err := machineCtx.IPAddr()
+	if err != nil {
+		return "", errors.Wrap(err, "error getting first IP address for machine")
 	}
 
-	return "", errors.New("unable to get control plane endpoint")
+	controlPlaneEndpoint := net.JoinHostPort(ipAddr, strconv.Itoa(int(machineCtx.BindPort())))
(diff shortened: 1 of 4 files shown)


pkg/cloud/vsphere/context/machine_context.go
```diff
@@ -193,15 +195,15 @@ func (c *MachineContext) IPAddr() string {
...
 		}
 
 		if preferredAPIServerCIDR.Contains(net.ParseIP(nodeAddr.Address)) {
-			return nodeAddr.Address
+			return nodeAddr.Address, nil
 		}
 	}
 
-	return ""
+	return "", ErrNoMachineIPAddr
```
andrewsykim: Technically don't think this error var is needed yet since any error from `IPAddr()` results in a requeue so we don't actually have to check the returned error value yet. With that said it does make the code path and unit tests easier to read so I kept it here

andrewsykim:
/assign @akutz 
cc @jpweber

pkg/cloud/vsphere/context/cluster_context.go
```diff
@@ -309,13 +309,18 @@ func (c *ClusterContext) ControlPlaneEndpoint() (string, error) {
...
-	if ipAddr := machineCtx.IPAddr(); ipAddr != "" {
-		controlPlaneEndpoint := net.JoinHostPort(ipAddr, strconv.Itoa(int(machineCtx.BindPort())))
-		machineCtx.Logger.V(2).Info("got control plane endpoint from machine", "control-plane-endpoint", controlPlaneEndpoint)
-		return controlPlaneEndpoint, nil
+	ipAddr, err := machineCtx.IPAddr()
+	if err != nil {
+		return "", errors.Wrap(err, "error getting first IP address for machine")
+	}
+
+	if ipAddr == "" {
```
akutz: Hi @andrewsykim,

This shouldn't be required since `ipAddr` should never empty without an error.

pkg/cloud/vsphere/context/machine_context_test.go
```diff
@@ -31,6 +31,7 @@ func Test_MachineContextIPAddr(t *testing.T) {
 		name   string
 		ctx    *MachineContext
 		ipAddr string
+		err    error
```
akutz: Hi @andrewsykim,

Please rename to `expectedErr` or `expectedError`.

codenrhoden:
/lgtm
/hold
Hold is for @sidharthsurana to have a chance to take a look if time allows.

Issue #314, opened 2019-06-17
Aligning CAPV with vSphere Cloud Provider


andrewsykim:
As it stands today, there are some behaviors between CAPV and both the in-tree and out-of-tree vSphere cloud provider that need to be aligned. Some examples of this below:

1) Networking configuration - in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/307, CAPV will assign a MAC addr for a VM network prior to the VM powering on. The MAC OUI prefix _needs_ to be acceptable for the cloud provider integration. For the in-tree case, that means the OUI _must_ have a prefix of `00:05:69`, `00:0c:29` , `00:1c:14` or `00:50:56`. For the out-of-tree case the mac addr assigned shouldn't matter.

2) Resource management - today the datacenter, datastore and resource pool a VM is created inside is determined at the machine spec level, however, the vSphere cloud provider may not always recognize the same values set by CAPV. For example, a machine spec may have been provisioned under a certain resource pool but the configured resource pool set on the cloud config file may differ. We need to ensure the requirements for these values between CAPV and the cloud provider are clear. Differing values between the two will break certain features in the cluster.

akutz:
I've spent the last several days working on a PR to address the second item called out in this issue description -- the configuration of the vSphere cloud provider when deployed with CAPV.

This work has solved problems, but also raised new points of discussion. I am writing this comment to set the context for the discussion of how to align the vSphere cloud controller manager and CAPV with respect to the cloud controller's configuration types.

### Configuring the cloud provider
PR #471 introduces a new type to CAPV, `CloudProviderConfiguration`. This may be defined in the CAPV cluster spec like so:

```yaml
apiVersion: "vsphere.cluster.k8s.io/v1alpha1"
kind: "VsphereClusterProviderSpec"
server: "my.vcenter.com"
username: "my.user"
password: "my.pass"
cloudProviderConfiguration:
  global:
    secretName: "cloud-provider-vsphere-credentials"
    secretNamespace: "kube-system"
  virtualCenter:
    my.vcenter.com:
  network:
    name: "sddc-cgw-network-5"
  workspace:
    server: "my.vcenter.com"
    datacenter: "SDDC-Datacenter"
    datastore: "WorkloadDatastore"
    resourcePool: "clusterapi"
    folder: "clusterapi"
```

This neatly aligns the configuration of the cloud provider at the cluster level, as it should be. However, this solution required the following:

[...]

PR #471, opened 2019-07-29
Centralized cloud provider configuration

### LINKED ISSUE #60: Cloud provider config should not come from master config
The cloud provider configuration to be setup for the cluster today is rendered using the machine variables used for the master node. This assumption is not correct as the machine variables are there to provide information about **where** and **what** to deploy as a master node and **_not_** to serve as the input for the cloud provider config that the target kubernetes will use.
We need to clearly abstract out the information intended to be used for generating the cloud provider configuration for the cluster.
One possible place where this cloud provider configuration information could live is in the `Cluster` object's `providerConfig`. Currently it only contains the VC endpoint and credentials. Thus we would need to expand that definition.

akutz:
Introduces support for a centralized cloud provider configuration

Fixes #60

Please see https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/314#issuecomment-516095579 for more information.

**Release note**:

```release-note
* Support for a centralized cloud provider configuration as part of the cluster spec
```


akutz:
Hi @frapposelli / @figo / @sidharthsurana / @andrewsykim,

Please lgtm this if there are no objections. FWIW, this includes the fix for the CP credentials.

figo:
@akutz do we need to make sure the resource management part of cloud provider configure is consistent with VM's resource management configure?  (same resource pool etc)

cc @andrewsykim

akutz:
Hi @figo,

> @akutz do we need to make sure the resource management part of cloud provider configure is consistent with VM's resource management configure? (same resource pool etc)
> 
> cc @andrewsykim

I spoke with Andrew, and the decision for the time being was to ensure that the VMs are located in the same workspace used by the cloud provider.

cmd/clusterctl/examples/vsphere/machineset.yaml.template
```diff
@@ -22,9 +22,6 @@ spec:
           apiVersion: vsphere.cluster.k8s.io/v1alpha1
           kind: VsphereMachineProviderSpec
           datacenter: "${VSPHERE_DATACENTER}"
```
figo: @akutz  it make sense to me now, thx for the explanation, could you help to clarify why `datacenter` kept in both machine spec and cloud provider configure?

akutz: Hi @figo,

Because you can configure multiple datacenter via multiple vCenter definitions in the cloud provider configuration. So technically you could have a K8s cluster with each of its control plane nodes in different datacenters as long as each datacenter had symmetrical pool/datastore/folder structures.

config/crds/vsphere_v1alpha1_vspheremachineproviderspec.yaml
```diff
@@ -342,7 +330,7 @@ spec:
           properties:
             devices:
               description: Devices is the list of network devices used by the virtual
-                machine.
+                machine. TODO(akutz) Make sure at least one network matches the             ClusterSpec.CloudProviderConfiguration.Network.Name
```
figo: remove the extra space?

pkg/apis/vsphere/v1alpha1/cloud/types.go
```diff
@@ -0,0 +1,199 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package cloud contains API types for the vSphere cloud provider.
+package cloud
+
+// Config is the vSphere cloud provider's configuration.
+type Config struct {
```
figo: this probably already been discussed: what is the plan to keep in-sync with cloud provider types? 
is vendor-in types from cloud provider repo in the long term plan? thanks

config/crds/vsphere_v1alpha1_vspheremachineproviderspec.yaml
akutz: That can be addressed in a future PR/patch. Thanks!

pkg/apis/vsphere/v1alpha1/cloud/types.go
akutz: That's correct. On both counts. Eventually these types will move into the CCM codebase.

figo:
/lgtm
/hold  
hold to give others chance to review, but feel free to cancel the hold

pkg/apis/vsphere/v1alpha1/cloud/types.go
frapposelli: Can you open an issue in the CCM repo for this?

andrewsykim:
Sorry I'm a bit late to the party, but this LGTM!

PR #476, opened 2019-08-01
Bugfix: Double base64 encoding of cloud provider credentials

thorsteinnth:
This PR contains a bugfix.

When creating the cloud provider credentials secret for target clusters, the username and password were being base64 encoded.

Since `StringData` is being used, this is not necessary. Instead, the already base64 encoded data was base64 encoded again.

This PR fixes this bug by passing the username and password directly to `StringData`.

### DIFF
--- a/pkg/cloud/vsphere/actuators/cluster/actuator.go
+++ b/pkg/cloud/vsphere/actuators/cluster/actuator.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cluster
 
 import (
-	"encoding/base64"
 	"fmt"
 	"net/url"
 	"strconv"
@@ -217,7 +216,7 @@ func (a *Actuator) reconcileCloudConfigSecret(ctx *context.ClusterContext) error
 		ctx.Logger.Error(err, "target cluster is not ready")
 		return &clusterErr.RequeueAfterError{RequeueAfter: config.DefaultRequeue}
 	}
-	// Define the kubeconfig secret for the target cluster.
+	// Define the cloud provider credentials secret for the target cluster.
 	secret := &apiv1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: constants.CloudProviderSecretNamespace,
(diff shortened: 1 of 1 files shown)


sidharthsurana:
@thorsteinnth Thanks for the fix.
But this also raises an interesting question, how has been the CI passing so far given that the VC credentials would have been messed up (double base64 encoded).
@andrewsykim or @figo does anyone of you shed light on the above question?

andrewsykim:
> @thorsteinnth Thanks for the fix.
But this also raises an interesting question, how has been the CI passing so far given that the VC credentials would have been messed up (double base64 encoded).
@andrewsykim or @figo does anyone of you shed light on the above question?

We don't test against target clusters yet in the e2e tests.

andrewsykim:
Interestingly.. the cloud provider is still working even though the creds are double encoded:
```
# kubectl --kubeconfig admin.conf get no -o yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: Node
  metadata:
    annotations:
      kubeadm.alpha.kubernetes.io/cri-socket: /var/run/containerd/containerd.sock
      node.alpha.kubernetes.io/ttl: "0"
      volumes.kubernetes.io/controller-managed-attach-detach: "true"
    creationTimestamp: "2019-07-26T15:17:11Z"
    labels:
      beta.kubernetes.io/arch: amd64
      beta.kubernetes.io/os: linux
      kubernetes.io/hostname: workload-cluster-1-controlplane-1
      node-role.kubernetes.io/master: ""
    name: workload-cluster-1-controlplane-1
    resourceVersion: "661131"
    selfLink: /api/v1/nodes/workload-cluster-1-controlplane-1
    uid: 70cc9a73-afb8-11e9-a98f-005056b0e8df
  spec:
    podCIDR: 100.96.0.0/24
    providerID: vsphere://423012f3-8a7a-49c3-540e-e42af2fb3a6e
```

andrewsykim:
/approve

Thank you @thorsteinnth

Issue #485, opened 2019-08-06
Cannot bind kube-apiserver to specific NIC in a machine with multiple valid interfaces


jpweber:
/kind bug

**What steps did you take and what happened:**
I have a VM with IPV4 interfaces. One Static one DHCP. Both on different Networks. Because of how `DataSourceVMwareGuestInfo.py` plumbs the IP information from cloud-init to the `kubeadm.conf template` I cannot force the kube-apiserver to come up on the address I want. 

I then took the approach of passing in kubeadm config properties through `machines.yaml` seen in this example:
```
        kubeadmConfiguration:
          init:
            localAPIEndpoint:
              advertiseAddress: "192.168.3.254"
              bindPort: 6443
```
But this data is not used, and instead is turned in to this:
```
      kubeadmConfiguration:
        init:
          localAPIEndpoint:
            advertiseAddress: '{{ ds.meta_data.local_ipv4 }}'
```
Which is because of 
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/6af70011c0e90b42d46c2387d05e49003970043d/pkg/cloud/vsphere/services/govmomi/create.go#L40
which is then used here (in the case of the init. It is also used for join but this issue is just describing the init phase)
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/6af70011c0e90b42d46c2387d05e49003970043d/pkg/cloud/vsphere/services/govmomi/create.go#L191
and here
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/6af70011c0e90b42d46c2387d05e49003970043d/pkg/cloud/vsphere/services/govmomi/create.go#L223

**What did you expect to happen:**
My kubeadm init phase configuration to be passed to the kubeadm.conf that is used on the machine when kubeadm init is run. 

**Anything else you would like to add:**
The scenario that brought this is up, is the machines in question will have one interface on a "management" network. The second interface will be where kubernetes and CNI networking will be used. 

This issue also manifests itself if the two NICs are IPv4/IPv6. The issue is _not_ IPv4/IPv4 only. 

**Environment:**

- Cluster-api-provider-vsphere version: 0.4.0-beta2
- Kubernetes version: (use `kubectl version`): 1.14.3
- OS (e.g. from `/etc/os-release`): ubuntu 18.04

andrewsykim:
[...]

PR #486, opened 2019-08-06
Added ability to override local IP advertise address for kubeadm init

jpweber:
Added ability to override local IP advertise address for kubeadm init. Method here is following suit of how the bindPort is being allowed to be overridden.

Signed-off-by: Jim Weber <jpweber@gmail.com>

Allows specifying the LocalAPI Advertise Address of the kubeadm init configuration to be used if it is present. 

**Which issue(s) this PR fixes** :
Partial solution for #485 

**Release note**:
```release-note
NONE
```


akutz:
/hold

I want to think about this a minute before implementing it as a quick-fix. I'll respond within 24h with more information.

jpweber:
Understood. I agree there may be a better way that solves the bigger picture of this issue than just solving this one small piece of it.

jpweber:
After having a chat with @akutz we are going to solve this problem through other means.

PR #487, opened 2019-08-06
Set GOPROXY for docker commands for Prow

codenrhoden:
Prow recently started setting GOPROXY for all its jobs by default. This
speeds up builds, but does not appy when using docker-in-docker. For our
jobs that build things in Docker, we need to supply GOPROXY ourselves.

Fixes #

This is a first attempt at this. Need to see the test output to see if it's working. If it is, builds should be considerably faster.

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,12 @@ SHELL := hack/shell-with-gopath.sh
 # Retrieves the git hash
 VERSION ?= $(shell git describe --always --dirty)
 
+# Set the goproxy for docker-in-docker if on Prow
+_goproxy="direct"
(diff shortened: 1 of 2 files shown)


codenrhoden:
/hold

This may not make sense for CAPV, since it vendors dependencies. Instead, I should check that all builds use `-mod vendor` where appropriate. They probably do already. Didn't even think about that -- instead I should use the on CSI and CPI...

codenrhoden:
Yeah, this is completely unnecessary for CAPV.

/close

PR #489, opened 2019-08-07
docs: remove controlPlane version field for worker…

zdyxry:

When I read docs/getting_started.md, I was confused about how to judge the ControlPlane node. 

I saw this in the code:

```go
// HasControlPlaneRole returns a flag indicating whether or not a machine has
// the control plane role.
func (c *MachineContext) HasControlPlaneRole() bool {
	if c.Machine == nil {
		return false
	}
	return clusterUtilv1.IsControlPlaneMachine(c.Machine)
}
```

```go
// IsControlPlaneMachine checks machine is a control plane node.
func IsControlPlaneMachine(machine *clusterv1.Machine) bool {
	return machine.Spec.Versions.ControlPlane != ""
}
```

So maybe need delete this field when adding a worker node?

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -239,7 +239,6 @@ spec:
           template: "ubuntu-1804-kube-v1.13.6"
       versions:
         kubelet: "1.13.6"
-        controlPlane: "1.13.6"
 ```
 
 Use `kubectl` with the `kubeconfig` for the management cluster to provision the new workload cluster:


zdyxry:
I signed it

akutz:
/close

There's a reason we're deferring the call to the upstream package. They've changed this check in v1a2. It used to be based on the version string, and now it is not.

akutz:
/reopen

I'm an idiot and misunderstood the PR.

akutz:
Thank you for this @zdyxry!

PR #502, opened 2019-08-15
[WIP] add support for external vSphere cloud provider

andrewsykim:

Adds opt-in for using the [external vSphere cloud provider](https://github.com/kubernetes/cloud-provider-vsphere/). 

If the `providerImage` field is set in the cluster cloud provider configuration, then we attempt to use the vSphere cloud controller manager with that image, otherwise we default to the in-tree cloud provider.

Fixes #

**Release note**:

```release-note
Add opt-in for external vSphere cloud provider
```


andrewsykim:
/hold 

Pending discussions re: versioning of the external provider and if we need to continue supporting the in-tree provider.

pkg/cloud/vsphere/services/cloudprovider/addons.go
```diff
@@ -0,0 +1,277 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	cloudProviderImage = "gcr.io/cloud-provider-vsphere/cpi/release/manager:latest"
```
andrewsykim: For the sake of simplicity I hardcoded this to `latest` but I think we should figure out how to pick a vSphere CCM version w.r.t to CAPV.

andrewsykim: Actually, in a conversation I had with @akutz we agreed to make the version configurable via the cloud configuration field.

pkg/cloud/vsphere/services/cloudprovider/addons.go
```diff
@@ -0,0 +1,277 @@
...
+							Effect: corev1.TaintEffectNoSchedule,
+						},
+					},
+					ServiceAccountName: "cloud-controller-manager",
+					Containers: []corev1.Container{
+						{
+							Name:  "vsphere-cloud-controller-manager",
+							Image: cloudProviderImage,
+							Args: []string{
+								"/bin/vsphere-cloud-controller-manager",
```
codenrhoden: The `/bin` part of this isn't necessary anymore because it's not party of the container image `ENTRYPOINT`.

The current examples have this removed as well: https://github.com/kubernetes/cloud-provider-vsphere/blob/551410442226d6e5d0754408917ead48c013c3eb/manifests/controller-manager/vsphere-cloud-controller-manager-pod.yaml#L22-L25

andrewsykim: Will update, thanks!

akutz:
Hi @andrewsykim,

Also, do you want to consider using the Controller Client instead of the client-go client? We'll need to refactor at some point, but I'm also fine deferring that for now to get this in.

examples/generate.sh
```diff
@@ -208,31 +220,29 @@ VSPHERE_B64ENCODED_PASSWORD="$(printf '%s' "${VSPHERE_PASSWORD}" | base64)"
...
-
-[Disk]
-scsicontrollertype = pvscsi
-
-[Network]
-public-network = "${VSPHERE_NETWORK}"
+if [ -n "${VSPHERE_PRE_67u3_SUPPORT}" ]; then
+  # Encode the cloud provider configuration.
+  CLOUD_CONFIG_B64ENCODED=$(cat <<EOF | { base64 -w0 2>/dev/null || base64; }
+  [Global]
```
akutz: Hi @andrewsykim,

Please unindent the heredoc so this reads:

```shell
if [ -n "${VSPHERE_PRE_67u3_SUPPORT}" ]; then
  # Encode the cloud provider configuration.
  CLOUD_CONFIG_B64ENCODED=$(cat <<EOF | { base64 -w0 2>/dev/null || base64; }
[Global]
secret-name = "cloud-provider-vsphere-credentials"
secret-namespace = "kube-system"
datacenters = "${VSPHERE_DATACENTER}"
insecure-flag = "1"

[VirtualCenter "${VSPHERE_SERVER}"]

[Workspace]
server = "${VSPHERE_SERVER}"
datacenter = "${VSPHERE_DATACENTER}"
folder = "${VSPHERE_FOLDER}"
default-datastore = "${VSPHERE_DATASTORE}"
resourcepool-path = "${VSPHERE_RESOURCE_POOL}"

[Disk]
scsicontrollertype = pvscsi

[Network]
public-network = "${VSPHERE_NETWORK}"
EOF
  )
  export CLOUD_CONFIG_B64ENCODED
fi
```

Otherwise the content is indented. Also, I reintroduced the empty lines between sections if that's okay. I think it makes it easier to read.

examples/pre-67u3/controlplane/controlplane.yaml
```diff
@@ -21,9 +21,6 @@ apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
 kind: VSphereMachine
 metadata:
   name: ${CLUSTER_NAME}-controlplane-0
-  labels:
```
akutz: Hi @andrewsykim,

Please add this back. @yastij fixed this in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/556 and it's consistent with CAPA.

PR #508, opened 2019-08-17
set clone status on error

### LINKED ISSUE #431: Mark machine in error if VM cannot be cloned because same name exists
/kind feature

If VM cannot be cloned because it already exists then mark the machine in error.

pickledrick:
Updates the status of a machine if a clone operation is unsuccessful due to duplicate machine name.
Fixes #431 

Hi,

I wanted to have a crack at this one. But wasn't sure on a few things.
- where to update the status? I have used annotations in this PR but happy to change to whatever is appropriate. I was wondering wether this should be `MachineRef` or a new field in the MachineStatus.
- Wether waiting for the clone operation is appropriate?

**Release note**:

```release-note
update machine status on failed create when machine name exists.
```

### DIFF
--- a/pkg/cloud/vsphere/services/govmomi/create.go
+++ b/pkg/cloud/vsphere/services/govmomi/create.go
@@ -21,7 +21,6 @@ import (
 	"net"
 
 	"github.com/pkg/errors"
-
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/constants"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/context"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/cloud/vsphere/services/certificates"
@@ -58,6 +57,7 @@ func Create(ctx *context.MachineContext, bootstrapToken string) error {
 	// does not exist, only when there's an error checking or when the op should
 	// be requeued, like when the VM has an in-flight task.
 	vm, err := lookupVM(ctx)
+
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/cloud/vsphere/services/govmomi/create.go
```diff
@@ -71,7 +71,13 @@ func Create(ctx *context.MachineContext, bootstrapToken string) error {
 	}
 
 	if ctx.Session.IsVC() {
-		return vcenter.Clone(ctx, userData)
+		err = vcenter.Clone(ctx, userData)
```
akutz: Hi @pickledrick,

Please change this to `if err := vcenter.Clone` to reduce the scope of `err`.

pkg/cloud/vsphere/services/govmomi/vcenter/clone.go
```diff
@@ -122,7 +121,10 @@ func Clone(ctx *context.MachineContext, userData []byte) error {
 	if err != nil {
 		return errors.Wrapf(err, "error trigging clone op for machine %q", ctx)
 	}
-
+	_, err = task.WaitForResult(ctx, nil)
```
akutz: Hi @pickledrick,

CAPV is reentrant and never blocks. If something isn't ready, it requeues the operation.

pkg/cloud/vsphere/services/govmomi/create.go
```diff
@@ -71,7 +71,13 @@ func Create(ctx *context.MachineContext, bootstrapToken string) error {
...
+		if err != nil {
+			annotations := make(map[string]string)
+			annotations["clone-machine-error"] = err.Error()
```
akutz: Don't set errors as annotations - https://github.com/kubernetes-sigs/cluster-api/blob/961efcc224d6c5b278ecf19aee8512e8c5a98283/cmd/clusterctl/clusterdeployer/clusterclient/clusterclient.go#L1005

akutz:
Hi @pickledrick,

I'm going to close this, but I'd like to have you open it as an issue. Then you can either work with @ykakarap on this as he's working on the v1a2 machine controller, or wait until his work is merged so you can start from there. We should define some type of errors that we can examine to know when something shouldn't be requeued or returned as an error, but when the machine itself should be set in error. That way we can handle things centrally.

/close

PR #538, opened 2019-08-29
introduce load balancer API for capv

### LINKED ISSUE #468: add support  for VMC
/kind feature

Machines deployed in VMC are not accessible from the public internet by default, and the ability to link an SDDC to AWS to utilize AWS services such as ELB provides a useful solution to this problem and more (e.g. HA setups).

we need to add the ability to provision an ELB  for clusters deployed to VMC on AWS

/assign

**Environment:**

yastij:
**What this PR does / why we need it**: This PR introduces a load balancing API to implement in-tree load balacing providers. This PR ships AWS as a first provider

**Which issue(s) this PR fixes** : Fixes  #468

what is missing

- [x] updating the manifest generation to include the AWS credentials secret (secret object + volumeMount for the manager), `loadbalancerRef` kustomization and an `AWSLoadBalancer` resource
- [ ] add test coverage for this (e2e + mock aws api to unit test the controller)

/assign @akutz 
/cc @andrewsykim 

**Release note**:

```release-note
introduce a load balancing API. AWS is the first provider to be implemented for CAPV
```


akutz:
Hi @yastij,

I started my review yesterday, but didn't have time to complete it. One thing I noticed was that the types are in `./api/v1alpha2/cloud`. That was just a suggestion about how to contain types in a specific directory -- it wasn't guidance to use **that** directory. I think the types should fall into a structure similar to something like this:

```shell
├── api
│   └── v1alpha2
│       ├── cloud
│       └── load-balancer
│           └── vmc
│               └── aws
```

or

```shell
├── api
│   ├── load-balancer
│   │   ├── v1alpha1
│   │   └── vmc
│   │       └── aws
│   └── vsphere
│       └── v1alpha2
│           └── cloud
```

Now obviously the second pattern would require not insignificant refactoring since it changes the location of the existing vSphere types, **but**, the second pattern **does** adhere to the Kubebuilder [recommended guidelines for multi-group projects](https://book.kubebuilder.io/migration/multi-group.html).

controllers/util/cluster.go
```diff
@@ -0,0 +1,63 @@
+package util
```
akutz: Hi @yastij,

I'm not sure it's worth introducing a new package for this one function. Unless there's a strong reason for this, let's just fold this back into the `controllers` package as a private function.

controllers/vspherecluster_controller.go
```diff
@@ -56,48 +55,13 @@ type VSphereClusterReconciler struct {
 
 // Reconcile ensures the back-end state reflects the Kubernetes resource state intent.
 func (r *VSphereClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error) {
```
akutz: Hi @yastij,

I'm not sure I want the context initialization of this controller refactored as part of *this* PR. I don't mind handling it later, but I'd like to reduce this PR to just what is required for the LB. What do you think?

controllers/vspherecluster_controller.go
```diff
@@ -130,7 +94,7 @@ func (r *VSphereClusterReconciler) reconcileNormal(ctx *context.ClusterContext)
 	//   * Downloading OVAs into the content library for any machines that
 	//     use them.
 	//   * Create any load balancers for VMC on AWS, etc.
-	ctx.VSphereCluster.Status.Ready = true
+	ctx.VSphereCluster.Status.Ready = ctx.VSphereCluster.Spec.LoadBalancerConfiguration == nil
 	ctx.Logger.V(6).Info("VSphereCluster is infrastructure-ready")
```
akutz: Hi @yastij,

Should this log statement be updated based on whether or not the cluster is ready? Before it was always, implicitly ready. Now it depends. I think the log statement should reflect that.

pkg/cloud/vsphere/services/interfaces.go
```diff
@@ -30,3 +30,14 @@ type VirtualMachineService interface {
 	// DestroyVM powers off and removes a VM from the inventory.
 	DestroyVM(ctx *context.MachineContext) (infrav1.VirtualMachine, error)
 }
+
+// LoadBalancerService is service that reconciliate load balancers
+type LoadBalancerService interface {
+
+	// Reconcile reconciles a load balancer for vsphereCluster using controlPlaneEndpoints
+	// to serve as api endpoint
+	Reconcile(vsphereCluster *infrav1.VSphereCluster, controlPlaneIPs []string) (infrav1.APIEndpoint, error)
```
akutz: Hi @yastij,

Shouldn't the first parameter be a `CAPI` cluster? Otherwise how is the service generic across infra providers?

controllers/loadbalancer_controller.go
```diff
@@ -0,0 +1,142 @@
...
+	client.Client
+	ProviderName string
+	Recorder     record.EventRecorder
+	Log          logr.Logger
+}
+
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=vsphereclusters,verbs=get;list;watch;
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=vsphereclusters/status,verbs=get;update;patch
+
+// Reconcile todo
```
akutz: Hi @yastij,

I suggest rewording this as:

```
// Reconcile ensures the back-end state reflects the Kubernetes state intent for a LoadBalancer resource.
```

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+	ns, err := session.NewSession(aws.NewConfig().WithRegion(region))
+	if err != nil {
+		return nil, err
+	}
+
+	sessionCache.Store(region, ns)
+	return ns, nil
+}
+
+// Service TODO
```
akutz: Hi @yastij,

I suggest rewording this as:

```
// Service is the AWS implementation of the vSphere LoadBalancer service and
// handles reconciling and deleting AWS elastic load balancers (ELBs) tied to
// vSphere clusters.
```

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+type Service struct {
+	ELB elbv2iface.ELBV2API
+}
+
+// NewProvider TODO
+func NewProvider(config *cloud.LoadBalancerConfig) (services.LoadBalancerService, error) {
```
akutz: Hi @yastij,

I suggest rewording this as:

```
// NewProvider returns a new instance of the AWS LoadBalancerService.
```

I would also consider renaming this function simply to `New` since it's in a package dedicated to this service.

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+	session, err := sessionForRegion(region)
+	if err != nil {
+		return nil, err
+	}
+	return &Service{
+		ELB: elbv2.New(session),
+	}, nil
+}
+
+func generateELBName(clusterName string) string {
```
akutz: Hi @yastij,

Please consider exporting this function so it's easy to determine the would-be name of an ELB. For example:

```golang
// GetLoadBalancerNameForCluster returns the name of the load balancer for a given cluster.
func GetLoadBalancerNameForCluster(clusterName string) string {
	return fmt.Sprintf("%s-endpoint", clusterName)
}
```

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+	return fmt.Sprintf("%s-endpoint", clusterName)
+}
+
+// Reconcile reconciles loadbalancing resources in aws
```
akutz: Hi @yastij,

I suggest rewording this as:

```
// Reconcile ensures the AWS load balancer for a vSphere cluster is synchronized with the intended state.
```

I would also consider renaming this function simply to `New` since it's in a package dedicated to this service.

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+func (svc *Service) Reconcile(vsphereCluster *infrav1.VSphereCluster, controlPlaneIPs []string) (infrav1.APIEndpoint, error) {
+	AwsProvider := vsphereCluster.Spec.LoadBalancerConfiguration.AwsProvider
```
akutz: Hi @yastij,

Variable names should not begin with capital letters.

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
akutz: Hi @yastij,

It might be useful to pass a `github.com/go-logr/logr.Logger` into `NewProvider` in order to facilitate logging inside the service.

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+
+	listenerPort, err := svc.reconcileListeners(loadBalancerArn, targetGroupArn)
+	if err != nil {
+		return infrav1.APIEndpoint{}, err
+	}
+	apiEndpoint := infrav1.APIEndpoint{
+		Host: *loadBalancerDNS,
+		Port: int(*listenerPort),
+	}
+	return apiEndpoint, nil
```
akutz: Hi @yastij,

It would be nice to be able to log things like the target group ARN, LB ARN, DNS, ports, etc. in this function. In an earlier comment I suggested passing a logger into the `NewProvider` function.

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+	actualTargetGroupInput := convertTargetGroupOutputToInput(describeTargetGroupOutput.TargetGroups[0])
+	if !reflect.DeepEqual(actualTargetGroupInput, desiredTargetGroupInput) {
+		modifyTargetGroup := &elbv2.ModifyTargetGroupInput{
+			HealthCheckEnabled:         desiredTargetGroupInput.HealthCheckEnabled,
+			HealthCheckIntervalSeconds: desiredTargetGroupInput.HealthCheckIntervalSeconds,
+			HealthCheckTimeoutSeconds:  desiredTargetGroupInput.HealthCheckTimeoutSeconds,
+			HealthyThresholdCount:      desiredTargetGroupInput.HealthyThresholdCount,
+			UnhealthyThresholdCount:    desiredTargetGroupInput.UnhealthyThresholdCount,
+		}
+		_, err := svc.ELB.ModifyTargetGroup(modifyTargetGroup)
```
akutz: Hi @yastij,

Please collapse this to reduce scope:

```golang
		if _, err := svc.ELB.ModifyTargetGroup(modifyTargetGroup); err != nil {
				return nil, err
		}
```

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+				Id:               aws.String(controlPlaneIP),
+				Port:             aws.Int64(int64(6443)),
+			})
+		}
+
+		registerTargetInput := &elbv2.RegisterTargetsInput{
+			TargetGroupArn: targetGroupArn,
+			Targets:        targetDescription,
+		}
+		_, err = svc.ELB.RegisterTargets(registerTargetInput)
```
akutz: Hi @yastij,

Please collapse this to reduce scope:

```golang
		if _, err := svc.ELB.RegisterTargets(registerTargetInput); err != nil {
				return nil, err
		}
```

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	return apiEndpoint, nil
+}
+
+func (svc *Service) reconcileTargetGroup(clusterName string, vpcID string, controlPlaneIPs []string) (*string, error) {
+	describeTargetGroupInput := &elbv2.DescribeTargetGroupsInput{}
+	describeTargetGroupInput.Names = append(describeTargetGroupInput.Names, aws.String(clusterName+"-controlPlane"))
+
+	describeTargetGroupOutput, err := svc.ELB.DescribeTargetGroups(describeTargetGroupInput)
+	desiredTargetGroupInput := &elbv2.CreateTargetGroupInput{
```
akutz: Hi @yastij,

The order of the code here is confusing. It's necessary, but confusing. I think it makes sense in this case to group the declarations first. For example:

```golang
func (svc *Service) reconcileTargetGroup(clusterName string, vpcID string, controlPlaneIPs []string) (*string, error) {
	var (
		// targetGroupArn will store the ID of the discovered target group
		targetGroupArn *string

		// describeTargetGroupsInput is used to get a list of the target groups
		describeTargetGroupsInput = &elbv2.DescribeTargetGroupsInput{
			Names: []*string{aws.String(clusterName+"-controlPlane")},
		}

		// desiredTargetGroupInput describes the target group we want to find
		desiredTargetGroupInput = &elbv2.CreateTargetGroupInput{
			Name:                       aws.String(clusterName + "-controlPlane"),
			HealthCheckEnabled:         aws.Bool(true),
			HealthCheckIntervalSeconds: aws.Int64(int64(30 * time.Second.Seconds())),
			HealthCheckTimeoutSeconds:  aws.Int64(int64(10 * time.Second.Seconds())),
			HealthyThresholdCount:      aws.Int64(int64(3)),
			UnhealthyThresholdCount:    aws.Int64(int64(3)),
			Port:                       aws.Int64(int64(6443)),
			Protocol:                   aws.String(elbv2.ProtocolEnumTcp),
			TargetType:                 aws.String(string("ip")),
			VpcId:                      &vpcID,
		}
	)

	// Get a list of the target groups. If no target groups are discovered then
	// a new target group is created and its ARN is stored in targetGroupArn.
	describeTargetGroupsOutput, err := svc.ELB.DescribeTargetGroups(describeTargetGroupsInput)
	if err != nil {
		if IsNotFound(err) {
			createTargetGroupOutput, err := svc.ELB.CreateTargetGroup(desiredTargetGroupsInput)
			if err != nil {
				return nil, err
			}
			targetGroupArn = createTargetGroupOutput.TargetGroups[0].TargetGroupArn
		} else {
			return nil, err
		}
	}

	// If the target group already existed then make sure its ID is assigned to
	// targetGroupArn. If the target group didn't already exist, then its ID was
	// assigned to targetGroupArn when the target group was created above.
	if targetGroupArn == nil {
		targetGroupArn = describeTargetGroupsOutput.TargetGroups[0].TargetGroupArn
	}

	// Ensure the actual state of the target group matches the intended state of
	// target group. If the actual and intended states do not match then modify the
	// target group so its actual state matches the intended state.
	actualTargetGroupInput := convertTargetGroupOutputToInput(describeTargetGroupsOutput.TargetGroups[0])
	if !reflect.DeepEqual(actualTargetGroupInput, desiredTargetGroupInput) {
		modifyTargetGroup := &elbv2.ModifyTargetGroupInput{
			HealthCheckEnabled:         desiredTargetGroupInput.HealthCheckEnabled,
			HealthCheckIntervalSeconds: desiredTargetGroupInput.HealthCheckIntervalSeconds,
			HealthCheckTimeoutSeconds:  desiredTargetGroupInput.HealthCheckTimeoutSeconds,
			HealthyThresholdCount:      desiredTargetGroupInput.HealthyThresholdCount,
			UnhealthyThresholdCount:    desiredTargetGroupInput.UnhealthyThresholdCount,
		}
		if _, err := svc.ELB.ModifyTargetGroup(modifyTargetGroup); err != nil {
			return nil, err
		}
	}

	// If there is at least one control plane IP then register the IP addresses
	// to the target group.
	if len(controlPlaneIPs) > 0 {
		targetDescription := []*elbv2.TargetDescription{}
		for _, ipAddr := range controlPlaneIPs {
			targetDescription = append(targetDescription, &elbv2.TargetDescription{
				AvailabilityZone: aws.String(string("all")),
				Id:               aws.String(ipAddr),
				Port:             aws.Int64(int64(6443)),
			})
		}
		registerTargetInput := &elbv2.RegisterTargetsInput{
			TargetGroupArn: targetGroupArn,
			Targets:        targetDescription,
		}
		if _, err := svc.ELB.RegisterTargets(registerTargetInput); err != nil {
			return nil, err
		}
	}

	return targetGroupArn, nil
}
```

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+		targetGroupArn = describeTargetGroupOutput.TargetGroups[0].TargetGroupArn
+	}
+
+	if len(controlPlaneIPs) != 0 {
+		targetDescription := []*elbv2.TargetDescription{}
+		for _, controlPlaneIP := range controlPlaneIPs {
+			targetDescription = append(targetDescription, &elbv2.TargetDescription{
+				AvailabilityZone: aws.String(string("all")),
+				Id:               aws.String(controlPlaneIP),
+				Port:             aws.Int64(int64(6443)),
```
akutz: Hi @yastij,

There's a constant for this, please use that. Thanks!

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		_, err = svc.ELB.RegisterTargets(registerTargetInput)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return targetGroupArn, nil
+}
+
+func convertTargetGroupOutputToInput(output *elbv2.TargetGroup) *elbv2.CreateTargetGroupInput {
```
akutz: Hi @yastij,

Unless it's necessary to pass by reference, it's generally advised to pass by copy.

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+		UnhealthyThresholdCount:    output.UnhealthyThresholdCount,
+		Port:                       output.Port,
+		Protocol:                   output.Protocol,
+		TargetType:                 output.TargetType,
+		VpcId:                      output.VpcId,
+	}
+}
+
+func (svc *Service) reconcileLoadBalancer(clusterName string, subnets []string) (*string, *string, error) {
+	var loadBalancerArn *string
```
akutz: Hi @yastij,

Please simplify as:

```golang
var (
	loadBalancerArn *string
	loadBalancerDNS *string
	describeLoadBalancersInput = &elbv2.DescribeLoadBalancersInput{
		Names: []*string{aws.String(generateELBName(clusterName))},
	}
)
```

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	loadBalancerArn = describeLoadBalancersOutput.LoadBalancers[0].LoadBalancerArn
+	loadBalancerDNS = describeLoadBalancersOutput.LoadBalancers[0].DNSName
+
+	return loadBalancerArn, loadBalancerDNS, nil
+}
+
+func (svc *Service) reconcileListeners(loadBalancerArn *string, targetGroupArn *string) (*int64, error) {
+	var listenerPort *int64
```
akutz: Hi @yastij,

Please simplify as:

```golang
var (
	listenerPort *int64
	describeListnerInput = &elbv2.DescribeListenersInput{
		LoadBalancerArn: loadBalancerArn,
	}
)
```

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+	loadBalancerArn := describeLoadBalancersOutput.LoadBalancers[0].LoadBalancerArn
+	deleteLoadBalancerInput := &elbv2.DeleteLoadBalancerInput{
+		LoadBalancerArn: loadBalancerArn,
+	}
+	_, err = svc.ELB.DeleteLoadBalancer(deleteLoadBalancerInput)
+	return err
+}
+
+func (svc *Service) deleteTargetGroup(clusterName string) error {
+	describeTargetGroupInput := &elbv2.DescribeTargetGroupsInput{}
```
akutz: Hi @yastij,

Please simplify as:

```golang
describeTargetGroupInput := &elbv2.DescribeTargetGroupsInput{
	Names: []*string{aws.String(clusterName+"-controlPlane")},
}
```

Also, please note that I recommended above to extract the building of this name into a distinct function.

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+	apiEndpoint := infrav1.APIEndpoint{
+		Host: *loadBalancerDNS,
+		Port: int(*listenerPort),
+	}
+	return apiEndpoint, nil
+}
+
+func (svc *Service) reconcileTargetGroup(clusterName string, vpcID string, controlPlaneIPs []string) (*string, error) {
+	describeTargetGroupInput := &elbv2.DescribeTargetGroupsInput{}
+	describeTargetGroupInput.Names = append(describeTargetGroupInput.Names, aws.String(clusterName+"-controlPlane"))
```
akutz: Hi @yastij,

I've seen `aws.String(clusterName+"-controlPlane")` at least twice. Due to the fact that this is required to find a resource, let's extract this into a function:

```golang
// GetTargetGroupNameForCluster returns the name of a target group for the provided cluster.
func GetTargetGroupNameForCluster(clusterName string) *string {
	return aws.String(clusterName+"-controlPlane")
}
```

pkg/cloud/vsphere/services/loadbalancer/aws/loadbalancers.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// Delete performs an ELB resource deletion
+func (svc *Service) Delete(vsphereCluster *infrav1.VSphereCluster) error {
+	clusterName := vsphereCluster.Name
+	if err := svc.deleteLoadBalancer(clusterName); err != nil {
+		return err
+	}
+
+	if err := svc.deleteTargetGroup(clusterName); err != nil {
```
akutz: Hi @yastij,

This could be simplified as:

```golang
return svc.deleteTargetGroup(clusterName)
```

Or please feel free to leave as-is if you think we may include additional logic beneath this call one day.

api/v1alpha2/vspherecluster_types.go
```diff
@@ -27,6 +27,11 @@ const (
 	// resources associated with VSphereCluster before removing it from the
 	// API server.
 	ClusterFinalizer = "vspherecluster.infrastructure.cluster.x-k8s.io"
+
+	// LoadBalancerFinalizer allows to clean up the load balancer
+	// associated with VSphereCluster before removing it from the
+	// API server.
+	LoadBalancerFinalizer = "loadbalancer.infrastructure.cluster.x-k8s.io"
```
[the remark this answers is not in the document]
yastij: I'm not sure, the finalizer is added to the cluster object. The cluster object only supports one provider.

api/v1alpha2/loadbalancer/types.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// AwsProvider is the name of the aws provider
+	AwsProvider = "aws"
```
andrewsykim: I think the provider name should be explicit to the LB implementation, not the cloud provider, so this should be "ELB", thoughts?

yastij: actually this should be removed. As the current PR uses `Kind` to store/retrieve the provider

api/v1alpha2/loadbalancer/types.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=awsloadbalancers,scope=Namespaced,categories=cluster-api
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
+
+// AWSLoadBalancer is the schema for the AWS Load balancer API
+type AWSLoadBalancer struct {
```
andrewsykim: Shouldn't the API types be generic and the implementations be provider specific?

akutz: No. Think of this as a KubeadmBootstrap provider. There will be a diff controller / API model for each implementation of the LB.

yastij: The issue is that there's no common API for loadbalancers, as each provider needs its own information. I don't want us to endup implementing annotations on-top to satisfy each provider's specifics.

Another way could be store the information on-disk/cm implement something `Service` (e.g. `MachineService` or `MachineLoadbalancer`) which would select machines based on a the selector and add them to the load balancing pool. A drawback is that a management cluster is restricted to provision against one Loadbalancing provider. Thoughts @akutz @andrewsykim ?

akutz: I don't think it's that complicated. Just like each `Cluster` resource today has a `ConfigRef` to a Infrastructure provider's `Cluster`, so to will there be a `ConfigRef` from a `Cluster` to a load balancer config, in this case, `AWSLoadBalancerConfig`.

There's some questions around how to get the IP information from the machines, but that's something we can work out. 

The benefit of this decoupled model is that it enables the introduction of a load balancer provider of any kind.

yastij: Yes, plus this wouldn't change much the implementation itself. thoughts @andrewsykim ?

andrewsykim: So my thoughts around this was that we introduce a generic `LoadBalancer` type with a `ProviderRef` field similar to what @akutz mentioned re: Cluster with a ConfigRef. That way, for each implementation you only have to implement a new ProviderRef and not the entire LoadBalancer API again. Maybe we already agree on this but it's not clear to me yet.

api/v1alpha2/loadbalancer_types.go
```diff
@@ -0,0 +1,132 @@
...
+	ProtocolTCP = Protocol("TCP")
+)
+
+// LoadBalancerStatus defines the observed state of the AWS load balancer
+type LoadBalancerStatus struct {
+
+	// APIEndpoint represents the endpoint to communicate with the load
+	// balancer.
+	// +optional
+	APIEndpoint APIEndpoint `json:"apiEndpoint,omitempty"`
```
ncdc: Not doing a full review on this right now (unless you'd like), but one thing I wanted to mention was the API endpoint needs to be in spec, not status. If you have a provider that generates an endpoint (e.g. a DNS name) **and** the endpoint is not something you can determine by querying, we would lose the data if we store it in status and we're trying to move the resource from one management cluster to another.

yastij: @ncdc - I'll rebase and ping you for an API review

PR #547, opened 2019-09-03
Add -b option to hack/release.sh

codenrhoden:
We need a way to have our automation build and push multiple minor versions of releases (e.g. 0.4.x, 0.5.x), but be able to specify which of these versions should also get tagged as `latest`. We are already managing this with branches. Thus...

The -b option allows one to specify a branch for use with the -l option.
This means that if -l is used (tag images as latest) the "latest" tag
will *only* be applied if the current commit is in/on the given branch.
Otherwise "latest" will not be applied.

This allows for CI to do a job like `hack/release.sh -l -p -b release-0.4` that runs on every commit, and only have the `latest` tag applied to images built on tags on the `release-0.4` branch.

Fixes #

This was a bit harder than I expected it to be because I had to take into account what the `git` setup would be when run from Prow. Take a look at the [logs](https://storage.googleapis.com/kubernetes-jenkins/logs/post-cluster-api-provider-vsphere-release/1168971198973874176/clone-log.txt) from the recent 0.5.0 build. That results in an environment with no remotes and no branches present. Therefore, I had the script create the remote only when needed (and removed it afterwards), and specifically look for the tag on that specific remote.

**Release note**:

```release-note
NONE
```

/assign @akutz

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -46,6 +46,9 @@ VERSION=$(git describe --dirty --always 2>/dev/null)
 GCR_KEY_FILE="${GCR_KEY_FILE:-}"
 
 BUILD_RELEASE_TYPE="${BUILD_RELEASE_TYPE-}"
+readonly ORIGIN_REPO="https://github.com/kubernetes-sigs/cluster-api-provider-vsphere.git"
+ORIGIN_REPO_REMOTE_NAME="${ORIGIN_REPO_REMOTE_NAME:-capv-origin}"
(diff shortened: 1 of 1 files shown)


codenrhoden:
I tested this by running the script to confirm the following:

`v0.4.1` only gets tagged latest if you provide `-l -b release-0.4` (or don't provide `-b` at all)
`v0.5.0` only gets tagged latest if you provide `-l -b master` (or don't provided `-b` at all)

PR #552, opened 2019-09-04
log using WithValues for key=value formatted log messagse

andrewsykim:

Right now we are logging WithName for key/value log messages. We should use WithValue which was meant for key/value formats. 

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/vspherecluster_controller.go
+++ b/controllers/vspherecluster_controller.go
@@ -59,8 +59,8 @@ func (r *VSphereClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, r
 	parentContext := goctx.Background()
 
 	logger := r.Log.WithName(controllerName).
-		WithName(fmt.Sprintf("namespace=%s", req.Namespace)).
-		WithName(fmt.Sprintf("vsphereCluster=%s", req.Name))
+		WithValues("namespace", req.Namespace).
+		WithValues("vSphereCluster", req.Name)
 
 	// Fetch the VSphereCluster instance
 	vsphereCluster := &infrav1.VSphereCluster{}
@@ -84,7 +84,7 @@ func (r *VSphereClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, r
 		return reconcile.Result{RequeueAfter: config.DefaultRequeue}, nil
 	}
 
-	logger = logger.WithName(fmt.Sprintf("cluster=%s", cluster.Name))
+	logger = logger.WithValues("cluster", cluster.Name)
 
 	// Create the context.
 	ctx, err := context.NewClusterContext(&context.ClusterContextParams{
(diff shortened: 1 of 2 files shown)


akutz:
Hi @andrewsykim,

Are we sure `WithValues` is the same as `WithName`? Per the [GoDocs](https://godoc.org/github.com/go-logr/logr):

```golang
    // WithValues adds some key-value pairs of context to a logger.
    // See Info for documentation on how key/value pairs work.
    WithValues(keysAndValues ...interface{}) Logger

    // WithName adds a new element to the logger's name.
    // Successive calls with WithName continue to append
    // suffixes to the logger's name.  It's strongly reccomended
    // that name segments contain only letters, digits, and hyphens
    // (see the package documentation for more information).
    WithName(name string) Logger
```

It sure seems like `WithName` has a different distinction. But, then again, the go-logr `Logger` interface is just that, an interface. Looking at an implementation in the controller-runtime's [`DelegatingLogger`](https://github.com/kubernetes-sigs/controller-runtime/blob/ab6131a999ca9e6ff5bf83dba70dc7751751135f/pkg/log/deleg.go#L36-L62), it also appears that `WithName` initializes separate fields than the values:

```golang
// WithName provides a new Logger with the name appended
func (p *loggerPromise) WithName(l *DelegatingLogger, name string) *loggerPromise {
	res := &loggerPromise{
		logger:       l,
		name:         &name,
		promisesLock: sync.Mutex{},
	}

	p.promisesLock.Lock()
	defer p.promisesLock.Unlock()
	p.childPromises = append(p.childPromises, res)
	return res
}

// WithValues provides a new Logger with the tags appended
func (p *loggerPromise) WithValues(l *DelegatingLogger, tags ...interface{}) *loggerPromise {
	res := &loggerPromise{
		logger:       l,
		tags:         tags,
		promisesLock: sync.Mutex{},
	}

	p.promisesLock.Lock()
	defer p.promisesLock.Unlock()
	p.childPromises = append(p.childPromises, res)
	return res
}
```

andrewsykim:
Name is what you see prefixed in a log line, but I don't think it makes sense to use names like `namespace=%s`.  

Also worth mentioning in the comments:
```
    // WithName adds a new element to the logger's name.
    // Successive calls with WithName continue to append
    // suffixes to the logger's name.  It's strongly reccomended
    // that name segments contain only letters, digits, and hyphens
    // (see the package documentation for more information).
    WithName(name string) Logger
```

akutz:
Hi @andrewsykim,

> Name is what you see prefixed in a log line, but I don't think it makes sense to use names like namespace=%s.

I merely copied from CAPA. However, I agree that it violates the commented guidelines. I also maintain that the information is necessary due to the fact that each log entry about a resource benefits from describing the resource's provenance. This is in line with the way CAPV logged items in v1a1. 

This PR doesn't address either though. If you would like to avoid using special chars in `WithName` then please close this PR and open a new one that brings the loggers back in line with the way were initialized in v1a1. For example:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/64600d54f5427a3cd5b14e794aab0530469a9880/pkg/cloud/vsphere/context/cluster_context.go#L103

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/64600d54f5427a3cd5b14e794aab0530469a9880/pkg/cloud/vsphere/context/machine_context.go#L90

andrewsykim:
will do! 

/close

PR #556, opened 2019-09-06
add label to be able to list vspheremachines that are part of cluster

yastij:

**What this PR does / why we need it**: This PR adds the following labels:

```
  labels:
    cluster.x-k8s.io/control-plane: "true"
    cluster.x-k8s.io/cluster-name: "${CLUSTER_NAME}"
```

to be able to list vspheremachines that are part of a cluster through the function `GetVSphereMachinesInCluster()`.
Without this change, in #538 the loadbalancer_controller is not able to list vspheremachines  which result in the creation of a load balancer without a corresponding pool of machines/ips attached to it

Fixes #

/assign @akutz 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/examples/controlplane/controlplane.yaml
+++ b/examples/controlplane/controlplane.yaml
@@ -21,6 +21,9 @@ apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
 kind: VSphereMachine
 metadata:
   name: ${CLUSTER_NAME}-controlplane-0
+  labels:
+    cluster.x-k8s.io/control-plane: "true"
+    cluster.x-k8s.io/cluster-name: "${CLUSTER_NAME}"
 spec:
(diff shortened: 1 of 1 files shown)


akutz:
Hi @yastij,

For an issue like this I think filing an issue in advance is worth it to have a discussion first. The reason is I'm not sure if the labels belong on the `VSphereMachine` resources. We don't want to duplicate all the information from the CAPI `Machine` resources on the analogous, provider machine resources. It's a slippery slope.

Once you have one resource it should be possible to get the other.

akutz:
Per our discussion in Slack, this looks consistent with the CAPI model. /lgtm

akutz:
Removing hold

/hold cancel

I was just checking to see if we needed to make the same change to the worker machines.

PR #558, opened 2019-09-07
Remove vendor & update Makefile

akutz:
This patch removes vendor and updates the Makefile to be in-line with other upstream infra providers. This PR does not include changes to the CAPV release process (or the bits in the Makefile to support it) as that will have to occur later.

Fixes #548

**Release note**:

```release-note

```


akutz:
cc @vincepri

PR #559, opened 2019-09-09
Add service domain to the cluster config

datianshi:
When kubeadconfig controller reconciles. It needs dnsDomain field. 
This documents in https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/master/README.md

image

The current manifest generation does not provide that field, which caused below error on cabpk

```validation failure list:\nspec.clusterConfiguration.networking.dnsDomain in body is required"  "controller"="kubeadmconfig" "request"={"Namespace":"default","Name":"management-cluster-controlplane-0”}```

This pull request will add that field and fix ^^ issue 

Fixes #

This PR does not change any image... 
But gcr.io/cluster-api-provider-vsphere/release/manifests:latest should be updated upon this change

**Release note**:

```release-note
the manifest generator will add Cluster.spec.clusterNetwork.serviceDomain field
```

### DIFF
--- a/examples/cluster/cluster.yaml
+++ b/examples/cluster/cluster.yaml
@@ -9,6 +9,7 @@ spec:
       cidrBlocks: ["${SERVICE_CIDR}"]
     pods:
       cidrBlocks: ["${CLUSTER_CIDR}"]
+    serviceDomain: ${DNS_DOMAIN}
   infrastructureRef:
     apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
     kind: VSphereCluster
--- a/examples/cluster/cluster.yaml
+++ b/examples/cluster/cluster.yaml
@@ -9,7 +9,7 @@ spec:
       cidrBlocks: ["${SERVICE_CIDR}"]
     pods:
       cidrBlocks: ["${CLUSTER_CIDR}"]
-    serviceDomain: ${DNS_DOMAIN}
+    serviceDomain: ${SERVICE_DOMAIN}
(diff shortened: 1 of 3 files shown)


akutz:
We've successfully deployed clusters with `master`, so either:

1. This error must not be fatal or
2. This requirement was added very recently to the bootstrapper

I'd like to figure out which it is in case we need to set up some means of figuring out when breaking changes occur in the new bootstrap provider.

examples/cluster/cluster.yaml
```diff
@@ -9,6 +9,7 @@ spec:
       cidrBlocks: ["${SERVICE_CIDR}"]
     pods:
       cidrBlocks: ["${CLUSTER_CIDR}"]
+    serviceDomain: ${DNS_DOMAIN}
```
akutz: Hi @datianshi,

Can you please rename the environment variable to `SERVICE_DOMAIN`?

examples/generate.sh
```diff
@@ -173,6 +173,7 @@ record_and_export() {
 record_and_export CLUSTER_NAME          ':-capv-mgmt-example'
 record_and_export SERVICE_CIDR          ':-100.64.0.0/13'
 record_and_export CLUSTER_CIDR          ':-100.96.0.0/11'
+record_and_export DNS_DOMAIN            ':-cluster.local'
```
akutz: Hi @datianshi,

See the above comment about `SERVICE_DOMAIN`

PR #574, opened 2019-09-14
The default version changed to 1.14.4 but docs does not change

datianshi:
The default version changed to 1.14.4: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/examples/generate.sh#L214

As I thought the default KUBERNETES_VERSION version is 1.13.6 without specifying it. It caused the error when kubeadm join api-server

```
Cluster info signature and contents are valid and TLS certificate validates against pinned roots, will use API Server "10.197.120.215:6443"
[discovery] Successfully established connection with API Server "10.197.120.215:6443"
[join] Reading configuration from the cluster...
[join] FYI: You can look at this config file with 'kubectl -n kube-system get cm kubeadm-config -oyaml'
[kubelet] Downloading configuration for the kubelet from the "kubelet-config-1.13" ConfigMap in the kube-system namespace
configmaps "kubelet-config-1.13" is forbidden: User "system:bootstrap:dtscpf" cannot get resource "configmaps" in API group "" in the namespace "kube-system"
```

Fixes #

Probably consider ```export KUBERNETES_VERSION=``` **not** an optional param

**Release note**:

```release-note

```

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -66,9 +66,9 @@ In order for `clusterctl` to bootstrap a management cluster on vSphere, it must 
 
 #### Uploading the CAPV Machine Image
 
-It is required that machines provisioned by CAPV use one of the official CAPV machine images as a VM template. The machine images are retrievable from public URLs. CAPV currently supports machine images based on Ubuntu 18.04 and CentOS 7. A list of published machine images is available [here](../README.md#kubernetes-versions-with-published-ovas). For this guide we'll be deploying Kubernetes v1.13.6 on Ubuntu 18.04 (link to [machine image](https://storage.googleapis.com/capv-images/release/v1.13.6/ubuntu-1804-kube-v1.13.6.ova)).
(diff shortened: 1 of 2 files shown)


datianshi:
Realized VSPHERE_TEMPLATE is required and I did provide that with 1.13.6

Readme needs this update anyway... since 

KUBERNETES_VERSION though is hidden in generate.sh, while think that should be exposed/configured/documented

datianshi:
which I mean :) : KUBERNETES_VERSION needs to match the VM template version. and probably should be mandatory instead of optional

akutz:
/hold

I will review in a bit. I'm not sure I understand the change...

akutz:
I just saw a second update. Thanks for the clarification. I'm about to run out the door, but I'll look at this later tonight. Thank you again!

akutz:
Hi @datianshi,

Okay, this looks fine. However, do we want to go ahead and update this to 1.15.3 while we're at it?

datianshi:
@akutz  that makes sense. Happy to update this pull request with 1.15.3

akutz:
Hi @datianshi,

Sounds great! Thank you again for doing this!

akutz:
Hi @datianshi,

Perfect! Do you mind squashing the commits, or would you prefer to keep them separate?

datianshi:
squashed, and it makes sense :)  @akutz

akutz:
/hold cancel
/approve
/lgtm

Thank you again @datianshi!

Issue #575, opened 2019-09-16
Support the vSphere CSI Driver for vSphere >= 6.7u3


andrewsykim:
/kind feature

**Describe the solution you'd like**
In https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/502, we added support to use the external vSphere cloud provider for vSphere >= 6.7u3. Along with that we should also be deploying the [vSphere CSI driver](https://github.com/kubernetes-sigs/vsphere-csi-driver) for any CAPV deployment that is using the external cloud provider as well (i.e. vSphere >= 6.7u3)

**Anything else you would like to add:**
cc @divyenpatel @sandeeppissay 

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

divyenpatel:
@andrewsykim If we can provide images for `CSI controller`, `node plugin`, `external-attacher`, `external-provisioner` and `metadata syncer/full sync` containers as a parameter it would be great.

andrewsykim:
@divyenpatel noted!

PR #578, opened 2019-09-20
Support vSphere CSI for vSphere >= 6.7u3

andrewsykim:

Just adding the Go types for now, will add the rest of the integration soon. 

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/575

**Release note**:

```release-note
Support vSphere CSI for vSphere >= 6.7u3
```


akutz:
/hold

Hi @andrewsykim,

I'm quite concerned about the potential for drift when it comes to the way we're creating the deployment resources for both this and CCM. I was okay with the CCM as-is, but now we're just introducing more of the same future, tech-debt, and [in the words](https://youtu.be/Jln3mi0vfJU) of Captain Jean-Luc Picard:

image

We need to explore one of the following options:

1. Stick with the way you've implemented the CCM and CNS deployment resources, but move the generation code to a package inside their respective repositories. That way the authoritative source of deployment generation for those services is located alongside those services.
2. Use the new controller-runtime client that supports [unstructured data](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#Client). It's astonishingly simple to use. This would provide the ability to pull in remote manifest YAML either at runtime or stored inside of the CAPW repository. The YAML would then be unmarshalled into a `map[string]interface{}` into an unstructured object like so:

    ```golang
    import (
    	"context"
    	"strings"

    	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
    	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
    	"sigs.k8s.io/controller-runtime/pkg/client"
    	"sigs.k8s.io/yaml"
    )

    // ...

    // Pretend the controller runtime client is initialized.
    var c client.Client

    // Pretend yamlString contains a multi-doc YAML payload that is one or
    // more Kubernetes YAML manifests.
    var yamlString string

    // Create a YAML reader to parse the mult-doc YAML string into discreet
    // YAML manifests.
    yamlReader := utilyaml.NewYAMLReader(strings.NewReader(yamlString))

    // Loop until the YAML reader has no more documents.
    for {
    	// Read a YAML doc.
    	buf, err := yamlReader.Read()
    	if err != nil {
    		if err == io.EOF {
    			return nil
    		}
    		return err
    	}

    	// Define an unstructured object.
    	obj := &unstructured.Unstructured{
    		Object: map[string]interface{},
    	}

    	// Unmarshal the YAML string into the unstructured object.
    	if err := yaml.Unmarshal(buf, obj.Object); err != nil {
    		return err
    	}

    	// Create the object on the API server.
    	if err := c.Create(context.Background(), obj); err != nil {
    		return err
    	}
    }
    ```

The fact of the matter is that it's unlikely we'll ever revisit this code until there's an issue. I'd rather measure twice and cut once now than be frustrated by a pressured situation later when we're responding to some bug due to an upstream change in resource configuration on the part of CNS or CCM.

akutz:
I forgot, to convert typed objects to unstructured ones, and back, you can use these helper functions I wrote:

```golang
func toUnstructured(name string, obj runtime.Object) *unstructuredv1.Unstructured {
       data, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
       Expect(err).NotTo(
               HaveOccurred(),
               "Error getting unstructured data for %s: %q",
               obj.GetObjectKind().GroupVersionKind(), name)
       return &unstructuredv1.Unstructured{Object: data}
}

func toStructured(name string, dst runtime.Object, src *unstructuredv1.Unstructured) {
       err := runtime.DefaultUnstructuredConverter.FromUnstructured(src.UnstructuredContent(), dst)
       Expect(err).NotTo(
               HaveOccurred(),
               "Error getting structured object for %s: %q",
               src.GroupVersionKind(), name)
}
```

Obviously these were for a test framework, but it's the same basic idea.

andrewsykim:
> Stick with the way you've implemented the CCM and CNS deployment resources, but move the generation code to a package inside their respective repositories. That way the authoritative source of deployment generation for those services is located alongside those services.

I plan to do this ^  as a follow-up, I think it would be useful for other consumers of those components as well anyways

akutz:
Hi @andrewsykim,

Any idea why the e2e test is failing? I'm issuing a retest now to see if it's ephemeral.

/retest

api/v1alpha2/cloud/types.go
```diff
@@ -51,9 +51,22 @@ type Config struct {
 // ProviderConfig defines any extra information used to configure
 // the vSphere external cloud provider
 type ProviderConfig struct {
+	CCM CCMConfig `json:"ccmConfig,omitempty"`
+	CNS CNSConfig `json:"cnsConfig,omitempty"`
+}
+
+type CCMConfig struct {
 	Image string `json:"image,omitempty"`
```
akutz: Hi @andrewsykim,

Please consider renaming `Image` to `ControllerImage` to match the `CNSConfig` or renaming `CNSConfig.ControllerImage` to `CNSConfig.Image`.

I personally think `CCMConfig.Image` should be `CCMConfig.ControllerImage` for consistency.

api/v1alpha2/cloud/types.go
```diff
@@ -51,9 +51,22 @@ type Config struct {
 // ProviderConfig defines any extra information used to configure
 // the vSphere external cloud provider
 type ProviderConfig struct {
+	CCM CCMConfig `json:"ccmConfig,omitempty"`
```
akutz: Hi @andrewsykim,

Please consider renaming field `CCM` to `Cloud` and field `CNS` to `Storage`.

controllers/vspherecluster_controller.go
```diff
@@ -161,7 +161,14 @@ func (r *VSphereClusterReconciler) reconcileNormal(ctx *context.ClusterContext)
 	// Create the external cloud provider addons
 	if err := r.reconcileCloudProvider(ctx); err != nil {
 		return reconcile.Result{}, errors.Wrapf(err,
-			"failed to reconcile cloud cloud provider for VSphereCluster %s/%s",
+			"failed to reconcile cloud provider for VSphereCluster %s/%s",
+			ctx.VSphereCluster.Namespace, ctx.VSphereCluster.Name)
+	}
+
+	// Create the vSphere CSI Driver addons
+	if err := r.reconcileCSIDriver(ctx); err != nil {
```
akutz: Hi @andrewsykim,

Maybe `reconcileStorageProvider`?

controllers/vspherecluster_controller.go
```diff
@@ -313,6 +320,59 @@ func (r *VSphereClusterReconciler) reconcileCloudProvider(ctx *context.ClusterCo
 	return nil
 }
 
+func (r *VSphereClusterReconciler) reconcileCSIDriver(ctx *context.ClusterContext) error {
```
akutz: Hi @andrewsykim,

As much as I debated you about the way you're doing this, it's damned beautiful code sir!

api/v1alpha2/cloud/types.go
andrewsykim: yeah I like that too, will update

api/v1alpha2/cloud/types.go
andrewsykim: sgtm

controllers/vspherecluster_controller.go
andrewsykim: sgtm

api/v1alpha2/cloud/types.go
andrewsykim: Going to use `CloudProvider` instead to avoid the package name stuter

andrewsykim: nvm, going to rename the package instead

api/v1alpha2/cloudprovider/doc.go
```diff
@@ -26,4 +26,4 @@ limitations under the License.
 // support reflecting a struct with a field of type "map[string]TYPE" to INI.
 //
 // +kubebuilder:object:generate=true
-package cloud
```
andrewsykim: renamed this package since golint complains about `cloud` stuffering with any other struct starting with `Cloud`, plus I think `cloudprovider` is a slighlty better name for this package anyways.

andrewsykim:
@akutz ready for another review. 

/assign @divyenpatel @codenrhoden @dvonthenen 

PTAL as well if you can

dvonthenen:
This looks fine to me based on what I know of the new CNS CSI deployment.

docs/getting_started.md
```diff
@@ -212,6 +212,15 @@ spec:
       secretNamespace: kube-system
     network:
       name: vm-network-1
+    providerConfig:
+      cloud:
+        controllerImage: gcr.io/cloud-provider-vsphere/cpi/release/manager:v1.0.0
```
akutz: Hi @andrewsykim,

In the `cluster.yaml` these values are quoted, but they aren't here. Just a small nit.

andrewsykim: I think this is true of all values though, something in the generator script (I think kustomize) removes the quotes on string values

PR #590, opened 2019-09-26
Allow user to set kubeadm clusterConfiguration imageRepository

figo:
kubeadm using "k8s.gcr.io" by default, we default it to same
default value, although this change allow user to set it to different value.

Fixes #

**Release note**:

```release-note
* Allow specification of a custom, default image repository
```

### DIFF
--- a/examples/generate.sh
+++ b/examples/generate.sh
@@ -33,6 +33,7 @@ ENV_VAR_REQ=':?required'
 CABPK_MANAGER_IMAGE="${CABPK_MANAGER_IMAGE:-us.gcr.io/k8s-artifacts-prod/capi-kubeadm/cluster-api-kubeadm-controller:v0.1.0}"
 CAPI_MANAGER_IMAGE="${CAPI_MANAGER_IMAGE:-us.gcr.io/k8s-artifacts-prod/cluster-api/cluster-api-controller:v0.2.1}"
 CAPV_MANAGER_IMAGE="${CAPV_MANAGER_IMAGE:-gcr.io/cluster-api-provider-vsphere/release/manager:latest}"
+K8S_IMAGE_REPOSITORY="${K8S_IMAGE_REPOSITORY:-k8s.gcr.io}"
 
 # Set the default log levels for the manager containers.
 CABPK_MANAGER_LOG_LEVEL="${CABPK_MANAGER_LOG_LEVEL:-4}"
@@ -56,6 +57,7 @@ FLAGS
   -i    input directory (default ${SRC_DIR})
   -m    capv manager image (default "${CAPV_MANAGER_IMAGE}")
   -M    capv manager log level (default "${CAPV_MANAGER_LOG_LEVEL}")
+  -r    kubernetes container image repository (default "${K8S_IMAGE_REPOSITORY}")
   -o    output directory (default ${OUT_DIR})
   -p    capi manager image (default "${CAPI_MANAGER_IMAGE}")
   -P    capi manager log level (default "${CAPI_MANAGER_LOG_LEVEL}")
@@ -63,7 +65,7 @@ FLAGS
 EOF
 }
 
-while getopts ':b:B:c:dfhi:m:M:o:p:P:u' opt; do
+while getopts ':b:B:c:dfhi:m:M:r:o:p:P:u' opt; do
   case "${opt}" in
   b)
     CABPK_MANAGER_IMAGE="${OPTARG}"
@@ -91,6 +93,9 @@ while getopts ':b:B:c:dfhi:m:M:o:p:P:u' opt; do
     ;;
   M)
     CAPV_MANAGER_LOG_LEVEL="${OPTARG}"
+    ;;
+  r)
+    K8S_IMAGE_REPOSITORY="${OPTARG}"
     ;;
   o)
     OUT_DIR="${OPTARG}"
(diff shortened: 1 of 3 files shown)


figo:
cc @fabianofranz @chuckha @akutz  i think this going to address the issue of https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/131 for capv, although each provider need to add same setting if user want custom image repository, let me know your thought here.

examples/generate.sh
```diff
@@ -287,4 +293,4 @@ echo "WARNING: ${PROVIDER_COMPONENTS_GENERATED_FILE} includes vSphere credential
 
 # If running in Docker then ensure the contents of the OUT_DIR have the
 # the same owner as the volume mounted to the /out directory.
-[ "${DOCKER_ENABLED-}" ] && chown -R "$(stat -c '%u:%g' /out)" "${OUT_DIR}"
+#[ "${DOCKER_ENABLED-}" ] && chown -R "$(stat -c '%u:%g' /out)" "${OUT_DIR}"
```
akutz: Hi @figo,

There is a reason this line is here. Why did you comment it out?

figo: sorry, thats just for my local testing, will restore it,

akutz:
Hi @figo,

So I don't think this PR fixes https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/131 the way @detiber, @chuckha, or I would like to see it fixed. I think the goal is to be able to submit a `KubeadmConfig` resource that CABPK can use as the _default_ for a cluster, and merge any values necessary into each machine's associated `KubeadmConfig` resource.

Still, for a temp fix, perhaps this is fine.

However, I believe the fix is incorrect. A [`KubeadmConfig`](https://godoc.org/github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2#KubeadmConfig) resource doesn't have a field `imageRespository` at the root of the type. It needs to be set in the spec's `clusterConfiguration`.

Also, doesn't this need to be set for etcd and CoreDNS as well? Something like:

```yaml
apiVersion: bootstrap.cluster.x-k8s.io/v1alpha2
kind: KubeadmConfig
metadata:
  labels:
    cluster.x-k8s.io/cluster-name: capv
    cluster.x-k8s.io/control-plane: "true"
  name: capv-control-plane-0
spec:
  clusterConfiguration:
    clusterName: capv
    imageRepository: k8s.io
    dns:
      imageRepository: k8s.io
      imageTag: v1.0.0
      type: CoreDNS
    etcd:
      local:
        dataDir: /var/lib/etcd
        imageRepository: k8s.io
        imageTag: v1.0.0
```

Please note that the `etc.dataDir` field is **required** or the API server will reject an attempt to specify a custom etcd image repository if the `dataDir` field is empty.

akutz:
/hold

(@figo, you can't place a hold if there's other text on the same line)

figo:
> Hi @figo,
> 
> So I don't think this PR fixes [kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm#131](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/131) the way @detiber, @chuckha, or I would like to see it fixed. I think the goal is to be able to submit a `KubeadmConfig` resource that CABPK can use as the _default_ for a cluster, and merge any values necessary into each machine's associated `KubeadmConfig` resource.
> 
> Still, for a temp fix, perhaps this is fine.
> 
> However, I believe the fix is incorrect. A [`KubeadmConfig`](https://godoc.org/github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2#KubeadmConfig) resource doesn't have a field `imageRespository` at the root of the type. It needs to be set in the spec's `clusterConfiguration`.
> 
> Also, doesn't this need to be set for etcd and CoreDNS as well? Something like:
> 
> ```yaml
> apiVersion: bootstrap.cluster.x-k8s.io/v1alpha2
> kind: KubeadmConfig
> metadata:
>   labels:
>     cluster.x-k8s.io/cluster-name: capv
>     cluster.x-k8s.io/control-plane: "true"
>   name: capv-control-plane-0
> spec:
>   clusterConfiguration:
>     clusterName: capv
>     imageRepository: k8s.gcr.io
>     dns:
>       imageRepository: k8s.gcr.io
>       imageTag: v1.0.0
>       type: CoreDNS
>     etcd:
>       local:
>         dataDir: /var/lib/etcd
>         imageRepository: k8s.gcr.io
>         imageTag: v1.0.0
> ```
> 
> Please note that the `etc.dataDir` field is **required** or the API server will reject an attempt to specify a custom etcd image repository if the `dataDir` field is empty.
> 
> Also please note that the etcd image repository is only valid if using a local etcd server, not when using an external etcd server.

You are right, i do tested this config
```
  clusterConfiguration:
    imageRepository: ${K8S_IMAGE_REPOSITORY}
``` 
Would like to learn more about `etcd.dataDir`,  kubeadm does looking for the custom repository of etcd, i will double check the api-server, thanks.

For external etcd server, it will not use local etcd image, correct?

akutz:
Hi @figo,

> For external etcd server, it will not use local etcd image, correct?

That is correct sir.

akutz:
Hi @figo,

Pending testing this looks good to me! Please feel free to remove the hold.

/lgtm

PR #593, opened 2019-09-26
Fix cloudprovider encoding tests

andrewsykim:
1) Adds new line after each section of the cloud config INI marshallar
2) fixes equal comparison in TestMarshalINI

Fixes #

**Release note**:

```release-note
NONE
```


api/v1alpha2/cloudprovider/encoding_test.go
```diff
@@ -281,7 +282,7 @@ func TestMarshalINI(t *testing.T) {
 				}
 			}
 
-			g.Expect(string(buf), tc.iniString,
```
andrewsykim: This doesn't actually do a string comparison so the test were always passing

PR #594, opened 2019-09-27
support cluster-id in cloud config which is required by vSphere CSI driver

andrewsykim:

Supports adding the `cluster-id` field in the cloud config which is required by vSphere CSI driver. Uses the format "<cluster-namespace>/<cluster-name>" which is guaranteed to be unique on the management cluster. 

**Release note**:

```release-note
Supports adding the `cluster-id` field in the cloud config which is required by vSphere CSI driver.
```

### DIFF
--- a/pkg/cloud/vsphere/services/cloudprovider/csi.go
+++ b/pkg/cloud/vsphere/services/cloudprovider/csi.go
@@ -14,6 +14,8 @@ limitations under the License.
 package cloudprovider
 
 import (
+	"fmt"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -562,6 +564,7 @@ func CSICloudConfigSecret(data string) *corev1.Secret {
 func ConfigForCSI(ctx *context.ClusterContext) *cloudprovider.Config {
 	config := &cloudprovider.Config{}
 
+	config.Global.ClusterID = fmt.Sprintf("%s/%s", ctx.Cluster.Namespace, ctx.Cluster.Name)
 	config.Global.Insecure = false
 	config.Network.Name = ctx.VSphereCluster.Spec.CloudProviderConfiguration.Network.Name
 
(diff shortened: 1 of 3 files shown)


pkg/cloud/vsphere/services/cloudprovider/csi.go
```diff
@@ -562,6 +564,7 @@ func CSICloudConfigSecret(data string) *corev1.Secret {
 func ConfigForCSI(ctx *context.ClusterContext) *cloudprovider.Config {
 	config := &cloudprovider.Config{}
 
+	config.Global.ClusterID = fmt.Sprintf("%s/%s", ctx.Cluster.Namespace, ctx.Cluster.Name)
```
akutz: Hi @andrewsykim,

You probably want to be careful here. If `ctx.Cluster.Namespace` is not set then you likely want to use `v1.DefaultNamespace` from K8s. Or just use the string `"default"`.

akutz: Hi @andrewsykim,

Also, either check for this to see if it's empty (and don't set it unless it is), or let's figure out a way to hide it from being assigned externally. As it is you've got as part of the CRD, and it shouldn't be if you're setting it at runtime no matter the value a user may have assigned.

andrewsykim: I don't think the namespace on a namespaced object be empty though

akutz: Okay, I guess the API server sets it to `default`. Either way, the config field shouldn't be public then.

andrewsykim: Which config field?

akutz: `ClusterID`

andrewsykim: That needs to be exported so we can set it from the cloudprovider package though.

andrewsykim: Spoke offline with Andrew, because `ClusterID` gets propagated as a field in `VSphereCluster`, at the very least we need a comment on the field that says it shouldn't be set by any other consumer of the API

PR #600, opened 2019-10-02
default insecure: true in the cloud provider since CAPV doesn't manage vCenter certificates yet

andrewsykim:

Sets `insecure: true` by default in the manifest generator, since CAPV doesn't really manage vCenter certificates yet. The option to set `insecure: false` still exists by updating it in the VSphereCluster resource.

Fixes #

**Release note**:

```release-note
default `insecure: true` in the cloud provider since CAPV doesn't manage vCenter certificates yet
```

### DIFF
--- a/examples/default/cluster/cluster.yaml
+++ b/examples/default/cluster/cluster.yaml
@@ -26,6 +26,7 @@ spec:
     global:
       secretName: "cloud-provider-vsphere-credentials"
       secretNamespace: "kube-system"
+      insecure: true
     virtualCenter:
       "${VSPHERE_SERVER}":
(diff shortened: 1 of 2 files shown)


examples/pre-67u3/cluster/cluster.yaml
```diff
@@ -25,6 +25,7 @@ spec:
     global:
       secretName: "cloud-provider-vsphere-credentials"
       secretNamespace: "kube-system"
+      insecure: true
```
akutz: Hi @andrewsykim,

I don't think this is required in the pre-67u3 config.

andrewsykim: good catch, updated

PR #605, opened 2019-10-04
add well-known tolerations to released CAPV manager manifest

andrewsykim:

Adds toleration for well-known taints like node-role.kubernetes.io/master to released CAPV manager manifest.

Fixes kubernetes-sigs/cluster-api#1486

**Release note**:

```release-note
Adds toleration for well-known taints like node-role.kubernetes.io/master to released CAPV manager manifest.
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -29,3 +29,6 @@ spec:
         imagePullPolicy: Always
         name: manager
       terminationGracePeriodSeconds: 10
+      tolerations:
+        - effect: NoSchedule
+          key: node-role.kubernetes.io/master


andrewsykim:
> Please add this as a patch in the same dir.

Is this necessary if all variations of this will have this toleration?

akutz:
> Is this necessary if all variations of this will have this toleration?

I only want to make sure we can replace `manager.yaml` with whatever upstream template gets updated over time. If the upstream template includes the toleration, then this is fine as is.

andrewsykim:
Yeah this should be consistent with upstream - https://github.com/kubernetes-sigs/cluster-api/pull/1487

PR #618, opened 2019-10-11
Parameterized livenessProbe image and registrar image

jiatongw:

There are two image repos that hard coded in `csi.go`: `livenessprobe` and `csi-node-driver-registrar`. This PR parameterized those two image repos so that we could pass them as a parameter in `cluster.yaml`. 

Tested by deploying a 3 worker nodes cluster, replaced `livenessProbeImage` and `registrarImage` fields to point to my personal image repo in `cluster.yaml` and it is working. Please double check

### DIFF
--- a/api/v1alpha2/cloudprovider/types.go
+++ b/api/v1alpha2/cloudprovider/types.go
@@ -65,6 +65,8 @@ type StorageConfig struct {
 	AttacherImage       string `json:"attacherImage,omitempty"`
 	ProvisionerImage    string `json:"provisionerImage,omitempty"`
 	MetadataSyncerImage string `json:"metadataSyncerImage,omitempty"`
+	LivenessProbeImage  string `json:"livenessProbeImage,omitempty"`
+	RegistrarImage      string `json:"registrarImage,omitempty"`
 }
 
 // unmarshallableConfig is used to unmarshal the INI data using the gcfg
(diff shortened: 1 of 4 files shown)


api/v1alpha2/cloudprovider/types.go
```diff
@@ -65,6 +65,8 @@ type StorageConfig struct {
 	AttacherImage       string `json:"attacherImage,omitempty"`
 	ProvisionerImage    string `json:"provisionerImage,omitempty"`
 	MetadataSyncerImage string `json:"metadataSyncerImage,omitempty"`
+	LivenessProbeImage  string `json:"livenessProbeImage,omitempty"`
```
akutz: Hi @jiatongw,

This line and the next should have `// +optional` as Kubernetes API guidelines state wherever there is `omitempty`,  the OpenAPI requires `// +optional` for said field.

Yes, I realize other fields are missing it, but that's because I need to go whip @andrewsykim with a wet piece of string and get him to update those fields :)

api/v1alpha2/cloudprovider/types.go
```diff
@@ -65,6 +65,8 @@ type StorageConfig struct {
...
+	RegistrarImage      string `json:"registrarImage,omitempty"`
```
akutz: Hi @jiatongw,

Please see the previous comment.

api/v1alpha2/cloudprovider/types.go
akutz: Also, please add GoDoc comments for exported fields. Regarding the other fields, see the note about A Kim and a piece of wet string :)

andrewsykim:
re: @akutz's comment, I can address those in a follow-up PR since the original mistake is on me. 
thanks @jiatongw 

/hold cancel
/approve
/lgtm

PR #631, opened 2019-10-15
[WIP] add defaults for CSI / CPI images

### LINKED ISSUE #629: Set sane defaults for CPI / CNS image repos
/kind feature

Now with CAPV deploying out-of-tree CPI / CSI by default, we should set sane defaults for those images if they are not specified in the cluster spec. Right now they are always required in the cluster spec or the CSI / CPI deployments do not work. 

**Environment:**

andrewsykim:

Set default images for CPI / CSI integrations 

Fixes #629 

**Release note**:

```release-note
Set default images for CSI / CPI integrations 
```


pkg/cloud/vsphere/services/cloudprovider/csi.go
```diff
@@ -27,6 +27,16 @@ import (
 
 // NOTE: the contents of this file are derived from https://github.com/kubernetes-sigs/vsphere-csi-driver/tree/master/manifests/1.14
 
+const (
+	DefaultCSIControllerImage     = "vmware/vsphere-block-csi-driver:v1.0.0"
+	DefaultCSINodeDriverImage     = "vmware/vsphere-block-csi-driver:v1.0.0"
```
mylesagray: Should we use v1.0.1 of the CSI images?

andrewsykim: yes -- this PR is just about setting defaults, I think image updates should happen in a separate PR

controllers/vspherecluster_controller.go
```diff
@@ -321,6 +333,49 @@ func (r *VSphereClusterReconciler) reconcileCloudProvider(ctx *context.ClusterCo
...
+		return nil
+	}
+
+	// if at least 1 field in the storage config is defined, assume CNS should be installed
+	// and use default images when not defined
+	if storageConfig.ControllerImage == "" {
+		storageConfig.ControllerImage = cloudprovider.DefaultCSIControllerImage
+	}
+
+	if storageConfig.NodeDriverImage == "" {
```
yastij: @andrewsykim  - maybe at some point we should consider shipping webhooks for both high-level validation and defaulting

andrewsykim:
> @andrewsykim If I understand correctly, I deleted storage field in cluster.yaml since this will be default deployed once this change applied.  

@jiatongw so if you delete the storage field entirely, this change skips deploying CSI. If you want the defaults to kick in, you have to set empty object like this:
```
storage: {}
```

OR 

specify at least one of the images for the defaults.

andrewsykim:
^ would appreciate input on if this API change makes sense. I think we want some way to indicate "don't deploy CSI" but maybe this is the wrong approach

yastij:
I agree, this is a case where we want to distinguish between not set and empty:

- field not set: skip deploying
- field empty: deploy the default

API-wise this is something fine IMO

controllers/vspherecluster_controller.go
akutz: HA! I was literally just mentioning this to Andrew yesterday :)

pkg/cloud/vsphere/services/cloudprovider/csi.go
```diff
@@ -138,7 +148,7 @@ func CSIDriver() *storagev1beta1.CSIDriver {
 	}
 }
 
-func VSphereCSINodeDaemonSet(storageConfig cloudprovider.StorageConfig) *appsv1.DaemonSet {
+func VSphereCSINodeDaemonSet(storageConfig *cloudprovider.StorageConfig) *appsv1.DaemonSet {
```
akutz: Why pass by reference? Unless changes are made, passing by copy is actually a lot easier on Go.

pkg/cloud/vsphere/services/cloudprovider/csi.go
```diff
@@ -343,7 +353,7 @@ func LivenessProbeForNodeContainer(image string) corev1.Container {
 	}
 }
 
-func CSIControllerStatefulSet(storageConfig cloudprovider.StorageConfig) *appsv1.StatefulSet {
+func CSIControllerStatefulSet(storageConfig *cloudprovider.StorageConfig) *appsv1.StatefulSet {
```
akutz: Why pass by reference? Unless changes are made, passing by copy is actually a lot easier on Go.

akutz:
/lgtm
/hold

In case anyone else has any final comments.

pkg/cloud/vsphere/services/cloudprovider/csi.go
andrewsykim: It's cause `StorageConfig` is now a pointer, otherwise we can't do nil checks to disable the storage config

akutz: You can always dereference when passing in.

andrewsykim: good point

controllers/vspherecluster_controller.go
yastij: I’ll open an issue for this one.

Issue #609, opened 2019-10-08
infrastructure-components.yaml should used tagged manager images


andrewsykim:
/kind bug

**What steps did you take and what happened:**
Currently, the output of `make release-manifests` (which is released as `infrastructure-components.yaml`) uses the `latest` tag for the CAPV manager. This is misleading since that means older releases of `infrastructure-components.yaml`, can end up pointing to new versions of CAPV. 

**What did you expect to happen:**
The released version of `infrastructure-components.yaml` should match the tagged version of the manager version in the manifests.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

PR #634, opened 2019-10-17
allow passing a VERSION when generating infrastructure-components.yaml

yastij:

**What this PR does / why we need it**: This PR allows us to pass a fixed version when generating `infrastructure-components.yaml`. it works using : `make release-manifests VERSION=x.y.z` 

**Which issue(s) this PR fixes** : Fixes https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/609

/assign @andrewsykim 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,9 +165,12 @@ generate-manifests: $(CONTROLLER_GEN) ## Generate manifests e.g. CRD, RBAC etc.
 
 .PHONY: release-manifests
 release-manifests: ## Builds the manifests to publish with a release
+ifndef VERSION
+	$(error VERSION is undefined)
+endif
 	@mkdir -p out
-	$(KUSTOMIZE) build config/default >out/infrastructure-components.yaml
-
+	cd config/manager/; ../../"$(KUSTOMIZE)" edit set image gcr.io/cluster-api-provider-vsphere/release/manager:"$(VERSION)"
+	"$(KUSTOMIZE)" build config/default > out/infrastructure-components.yaml
 ## --------------------------------------
 ## Cleanup / Verification
 ## --------------------------------------


Makefile
```diff
@@ -166,6 +166,7 @@ generate-manifests: $(CONTROLLER_GEN) ## Generate manifests e.g. CRD, RBAC etc.
 .PHONY: release-manifests
 release-manifests: ## Builds the manifests to publish with a release
 	@mkdir -p out
+	cd config/manager/; ../../$(KUSTOMIZE) edit set image gcr.io/cluster-api-provider-vsphere/release/manager=gcr.io/cluster-api-provider-vsphere/release/manager:$(VERSION)
```
andrewsykim: need to error if VERSION isn't set

andrewsykim: Don't have much experience with kustomize, but I think this works too and seems cleaner:
```
cd config/manager/; ../../$(KUSTOMIZE) edit set image gcr.io/cluster-api-provider-vsphere/release/managerr:$(VERSION)
```

hack/release-manifests.sh
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o pipefail
+
+# TODO: remove this script once we have a stable.txt
+if [ -z "$2" ]
+then
+    echo "usage make release-manifests VERSION=x.y.z"
```
andrewsykim: Prefer this to be 
```
VERSION=x.y.z make release-manifests
```

hack/release-manifests.sh
```diff
@@ -0,0 +1,30 @@
...
+    exit 1
+fi
+mkdir -p out
```
akutz: Some comments on these lines would be nice.

Makefile
```diff
@@ -165,9 +165,12 @@ generate-manifests: $(CONTROLLER_GEN) ## Generate manifests e.g. CRD, RBAC etc.
 
 .PHONY: release-manifests
 release-manifests: ## Builds the manifests to publish with a release
+    ifndef VERSION
```
andrewsykim: indent here is still off. I think the `ifndef VESION` can be same indent as `release-manifest`, but the error line has to match indent with the rest of the target

PR #637, opened 2019-10-21
CAPI 0.2.6

akutz:
This patch updates the CAPI dependency to 0.2.6 and CABPK to 0.1.4.

/assign @andrewsykim


examples/generate.sh
```diff
@@ -30,8 +30,8 @@ OVERWRITE=
 CLUSTER_NAME="${CLUSTER_NAME:-capv-mgmt-example}"
```
andrewsykim: line 330 and 335 also need an update here

PR #640, opened 2019-10-23
Update CSI default images to v1.0.1 on gcr.io

codenrhoden:
Update the default CSI images used to the most recent release from GitHub repo, published on gcr.io.

Fixes #

**Release note**:

```release-note

```

/assign @andrewsykim 
/assign @akutz

### DIFF
--- a/examples/default/cluster/cluster.yaml
+++ b/examples/default/cluster/cluster.yaml
@@ -42,10 +42,10 @@ spec:
       cloud:
         controllerImage: "gcr.io/cloud-provider-vsphere/cpi/release/manager:v1.0.0"
       storage:
-        controllerImage: "vmware/vsphere-block-csi-driver:v1.0.0"
-        nodeDriverImage: "vmware/vsphere-block-csi-driver:v1.0.0"
+        controllerImage: "gcr.io/cloud-provider-vsphere/csi/release/driver:v1.0.1"
+        nodeDriverImage: "gcr.io/cloud-provider-vsphere/csi/release/driver:v1.0.1"
         attacherImage: "quay.io/k8scsi/csi-attacher:v1.1.1"
         provisionerImage: "quay.io/k8scsi/csi-provisioner:v1.2.1"
-        metadataSyncerImage: "vmware/volume-metadata-syncer:v1.0.0"
+        metadataSyncerImage: "gcr.io/cloud-provider-vsphere/csi/release/syncer:v1.0.1"
         livenessProbeImage: "quay.io/k8scsi/livenessprobe:v1.1.0"
         registrarImage: "quay.io/k8scsi/csi-node-driver-registrar:v1.1.0"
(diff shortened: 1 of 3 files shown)


examples/default/cluster/cluster.yaml
```diff
@@ -42,10 +42,10 @@ spec:
       cloud:
         controllerImage: "gcr.io/cloud-provider-vsphere/cpi/release/manager:v1.0.0"
       storage:
-        controllerImage: "vmware/vsphere-block-csi-driver:v1.0.0"
-        nodeDriverImage: "vmware/vsphere-block-csi-driver:v1.0.0"
+        controllerImage: "gcr.io/cloud-provider-vsphere/csi/release/driver:v1.0.1"
```
andrewsykim: This release has fix for https://github.com/kubernetes-sigs/vsphere-csi-driver/issues/70 right? We probably want to remove `hostNetwork: true` in our configs now? Can you do that here or in a separate PR?

codenrhoden: It does... I can add it to this PR. And it needs the `NODE_NAME` stuff if it isn't already there.

andrewsykim: yup, the liveness probes also use custom ports (because they conflicted with host network), we can remove those as well

codenrhoden: @andrewsykim Okay, I pulled in the changes from https://github.com/kubernetes-sigs/vsphere-csi-driver/pull/71/files.

Take a look if I did it right. :)

codenrhoden:
The CRD test failure seems unrelated. I may fix that in a separate PR first... Looks like it came from a dep revision.

pkg/cloud/vsphere/services/cloudprovider/csi.go
```diff
@@ -171,7 +171,7 @@ func VSphereCSINodeDaemonSet(storageConfig *cloudprovider.StorageConfig) *appsv1
 					},
 				},
 				Spec: corev1.PodSpec{
-					HostNetwork: true,
+					DNSPolicy: corev1.DNSDefault,
```
andrewsykim: I'm still curious why we need to use this DNS policy. @divyenpatel

examples/default/cluster/cluster.yaml
codenrhoden: oh, I still need to do those liveness probe ports like you said... Hopefully can do that in a bit.

pkg/cloud/vsphere/services/cloudprovider/csi.go
andrewsykim: No (and this is one of those bad API designs in PodSpec) -- the default value is actually "ClusterFirst" (check query kube-dns). "Default" actually refers to using the node's default nameserver for DNS, but it is not the default value.

codenrhoden:
@andrewsykim Okay, I pushed a new version that removes those custom ports. I rebased, so tests hopefully pass now.

I'm out of time for right now because I have to pickup a sick kiddo, but I did notice when doing the update for the custom ports that we are missing the `liveness probe` config and port exposure: https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/7fff38993a2dc8afd7c0e8c41e3751ab9f25cda4/manifests/1.14/deploy/vsphere-csi-node-ds.yaml#L78-L89

For both the node and controller. Probably need/want to add that in as well. If nobody gets to it, I will do it tonight.

andrewsykim:
Thanks @codenrhoden I'll add it in a follow-up PR :)

PR #641, opened 2019-10-23
Fix idempotency convention link

codenrhoden:
**What this PR does / why we need it**: The CRD test was failing because the generated CRDs have an updated link
to idempotency conventions.

Fixes #

**Release note**:

```release-note

```

/assign @andrewsykim

### DIFF
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachinetemplates.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachinetemplates.yaml
@@ -74,7 +74,7 @@ spec:
                         with Reason ServerTimeout indicating a unique name could not
                         be found in the time allotted, and the client should retry
                         (optionally after the time indicated in the Retry-After header).
-                        \n Applied only if Name is not specified. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#idempotency"
+                        \n Applied only if Name is not specified. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#idempotency"
                       type: string
                     labels:
                       additionalProperties:


config/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachinetemplates.yaml
```diff
@@ -74,7 +74,7 @@ spec:
                         with Reason ServerTimeout indicating a unique name could not
                         be found in the time allotted, and the client should retry
                         (optionally after the time indicated in the Retry-After header).
-                        \n Applied only if Name is not specified. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#idempotency"
+                        \n Applied only if Name is not specified. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#idempotency"
```
andrewsykim: Weird how this broke now 🤔 should have been caught in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/637 if vendor updates is what changed the expected generated output

codenrhoden: Yeah, I agree it's weird. was trying to dig into it...

Looks like this comes from `controller-tools`, which is not even listed as a dependency in `go.mod`. It's build directly, [here](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/b0cacda1bc84029045e2be4aa7a8e4c50b3ef134/hack/tools/Makefile#L53-L55), which looks like it would always pull in current master (or latest tagged version with Go Modules, I believe).

Oh, but it is versioned because there is a `go.mod` file in that directory, so it's a different module entirely. Let me keep looking. :)

andrewsykim: okay, no need to block this PR on that if CI is green :)

Issue #622, opened 2019-10-12
If you delete a VM and then delete the cluster, the cluster won't delete


voor:
/kind bug

Reading through some historical issues regarding externally deleted VMs.  What is the intended behavior currently?  I deleted a VM from the vSphere UI, but also deleted the cluster from the mission control cluster, and now things are kind of weird.  capi-controller-manager is saying `Cluster default/skew-timer still has 1 children - deleting them first` 

**What steps did you take and what happened:**

- Create a new cluster.
- Delete the control plane VM in the vSphere UI.
- Delete the CRD cluster in the management cluster.
- CAPI controller does not delete cluster since it still has `Cluster default/skew-timer still has 1 children - deleting them first` 
- `kubectl get machines`
```
NAME                                 PROVIDERID                                       PHASE
management-cluster-controlplane-0    vsphere://4223a39e-5e3f-d729-172c-b36f382508db   running
skew-timer-controlplane-0                                                             deleting
```
- CAPV says the following:
```
capv-controller-manager-6589478cb7-6lh8r manager I1012 00:58:45.419279       1 vspherecluster_controller.go:129] controllers/VSphereCluster/vspherecluster-controller/namespace=default/vsphereCluster=skew-timer/infrastructure.cluster.x-k8s.io/v1alpha2/cluster=skew-timer/cluster.x-k8s.io/v1alpha2/default/skew-timer "level"=0 "msg"="Reconciling VSphereCluster"  
capv-controller-manager-6589478cb7-6lh8r manager E1012 00:58:45.434994       1 controller.go:218] controller-runtime/controller "msg"="Reconciler error" "error"="failed to reconcile API endpoints for VSphereCluster default/skew-timer: failed to get KubeadmConfig skew-timer-controlplane-0/default for Machine default//skew-timer-controlplane-0: KubeadmConfig.bootstrap.cluster.x-k8s.io \"skew-timer-controlplane-0\" not found"  "controller"="vspherecluster" "request"={"Namespace":"default","Name":"skew-timer"}
[...]

PR #643, opened 2019-10-24
check vm existence even if machineRef is not set

yastij:

**What this PR does / why we need it**: This fix checks vm existence even if the machineRef is not set, this covers the case where a vm delete happens fast enough through vcenter

**Which issue(s) this PR fixes** : Fixes https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/622

/assign @andrewsykim @akutz 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/vsphere/services/govmomi/service.go
+++ b/pkg/cloud/vsphere/services/govmomi/service.go
@@ -135,19 +135,17 @@ func (vms *VMService) DestroyVM(ctx *context.MachineContext) (infrav1.VirtualMac
 		return vm, err
 	}
 
-	// check if VM actually exists
-	if ctx.VSphereMachine.Spec.MachineRef != "" {
-		moRefID, err := findVMByInstanceUUID(ctx)
-		if err != nil {
-			return vm, err
-		}
-		if moRefID == "" {
-			// No vm exists
-			// remove the MachineRef and set the vm state to notfound
-			ctx.VSphereMachine.Spec.MachineRef = ""
-			vm.State = infrav1.VirtualMachineStateNotFound
-			return vm, nil
-		}
+	// check if the vm existts
+	moRefID, err := findVMByInstanceUUID(ctx)
+	if err != nil {
+		return vm, err
+	}
+	if moRefID == "" {
+		// No vm exists
+		// remove the MachineRef and set the vm state to notfound
+		ctx.VSphereMachine.Spec.MachineRef = ""
+		vm.State = infrav1.VirtualMachineStateNotFound
+		return vm, nil
 	}
 
 	// VM actually exists


pkg/cloud/vsphere/services/govmomi/service.go
```diff
@@ -135,19 +135,17 @@ func (vms *VMService) DestroyVM(ctx *context.MachineContext) (infrav1.VirtualMac
...
-			return vm, err
-		}
-		if moRefID == "" {
-			// No vm exists
-			// remove the MachineRef and set the vm state to notfound
-			ctx.VSphereMachine.Spec.MachineRef = ""
-			vm.State = infrav1.VirtualMachineStateNotFound
-			return vm, nil
-		}
+	// check if the vm existts
```
sidharthsurana: Actually seems like there is an additional issue here
if the `ctx.VSphereMachine.Spec.MachineRef` is set then we should be getting the VM via the `getVMfromMachineRef` and not the `findVMByInstanceUUID(ctx)` directly. This is because in the case of pivot the `MachineRef` is the accurate reference, and in this situation, the `findVMByInstanceUUID` might not be the correct way to find the VM.

pkg/cloud/vsphere/services/govmomi/service.go
```diff
@@ -135,19 +135,17 @@ func (vms *VMService) DestroyVM(ctx *context.MachineContext) (infrav1.VirtualMac
...
+	moRefID, err := findVMByInstanceUUID(ctx)
```
andrewsykim: Does findVMByInstanceUUID return non-nil err if instance is not found?

sidharthsurana: yes that is the behavior
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/20c489391b13de471fb5e86e5e9de3621dc77982/pkg/cloud/vsphere/services/govmomi/util.go#L44-L55

andrewsykim: If so, then the early return in the next line seems like the problem?

andrewsykim: Right, but `FindByInstanceUUID` can still return non-nil error if not found?

pkg/cloud/vsphere/services/govmomi/service.go
akutz: Hi @sidharthsurana,

We're specifically avoiding the MoRef in this case to always do the instance UUID lookup. It's because this code removed the safeguards the old [`lookupVM` function](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/release-0.3/pkg/cloud/vsphere/services/govmomi/util.go#L171-L233) had to assert that a MoRef actually pointed to a real VM. The code @yastij wrote above bypasses the MoRef altogether to be certain of whether the VM still exists or not. This is an iterative step to removing the MoRef as the token in the `MachineRef` altogether. 

This was part of the service refactor, and I just didn't catch it when I reviewed that PR. We need to reintroduce the safeguards of `lookupVM`. If a MorRef was present, it was always checked to make sure it actually existed. It didn't just assume it did.

pkg/cloud/vsphere/services/govmomi/service.go
andrewsykim: Oh, you are saying it does return non-nil error if not found

pkg/cloud/vsphere/services/govmomi/service.go
sidharthsurana: Agreed! the previous logic in the lookupVM needs to be re-introduced!

pkg/cloud/vsphere/services/govmomi/service.go
sidharthsurana: yup

akutz: > If so, then the early return in the next line seems like the problem?

Which early return?

andrewsykim:
```
moRefID, err := findVMByInstanceUUID(ctx)
if err != nil {
      return vm, err // THIS ONE
}
if moRefID == "" {
    // No vm exists
    // remove the MachineRef and set the vm state to notfound
    ctx.VSphereMachine.Spec.MachineRef = ""    // THIS NEVER HAPPENS
    vm.State = infrav1.VirtualMachineStateNotFound
    return vm, nil
}
```

^ if err is not nil from findVMByInstanceUUID when a VM doesn't exist, then we never end up unset MachineRef

akutz: The fact that it used to work tells me no error is returned if the item isn't found.

akutz: No error is returned by `findVMByInstanceUUID` if the VM is simply not found. An error is only returned if there was an error finding the VM. See https://github.com/vmware/govmomi/blob/f01505d273e85c71261f91457191e65fcb971be3/object/search_index.go#L122-L144 for the call that `findVMByInstanceUUID` makes.

This is tested, existing code. We're fine.

andrewsykim: > The fact that it used to work tells me no error is returned if the item isn't found.

I didn't know this was the case, which version do we know this was working? Btw it's not machine delete that's not working, it's machine delete on an already deleted VM.

akutz: Hi @yastij,

I think @andrewsykim raises a good point. Let's add a comment summarizing the above behavior to **our** code so it's clear why `findByInstanceUUID` doesn't return an error if the VM isn't found.

akutz: > The fact that it used to work tells me no error is returned if the item isn't found.

I'm specifically referring to https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/release-0.3/pkg/cloud/vsphere/services/govmomi/util.go#L171-L233.

andrewsykim: `methods.FindByUuid(ctx, s.c, &req)` calls [RoundTrip](https://github.com/vmware/govmomi/blob/master/vim25/soap/client.go#L535-L606) which has this snippet of code 
```
switch res.StatusCode {
case http.StatusOK:
 // OK
case http.StatusInternalServerError:
// Error, but typically includes a body explaining the error
default:
    return errors.New(res.Status)
}
```
there's no case for a 404 which was why I was asking but I'm not sure if govmomi is expected to return 404 for this request anyways

akutz: Hi @andrewsykim,

This activity diagram is the best way I know to explain the old model:

machine-reconcile

andrewsykim: @dougm  do you know if SearchIndex `FindByUuid` returns a non-nil err if the VM doesn't exist?

akutz: Hi @andrewsykim,

> there's no case for a 404

Because you're not accessing a VM with the call, you're accessing the search index, which will never 404 unless the SDK server has a massive meltdown. That's the point of the MoRef. It's the basis of a functional API. Most functions are linked to services, ex. SearchIndex, and the first param of most calls is a MoRef used to indicate the object that is the subject of the operation.

andrewsykim: Thanks I was missing that context

akutz: FWIW, please see https://pubs.vmware.com/vi3/sdk/ReferenceGuide/vim.SearchIndex.html#findByUuid

> The virtual machine or host managed entity that is found. If no managed entities are found, null is returned. Only a single entity is returned, even if there are multiple matches.

If no object is found it isn't an error -- it just means a nil value is returned by that function. Which is good, because otherwise it would mean we'd need to check for a specific error. Although I suppose that at least is more obvious and in-line with Go. I wonder if that makes more sense? @dougm, what do you think? I suppose a change like that though could have massive consequences for existing code that uses the search index via govmomi.

andrewsykim:
Overall PR looks good. 

I was able to test/reproduce the bug as specified in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/622 and ran into an interesting case. 

Deleted the VM directly like so:
```
$ govc vm.destroy target-cluster01-md-0-7d99bd4955-snc8h
```

Check the `vspheremachine` resource:
```
$ kubectl get vspheremachine target-cluster01-md-0-hsjmv -o yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
kind: VSphereMachine
metadata:
  creationTimestamp: "2019-10-25T02:36:54Z"
  finalizers:
  - vspheremachine.infrastructure.cluster.x-k8s.io
  generateName: target-cluster01-md-0-
  generation: 3
  name: target-cluster01-md-0-hsjmv
  namespace: default
  ownerReferences:
  - apiVersion: cluster.x-k8s.io/v1alpha2
    kind: Machine
    name: target-cluster01-md-0-7d99bd4955-snc8h
    uid: 59aa2223-19e7-457c-b47e-6ff14b082cc3
  resourceVersion: "355425"
  selfLink: /apis/infrastructure.cluster.x-k8s.io/v1alpha2/namespaces/default/vspheremachines/target-cluster01-md-0-hsjmv
  uid: bd3bb10a-7202-4a2e-bc43-c84f5a78fb1f
spec:
  datacenter: SDDC-Datacenter
  diskGiB: 50
  machineRef: vm-36065
  memoryMiB: 2048
  network:
    devices:
    - dhcp4: true
      dhcp6: false
      networkName: sddc-cgw-network-3
  numCPUs: 2
  providerID: vsphere://4230467f-656c-ebb4-06ef-95852af30a7e
  template: ubuntu-1804-kube-v1.16.2
status:
  addresses:
  - address: 192.168.3.192
    type: InternalIP
  networkStatus:
  - connected: true
    ipAddrs:
    - 192.168.3.192
    macAddr: 00:50:56:b0:7f:60
  ready: true
  taskRef: task-266925
```
Looks pretty normal. 

Then I delete the machine object:
```
$ kubectl delete machine target-cluster01-md-0-7d99bd4955-snc8h
```

andrewsykim:
I also see that taskRef never get unset even though [hasInFlightTask](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/pkg/cloud/vsphere/services/govmomi/util.go) eventually unsets it. Wondering if that's related

akutz:
Possibly. It’s why it used to be all part of the same function. Still, the whole thing is supposed to be treated as a reentrant. Some ops only occur after functions are reentered.

akutz:
What do you mean it isn’t unset? The function you referenced does unset it, and it’s part of the destroy call.

akutz:
I encourage you to review the reconcile diagram I posted. The model is based on a reconcile loop. It doesn’t expect the final state to be achieved on the first run through. This is how it’s worked since @sidharthsurana and I first worked out the loop during the first refactor.

akutz:
As you said, unsetting the MachineRef happens and then the next time through the loop the finalized is removed.

andrewsykim:
> What do you mean it isn’t unset? 

Referring to **tasRef**, see my comment https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/643#issuecomment-546178341
```
$ kubectl get vspheremachine target-cluster01-md-0-hsjmv -o yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
kind: VSphereMachine
metadata:
  creationTimestamp: "2019-10-25T02:36:54Z"
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2019-10-25T02:42:36Z"
  finalizers:
  - vspheremachine.infrastructure.cluster.x-k8s.io
  generateName: target-cluster01-md-0-
  generation: 5
  name: target-cluster01-md-0-hsjmv
  namespace: default
  ownerReferences:
  - apiVersion: cluster.x-k8s.io/v1alpha2
    kind: Machine
    name: target-cluster01-md-0-7d99bd4955-snc8h
    uid: 59aa2223-19e7-457c-b47e-6ff14b082cc3
  resourceVersion: "356239"
  selfLink: /apis/infrastructure.cluster.x-k8s.io/v1alpha2/namespaces/default/vspheremachines/target-cluster01-md-0-hsjmv
  uid: bd3bb10a-7202-4a2e-bc43-c84f5a78fb1f
spec:
  datacenter: SDDC-Datacenter
  diskGiB: 50
  memoryMiB: 2048
  network:
    devices:
    - dhcp4: true
      dhcp6: false
      networkName: sddc-cgw-network-3
  numCPUs: 2
  providerID: vsphere://4230467f-656c-ebb4-06ef-95852af30a7e
  template: ubuntu-1804-kube-v1.16.2
status:
  addresses:
  - address: 192.168.3.192
    type: InternalIP
  networkStatus:
  - connected: true
    ipAddrs:
    - 192.168.3.192
    macAddr: 00:50:56:b0:7f:60
  ready: true
  taskRef: task-266925  // TASK REF NOT SET
```

andrewsykim:
> As you said, unsetting the MachineRef happens and then the next time through the loop the finalized is removed.

Yup, just pointing out the odd behavior that we are returning nil error in `DestroyVM` once but that isn't removing the machine finalizers.

akutz:
Isn’t that because the VM isn’t marked as not found until the call is entered and the moref and task are both unset? https://github.com/yastij/cluster-api-provider-vsphere/blob/4e2cead625a8fa0e146a04d7ce8a03583d96d65c/pkg/cloud/vsphere/services/govmomi/service.go#L129

There’s no reason to think a non nil error means the op is successful or complete. You have the (book, error) pattern a lot. In this case the error means the state couldn't be ascertained. It’s the VM object that dictates when finalizers are removed.

akutz:
Plus I think it’s probably best not to infer what the reconcile loop should do based on an internal service. It’s probably best to look at the reconcile loop on the controller. https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/98a597b74d457e9b973827217373dc1c8cea0377/controllers/vspheremachine_controller.go#L173

> 	// Requeue the operation until the VM is "notfound".

akutz:
FWIW, I intend to get rid of the requeues entirely and implement a watch on the vCenter task manager using the external resource trigger. That way reconciles are triggered when task events occur for known VMs.

andrewsykim:
> Isn’t that because the VM isn’t marked as not found until the call is entered and the moref and task are both unset?

But we also mark the VM as not found when we initially unset the machineRef (which we know happens) https://github.com/yastij/cluster-api-provider-vsphere/blob/4e2cead625a8fa0e146a04d7ce8a03583d96d65c/pkg/cloud/vsphere/services/govmomi/service.go#L146-L147

> There’s no reason to think a non nil error means the op is successful or complete. You have the (book, error) pattern a lot. In this case the error means the state couldn't be ascertained. It’s the VM object that dictates when finalizers are removed.

Hmm.. maybe I'm missing context -- in this case it seems like machine deletion is dependant on the error by DestroyVM since that determines if we delete the finalizer of `VSphereMachine.

andrewsykim:
> FWIW, I intend to get rid of the requeues entirely and implement a watch on the vCenter task manager using the external resource trigger. That way reconciles are triggered when task events occur for known VMs.

This sounds great :)

akutz:
I think we just disagree what the purpose of an error is. There’s no reason to think the absence of an error means an action was completed successfully. It just means the action was performed successfully. Whether the result is complete or what you want, we don’t know yet. Remember, interactions with vSphere aren’t synchronous. Thus we’re just acting and reacting each time through the reconcile loop based on the current state as we’re able to determine it.

akutz:
The error from destroyVM is not the deciding factor whether we delete the finalizer. The state of the VM object determines that. If there is a nil error and VM.state == notfound, we remove finalizer.

andrewsykim:
> The error from destroyVM is not the deciding factor whether we delete the finalizer. The state of the VM object determines that. If there is a nil error and VM.state == notfound, we remove finalizer.

Right, and my point of confusion is that we do ste `vm.State = infrav1.VirtualMachineStateNotFound` in the only place we unset MachineRef (which we know happens), so why is the VM not deleting at that point?

akutz:
Regarding “but we also..”

That’s the artifact of the refactor away from a single function. I encourage you to go look at the old CRUD model with “lookupVM” at top of each call. It was more obvious what was occurring. I should probably put that back.

akutz:
Aw crap, is this the new code from Yassine? You’re saying that the link you provided still isn’t causing the finalizer to be removed upon return?

akutz:
Re https://github.com/yastij/cluster-api-provider-vsphere/blob/4e2cead625a8fa0e146a04d7ce8a03583d96d65c/pkg/cloud/vsphere/services/govmomi/service.go#L146-L147

I agree. Upon this return the finalizer should be removed.

yastij:
@akutz @andrewsykim - maybe I'm missing something but the finalizer is removed anyway in the machine_controller no ?: 

- I confirmed with @dougm that all methods of a `SearchIndex` returns `nil,nil` when an object is not found, so this should go through here: https://github.com/yastij/cluster-api-provider-vsphere/blob/4e2cead625a8fa0e146a04d7ce8a03583d96d65c/pkg/cloud/vsphere/services/govmomi/service.go#L148

- then in the machine_controller we should go through this https://github.com/yastij/cluster-api-provider-vsphere/blob/4e2cead625a8fa0e146a04d7ce8a03583d96d65c/controllers/vspheremachine_controller.go#L180 since the vm state is not found

andrewsykim:
> maybe I'm missing something but the finalizer is removed anyway in the machine_controller 

> then in the machine_controller we should go through this https://github.com/yastij/cluster-api-provider-vsphere/blob/4e2cead625a8fa0e146a04d7ce8a03583d96d65c/controllers/vspheremachine_controller.go#L180 since the vm state is not found

That's exactly it @yastij -- before this PR we are already removing the machine ref, which means we also [set the vm state to not found](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/pkg/cloud/vsphere/services/govmomi/service.go#L146-L147) but the finalizer is not removed until a 2nd (or possibly 3rd) pass at DestroyVM. See my test case in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/643#issuecomment-546178341. This PR seems to fix the problem because we allow more attempts to findVMByInstanceUUID but it sounds like there's a Patch error or something happening else that is the root cause.

PR #646, opened 2019-10-25
only document latest Kubernetes patch versions

andrewsykim:

Given the recent Kubernetes CVE, I don't think there's much value in documenting OVAs for older Kubernetes versions 

Fixes #

**Release note**:

```release-note

```


codenrhoden:
I think documenting just the latest images makes sense.

I'm not sure if I'd tie it to CVE's though. Meaning that even with just the two latest in the list now, the "older" version has the CVE, and newer one does not.

andrewsykim:
Fine with me! Will update shortly

PR #650, opened 2019-10-29
Update CAPBK to v0.1.5

andrewsykim:

Update CAPBK to v0.1.5 

Fixes #

**Release note**:

```release-note
Update CAPBK to v0.1.5
```


akutz:
Please update the PR with a release note, and then feel free to remove the hold.

andrewsykim:
fwiw, we don't really parse the release notes there, we really should for generating CHANGELOG :)

akutz:
> fwiw, we don't really parse the release notes there, we really should for generating CHANGELOG :)

We don't **yet**, but as you said, we should, and it's something I've wanted to do for a while. Given the desire, and the small time it takes to add them as we create PRs, it makes sense to have a populated backlog so we can verify the generation when we get around to it.

andrewsykim:
>We don't yet, but as you said, we should, and it's something I've wanted to do for a while. Given the desire, and the small time it takes to add them as we create PRs, it makes sense to have a populated backlog so we can verify the generation when we get around to it.

+ 1000

PR #658, opened 2019-11-05
Bump default golang version to 1.12.10 to address CVE-2019-16276

figo:
Bump default golang version to 1.12.10 to address CVE-2019-16276

Set golang_version argument, user can always build with specific
golang version by passing --build-arg golang_version:golang:1.x.x

N/A
N/A

**Release note**:

```release-note

```

cc @akutz @andrewsykim @jiatongw @frapposelli

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 # Build the manager binary
-FROM golang:1.12.9 as builder
+ARG GOLANG_VERSION=golang:1.12.10
+FROM $GOLANG_VERSION as builder
 WORKDIR /workspace
 
 # Run this with docker build --build_arg $(go env GOPROXY) to override the goproxy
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 # Build the manager binary
-FROM golang:1.12.9 as builder
+ARG golang_version=golang:1.12.10
```
andrewsykim: Just out of curiosity, why make this an ARG? Do we want to encourage users trying different golang versions for CAPV?

akutz: It's more for us and our testing. If you end up keeping it, please make the arg ALL CAPS. Please see [this Dockerfile](https://github.com/kubernetes/cloud-provider-vsphere/blob/master/hack/images/ci/Dockerfile) for an example.

figo: @andrewsykim it is for downstream release process to define golang version without need to change upstream.

figo: @akutz done, thx

PR #665, opened 2019-11-19
Simplify package layout

akutz:
This patch simplifies the package layout by moving all of the packages located in `pkg/cloud/vsphere` directly into `pkg`. The previous layout is a vestigial artifact of when the providers all lived in a single repository, and that is no longer the case.

/assign @andrewsykim 

**Release note**:

```release-note
NONE
```


akutz:
Hi @vincepri / @ncdc,

I'd also appreciate your feedback on this PR. Thank you!

ncdc:
For comparison, CAPA has pkg/cloud/..., CAPG has cloud (no pkg). I think it's largely up to you / we don't have consistency across the CAP* projects at the moment. This seems like a reasonable change.

sidharthsurana:
looks like a reasonable change to me too
/lgtm

PR #668, opened 2019-11-20
Replace managed object reference with provider ID

### LINKED ISSUE #661: Refactor CAPV's machine identity logic
/kind feature

Currently CAPV's machine identity logic looks like this:

1. Create VM with an `InstanceUUID` based on the CAPI `Machine` resource's `UID` value
1. Wait until the `CloneVM_Task` is complete
1. Query the VM based on its `InstanceUUID` using the CAPI `Machine` resource's `UID` value
1. Store the VM's managed object reference (`MoRef`) in the CAPV `VSphereMachine` resource's `VSphereMachine.Spec.MachineRef` property
1. Subsequent reconcile requests use the `VSphereMachine.Spec.MachineRef`, the VM's `MoRef`, to reconcile the VM

At this point, future queries are based on the value of `VSphereMachine.Spec.MachineRef`, a `MoRef`, and managed object references are:

* Only unique per vCenter instance
* May be re-used

For these reasons, a VM's `MoRef` should not be stored as part of the `VSphereMachine` spec, but instead be resolved at the beginning of each reconcile request or even optionally stored as part of a `VSphereMachine`'s status.

Additionally, the `VSphereMachine.Spec.MachineRef` field should be removed as there is already a field in CAPV v1a2 that contains information that may be used to query the VM --  `VSphereMachine.Spec.ProviderID`. This value is assigned the value of the VM's hex-encoded BIOS UUID with the prefix `vsphere://`. Thus the BIOS UUID can be obtained from `VSphereMachine.Spec.ProviderID` and used to query the VM's `MoRef` at the beginning of each reconcile request.

In other words, the new CAPV identity logic will be:

1. Create VM with an `InstanceUUID` based on the CAPI `Machine` resource's `UID` value
1. Wait until the `CloneVM_Task` is complete
1. Query the VM based on its `InstanceUUID` using the CAPI `Machine` resource's `UID` value
1. Get the VM's BIOS UUID and store it as `vsphere://BIOS_UUID` in `VSphereMachine.Spec.ProviderID`
1. Subsequent reconcile requests use the `VSphereMachine.Spec.ProviderID` value to query the VM based on its BIOS UUID and get the VM's `MoRef` to reconcile the VM

This will also fix an issue with the CAPI v1alpha2 upgrade tool not working with CAPV due to the fact that the `VSphereMachine.Spec.MachineRef` field is still present when the upgrade tool copies the resources. @andrewsykim can provide more information on this. Because the upgrade tool knows to remove the infrastructure machine resource's `ProviderID`, the copied resource will result in the creation of a new VM.

cc @sidharthsurana

akutz:
This patch removes the storage of the VM's managed object reference as part of a VSphereMachine resource's spec. Instead, the provider ID is used at the beginning of a reconcile request to resolve the VM's managed object reference. The new machine identity workflow is described in [a sequence diagram](https://raw.githubusercontent.com/akutz/cluster-api-provider-vsphere/feature/rm-moref/docs/design/machine-controller-reconcile.puml):

machine-controller-reconcile

Fixes #661

This PR should unblock support for the [Cluster API upgrade tool](https://github.com/vmware/cluster-api-upgrade-tool) with CAPV. @frapposelli had asked @andrewsykim to look into the upgrade tool and CAPV, and Andrew discovered an issue that prevented support for the tool. This PR should resolve that. For more information, please see #661.

**Release note**:

```release-note
* ManagedObjectReference values are no longer stored as part of VSphereMachine resource specs.
```


pkg/util/machines.go
```diff
@@ -179,3 +170,43 @@ func GetMachineMetadata(hostname string, machine infrav1.VSphereMachine, network
...
+// A valid UUID should adhere to the format specified by UUIDPattern.
+func ConvertUUIDToProviderID(uuid string) string {
+	if uuid == "" {
+		return ""
+	}
+	pattern := regexp.MustCompile(UUIDPattern)
+	if !pattern.MatchString(uuid) {
+		return ""
+	}
+	return "vsphere://" + uuid
```
figo: declare "vsphere://" as const?

akutz: Hi @figo,

Perhaps, but this is the only place it is used outside of the regex. I suppose declaring it a const and then building the regex with it isn't a bad idea. Thanks!

pkg/context/session.go
```diff
@@ -95,26 +95,20 @@ func getOrCreateCachedSession(ctx *MachineContext) (*Session, error) {
...
+	return s.findByUUID(ctx, uuid, true)
 }
 
-// FindByUUID finds an object by its UUID.
+// FindByUUID finds an object by its BIOS UUID.
 func (s *Session) FindByUUID(ctx context.Context, uuid string) (object.Reference, error) {
+	return s.findByUUID(ctx, uuid, false)
+}
+
+func (s *Session) findByUUID(ctx context.Context, uuid string, findByInstanceUUID bool) (object.Reference, error) {
```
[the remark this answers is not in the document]
akutz: Hi @frapposelli,

Fair. I'll update it.

pkg/services/govmomi/util.go
```diff
@@ -66,67 +85,40 @@ func getTask(ctx *context.MachineContext) *mo.Task {
...
 
-func getVMObject(ctx *context.MachineContext) (mo.VirtualMachine, error) {
-	moRef := types.ManagedObjectReference{
-		Type:  "VirtualMachine",
-		Value: ctx.VSphereMachine.Spec.MachineRef,
-	}
-	var obj mo.VirtualMachine
-	err := ctx.Session.RetrieveOne(ctx, moRef, []string{"name"}, &obj)
-	return obj, err
+func isStringNilOrEmpty(s *string) bool {
```
figo: i feel it is slightly easier to read the code if not creating this function, but i am good with both ways

akutz: Interesting. I wrote it without the helper function the first time, but I thought the function name made it easier to understand. I'll think it over some more. Thanks Hui!

akutz: Hi @figo,

Good call. I forgot that this function is no longer called in multiple locations. You're right, now it's just overkill.

pkg/util/machines.go
akutz: Done and done. Thanks!

akutz:
Hi @figo and @frapposelli,

I've addressed all of your feedback. Thank you both!

akutz:
FWIW, I'm rebasing this PR to squash the commits that were just to address feedback.

Issue #692, opened 2019-12-16
machine reconcilation never completes


ntfrnzn:
/kind bug

In creating a set of machines, the first VM is created, but not others.  On looking at the logs
I see repeated logging from the deferred patch function, "resource is not patched", seemingly
waiting forever for the patch to be applied.

A little debugging shows that the two status objects are different, one with an ip-address slice of nil, the other of []string{}.

**Anything else you would like to add:**
See https://github.com/NetApp/cluster-api-provider-vsphere/commit/c275e5ecbdc920de227098115b1c2484b78501b3 for a proposed fix, changing the comparison
to equate empty and nil.

akutz:
Hi @ntfrnzn,

Thank you for finding this. Also, dammit, why did you find this? :) 

Your fix looks fine (thank you for that), but for one thing -- the diff function:

1. `Diff` will return a non-empty string **only** if the `Equal` call returns false. Thus, don't do both the `Equal` **and** `Diff` calls. Just do the `Diff` and save the output and let `diff == ""` determine if the objects are equal.
2. Don't print the diff in the logger. I specifically did **not** do that because the logger munges the diff and makes it useless to read. If you want an example of that work before I removed the diffs and how they were printed, please see https://github.com/akutz/cluster-api-provider-vsphere/blob/feature/task-source-logging/controllers/vspheremachine_controller.go#L190-L270.

ntfrnzn:
Leaving the diff out is fine by me. It's not very readable. For my part, I copied the munged text out of the log and ran it through printf, 
```
  v1alpha2.VSphereMachineStatus{
  	Ready:     true,
  	Addresses: []v1.NodeAddress{{Type: "InternalIP", Address: "10.68.192.84"}},
  	TaskRef:   "",
  	Network: []v1alpha2.NetworkStatus{
  		{Connected: true, IPAddrs: []string{"10.68.192.84"}, MACAddr: "00:50:56:ae:49:31", NetworkName: "VM_Network"},
  		{
  			Connected:   true,
- 			IPAddrs:     []string{},
+ 			IPAddrs:     nil,
  			MACAddr:     "00:50:56:ae:b6:51",
  			NetworkName: "iSCSI-B",
  		},
  	},
  	ErrorReason:  nil,
  	ErrorMessage: nil,
  }
```
So the problem here is that there's a second storage NIC which, for whatever reason, is not getting an IP. Probably an unusual scenario.

[...]

PR #693, opened 2019-12-16
use equate-empty in comparing machine status

ntfrnzn:
fixes https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/692


akutz:
Merging manually since e2e pre-submits are broken until a later date yet to be determined.

PR #698, opened 2019-12-18
Support a control plane endpoint annotation for v1a2

### LINKED ISSUE #697: Add support for an annotation that specifies the control plane endpoint in the VSphereCluster resource in v1a2
/kind feature

After speaking with @rsmitty on Slack, he needs to support a non-kubeadm bootstrap provider in CAPV. In v1a3 this is possible because the control plane endpoint is part of the cluster spec. In v1a2 CAPV explicitly searches for a `KubeadmBootstrapConfig` resource for the information. This does not help if not using the Kubeadm bootstrap controller. To support the [Talos bootstrap provider](https://github.com/talos-systems/cluster-api-bootstrap-provider-talos), CAPV v1a2 should support the annotation `capv.infrastructure.cluster.x-k8s.io/control-plane-endpoint` on a `VSphereCluster` resource.

The plan is for this patch to be the last change that happens on `release-0.5` (#696) before releasing `0.5.4`.

akutz:
/kind feature

This patch supports a control plane endpoint annotation for v1alpha2.

cc @rsmitty

Fixes #697

**Release note**:

```release-note

```

### DIFF
--- a/api/v1alpha2/types.go
+++ b/api/v1alpha2/types.go
@@ -26,6 +26,11 @@ const (
 	// infrastructure sources are ready and machines may be created.
 	AnnotationClusterInfrastructureReady = "vsphere.infrastructure.cluster.x-k8s.io/infrastructure-ready"
 
+	// AnnotationControlPlaneEndpoint is used only in v1alpha2 by other
+	// bootstrap providers to specify the control plane endpoint prior to
+	// the creation of the cluster.
+	AnnotationControlPlaneEndpoint = "capv.infrastructure.cluster.x-k8s.io/control-plane-endpoint"
+
 	// AnnotationControlPlaneReady indicates the cluster's control plane is
 	// ready.
 	AnnotationControlPlaneReady = "vsphere.infrastructure.cluster.x-k8s.io/control-plane-ready"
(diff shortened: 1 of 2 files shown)


akutz:
Hey @rsmitty, this at least doesn't break the current e2e. Can you please build an image and see if it works the way we discussed for your use case?

akutz:
Hi @yastij,

I'd like a review and lgtm, even though the test technically fails. Please see this [Slack thread](https://kubernetes.slack.com/archives/C8TSNPY4T/p1576636575069800) for more information. The fact is, CAPV **is** working. It's something to do with k/k or the cloud-provider as to why the second control plane node isn't being assigned a provider ID in the `Node` resource. As far as the k/k cluster is concerned, CAPV **is** standing up a multi-control plane node cluster. The linked thread includes a gist with logs that show:

```shell
$ kubectl --kubeconfig /etc/kubernetes/admin.conf get nodes
NAME                   STATUS   ROLES    AGE     VERSION
sakr5-controlplane-0   Ready    master   11m     v1.16.2
sakr5-controlplane-1   Ready    master   8m43s   v1.16.2
```

akutz:
Hi @yastij,

> we'd need to account for this if we have conversion webhooks between v1alpha2 and v1alpha3

Perhaps. It would be easy enough to do, but I also added this only to support @rsmitty's use case over at Talos. They don't use the Kubeadm bootstrapper, and so the previous solution didn't work for them.

PR #705, opened 2019-12-21
Support for haproxy as a control-plane load balancer

### LINKED ISSUE #216: Feature: Support creating linked clone from snapshots
Today we are doing a full clone of the template when creating the nodes.
We should support doing a linked clone as well. That would tremendously help speed up the VM
creation. Linked clone should be presented as an option for the user creating the `Machine` if they want to use it.

### LINKED ISSUE #245: Feature: Add support for load balancer for HA
In environments that have load balancers, we need to add some programmatic support to set that up before we create HA clusters.  For environments that do not have load balancers, we need to set up separate load balancers (e.g. running in a separate VMs).

akutz:
/kind feature

This PR introduces support for HAProxy as the de facto load-balancer solution for CAPV. Patches include:

- A new CRD, `VSphereVM`, and controller that may be used to deploy VMs to vSphere without any knowledge or relationship to Kubernetes. The `VSphereMachine` controller has been refactored to leverage this new CRD. The new CRD will be used to deploy the HAProxy load balancer OVA.
- A directory `hack/tools/haproxy` that contains an example of the HAProxy load-balancer using a PhotonOS 3 based container image as well as a framework for building a PhotonOS 3 OVA with HAProxy and its dataplane API.
- A new CRD, `HAProxyLoadBalancer` that may be assigned to `VSphereCluster.Spec.LoadBalancerRef` in order to use an existing or provision a new HAProxy load balancer that can be remotely configured via its [dataplane API server](https://www.haproxy.com/documentation/hapee/1-9r1/configuration/dataplaneapi/).
- Supports linked mode clones to dramatically speed up the deployment process of a VM

Many thanks to @moshloop for helping to make this possible with his work on the machine load balancer.

Fixes #216, Fixes #245

All, 

* The e2e pre-submit will fail until the Prow job is updated to use the new e2e framework. Please run `make e2e` ([docs](https://github.com/akutz/cluster-api-provider-vsphere/tree/feature/haproxy/test/e2e)) locally to validate. Here's [output](https://gist.github.com/akutz/f89f8417e05e447489b1f8c0641af433) from a successful e2e run
* It’s easy to build the HAProxy OVA from `hack/tools/haproxy`, but I also uploaded an [HAProxy load balancer OVA](http://storage.googleapis.com/capv-images/extra/haproxy/release/v0.5.3-77-g224e0ef6/capv-haproxy-v0.5.3-77-g224e0ef6.ova). You can read more about downloading the HAProxy OVA [here](https://github.com/akutz/cluster-api-provider-vsphere/tree/feature/haproxy/hack/tools/haproxy#downloading-the-ova).

cc @yastij 

cc @sidharthsurana @figo This PR adds support for linked mode clones per #216. While not called out in any documentation yet, linked mode is now the default clone mode as long as the source template has at least a single snapshot. To create a snapshot on a template, you need to:
1. Convert the template to a VM
2. Create a snapshot of the template
3. Convert the VM back to a template

**Release note**:

```release-note
* Support for using HAProxy as a control-plane load balancer
```


akutz:
This PR's verify-markdown pre-submit will fail until https://github.com/kubernetes/test-infra/pull/15718 is merged.

cc @yastij @figo @codenrhoden

api/v1alpha3/vspherevm_types.go
```diff
@@ -0,0 +1,112 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
tedyu: Times goes by fast.
If you have time, you can replace the year in all new files.

api/v1alpha3/zz_generated.deepcopy.go
```diff
@@ -448,6 +569,113 @@ func (in *VSphereMachineTemplateSpec) DeepCopy() *VSphereMachineTemplateSpec {
...
+		return nil
+	}
+	out := new(VSphereVM)
+	in.DeepCopyInto(out)
+	return out
+}
+
+// DeepCopyObject is an autogenerated deepcopy function, copying the receiver, creating a new runtime.Object.
+func (in *VSphereVM) DeepCopyObject() runtime.Object {
+	if c := in.DeepCopy(); c != nil {
```
tedyu: Looking at line 596, it seems the nil check is not needed

edit: though this is generated code

controllers/vspherevm_controller.go
```diff
@@ -0,0 +1,297 @@
...
+
+				// There was an issue getting the remote resource. Sleep for a
+				// second and try again.
+				vmContext.Logger.Error(err, "failed to get VSphereVM while exiting reconcile")
+				return false, nil
+			}
+
+			// If the remote resource version is not the same as the local
+			// resource version, then it means we were able to get a resource
+			// newer than the one we already had.
```
tedyu: To match the comment, shouldn't the condition below be changed to:

localObj.ResourceVersion < remoteObj.ResourceVersion

controllers/vspherevm_controller.go
```diff
@@ -0,0 +1,297 @@
...
+				"resource is not patched",
+				"local-resource-version", localObj.ResourceVersion,
+				"remote-resource-version", remoteObj.ResourceVersion)
+			return false, nil
+		})
+	}()
+
+	// Handle deleted machines
+	if !vsphereVM.ObjectMeta.DeletionTimestamp.IsZero() {
+		return r.reconcileDelete(vmContext)
```
tedyu: nit: reconcileDelete -> reconcileDeletion

api/v1alpha3/vspherevm_types.go
akutz: Hi @tedyu,

Technically it was 2019 when that file was created. I think it's probably not necessary to replace dates on files in PRs that span the new year.

api/v1alpha3/zz_generated.deepcopy.go
akutz: Hi @tedyu,

Yep, generated code. Resolving this comment. Thanks though!

controllers/vspherevm_controller.go
akutz: Hi @tedyu,

One would think, but see the above comment:

> ```golang
> 		// FYI - resource versions are opaque, numeric strings and should not
>  		// be compared with < or >, only for equality -
>  		// https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions.
>  		//
>  		// Since CAPV is currently deployed with a single replica, and this
>  		// controller has a max concurrency of one, the only agent updating the
>  		// VSphereVM resource should be this controller.
>  		//
>  		// So if the remote resource's ResourceVersion is different than the
>  		// ResourceVersion of the resource fetched at the beginning of this
>  		// reconcile request, then that means the remote resource should be
>  		// newer than the local resource.
> ```

controllers/vspherevm_controller.go
akutz: Hi @tedyu,

I think that's right? Regardless, this pattern is used across all CAPI providers, so I'll likely leave this alone since it's the same in CAPI, CAPA, CAPG, CAPD, etc.

tedyu: Alright.
I was not aware of the other places where this name is used.

controllers/vspherecluster_controller.go
```diff
@@ -290,56 +455,118 @@ func (r clusterReconciler) reconcileControlPlaneEndpoint(ctx *context.ClusterCon
...
+			// plane endpoint information is logged, it is necessary to fetch
+			// an up-to-date Cluster resource. If this fails, then set the
+			// control plane endpoint information to the values from the
+			// VSphereCluster resource, as it must have the correct information
+			// if the API server is online.
+			cluster := &clusterv1.Cluster{}
+			clusterKey := client.ObjectKey{Namespace: ctx.Cluster.Namespace, Name: ctx.Cluster.Name}
+			if err := ctx.Client.Get(ctx, clusterKey, cluster); err != nil {
+				cluster = ctx.Cluster.DeepCopy()
+				cluster.Spec.ControlPlaneEndpoint.Host = ctx.VSphereCluster.Spec.ControlPlaneEndpoint.Host
```
yastij: why are we setting these ? as cluster is local, maybe ctx.Cluster ?

akutz: Hi @yastij,

Strictly for the logging statement below.

controllers/vspherecluster_controller.go
```diff
@@ -290,56 +455,118 @@ func (r clusterReconciler) reconcileControlPlaneEndpoint(ctx *context.ClusterCon
...
+		return
+	}
+	apiServerTriggers[ctx.Cluster.UID] = struct{}{}
+	go func() {
+		// Block until the target API server is online.
+		ctx.Logger.Info("start polling API server for online check")
+		wait.PollImmediateInfinite(time.Second*1, func() (bool, error) { return r.isAPIServerOnline(ctx), nil })
+		ctx.Logger.Info("stop polling API server for online check")
+		ctx.Logger.Info("triggering GenericEvent", "reason", "api-server-online")
+		eventChannel := ctx.GetGenericEventChannelFor(ctx.VSphereCluster.GetObjectKind().GroupVersionKind())
```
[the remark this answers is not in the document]
akutz: Hi @yastij,

Because it's still required for deployments without an LB. This:

* Causes the `VSphereCluster` to set its `ControlPlaneEndpoint` from a `VSphereMachine`
* Which then causes the CAPI `Cluster` to reconcile since it watches its infrastructure cluster.

contrib/haproxy/Makefile
```diff
@@ -0,0 +1,187 @@
...
+	$(MAKE) --no-print-directory fix-backend-model
+	$(MAKE) --no-print-directory fix-frontend-model
+
+# Please see https://github.com/OpenAPITools/openapi-generator/issues/741
+# for more information on the issue of duplicate Opt structures.
+.PHONY: remove-dupe-opts
+remove-dupe-opts: | $(GOLANGCI_LINT)
+remove-dupe-opts: ## Removes duplicate Opt structs from the generated code
+	@for struct in $$(grep -h 'type .\{1,\} struct' $(OUTPUT_DIR)/*.go | grep Opts  | sort | uniq -c | grep -v '^      1' | awk '{print $$3}'); do \
+	  for f in $$(/bin/ls $(OUTPUT_DIR)/*.go); do \
```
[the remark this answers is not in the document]
akutz: Hi @yastij,

We *could*, but since this is in its own `Makefile` I don't think it's strictly necessary for now. However, I agree that we could iterate on this in the future and create `contrib/haproxy/hack` for some of these functions.

akutz:
This PR has all the required approvals and lgtms, and so I'm merging this manually since the e2e pre-submit is failing until we fix that on Prow.

PR #724, opened 2020-01-15
Rename clusterctl-settings file extension

vuil:
XRef https://github.com/kubernetes-sigs/cluster-api/pull/2027
a file with name clusterctl-settings.json is what is expected

This PR adds the clusterctl-settings.json so the vSphere provider can be tested with clusterctl even if now there are no published releases for v1alpha3

```release-note
None
```


akutz:
Huh. Fabrizio opened the original PR. Seems odd he’d use the wrong extension since he’s the author of the tool.

PR #727, opened 2020-01-20
Update CAPV to use CAPI e2e latest

akutz:
/kind feature

This patch:

* Updates the dependency on Kind to v0.7.0
* Updates CAPV to use the latest CAPI e2e bits
* Introduces a configuration resource for the management cluster helper for configuring the e2e test environment via a configuration file:

**Example `e2e.conf`**

```yaml
---
images:
- capv-manager:e2e

components:

# Load the certificate manager and wait for all of its pods and service to
# become available.
- name:    cert-manager
  sources:
  - type:  url
    value: https://github.com/jetstack/cert-manager/releases/download/v0.11.1/cert-manager.yaml
  waiters:
  - type:  service
    value: v1beta1.webhook.cert-manager.io
  - value: cert-manager

# Load CAPI core and wait for its pods to become available.
- name:    capi
  sources:
  - value: https://github.com/kubernetes-sigs/cluster-api//config/default?ref=master
    replacements:
    - old: "imagePullPolicy: Always"
      new: "imagePullPolicy: IfNotPresent"
  waiters:
  - value: capi-system

# Load the CAPI kubeadm bootstrapper and wait for its pods to become available.
- name:    capi-kubeadm-bootstrap
  sources:
  - value: https://github.com/kubernetes-sigs/cluster-api//bootstrap/kubeadm/config/default?ref=master
    replacements:
    - old: "imagePullPolicy: Always"
      new: "imagePullPolicy: IfNotPresent"
  waiters:
  - value: capi-kubeadm-bootstrap-system

# Load the CAPI kubeadm control plane and wait for its pods to become available.
- name:    capi-kubeadm-control-plane
  sources:
  - value: https://github.com/kubernetes-sigs/cluster-api//controlplane/kubeadm/config/default?ref=master
    replacements:
    - old: "imagePullPolicy: Always"
      new: "imagePullPolicy: IfNotPresent"
  waiters:
  - value: capi-kubeadm-control-plane-system

# Load CAPV and wait for its pods to become available.
- name:    capv
  sources:
  - value: ../../config/default
    replacements:
    - old: gcr.io/cluster-api-provider-vsphere/release/manager:latest
      new: capv-manager:e2e
    - old: "imagePullPolicy: Always"
      new: "imagePullPolicy: IfNotPresent"
  waiters:
  - value: capv-system
```

For more information, please see [`test/e2e/framework/config.go`](https://github.com/akutz/cluster-api-provider-vsphere/blob/feature/switch-back-to-latest-capi-for-e2e/test/e2e/framework/config.go) and its test, [`test/e2e/framework/config_test.go`](https://github.com/akutz/cluster-api-provider-vsphere/blob/feature/switch-back-to-latest-capi-for-e2e/test/e2e/framework/config_test.go).

Fixes #

cc @chuckha The `e2e.conf` and its sources is the next bit I plan to upstream. I've done away with flags and env vars, and rely solely on the config file now.

**Release note**:

```release-note

```


akutz:
/assign @yastij 

This should be ready for review. Please note that this isn't a super-simple PR. This is due to CAPI splitting the core controllers into three, distinct controller managers. I opted to allow for customization that will handle future, possible changes as well.

test/e2e/framework/config.go
```diff
@@ -0,0 +1,299 @@
...
+	// condition of Ready.
+	PodsWaiter WaiterType = "pods"
+)
+
+// ComponentWaiter contains information to help determine whether installed
+// components are ready.
+type ComponentWaiter struct {
+	// Value varies depending on the specified check Type, ex. when Type is
+	// "service" then Value is the name of a Service resource.
+	Value string `json:"value"`
```
yastij: I see what is the goal for this. But the values of the « Value » field doesn’t make sense unless we document for each type what is the expected value (for pods and service types)

test/e2e/framework/config_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+# Load CAPV and wait for its pods to become available.
+- name:    capv
+  sources:
+  - value: ../../config/default
+    replacements:
+    - old: gcr.io/cluster-api-provider-vsphere/release/manager:latest
+      new: capv-manager:e2e
+  waiters:
+  - value: capv-system
```
yastij: Does this mean we are waiting for the whole pods of the namespace to be ready ? Is value accepting pods or namespace values ?

test/e2e/framework/control_plane.go
```diff
@@ -84,40 +60,53 @@ func SingleNodeControlPlane(input *SingleNodeControlPlaneInput) {
...
+
+		By(fmt.Sprintf("creating control plane resource %d: BootstrapConfig", i+1))
+		Expect(mgmtClient.Create(ctx, node.BootstrapConfig)).To(Succeed())
+
+		By(fmt.Sprintf("creating control plane resource %d: Machine", i+1))
+		Expect(mgmtClient.Create(ctx, node.Machine)).To(Succeed())
+
+		// If this is the first node then block until the control plane is
+		// initialized.
+		if i == 0 {
```
yastij: I’d store `cluster.Status.ControlPlaneInitialized` and test against it here

hack/e2e.sh
```diff
@@ -19,7 +19,7 @@ set -o errexit   # exits immediately on any unexpected error (does not bypass tr
 set -o nounset   # will error if variables are used without first being defined
 set -o pipefail  # any non-zero exit code in a piped command causes the pipeline to fail with that code
 
-KIND_VERSION="v0.5.1"
+KIND_VERSION="v0.7.0"
```
yastij: Was the KIND bump brought in through the CAPI bump ? if not I’d rather see this in a specific commit

akutz: Hi @yastij,

You hit the nail on the head. The updated CAPI e2e kind bits required the newer version of Kind.

test/e2e/framework/config_test.go
akutz: I'll document this.

test/e2e/framework/control_plane.go
akutz: Hi @yastij,

I'm not sure what you mean. That's not how `Eventually` works. What this does is after the first node is deployed, blocks until the control plane is online to deploy additional nodes.

test/e2e/framework/config_test.go
akutz: Hi @yastij,

I don't think it's necessary to document these values every place they're used. I'm documenting them in the `ComponentWaiterTypes` in source. The documentation for the component also indicates that the logic is blocked contingent upon the pods being available.

akutz:
Hi @yastij,

Please take another look. I addressed your feedback.

akutz:
Hi @chuckha,

Thank you for the review! So, a few things:

> This config is a little bit more of a general abstraction than I was hoping for. 

I was trying to cover two primary use cases:

1. Replacing images with custom builds but using existing YAML (see below re: `ComponentReplaceable`)
2. Replacing pluggable components, such as the bootstrapper, with non-standard variants

> I believe the config work done in this PR would close the issue linked above, however the ComponentReplacement requires the author to be aware of the yaml being generated which feels a bit like an abstraction leak to me.

I don't know if I agree? The current `ComponentGenerator` implementations for [CAPI core](https://github.com/kubernetes-sigs/cluster-api/blob/db823b790de0e73614205895dbfbf2b4402bb833/test/framework/generators/capi.go#L67), the [Kubeadm Bootstrapper](https://github.com/kubernetes-sigs/cluster-api/blob/db823b790de0e73614205895dbfbf2b4402bb833/test/framework/generators/kubeadm-bootstrap.go#L67), and the [Kubeadm ControlPlane](https://github.com/kubernetes-sigs/cluster-api/blob/db823b790de0e73614205895dbfbf2b4402bb833/test/framework/generators/kubeadm-control-plane.go#L67) all provide replacements, but they're hardcoded. Making it configurable makes sense to me for a few reasons:

* Replacing an `image` with a custom version in produced YAML
* Changing the `imagePullPolicy` (as per the above examples)
* Keeping things simple and not reimplementing Kustomize

Thank you again for the feedback!

akutz:
Thanks @yastij! I'm cancelling the hold to let this merge after the patch with the updated comments passes e2e.

/hold cancel

PR #738, opened 2020-01-29
bump csi version to the latest release

yastij:

**What this PR does / why we need it**: This PR bumps the CSI version to the latest release. This should avoid breaking setups in VMC, as its SDDCs are being upgraded to vsphere 7.0 which is not compatible with the current CSI version.

**Which issue(s) this PR fixes** Fixes #

/assign @akutz 

**Release note**:

```release-note
bump the vsphere csi driver to the newest release 1.0.2
```


yastij:
ref https://github.com/kubernetes-sigs/vsphere-csi-driver/pull/127

PR #740, opened 2020-01-30
wait for KCP nodes to be provisioned

yastij:

**What this PR does / why we need it**: This PR moves from waiting KCP nodes to be initialized to checking if they're provisioned 

**Release note**:

```release-note

```

### DIFF
--- a/test/e2e/capv_test.go
+++ b/test/e2e/capv_test.go
@@ -43,7 +43,7 @@ var _ = Describe("CAPV", func() {
 			machineDeploymentGen   MachineDeploymentGenerator
 			controlPlaneNodeGen    ControlPlaneNodeGenerator
 			kubeadmControlPlaneGen *KubeadmControlPlaneGenerator
-			pollTimeout            = 10 * time.Minute
+			pollTimeout            = 15 * time.Minute
 			pollInterval           = 10 * time.Second
 
 			numControlPlaneMachines int32
--- a/test/e2e/framework/control_plane.go
+++ b/test/e2e/framework/control_plane.go
@@ -158,13 +158,13 @@ func waitForControlPlaneInitialized(ctx context.Context, input *ControlplaneClus
 		Namespace: input.Cluster.GetNamespace(),
 		Name:      input.Cluster.GetName(),
 	}
-	Eventually(func() (bool, error) {
+	Eventually(func() (string, error) {
 		cluster := &clusterv1.Cluster{}
 		if err := mgmtClient.Get(ctx, clusterKey, cluster); err != nil {
-			return false, err
+			return "", err
 		}
-		return cluster.Status.ControlPlaneInitialized, nil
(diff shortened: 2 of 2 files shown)


akutz:
/lgtm
/approve
/hold

The hold will be removed once e2e passes a second time. This appears to have caused the flake. Essentially the upstream test harness was using `is control plane provisioned` as a check before proceeding, and CAPV was using `is control plane initialized`. This is because a multi-node CP in CAPV used to be multiple control plane nodes, not a KCP. @chuckha also ran into the flake where applying additional CP nodes before the CP was initialized would sometimes cause `kubeadm join` calls to fail on the additional CP nodes. So CAPV switched to `is control plane initialized` whereas Chuck and CAPD moved onto KCP. 

Since CAPV is using KCP now as well, we can switch back to `is control plane provisioned` and no longer hit the timeout caused by waiting for the CP to be initialized.

akutz:
Ugh. Except it failed due to an inability to talk to Kind. @detiber, didn’t you saw you something similar? Like Prow killing kind?

akutz:
So it looks like a second change hit this PR that caused the build to fail, but that was in itself in response to a failed retest to assert the first green wasn't a false positive. Yassine is reverting the second change as it was unnecessary, and the failed retest is because of an error I've never seen before:

```shell
STEP: waiting for the workload nodes to exist
924
E0130 20:24:07.031941   11750 request.go:879] Unexpected error when reading response body: read tcp 10.8.0.7:45678->192.168.5.87:6443: read: connection reset by peer
925
E0130 20:24:07.031950   11750 request.go:879] Unexpected error when reading response body: read tcp 10.8.0.7:45678->192.168.5.87:6443: read: connection reset by peer
926
E0130 20:24:07.032376   11750 request.go:879] Unexpected error when reading response body: read tcp 10.8.0.7:45678->192.168.5.87:6443: read: connection reset by peer
927
E0130 20:24:07.032421   11750 request.go:879] Unexpected error when reading response body: read tcp 10.8.0.7:45678->192.168.5.87:6443: read: connection reset by peer
928
E0130 20:24:07.032439   11750 request.go:879] Unexpected error when reading response body: read tcp 10.8.0.7:45678->192.168.5.87:6443: read: connection reset by peer
929
E0130 20:24:07.032450   11750 request.go:879] Unexpected error when reading response body: read tcp 10.8.0.7:45678->192.168.5.87:6443: read: connection reset by peer
930
E0130 20:24:07.032450   11750 request.go:879] Unexpected error when reading response body: read tcp 10.8.0.7:45678->192.168.5.87:6443: read: connection reset by peer
```

So yeah, it's unclear.

What *is* clear is that we may also need to reevaluate our timeouts on Prow. The e2e framework defaults to 10m, but that's based on CAPD, not real infrastructure. Ten minutes may simply not be enough time for CAPV, CAPA, or other providers. We're continuing to monitor the situation.

Issue #729, opened 2020-01-23
Support for adding created VM to DRS group


rhockenbury:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
The user is able to specify the name of a vsphere DRS group, and the created vms are added to this DRS group.

This would grant a user better control over VM placement since the DRS group could be used to specify VM-host affinity rules.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
Related - https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/175

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

MaxRink:
We would be interested in such a feature as well.
We run multiple vCenters, some of which in georedundant datacenter setup.
So spanning across multiple Clusters within one vCenter would be interesting for us, but also a per-esxi-host approach would allow us to basically make sure that not all masters of a cluster get scheduled on one bladecenter frame for example.

randomvariable:
This has been implemented for some time with VM groups and failure domains.

/close

Feel free to reopen if the current feature set doesn't address your concerns.

PR #752, opened 2020-02-07
Failure domains API to CAPV

yastij:
**What this PR does / why we need it**: This PR adds the expected fields by CAPI for failure domains support.

**Which issue(s) this PR fixes** : part of #729

/assign @ncdc 
/cc @randomvariable 

**Release note**:

```release-note

```


api/v1alpha3/vspherecluster_types.go
```diff
@@ -60,6 +61,9 @@ type VSphereClusterSpec struct {
 // VSphereClusterStatus defines the observed state of VSphereClusterSpec
 type VSphereClusterStatus struct {
 	Ready bool `json:"ready"`
+
+	// FailureDomains is a slice of failure domain objects synced from the infrastructure provider.
```
ncdc: This is the infra provider so this comment needs adjusting

yastij: /hold

let's leave the PR open until the implementation

randomvariable:
don't think we can land this in v1alpha3 as it requires CRD changes to start untying vSphereMachine and vSphereCluster to properly represent real life variation in vsphere.

Suggest we close, pending discussion of requirements and use cases?

yastij:
@randomvariable - adding this to next. Let's leaving it open, I will circle back to this post-v1a3

randomvariable:
ACK

api/v1alpha3/vspherecluster_types.go
jzhoucliqr: Curious to know is there any thoughts around the scope that failure domain will, or plan to, cover, like across multiple datacenters ? or across multiple clusters within one datacenter? Will it support different datastores, networks ?

ncdc: @jzhoucliqr do you have any ideas on how you'd expect it to work, and what you'd want it to cover?

jzhoucliqr: one of the use case that we have is to across multiple compute clusters within one datacenter, with non-shared datastores and different networks. I guess this might be a common requirement in terms of HA for control plane nodes.

jzhoucliqr:
@yastij do we have this in roadmap?

jzhoucliqr:
@yastij @ncdc I'll pick up and add implementation for this.

PR #773, opened 2020-02-25
tilt: Add tilt configuration

randomvariable:
Signed-off-by: Naadir Jeewa <jeewan@vmware.com>

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/config/default/manager_credentials_patch.yaml
+++ b/config/default/manager_credentials_patch.yaml
@@ -12,11 +12,11 @@ spec:
         - name: VSPHERE_USERNAME
           valueFrom:
             secretKeyRef:
-              name: capv-manager-bootstrap-credentials
+              name: manager-bootstrap-credentials
               key: username
         - name: VSPHERE_PASSWORD
           valueFrom:
             secretKeyRef:
-              name: capv-manager-bootstrap-credentials
+              name: manager-bootstrap-credentials
               key: password
 
--- a/config/webhook/manager_webhook_patch.yaml
+++ b/config/webhook/manager_webhook_patch.yaml
@@ -23,4 +23,4 @@ spec:
       - name: cert
         secret:
           defaultMode: 420
-          secretName: $(SERVICE_NAME)-cert
+          secretName: capv-webhook-service-cert
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ out/
 
 # Ignore Goland files
 .idea/
+
+# Tilt
+.tiltbuild/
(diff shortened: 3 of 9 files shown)


ncdc:
I was fiddling around with this myself, and when I tried to `tilt up`, it complained about not having a secret that CAPV needs. Did you run into that?

yastij:
@ncdc - CAPV requires a pre-existing secret with credentials for the `capv-controller-manager`

randomvariable:
I've added that secret, in line with what CAPA does.

https://github.com/randomvariable/cluster-api-provider-vsphere/blob/tilt/docs/development.md

config/default/manager_credentials_patch.yaml
```diff
@@ -12,11 +12,11 @@ spec:
         - name: VSPHERE_USERNAME
           valueFrom:
             secretKeyRef:
-              name: capv-manager-bootstrap-credentials
+              name: manager-bootstrap-credentials
               key: username
         - name: VSPHERE_PASSWORD
           valueFrom:
             secretKeyRef:
-              name: capv-manager-bootstrap-credentials
```
yastij: @randomvariable - can we document this change on the release note ?

config/webhook/manager_webhook_patch.yaml
```diff
@@ -23,4 +23,4 @@ spec:
       - name: cert
         secret:
           defaultMode: 420
-          secretName: $(SERVICE_NAME)-cert
```
yastij: @randomvariable - shouldn't this be subject varReference substitution ?

randomvariable: Doesn't work with Tilt, because it injects another volume to support hot-reloading, and once that happens, kustomize can no longer do substitutions, because it doesn't support indexing into an array for variable substitution.

yastij: sgtm

config/default/manager_credentials_patch.yaml
randomvariable: Don't think this is a change rather than a fixup. Because of the nested kustomize runs, this was coming out as `capv-capv-manager-bootstrap-credentials`, kustomize now correctly prepends `capv-` once

randomvariable:
Tests are probably failing because of the webhooks?

yastij:
@randomvariable - nope I think it's mostly due to connectivity issues, as other PRs do pass tests

randomvariable:
Going to omnibus this into a PR which has more manifest fixes
/close

PR #775, opened 2020-02-25
remove deadcode in kubeadm.go

### LINKED ISSUE #769: Creating control plane machine following the quick start guide fails with "failed to retrieve kubeconfig secret"
/kind bug

After following the instructions in the quick start guide and using the `Machine` object definition defined under the [usage](https://cluster-api.sigs.k8s.io/user/quick-start.html#usage) section for vSphere the VM is created successfully but the <cluster>-kubeconfig secret is not created. On further troubleshooting, the `capi-controller-manager` logs clearly show that it is unable to retrieve this secret

```shell
root@cli-vm:~# k logs capi-controller-manager-6c64c695bb-gwkj5 -n capi-system

E0220 15:36:11.664898       1 controller.go:218] controller-runtime/controller "msg"="Reconciler error" 
"error"="failed to retrieve kubeconfig secret for Cluster \"capi-quickstart\" in namespace \"default\": Secret
 \"capi-quickstart-kubeconfig\" not found"  "controller"="machine" "request"=
{"Namespace":"default","Name":"capi-quickstart-controlplane-0"}
```
From the `capv-controller-manager` logs the error message says that the `KubeadmConfig.bootstrap.cluster.x-k8s.io \"capi-quickstart-controlplane-0\" not found" ` which is very strange as the object is created successfully.

```shell
k logs -n capv-system capv-controller-manager-88f646758-pr6wj

E0220 15:27:23.411515       1 controller.go:218] controller-runtime/controller "msg"="Reconciler error" 
"error"="failed to reconcile API endpoints for VSphereCluster default/capi-quickstart: failed to get 
KubeadmConfig capi-quickstart-controlplane-0/ for Machine default//capi-quickstart-controlplane-0: 
KubeadmConfig.bootstrap.cluster.x-k8s.io \"capi-quickstart-controlplane-0\" not found"  
"controller"="vspherecluster" "request"={"Namespace":"default","Name":"capi-quickstart"}
```

Once the cluster and machine objects are created, the `<cluster-name>-kubeconfig` secret should also be created successfully.

The issue was originally opened against the CAPI repo to fix the docs temporarily but after some troubleshooting and discussions this looks like a bug with CAPV as per https://github.com/kubernetes-sigs/cluster-api/issues/2403#issuecomment-590513263

I can take this up as I already have context on this issue.

**Environment:**

- Cluster-api-provider-vsphere version: v0.5.4
- Kubernetes version: (use `kubectl version`): 1.16.3
- OS (e.g. from `/etc/os-release`): capv-images/release/v1.16.3/ubuntu-1804-kube-v1.16.3.ova

yastij:

**What this PR does / why we need it**: This PR removes some deadcode not needed in v1alpha3

/assign @ncdc 

**Release note**:

```release-note
NONE
```


pkg/util/kubeadm.go
```diff
@@ -38,9 +38,13 @@ func GetKubeadmConfigForMachine(
 	machine *clusterv1.Machine) (*bootstrapv1.KubeadmConfig, error) {
 
 	kubeadmConfig := &bootstrapv1.KubeadmConfig{}
+	kubeadmConfigNamespace := machine.Namespace
+	if machine.Spec.Bootstrap.ConfigRef.Namespace != "" {
```
randomvariable: Which one do we want to win out? Is there an interaction with the conversion webhooks here? cc @ncdc

yastij: I'd say, we need to preserve the user's intent, If not we default to `machine.Namespace`

randomvariable: Liggitt suggested yesterday in the context of the CAPI main controllers we explicitly don't use the configref namespace if our intention is to delete it, hence the ask.

It's also an interesting attack vector to steal kubeadmsecrets from namespaces the user might not have access to.

randomvariable:
How is this function consumed? Can't find references to it?

pkg/util/kubeadm.go
ncdc: We just want to use `machine.Namespace`. We do not allow cross-namespace references. I liked Jordan's suggestion to forcibly reject (validation webhook in CAPI) if the ref's namespace is set, as we want to remove the field entirely in the future.

yastij:
@randomvariable - actually, this isn't used anymore in v1alpha3. I'll just remove the function here, and move the fix to a new PR against `release-0.5` branch

randomvariable:
rename the PR and remove the fixes issue bit, but otherwise
/lgtm

PR #807, opened 2020-03-06
Simplify getting started

randomvariable:
Now there's actual releases, don't need to specify the versions in the clusterctl commands.

Also, adds a note that other tools other than `govc` can be used


randomvariable:
linter appeased

PR #812, opened 2020-03-09
Fixed the value of DefaultCSIControllerImage pointing to the right image

georgegoh:
Fixed the value of DefaultCSIControllerImage to point to the right image for the vsphere-csi-controller container.

Reference manifest at https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/master/manifests/1.14/deploy/vsphere-csi-controller-ss.yaml#L42.

Signed-off-by: George Goh <gohge@vmware.com>

The value for DefaultCSIControllerImage should point to the CSI driver, and not syncer.

Fixes #

**Release note**:
 NONE


yastij:
/hold

until fixing CI

PR #826, opened 2020-03-12
inject CAPV credentials as a file

### LINKED ISSUE #776: CAPV should support a credential file rather than env vars
/kind bug
/area security

Providing credentials via environment variables is considered an anti-pattern, as this can be read from `/proc`.

CAPV should take a credential filename as an argument, that could then be provided as a Kubernetes secret mount.

**Environment:**

yastij:

**What this PR does / why we need it**: This PR moves `capv-controller-manager` to file-based secret injection

**Which issue(s) this PR fixes** : Fixes #776

/assign @randomvariable 

**Release note**:

```release-note
capv-controller-manager now reads credentials from a file. default path is /etc/capv/credentials.yaml
```

### DIFF
--- a/pkg/manager/options.go
+++ b/pkg/manager/options.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"time"
 
+	"sigs.k8s.io/yaml"
+
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
@@ -99,6 +101,9 @@ type Options struct {
 	// WebhookPort is the port that the webhook server serves at.
 	WebhookPort int
 
+	// CredentialsFile is the file that contains credentials of CAPV
+	CredentialsFile string
+
 	Logger     logr.Logger
 	KubeConfig *rest.Config
 	Scheme     *runtime.Scheme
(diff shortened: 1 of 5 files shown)


pkg/manager/options.go
```diff
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"time"
 
+	"gopkg.in/yaml.v2"
```
ncdc: Use sigs.k8s.io/yaml?

pkg/manager/options.go
```diff
@@ -99,6 +101,9 @@ type Options struct {
 	// WebhookPort is the port that the webhook server serves at.
 	WebhookPort int
 
+	// credentialsFile is the file that contains credentials of CAPV
```
ncdc:
```suggestion
	// CredentialsFile is the file that contains credentials of CAPV
```

pkg/manager/options.go
```diff
@@ -152,3 +156,17 @@ func (o *Options) defaults() {
 		o.PodNamespace = DefaultPodNamespace
 	}
 }
+
+func (o *Options) getCredentials() map[string]string {
```
ncdc: I would have this return an `error` too instead of calling `os.Exit` from within

yastij:
credentials stalled. @akutz is taking a look. meanwhile this passes locally.

PR #830, opened 2020-03-13
fix file-based credentials injection

yastij:

**What this PR does / why we need it**: This PR fixes an unmarshaling issues and adds errors swallowing so that the `capv-controller-manager` can start in a webhook mode without credentials

/assign @randomvariable 

**Release note**:

```release-note

```


randomvariable:
Swallowing errors will cause the linter to fail. Will need to do something

PR #833, opened 2020-03-17
read failureReason and failureMessage from the vsphereVM object to the vspheremachine

### LINKED ISSUE #805: Deleting a VM from a cluster ends up in a continual Loop and fails to recreate the VM
/kind bug

When a VM is deleted from vCenter, CAPI/CAPV should clone a new VM from template and add it back to the cluster to account for the VM that was removed from the cluster. Instead, it ends up in a continual loop unable to create the VM.

Turn Off the VM
Screen Shot 2020-03-05 at 5 18 29 PM

Delete the VM. 
Screen Shot 2020-03-05 at 5 18 38 PM

Wait upwards of 10 minutes for the watch to see the VM is no longer in the vCenter inventory.
CAPI/CAPV will attempt to clone and recreate the VM and add it back to the cluster
The VM fails because there is a continual process trying to recreate the VM and it says there is already one with a duplicate name
Screen Shot 2020-03-05 at 5 32 23 PM

The only way to get this to stop was I had to completely delete the cluster using the Local Control Plane with `kubectl delete cluster workload-cluster`
Screen Shot 2020-03-05 at 5 32 30 PM

The VM to be cloned and added back to the cluster with no errors.

**Environment:**

- Cluster-api-provider-vsphere version: v0.3.0-rc.2
- Kubernetes version: (use `kubectl version`): 1.17.3
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

yastij:
**What this PR does / why we need it**: This PR ensures that we set the machines as failed  when they are deleted from vcenter directly. remediation should happen through `MachineHealthChecks`

**Which issue(s) this PR fixes** : Fixes #805

/cc @akutz @randomvariable 

**Release note**:

```release-note
set machines as failed if they are deleted directly through vcenter
```


api/v1alpha3/vspherevm_types.go
```diff
@@ -82,6 +83,44 @@ type VSphereVMStatus struct {
...
+	// the controller, and that manual intervention is required. Examples
+	// of terminal errors would be invalid combinations of settings in the
+	// spec, values that are unsupported by the controller, or the
+	// responsible controller itself being critically misconfigured.
+	//
+	// Any transient errors that occur during the reconciliation of Machines
+	// can be added as events to the Machine object and/or logged in the
+	// controller's output.
+	// +optional
+	FailureReason *errors.MachineStatusError `json:"failureReason,omitempty"`
```
akutz: Hi @yastij,

The comments are incorrect. They should refer to `VSphereVM`, not `Machine`.

api/v1alpha3/vspherevm_types.go
```diff
@@ -82,6 +83,44 @@ type VSphereVMStatus struct {
...
+	// the controller, and that manual intervention is required. Examples
+	// of terminal errors would be invalid combinations of settings in the
+	// spec, values that are unsupported by the controller, or the
+	// responsible controller itself being critically misconfigured.
+	//
+	// Any transient errors that occur during the reconciliation of Machines
+	// can be added as events to the Machine object and/or logged in the
+	// controller's output.
+	// +optional
+	FailureMessage *string `json:"failureMessage,omitempty"`
```
akutz: Hi @yastij,

The comments are incorrect. They should refer to `VSphereVM`, not `Machine`.

pkg/services/govmomi/service.go
```diff
@@ -69,8 +71,15 @@ func (vms *VMService) ReconcileVM(ctx *context.VMContext) (vm infrav1.VirtualMac
 		if !isNotFound(err) {
 			return vm, err
 		}
-		// If VM's MoRef could not be found then the VM does not exist,
-		// and the VM should be created.
+
+		// If the machine was not found by BIOS UUID it means that it got deleted from vcenter directly
+		if wasNotFoundByBIOSUUID(err) {
+			ctx.VSphereVM.Status.FailureReason = capierrors.MachineStatusErrorPtr(capierrors.UpdateMachineError)
+			ctx.VSphereVM.Status.FailureMessage = pointer.StringPtr("Unable to find VM by BIOS UUID. The vm was removed from infra")
```
akutz: Hi @yastij,

Should the UUID be included in the error message?

controllers/vspherevm_controller.go
```diff
@@ -100,6 +100,11 @@ func (r vmReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error)
 		return reconcile.Result{}, err
 	}
 
+	if vsphereVM.Status.FailureReason != nil || vsphereVM.Status.FailureMessage != nil {
```
ncdc: Having this here will prevent the deletion from proceeding (to eventually remove the finalizer) - won't it?

yastij: +1 this should be in `reconcileNormal`

controllers/vspheremachine_controller.go
```diff
@@ -274,7 +269,37 @@ func (r machineReconciler) reconcileDeleteVMPre7(ctx *context.MachineContext) er
...
+		Namespace: ctx.VSphereMachine.Namespace,
+		Name:      ctx.Machine.Name,
+	}
+	// Attempt to find the associated VSphereVM resource.
+	if err := ctx.Client.Get(ctx, vmKey, vm); err != nil {
+		// If an error occurs finding the VSphereVM resource other than
+		// IsNotFound, then return the error. Otherwise it means the VSphereVM
+		// is already deleted, and that's okay.
+		if !apierrors.IsNotFound(err) {
+			return vm, errors.Wrapf(err, "failed to get VSphereVM %s", vmKey)
```
akutz: Hi @yastij,

You're not relying on `apierrors` to inspect this error any further, right? If you are then wrapping it will cause an inspection to fail.

controllers/vspheremachine_controller.go
```diff
@@ -274,7 +269,37 @@ func (r machineReconciler) reconcileDeleteVMPre7(ctx *context.MachineContext) er
...
+		}
+	}
+	return vm, nil
+}
+
 func (r machineReconciler) reconcileNormal(ctx *context.MachineContext) (reconcile.Result, error) {
+	vsphereVM, err := r.findVMPre7(ctx)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
```
andrewsykim: `findVMPre7` will never return the IsNotFound error but we still check it here, is that expected?

akutz: +1. This is related to my comment about the error being wrapped.

controllers/vspheremachine_controller.go
```diff
@@ -274,7 +269,37 @@ func (r machineReconciler) reconcileDeleteVMPre7(ctx *context.MachineContext) er
 	return nil
 }
 
+func (r machineReconciler) findVMPre7(ctx *context.MachineContext) (*infrav1.VSphereVM, error) {
```
andrewsykim: Curious what makes this method "pre vSphere 7", can we add a comment to clarify?

yastij: @akutz - IIRC we introducted this (xxxPre7, e.g. `reconcileNormalPre7`) in case there were different ways to deal with vms

akutz:
/lgtm
/hold

Hi @yastij,

Holding in case anyone else wants to make a final pass.

PR #847, opened 2020-03-19
watch for paused field in the vspherevm controller

yastij:

**What this PR does / why we need it**: This PR adds watches for paused field in the vspherevm controller, otherwise this potentially leads to lost status after `clusterctl move`

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/controllers/vspherevm_controller.go
+++ b/controllers/vspherevm_controller.go
@@ -17,6 +17,7 @@ limitations under the License.
 package controllers
 
 import (
+	goctx "context"
 	"fmt"
 	"reflect"
 	"strings"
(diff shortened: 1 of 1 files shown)


controllers/vspherevm_controller.go
```diff
@@ -80,7 +86,34 @@ func AddVMControllerToManager(ctx *context.ControllerManagerContext, mgr manager
...
 			&handler.EnqueueRequestForObject{},
 		).
-		Complete(vmReconciler{ControllerContext: controllerContext})
+		Build(r)
+
+	if err != nil {
+		return err
+	}
+
+	err = controller.Watch(
```
akutz: nit, this should be `if err := controller.Watch` to descope the error.

detiber: Can probably also just simplify to `return controller.Watch(...)`, since we are just returning the `err` in the non-nil case and nil otherwise.

PR #856, opened 2020-03-22
Clarifying CRD schema documentation for `ipAddrs` param that CIDR notation is required.

chipzoller:
**What this PR does / why we need it**: Clarifies CRD schema documentation for the `ipAddrs` parameter to state that the format must be in CIDR notation as this is not stated anywhere.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_haproxyloadbalancers.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_haproxyloadbalancers.yaml
@@ -129,7 +129,7 @@ spec:
                             ipAddrs:
                               description: IPAddrs is a list of one or more IPv4 and/or
                                 IPv6 addresses to assign to this device. Required
-                                when DHCP4 and DHCP6 are both false.
+                                when DHCP4 and DHCP6 are both false. Written in CIDR notation.
                               items:
                                 type: string
                               type: array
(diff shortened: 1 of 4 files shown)


config/crd/bases/infrastructure.cluster.x-k8s.io_haproxyloadbalancers.yaml
```diff
@@ -129,7 +129,7 @@ spec:
                             ipAddrs:
                               description: IPAddrs is a list of one or more IPv4 and/or
                                 IPv6 addresses to assign to this device. Required
-                                when DHCP4 and DHCP6 are both false.
+                                when DHCP4 and DHCP6 are both false. Written in CIDR notation.
```
yastij: comments needs to be updated direcly on the types. then running `make generate` will update the CRDs automatically

chipzoller: Would this only be in api/v1alpha3/types.go ?

yastij: Yes

ncdc:
@chipzoller do you think it's still worth updating the API docs?

chipzoller:
It probably is. I closed this because I needed to update the correct file but never followed up on it. It's of minimal value, but helpful for completeness.

PR #881, opened 2020-03-31
haproxy: set PhotonOS password expiry to never

codenrhoden:
This patch sets the default password expiry to -1, which makes it be equivalent to never. The default in PhotonOS is 90 days, which would lead to long-term issues with both the root account and any accounts created by cloud-init. Even though passwords are not used to access the image, an expired password means that admins logging in (even via SSH keypair, the only way) are prompted to change a password.

We also set the PhotonOS root account back to its [default](https://github.com/vmware/photon/blob/7b2f140e6496c7baaae4c99df9c1c9fdadabb9d3/installer/ks_config.txt#L156-L175). Though it's kind of a funny "default" -- if you don't set text or crypted fields in the kickstart, the installer aborts as they are required fields. 🤷‍♂
There is no reason to set an actual password for the root account here since it is never used - it was already locked.

Note that setting age: -1 also [changes the default value](https://github.com/vmware/photon/blob/3.0/installer/modules/m_updaterootpassword.py#L36-L38) in /etc/login.defs, meaning that it applies to any user when it is created during initial boot via cloud-init.

Fixes #

This is a copy of https://github.com/kubernetes-sigs/image-builder/pull/192 in image-builder.

**Release note**:

```release-note

```

### DIFF
--- a/hack/tools/haproxy/kickstart.json
+++ b/hack/tools/haproxy/kickstart.json
@@ -1,8 +1,9 @@
 {
   "hostname": "localhost",
   "password": {
-    "crypted": false,
-    "text": "photon"
+    "crypted": true,
+    "text": "*",
+    "age": -1
   },
   "disk": "/dev/sda",
   "packages": [


voor:
@codenrhoden Looking back at this issue, what would be the implication of having something like a cron job that just randomized the password every XX amount of time?

PR #893, opened 2020-04-09
Support for load balancing extra services through HAProxy

shyaamsn:
Adds extra services to HAProxy , to loadbalance certain core services in the cluster apart from loadbalancing kubernetes API server. Certain core services like authentication requires a stable endpoint that other cluster can point to. In order to address this requirement, extra services is added in HAProxy API which will loadbalance to the backends specified in the extra service spec.

Fixes #

Manual tests done on kubernetes 1.17.3 cluster and CAPV from master
Added  the following to HAProxyLoadBalancer spec and verified HAProxy cfg takes effect, stats show the service is up and running and HAproxy is actually able to loadbalance to the backends.
```
extraServices:
    - machineSelector:
        cluster.x-k8s.io/control-plane: ""
      name: dex
      ports:
      - port: 30167
      healthCheck:
        httpCheck:
          method: GET
          uri: /.well-known/openid-configuration
          scheme: HTTPS
          response:
            status: "200"
```

Configuration updated in HAproxy:
(node name and node IP removed for security purposes)

```
frontend svc_dex-30167
  mode tcp
  bind *:30167 name lb
  option tcplog
  default_backend svc_dex-30167

backend svc_dex-30167
  mode tcp
  balance roundrobin
  option httpchk GET /.well-known/openid-configuration
  default-server inter 10s downinter 10s rise 5 fall 3 slowstart 120s maxconn 1000 maxqueue 256 weight 100
  server <node-name> <node-ip>:30167 check check-ssl verify none
  server <node-name> <node-ip>:30167 check check-ssl verify none
  server <node-name> <node-ip>:30167 check check-ssl verify none
  http-check expect status 200
```

Status updated in HAProxy:
For each backend the stats are updated as follows.
```
{
        "backend_name": "svc_dex-30167",
        "name": "node-name",
        "stats": {
          "act": 1,
          "addr": "node-ip:30167",
          "bck": 0,
          "bin": 0,
          "bout": 0,
          "check_code": 200,
          "check_desc": "Layer7 check passed",
          "check_duration": 10,
          "check_fall": 3,
          "check_health": 7,
          "check_rise": 5,
          "check_status": "L7OK",
          "chkdown": 0,
          "chkfail": 0,
          "cli_abrt": 0,
          "ctime": 0,
          "downtime": 0,
          "dresp": 0,
          "econ": 0,
          "eresp": 0,
          "iid": 7,
          "lastchg": 164,
          "lastsess": -1,
          "lbtot": 0,
          "mode": "tcp",
          "pid": 1,
          "qcur": 0,
          "qlimit": 256,
          "qmax": 0,
          "qtime": 0,
          "rate": 0,
          "rate_max": 0,
          "rtime": 0,
          "scur": 0,
          "sid": 2,
          "slim": 1000,
          "smax": 0,
          "srv_abrt": 0,
          "status": "UP",
          "stot": 0,
          "ttime": 0,
          "weight": 100,
          "wredis": 0,
          "wretr": 0
        },
        "type": "server"
      }
```

**Release note**:
```release-note
Adds extra services to HAProxy, to loadbalance certain core services in the cluster apart from loadbalancing kubernetes API server.
```


api/v1alpha3/haproxyloadbalancer_types.go
```diff
@@ -37,6 +38,86 @@ type HAProxyLoadBalancerSpec struct {
...
+type ExtraService struct {
+	// The name of this service
+	Name string `json:"name"`
+
+	// The list of ports that are exposed by this service.
+	Ports []ExtraServicePort `json:"ports"`
+
+	// Label selector for backend servers. Machines matching
+	// the labels will be selected as targets for the backend servers
+	// +optional
```
vijaykatam, line 57: This should be required, otherwise you end up with empty backends.

api/v1alpha3/haproxyloadbalancer_types.go
```diff
@@ -37,6 +38,86 @@ type HAProxyLoadBalancerSpec struct {
...
+	MachineSelector map[string]string `json:"machineSelector,omitempty"`
+
+	// HealthCheck options
+	// +optional
+	HealthCheck *ExtraServiceHealthCheck `json:"healthCheck,omitempty"`
```
vijaykatam, line 62: The type could be just `HealthCheck`.

api/v1alpha3/haproxyloadbalancer_types.go
```diff
@@ -37,6 +38,86 @@ type HAProxyLoadBalancerSpec struct {
...
+}
+
+// ExtraServiceHealthCheck defines the health check
+type ExtraServiceHealthCheck struct {
+	// HealthCheckOption used to perform health check
+	HealthCheckOption `json:",inline"`
+}
+
+// HealthCheckOption defines the different health checking options available
+type HealthCheckOption struct {
```
vijaykatam, line 82: Are there haproxy types we could reuse?

controllers/haproxyloadbalancer_controller.go
```diff
@@ -369,25 +368,118 @@ func (r haproxylbReconciler) BackEndpointsForCluster(ctx *context.HAProxyLoadBal
...
+
+	for _, extraService := range ctx.HAProxyLoadBalancer.Spec.ExtraServices {
+		// Get the CAPI Machine resources for the cluster.
+		machineList := &clusterv1.MachineList{}
+		if err := ctx.Client.List(
+			ctx, machineList,
+			ctrlclient.InNamespace(ctx.Cluster.Namespace),
+			ctrlclient.MatchingLabels(extraService.MachineSelector),
+		); err != nil {
+			return nil, errors.Wrap(err, "Failed to get machines for cluster")
```
vijaykatam, line 407: Additional info in error would be helpful. `"failed to get machines for cluster with labels %s"`

api/v1alpha3/haproxyloadbalancer_types.go
```diff
@@ -37,6 +38,86 @@ type HAProxyLoadBalancerSpec struct {
...
+	// +optional
+	URI string `json:"uri,omitempty"`
+
+	// HTTP version to use for health check
+	// +optional
+	Version string `json:"version,omitempty"`
+
+	// HTTP scheme to use (HTTP or HTTPS)
+	// +optional
+	Scheme corev1.URIScheme `json:"scheme,omitempty"`
```
vijaykatam, line 104: I am wondering if should have an option for turning off ssl checks for instances where self signed certs are used.

pkg/haproxy/config.go
```diff
@@ -85,15 +85,29 @@ frontend stats
   stats refresh 500ms
   stats hide-version
   stats show-legends
-{{ $port := .Port | printf "%d" }}
+{{range .ExtraServices}}
+frontend svc_{{.Name}}
+  mode tcp
+  bind *:{{.Port | printf "%d"}} name lb
```
vijaykatam, line 91: Should the name match frontend?

pkg/haproxy/config.go
```diff
@@ -85,15 +85,29 @@ frontend stats
...
   balance first
   option httpchk GET /readyz
   default-server inter 10s downinter 10s rise 5 fall 3 slowstart 120s maxconn 1000 maxqueue 256 weight 100{{range .Addresses}}
   server {{ .NodeName }} {{ .IP }}:{{ $port }} check check-ssl verify none{{end}}
   http-check expect status 200
-
+{{ range .ExtraServices }}{{ $backendPort := .Backend.Port | printf "%d" }}{{ $backendHealthCheckConfig := .Backend.HealthCheckConfig }}
+backend svc_{{.Name}}
+  mode tcp
+  balance roundrobin{{if $backendHealthCheckConfig.Option}}
```
vijaykatam, line 105: Perhaps move the if conditional into next line? You can always use `-` to suppress blank lines.

controllers/haproxyloadbalancer_controller.go
```diff
@@ -369,25 +368,118 @@ func (r haproxylbReconciler) BackEndpointsForCluster(ctx *context.HAProxyLoadBal
...
+				IP:       addr.Address,
+			}
+			machineEndpoints = append(machineEndpoints, endpoint)
+		}
+	}
+	return machineEndpoints
+}
+
+// extraServicesForCluster returns the extra services that the loadbalancer proxies to
+func (r haproxylbReconciler) extraServicesForCluster(ctx *context.HAProxyLoadBalancerContext) ([]haproxy.Frontend, error) {
```
vijaykatam, line 396: Slightly confused why this method is returning just frontends as opposed to both front and backends.

pkg/haproxy/config.go
```diff
@@ -222,6 +236,83 @@ type RenderConfiguration struct {
 
 	// The load balancer port. Is not currently configurable.
 	Port uint32
+
+	// ExtraServices is the additional services that loadbalancer proxies to
+	ExtraServices []Frontend
+}
+
+// Frontend represent the loadbalancer endpoint that will  loadbalance the request to the backend servers
+type Frontend struct {
```
vijaykatam, line 245: This is confusing, since intuitively I am mapping `Frontend` to haproxy frontend, whereas this is more like the entire service object since it has the backend properties as well. I would instead rename `Frontend` to `Service` as it has both front end and back end attributes.

pkg/haproxy/config.go
```diff
@@ -222,6 +236,83 @@ type RenderConfiguration struct {
...
+	Port uint32
+
+	// Addresses of the machines for this backend
+	Addresses []corev1.EndpointAddress
+
+	// HealthCheckConfig for this backend
+	HealthCheckConfig *BackendHealthCheckConfig
+}
+
+type BackendHealthCheckConfig struct {
```
vijaykatam, line 268: Why do we need to redefine the health check types?

pkg/haproxy/config.go
```diff
@@ -312,6 +411,57 @@ func (c *RenderConfiguration) RenderHAProxyConfiguration() (string, error) {
...
+		option = option + " " + httpCheckConfig.Method
+	}
+	if httpCheckConfig.URI != "" {
+		option = option + " " + httpCheckConfig.URI
+	}
+	if httpCheckConfig.Version != "" {
+		option = option + " " + httpCheckConfig.Version
+	}
+	if httpCheckConfig.SSL {
+		config += " check-ssl verify none"
```
vijaykatam, line 455: I think verify should be configurable.

pkg/haproxy/haproxy_test.go
```diff
@@ -151,6 +151,56 @@ func TestCreateLoadBalancer(t *testing.T) {
...
+							IP:       http1Addr,
+							NodeName: pointer.StringPtr("http1-extra-service-2"),
+						},
+						{
+							IP:       http2Addr,
+							NodeName: pointer.StringPtr("http2-extra-service-2"),
+						},
+					},
+				},
+			},
```
vijaykatam, line 203: Are we asserting the rendered config against something?

api/v1alpha3/haproxyloadbalancer_types.go
shyaamsn: Changed

api/v1alpha3/haproxyloadbalancer_types.go
shyaamsn: By default if the labels are empty it would pick up all the machines. Is this behavior undesirable?

api/v1alpha3/haproxyloadbalancer_types.go
shyaamsn: There is one type HTTPCheck that i have reused.

controllers/haproxyloadbalancer_controller.go
shyaamsn: Done

api/v1alpha3/haproxyloadbalancer_types.go
shyaamsn: Verify is now configurable. And verify false is the default.

pkg/haproxy/config.go
shyaamsn: I dont think it needs to be match. But for consistency sake keeping it same

pkg/haproxy/config.go
shyaamsn: Thanks for the tip. Didnt know about "-".

controllers/haproxyloadbalancer_controller.go
shyaamsn: Changed the struct name to be Service

pkg/haproxy/config.go
shyaamsn: Since haproxy configuration actually has backend reference in the frontend section, I kept it that way.
e.g.:
frontend svc_dex
default_backend "svc_dex_backend"

But, I agree for reader it might be confusing. Changed it to Service as per your recommendation

pkg/haproxy/config.go
shyaamsn: I initially thought health check types here might have some different configs than the types in API that we are exposing. But, changed it now to use the same type  as it is good to be consistent

pkg/haproxy/config.go
shyaamsn: Done

pkg/haproxy/haproxy_test.go
shyaamsn: It was only verifying if config update is successful. But added assertion now

akutz:
I am not sure this is a desired PR. This is for load balancing services, not the control plane. Either way this should likely be handled via the load balancer API we have discussed. 

cc @ncdc

ncdc:
I agree with @akutz - let's set up a time to chat early this week

vijaykatam:
The motivation for this PR is to satisfy 2 requirements.
1. Stable endpoint for certain services on management cluster that doesn't change on upgrades and or recreation.
2. Endpoint that is resolvable by clients outside the cluster without requiring service projection or equivalent. 

Lets discuss more if the load balancer API can cover these requirements.

shyaamsn:
Closing this after discussions with @akutz @ncdc @vijaykatam . This will be revisited later after loadbalancer API is available.

PR #915, opened 2020-05-13
upgrade csi to v2.0.0 for newly created clusters

yastij:

**What this PR does / why we need it**: This PR bumps the vsphere CSI driver to v2.0.0 for existing clusters

changes are based on https://github.com/kubernetes-sigs/vsphere-csi-driver/tree/master/manifests/v2.0.0/vsphere-67u3/vanilla

/assign @randomvariable 

**Release note**:

```release-note
upgrade vsphere CSI driver to v2.0.0
```

### DIFF
--- a/pkg/services/cloudprovider/csi.go
+++ b/pkg/services/cloudprovider/csi.go
@@ -40,13 +40,15 @@ const (
 	DefaultCSIMetadataSyncerImage = "gcr.io/cloud-provider-vsphere/csi/release/syncer:v1.0.2"
 	DefaultCSILivenessProbeImage  = "quay.io/k8scsi/livenessprobe:v1.1.0"
 	DefaultCSIRegistrarImage      = "quay.io/k8scsi/csi-node-driver-registrar:v1.1.0"
+	CSINamespace                  = metav1.NamespaceSystem
+	CSIControllerName             = "vsphere-csi-controller"
 )
 
 func CSIControllerServiceAccount() *corev1.ServiceAccount {
 	return &corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "vsphere-csi-controller",
-			Namespace: "kube-system",
+			Name:      CSIControllerName,
+			Namespace: CSINamespace,
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/services/cloudprovider/csi.go
```diff
@@ -40,13 +40,15 @@ const (
 	DefaultCSIMetadataSyncerImage = "gcr.io/cloud-provider-vsphere/csi/release/syncer:v1.0.2"
 	DefaultCSILivenessProbeImage  = "quay.io/k8scsi/livenessprobe:v1.1.0"
 	DefaultCSIRegistrarImage      = "quay.io/k8scsi/csi-node-driver-registrar:v1.1.0"
+	CSINamespace                  = "kube-system"
```
randomvariable, line 43:
```suggestion
	CSINamespace                  = metav1.NamespaceSystem
```

pkg/services/cloudprovider/csi.go
```diff
@@ -306,6 +302,14 @@ func VSphereCSINodeContainer(image string) corev1.Container {
 				Name:  "VSPHERE_CSI_CONFIG",
 				Value: "/etc/cloud/csi-vsphere.conf",
 			},
+			{
+				Name:  "LOGGER_LEVEL",
+				Value: "PRODUCTION",
+			},
+			{
+				Name:  "X_CSI_DEBUG",
```
randomvariable, line 310: Do we want debug on by default?

yastij: IIRC `X_CSI_DEBUG` enables logging entries of GoCSI. perhaps @codenrhoden can confirm

codenrhoden: yes and no. setting `X_CSI_DEBUG=true` would set the log level to `DEBUG`, which will enable a *lot* of stuff, and also enables request/response logging. 

From the [GoCSI](https://github.com/rexray/gocsi) README:
```
A true value is equivalent to:
X_CSI_LOG_LEVEL=debug
X_CSI_REQ_LOGGING=true
X_CSI_REP_LOGGING=true
```
So it depends on if you want all those things turned on.
Request and response logging can be independentently controlled, and the default `X_CSI_LOG_LEVEL` is `WARN`. This could be changed to `INFO` if you want it more verbose by default.

randomvariable: OK, let's leave DEBUG=false
and set X_CSI_LOG_LEVEL=info instead.

codenrhoden: If you are doing that, which I think makes sense, I would also recommend setting it for the controller as well as the node. Looks like it is only set for the node right now.

yastij: done

PR #917, opened 2020-05-25
Get insecure flag value from provider config for session creation

alexander-demichev:
Currently, an insecure flag for creating sessions is hardcoded to `true`. This PR introduces ability to get it from the provider config.

### DIFF
--- a/api/v1alpha3/vspherevm_types.go
+++ b/api/v1alpha3/vspherevm_types.go
@@ -46,6 +46,9 @@ type VSphereVMSpec struct {
 	// this CRD as unstructured data.
 	// +optional
 	BiosUUID string `json:"biosUUID,omitempty"`
+
+	// Insecure is a flag that disables TLS peer verification.
+	Insecure bool `json:"insecure,omitempty"`
 }
 
 // VSphereVMStatus defines the observed state of VSphereVM
--- a/controllers/vspheremachine_controller.go
+++ b/controllers/vspheremachine_controller.go
@@ -459,6 +459,13 @@ func (r machineReconciler) reconcileNormalPre7(ctx *context.MachineContext, vsph
 		if vsphereVM != nil {
 			vm.Spec.BiosUUID = vsphereVM.Spec.BiosUUID
 		}
+
+		if ctx.VSphereCluster != nil {
+			vm.Spec.Insecure = ctx.VSphereCluster.Spec.CloudProviderConfiguration.Global.Insecure
+		} else {
+			vm.Spec.Insecure = true
+		}
+
 		return nil
 	}
 	if _, err := ctrlutil.CreateOrUpdate(ctx, ctx.Client, vm, mutateFn); err != nil {
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_vspherevms.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_vspherevms.yaml
@@ -107,6 +107,9 @@ spec:
                 description: Folder is the name or inventory path of the folder in
                   which the virtual machine is created/located.
                 type: string
+              insecure:
+                description: Insecure is a flag that disables TLS peer verification.
+                type: boolean
               memoryMiB:
                 description: MemoryMiB is the size of a virtual machine's memory,
                   in MiB. Defaults to the eponymous property value in the template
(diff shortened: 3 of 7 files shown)


ncdc:
I believe @yastij was considering removing insecure support. Let's wait for him to weigh in.
/hold

api/v1alpha3/vspherevm_types.go
```diff
@@ -44,6 +44,9 @@ type VSphereVMSpec struct {
 	// this CRD as unstructured data.
 	// +optional
 	BiosUUID string `json:"biosUUID,omitempty"`
+
+	// Insecure is a flag that disables TLS peer verification.
+	Insecure bool `json:"insecure,omitempty"`
```
yastij, line 49: @alexander-demichev - we're looking at deprecating all of the `insecure` knobs. maybe you'd like to tackle this insead ?

randomvariable:
Getting rid of this entirely is pretty tricky from a UX perspective, particularly for quickstart scenarios. You would need the CA that's signed the vcenter cert, normally self-signed mounted in the CAPV pod at the correct location. I'd be more in favour of allowing this change now, but with a better named field, and then fixing this more broadly as part of a multitenancy solution similar to CAPA's.

alexander-demichev:
I agree about UX with @randomvariable, the change will default to secure connection anyway.
@randomvariable I can change the naming, no problem here.

api/v1alpha3/vspherevm_types.go
```diff
@@ -44,6 +44,9 @@ type VSphereVMSpec struct {
 	// this CRD as unstructured data.
 	// +optional
 	BiosUUID string `json:"biosUUID,omitempty"`
+
+	// EnableSecureVCenterTransport is a flag that disables TLS peer verification.
+	EnableSecureVCenterTransport bool `json:"insecure,omitempty"`
```
yastij, line 49: Hi @alexander-demichev - sorry to ask you this again, but after some discussion we feel like having `insecure` here (i.e. same name as `VSphereCluster` is easier. could you revert the rename ?

controllers/vspheremachine_controller.go
```diff
@@ -414,6 +414,9 @@ func (r machineReconciler) reconcileNormalPre7(ctx *context.MachineContext) (run
 		if vm.Spec.ResourcePool == "" {
 			vm.Spec.ResourcePool = vsphereCloudConfig.ResourcePool
 		}
+
+		vm.Spec.EnableSecureVCenterTransport = ctx.VSphereCluster.Spec.CloudProviderConfiguration.Global.Insecure
```
yastij, line 418: if nothing is set on `VSphereCluster` this should default to true for backward-compatibility

api/v1alpha3/vspherevm_types.go
alexander-demichev: :+1:

controllers/vspheremachine_controller.go
alexander-demichev: Yeah, since `Insecure` is an optional field we should probably make it a pointer

yastij: Hmm that would be breaking change though ? let's summarize

- if vspherecluster insecure is true adding this code shouldn't break anyone
- if vspherecluster insecure is set to false adding this code is still okay, as the CPI wouldn't be working if there was an issue with `insecure: false`, so vm provisioning should be the same

so no need to change something here

alexander-demichev: right, I pushed some changes PTAL

yastij:
@alexander-demichev - could you rebase this ?

alexander-demichev:
@yastij done

yastij:
@alexander-demichev -  It seems like the linter is complaining

yastij:
@alexander-demichev - sorry for the latency, after some thoughts around the UX for this, I think it might more user-friendly to rely on certificate thumbprint, instead of letting users mount the ca.crt and trust it e.g. https://github.com/kubernetes/kubernetes/blob/323f34858de18b862d43c40b2cced65ad8e24052/staging/src/k8s.io/legacy-cloud-providers/vsphere/vclib/connection.go#L191-L192 what do you think ?

ncdc:
@yastij would it make sense to support both approaches?

yastij:
@ncdc - yeah that would make sense

yastij:
ping @alexander-demichev do you want to explore the thumbprint approach ? Ideally If the thumbprint field is set we should use it, if not we should expect a CA to be mounted an trusted ?

yastij:
closing in favour of https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1053

Issue #903, opened 2020-04-27
Move e2e tests to clusterctl framework


vincepri:
With the introduction of the new clusterctl test framework, we should draw a plan of action and publish a gap-analysis of the current e2e tests CAPV has today, and how we can port them to the new framework.

/kind cleanup
/assign @sedefsavas 
/cc @fabriziopandini @yastij

yastij:
@vincepri - that's on my plate of things to do this week.

sedefsavas:
This CAPV config is working with the new e2e, may be useful: https://github.com/kubernetes-sigs/cluster-api/pull/2950

fabriziopandini:
happy to help if required!

PR #922, opened 2020-06-01
Refactor e2e tests to utilize the new CAPI test framework

sedefsavas:
This PR moves e2e test to use the new CAPI e2e frameworks that utilizes clusterctl.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/903


test/e2e/data/infrastructure-vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,211 @@
+---
+apiVersion: cluster.x-k8s.io/v1alpha3
+kind: Cluster
```
yastij, line 3: we have a cli that generates the templates for us, I'd rather run it before `make e2e`, this will avoid having to update things into multiple places

sedefsavas: @yastij I can squash if changes look good.

yastij: @sedefsavas - looks good, we might want to remove this file though ?

sedefsavas: At least CAPZ provider is keeping the generated manifest files for e2e. It may be helpful to see what we are testing.

yastij: Let's keep it for now. it's not the source of truth so we'll need to update it whenever we change the templates

PR #926, opened 2020-06-03
Fix race condition that leads to dropping the biosuuid

yastij:

**What this PR does / why we need it**: This PR fixes some rare race condition(s) where you could potentially end up overwriting the `spec.biosUUID` field, leading to a cloning loop

/hold

holding for some more testing

**Release note**:

```release-note
fix a race condition that leads to drop the bios uuid
```

### DIFF
--- a/controllers/vspheremachine_controller.go
+++ b/controllers/vspheremachine_controller.go
@@ -315,7 +315,7 @@ func (r machineReconciler) reconcileNormal(ctx *context.MachineContext) (reconci
 	}
 
 	// TODO(akutz) Determine the version of vSphere.
-	vm, err := r.reconcileNormalPre7(ctx)
+	vm, err := r.reconcileNormalPre7(ctx, vsphereVM)
 	if err != nil {
 		if apierrors.IsAlreadyExists(err) {
 			return reconcile.Result{}, nil
@@ -368,7 +368,7 @@ func (r machineReconciler) reconcileNormal(ctx *context.MachineContext) (reconci
 	return reconcile.Result{}, nil
 }
 
-func (r machineReconciler) reconcileNormalPre7(ctx *context.MachineContext) (runtime.Object, error) {
+func (r machineReconciler) reconcileNormalPre7(ctx *context.MachineContext, vsphereVM *infrav1.VSphereVM) (runtime.Object, error) {
 	// Create or update the VSphereVM resource.
 	vm := &infrav1.VSphereVM{
 		ObjectMeta: metav1.ObjectMeta{
@@ -432,6 +432,9 @@ func (r machineReconciler) reconcileNormalPre7(ctx *context.MachineContext) (run
 		}
 		if vm.Spec.ResourcePool == "" {
 			vm.Spec.ResourcePool = vsphereCloudConfig.ResourcePool
+		}
+		if vsphereVM != nil {
+			vm.Spec.BiosUUID = vsphereVM.Spec.BiosUUID
 		}
 		return nil
 	}
--- a/controllers/vspherevm_controller.go
+++ b/controllers/vspherevm_controller.go
@@ -222,7 +222,6 @@ func (r vmReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error) 
 				vmContext.Logger.Error(err, "failed to get VSphereVM while exiting reconcile")
 				return false, nil
 			}
-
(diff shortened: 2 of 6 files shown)


pkg/services/govmomi/util.go
```diff
@@ -49,26 +50,45 @@ func sanitizeIPAddrs(ctx *context.VMContext, ipAddrs []string) []string {
...
+		inventoryPath := path.Join(folder.InventoryPath, ctx.VSphereVM.Name)
+		ctx.Logger.Info("using inventory path to find vm", "path", inventoryPath)
+		vm, err := ctx.Session.Finder.VirtualMachine(ctx, inventoryPath)
+		if err != nil {
+			if isVirtualMachineNotFound(err) {
+				return types.ManagedObjectReference{}, errNotFound{byName: inventoryPath}
+			}
+			return types.ManagedObjectReference{}, err
+		}
+		ctx.Logger.Info("vm found by name", "vmref", vm.Reference())
```
ncdc, line 88: We need to make sure that if `ctx.VSphereVM.Spec.BiosUUID` has a value, we compare that value with the VM we just found by reference. If they do not match, we cannot consider the VM as a match.

controllers/vspherevm_controller.go
```diff
@@ -324,7 +325,19 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
 	}
 
 	// Update the VSphereVM's BIOS UUID.
-	ctx.VSphereVM.Spec.BiosUUID = vm.BiosUUID
+	ctx.Logger.Info("vm bios-uuid", "biosuuid", vm.BiosUUID)
```
akutz, line 328: Hi @yastij,

I think this check should validate the `BiosUUID` is non-empty. It should *never* be empty, and thus an error should be returned here, causing the reconcile event to requeue.

pkg/services/govmomi/util.go
yastij: If we're here it means that `ctx.VSphereVM.Spec.BiosUUID` is empty

controllers/vspherevm_controller.go
yastij: @akutz - do you mean something like this ?:

```go
	if vm.BiosUUID != "" {
		ctx.VSphereVM.Spec.BiosUUID = vm.BiosUUID
	} else {
         // return err and requeue
         ....
        }

File: controllers/vspheremachine_controller.go [hunk 3]
```diff
@@ -433,6 +433,9 @@ func (r machineReconciler) reconcileNormalPre7(ctx *context.MachineContext) (run
 		if vm.Spec.ResourcePool == "" {
 			vm.Spec.ResourcePool = vsphereCloudConfig.ResourcePool
 		}
+		if vsphereVM != nil {
```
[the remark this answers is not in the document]
yastij: @ncdc - it's created here https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/controllers/vspheremachine_controller.go#L373 and the bios uuid is never set

pkg/services/govmomi/util.go
```diff
@@ -49,26 +50,45 @@ func sanitizeIPAddrs(ctx *context.VMContext, ipAddrs []string) []string {
...
 	objRef, err := ctx.Session.FindByInstanceUUID(ctx, instanceUUID)
 	if err != nil {
 		return types.ManagedObjectReference{}, err
 	}
 	if objRef == nil {
-		return types.ManagedObjectReference{}, errNotFound{instanceUUID: true, uuid: instanceUUID}
+		// fallback to use inventory paths
+		folder, err := ctx.Session.Finder.FolderOrDefault(ctx, ctx.VSphereVM.Spec.Folder)
+		if err != nil {
+			return types.ManagedObjectReference{}, errors.Wrapf(err, "unable to get folder for %q", ctx)
```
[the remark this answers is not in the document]
yastij: fixed

controllers/vspheremachine_controller.go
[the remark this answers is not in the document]
ncdc: 😄 I'm talking about `vsphereVM` which is being passed in and came from a `client.Get()` call, not `vm`.

ncdc:
Did a live Zoom review with @yastij - this is good to go.
/lgtm

PR #932, opened 2020-06-15
Update go.mod

### LINKED ISSUE #931: Support for hierarchical networks
/kind bug

Note the fix to this is easy, just upgrade to govmami - testing it will be tricky though as the environment to reproduce this requires a very specific network scenario. 

In certain newer VSphere environments using CVDS related features a network name which exists in two places needs a / to disambiguate it.  i.e. 

```
duplicate switches in separate DVS's (solvable)
dvs1
  sw1 ] id = 1 (dvs1/sw1)
dvs2
  sw1  id = 1  (dvs2/sw1
```

In the above scenario the VM_NETWORK id would be = dvs1/sw1 as opposed to sw1.

VM_NETWORK would support slashes

- I havent reproduced this bug but know it exists in certain environments from anecdotal evidence.  The fix is just to upgrade Govmami, putatively, to 23.0.
- In general, I have validated that cvDS is not a problem at all for CAPV.

**Environment:**

- Cluster-api-provider-vsphere version: master
- Kubernetes version: (use `kubectl version`): any
- OS (e.g. from `/etc/os-release`): any

jayunit100:
fixes #931

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/onsi/gomega v1.9.0
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v0.0.6
-	github.com/vmware/govmomi v0.22.2
+	github.com/vmware/govmomi v0.23.0
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
 	gopkg.in/gcfg.v1 v1.2.3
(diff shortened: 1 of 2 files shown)


yastij:
@jayunit100 - can you squash your commits into one ? Also you have a typo s/govmami/govmomi/g

PR #935, opened 2020-06-15
Add VSphereVM conditions

fabriziopandini:
This PR adds conditions to the VSphereVM kind

This PR builds on top of https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/934, so please consider only the last commit

**Release note**:
```release-note
The VSphereVM kind now has a `Status.Conditions` field providing evidence about the current state of the objects.
```

/hold
/assign @yastij


pkg/services/govmomi/service.go
```diff
@@ -118,6 +130,7 @@ func (vms *VMService) ReconcileVM(ctx *context.VMContext) (vm infrav1.VirtualMac
 		return vm, err
 	}
 
+	conditions.MarkTrue(ctx.VSphereVM, infrav1.VMProvisionedCondition)
```
fabriziopandini, line 133: This is probably redundant given we are doing the same in `reconcileNormal`...

fabriziopandini: removed

fabriziopandini:
/hold cancel
rebased after #934 merged

api/v1alpha3/condition_consts.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Conditions and condition Reasons for the VSphereVM object
+
+const (
+
+	// VMProvisionedCondition documents the status of the provisioning of a VSphereVM.
+	VMProvisionedCondition clusterv1.ConditionType = "VMProvisioned"
+
+	// CloningReason documents (Severity=Info) a VSphereVM currently executing the clone operation.
+	CloningReason = "Cloning"
```
yastij, line 29: one question, do we want condition to be on going or in the past

api/v1alpha3/condition_consts.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// CloningFailedReason (Severity=Warning) documents a VSphereVM controller detecting
+	// an error while provisioning; the reconcile loop will automatically retry the operation,
+	// but a user intervention might be required to fix the problem.
+	CloningFailedReason = "CloningFailed"
+
+	// PoweringOnReason documents (Severity=Info) a VSphereVM currently executing the power on sequence.
+	PoweringOnReason = "PoweringOn"
```
yastij, line 37: same

api/v1alpha3/condition_consts.go
fabriziopandini: VMProvisining happens only one time at the beginning of the machine lifecycle, so IMO it is fine to use past

fabriziopandini:
@yastij rebased

PR #939, opened 2020-06-17
Add VSphereMachine conditions

fabriziopandini:
This PR adds conditions to the VSphereMachine kind

Being the VSphereMachine a wrapper of VSphereVM, for sake of consistency the two objects are sharing the same condition and some of the condition reason.

**Release note**:
```release-note
The VSphereMachine kind now has a `Status.Conditions` field providing evidence about the current state of the objects.
```


fabriziopandini:
@yastij rebased

Issue #854, opened 2020-03-22
Static IP assignment failure in pre-built Ubuntu 18.04 CAPV templates (1.16.3 and 1.17.4)


chipzoller:
/kind bug

**What steps did you take and what happened:**
Define a VSphereMachineTemplate spec with static IPs. Deploy the manifest. IP is attempted to be assigned in controller but is never applied with cloud-init.

**What did you expect to happen:**
Static IP is assigned.

**Anything else you would like to add:**
IP is pulled from range
```
I0322 15:56:33.843461       1 util.go:501] capv-controller-manager/vspherevm-controller/default/czstatic-25tdz "msg"="the VM is missing the requested IP address"  "addressType"="static" "addressValue"="192.168.1.95"
```
but cloud-init never assigns IP.

**Environment:**

- Cluster-api-provider-vsphere version: 0.6.2
- Kubernetes version: (use `kubectl version`): 
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.4", GitCommit:"8d8aa39598534325ad77120c120a22b3a990b5ea", GitTreeState:"clean", BuildDate:"2020-03-12T21:03:42Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

chipzoller:
See also issue https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/853

chipzoller:
Values for IP addresses need to be given in CIDR format.

Issue #853, opened 2020-03-22
Validation needed for VSphereMachineTemplate.spec.template.spec.network.devices.ipAddrs


chipzoller:
****Have not inspected the code so unsure if bug or feature request.**

**What steps did you take and what happened:**
Populate the `devices` object with invalid string data for param `ipAddrs` specifying range.

```yaml
devices:
        - dhcp4: false
          networkName: VLAN1
          deviceName: ens150
          ipAddrs: [192.168.1.170-192.168.1.172]
          gateway4: 192.168.1.1
          nameservers: [192.168.1.5]
          searchDomains: [domain.local]
```

Object is accepted and cluster builds. Invalid data causes first deployed control plane to have IP address unset causing a hang preventing further cluster build-out.

**What did you expect to happen:**
Validation webhook error stating `ipAddrs` is invalid array of strings (pattern match failure).

**Anything else you would like to add:**
Two possible desired outcomes:

1. Validation controller rejects the request because array of strings for IP address type is invalid.
2. (preferable) Support range of IP addresses if contiguous addresses can be supplied.
3. (irrespective of 1 or 2) Ensure number of supplied IP addresses (derived from range or observed directly from explicit strings) matches total number of machines in the referenced MachineTemplate and KubeadmControlPlane (again, code not inspected to verify if this currently is not implemented).

**Environment:**

- Cluster-api-provider-vsphere version:  0.6.2
- Kubernetes version: (use `kubectl version`): 
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.4", GitCommit:"8d8aa39598534325ad77120c120a22b3a990b5ea", GitTreeState:"clean", BuildDate:"2020-03-12T21:03:42Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

chipzoller:
***Edit just to be clear:**  I'm well aware I supplied the value in an invalid way and that the IPs should have been added individually. This was only a test.

[...]

PR #320, opened 2019-06-19
Update to CAPI 0.1.3

akutz:
This patch updates the CAPI dependency to 0.1.3.

PR #957, opened 2020-07-08
wait for staticIP to be set in VSphereVM

### LINKED ISSUE #728: Static IP pool mechanism
More of a general inquiry about how/if static IP pools can be assigned/consumed by CAPV. Currently, it appears to be DHCP only. As this project matures and is consumed by various automations, the desire to either pull from a vCenter static IP pool or external IPAM integration will be required. Wondering if this is being thought about.

abhinavnagaraj:
CAPV has support to assign static IP to VSphereMachine, but the static IP itself can be assigned by different means. For eg., one such implementation can be using a third party controller to modify VSphereMachine to add static IP after the object is created.
Irrespective of the implementation choice to assign static IP, if DHCP is disabled, the CAPV controllers have to wait for static IPs to be available and then continue to provision VMs.
Currently, if both DHCP4 and DHCP6 are disabled and a staticIP is set in HAProxyLB's Network.Devices.IpAddrs, the VSphereVM for HAProxyLB goes to 'Ready' without any IPAddress in the status. The HAProxyLB then never goes to 'Ready' since its indefinitely waiting for the VSphereVM's IPAddress.
This PR checks if both DHCP4 and DHCP6 are disabled for all network devices(NetworkSpec.devices) and if there is no IP provided in the IPAddrs list, then the VSphereVM controller waits for the static IP to be available/assigned.

Fixes #728 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/vspherevm_controller.go
+++ b/controllers/vspherevm_controller.go
@@ -315,6 +315,11 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
 	// TODO(akutz) Implement selection of VM service based on vSphere version
 	var vmService services.VirtualMachineService = &govmomi.VMService{}
 
+	if r.isWaitingForStaticIPAllocation(ctx) {
+		ctx.Logger.Info("vm is waiting for static ip to be available")
+		return reconcile.Result{}, nil
+	}
+
 	// Get or create the VM.
 	vm, err := vmService.ReconcileVM(ctx)
 	if err != nil {
@@ -349,6 +354,18 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
 	ctx.Logger.Info("VSphereVM is ready")
 
 	return reconcile.Result{}, nil
+}
+
+func (r vmReconciler) isWaitingForStaticIPAllocation(ctx *context.VMContext) bool {
+	devices := ctx.VSphereVM.Spec.Network.Devices
+	for _, dev := range devices {
+		if !dev.DHCP4 && !dev.DHCP6 && len(dev.IPAddrs) == 0 {
+			// Static IP is not available yet
+			return true
+		}
+	}
+
+	return false
 }
 
 func (r vmReconciler) reconcileNetwork(ctx *context.VMContext, vm infrav1.VirtualMachine) {
--- a/pkg/services/govmomi/util.go
+++ b/pkg/services/govmomi/util.go
@@ -440,7 +440,9 @@ func waitForIPAddresses(
 					// spec's static IP addresses.
 					isStatic := false
 					for _, specIP := range deviceSpec.IPAddrs {
-						if discoveredIP == specIP {
+						// The static IP assigned to the VM is required in the CIDR format
+						ip, _, _ := gonet.ParseCIDR(specIP)
+						if discoveredIP == ip.String() {
 							isStatic = true
 							break
 						}
@@ -525,7 +527,8 @@ func waitForIPAddresses(
 			// If the device spec requires static IP addresses, the wait
 			// is not over if the device lacks one of those addresses.
 			for _, specIP := range deviceSpec.IPAddrs {
-				if _, ok := macToHasStaticIP[mac][specIP]; !ok {
+				ip, _, _ := gonet.ParseCIDR(specIP)
+				if _, ok := macToHasStaticIP[mac][ip.String()]; !ok {
 					ctx.Logger.Info(
 						"the VM is missing the requested IP address",
 						"addressType", "static",


ncdc:
@abhinavnagaraj thanks for this! We are in the middle of a few things, so it may take us a week or so before we can review this, but it's definitely on our to-do list. Feel free to reach out the week of July 20th if you don't hear back from us by then. Thanks for your patience.

controllers/vspherevm_controller.go
```diff
@@ -351,13 +363,34 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
...
 
-func (r vmReconciler) reconcileNetwork(ctx *context.VMContext, vm infrav1.VirtualMachine) {
+func (r vmReconciler) isWaitingForStaticIPAllocation(ctx *context.VMContext) bool {
+	waitForIP := false
+	devices := ctx.VSphereVM.Spec.Network.Devices
+
+	for _, dev := range devices {
+		if !dev.DHCP4 && !dev.DHCP6 && len(dev.IPAddrs) == 0 {
+			// Static IP is not available yet
+			waitForIP = true
```
ncdc, line 373:
```suggestion
			return true
```

controllers/vspherevm_controller.go
```diff
@@ -351,13 +363,34 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
...
+		}
+	}
+
+	return waitForIP
```
ncdc, line 377:
```suggestion
	return false
```

controllers/vspherevm_controller.go
```diff
@@ -351,13 +363,34 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
...
+}
+
+func (r vmReconciler) reconcileNetwork(ctx *context.VMContext, vm infrav1.VirtualMachine) (bool, error) {
```
ncdc, line 380: This function never returns a non-nil error. Please change the signature to only return a boolean.

ncdc, line 380: Please add godoc that indicates what the returned boolean is for.

controllers/vspherevm_controller.go
```diff
@@ -341,7 +346,14 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
...
 
 	// Update the VSphereVM's network status.
-	r.reconcileNetwork(ctx, vm)
+	if ok, err := r.reconcileNetwork(ctx, vm); !ok {
+		if err != nil {
+			return reconcile.Result{}, errors.Wrapf(err,
+				"unexpected error while reconciling network for %s", ctx)
+		}
+		ctx.Logger.Info("network is not reconciled")
+		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
```
ncdc, line 355: This does change the behavior of when the VSphereVM's status.ready field is set to true, along with the provisioned condition. It now waits until guest tools reports at least one IP address before doing this. Will this have any negative repercussions with the delay in setting ready/provisioned? Before, CAPV considered the VSphereVM and therefore the VSphereMachine as ready without needing guest tools to report an IP. Now, this adds delays in 30-second increments.

Do you have a strong need to delay considering the VSphereVM/VSphereMachine ready until guest tools reports an IP?

I'm tempted to say we don't include this change as part of this PR, and instead only do the `isWaitingForStaticIPAllocation` part. WDYT?

controllers/vspherevm_controller.go
```diff
@@ -341,7 +346,14 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
 	}
 
 	// Update the VSphereVM's network status.
-	r.reconcileNetwork(ctx, vm)
+	if ok, err := r.reconcileNetwork(ctx, vm); !ok {
```
ncdc, line 349: Once you change the signature as requested below, this can become `if !r.reconcileNetwork(ctx, vm) {`

controllers/vspherevm_controller.go
```diff
@@ -341,7 +346,14 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
...
+		if err != nil {
+			return reconcile.Result{}, errors.Wrapf(err,
+				"unexpected error while reconciling network for %s", ctx)
+		}
+		ctx.Logger.Info("network is not reconciled")
```
ncdc, line 354: I would remove this log message - you have the one below in reconcileNetwork that logs that it's waiting on ip addresses.

controllers/vspherevm_controller.go
abhinavnagaraj: Ack

controllers/vspherevm_controller.go
abhinavnagaraj: Ack

controllers/vspherevm_controller.go
abhinavnagaraj: Ack

abhinavnagaraj: Ack

controllers/vspherevm_controller.go
abhinavnagaraj: Ack

controllers/vspherevm_controller.go
abhinavnagaraj: Ack

controllers/vspherevm_controller.go
abhinavnagaraj: No, there is no specific requirement to delay setting the VSphereVM to 'Ready'. Without the 'Requeue', for some reason the 'PollImmediateInfinite' is indefinitely stuck at "resource is not patched". I need to figure out the root cause. Otherwise, we can restrict this PR to only do the "isWaitingForStaticIPAllocation".

jzhoucliqr: just my understanding, let me know if it's correct. without this requeue, since vspheremachine controller is waiting for IP in vspherevm.status, so until next resync reconciliation loop (default 10m), vspheremachine will not be ready. 

for the same reason, even line #320, we may need to do requeue as well.  @abhinavnagaraj did some tests, it took 20m for LB+CP to come up. 

Unless there is other places that we need to change to wait for the IPs synchronously.

ncdc: I'm about to run into a meeting, but I think the vmService.ReconcileVM() call which is above this line kicks off a goroutine that waits for all IPs to be seen in the VM, and then it reenqueues the VSphereVM. I'm surprised this isn't working and is causing the resync delay... or maybe I'm misreading it. I'll check more later when I have more time. Thanks.

jzhoucliqr: Thanks. I'll double check.

abhinavnagaraj: We figured the issue is with line https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/4ee7445b4b9fcebc660a486a40bb05105a239c46/pkg/services/govmomi/util.go#L443. Since static IP is required to be in the CIDR format(https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/854), we need to parse the 'specIP' which is in CIDR format here, before comparing it with the discoveredIP. I was able to verify the PR after fixing this.

pkg/services/govmomi/util.go
```diff
@@ -525,7 +533,14 @@ func waitForIPAddresses(
 			// If the device spec requires static IP addresses, the wait
 			// is not over if the device lacks one of those addresses.
 			for _, specIP := range deviceSpec.IPAddrs {
-				if _, ok := macToHasStaticIP[mac][specIP]; !ok {
+				ip, _, err := gonet.ParseCIDR(specIP)
+				if err != nil {
+					ctx.Logger.Info(
```
jzhoucliqr, line 538: shall we use log.Error() ?

controllers/vspherevm_controller.go
ncdc: Great find!

pkg/services/govmomi/util.go
```diff
@@ -440,7 +440,15 @@ func waitForIPAddresses(
 					// spec's static IP addresses.
 					isStatic := false
 					for _, specIP := range deviceSpec.IPAddrs {
-						if discoveredIP == specIP {
+						// The static IP assigned to the VM is required in the CIDR format
+						ip, _, err := gonet.ParseCIDR(specIP)
+						if err != nil {
+							ctx.Logger.Info(
```
ncdc, line 446: If we do #853, getting an error here should be impossible

pkg/services/govmomi/util.go
```diff
@@ -525,7 +533,14 @@ func waitForIPAddresses(
 			// If the device spec requires static IP addresses, the wait
 			// is not over if the device lacks one of those addresses.
 			for _, specIP := range deviceSpec.IPAddrs {
-				if _, ok := macToHasStaticIP[mac][specIP]; !ok {
+				ip, _, err := gonet.ParseCIDR(specIP)
+				if err != nil {
```
ncdc, line 537: If we do #853, getting an error here should be impossible

pkg/services/govmomi/util.go
abhinavnagaraj: Ack. I'll remove the err check.

ncdc: Well.... we haven't done #853 yet, so I'd leave it in for now.

pkg/services/govmomi/util.go
ncdc: Error is probably better. It's also important to remember that these logs are not typically user-facing.

controllers/vspherevm_controller.go
```diff
@@ -351,6 +356,18 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
 	return reconcile.Result{}, nil
 }
 
+func (r vmReconciler) isWaitingForStaticIPAllocation(ctx *context.VMContext) bool {
+	devices := ctx.VSphereVM.Spec.Network.Devices
+	for _, dev := range devices {
+		if !dev.DHCP4 && !dev.DHCP6 && len(dev.IPAddrs) == 0 {
```
yastij, line 362: I'd rather have this as:
```
if dev.DHCP4 || dev.DHCP6 || len(dev.IPAddrs) != 0 {
       return false
}
```
easier to reason about

pkg/services/govmomi/util.go
yastij: #853 is going to be merged today, so I'd rather just remove this

pkg/services/govmomi/util.go
yastij: #853 is going to be merged today, so I'd rather just remove this

pkg/services/govmomi/util.go
yastij: I agree here

pkg/services/govmomi/util.go
ncdc: The PR for this is in (#970) - should we update this to not check the error any more?

abhinavnagaraj: Ack, will remove the error check.

pkg/services/govmomi/util.go
abhinavnagaraj: Ack, will remove the error check.

controllers/vspherevm_controller.go
abhinavnagaraj: This won't work if more than one device is waiting for static IP. We want to wait till all devices have the static IP.

yastij: after giving some thoughts, this might be the more straight-forward way of thinking about this

yastij:
/lgtm
/hold

@ncdc - this is good to go, feel free to remove the hold

ncdc:
@abhinavnagaraj could you please squash your commits?

ncdc:
/lgtm cancel
while we wait for squashing

ncdc:
I think you have a bad rebase?

PR #958, opened 2020-07-08
Log whole object

jayunit100:
Quick attempt at logging the VSPHERE_NETWORK and other potentially important spec informations when the network not ready yet error message occurs.


randomvariable:
This is a security risk if we print GuestInfo to the logs, which includes sensitive information.

ncdc:
Closing because this is a security risk.
/close

PR #959, opened 2020-07-08
add kube-vip support to CAPV

### LINKED ISSUE #819: Internal Load Balancer for ControlPlane HA (other types)
/kind feature

I fount it difficult to controlplane HA because the built-in load balancer is not in vsphere.
And I also found some attempts to solve this.
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/705 (a separate LB)
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/722 (nsx-t LB)

However, a separate LB do not guarantee HA for LB(haproxy vm failure) and need additional resource for this.
And nsx-t LB can only be configured in certain supported environments.

I want a load balancer with no SPOF due to LB failure, no additional resources and no environmental dependencies.

This is for reference and it's actually the way I'm using it on-prem.
Run `haproxy` and `keepalived` on the controlplane to make internal LB.
It only requires to allocate `vip` that can be routed on all cluster nodes.
The concept architecture is shown below and has been tested for node failure.

architecture-1
architecture-2

In my on-prem, I wrote static ip manually in `haproxy.cfg` and `keepalived.conf` and then deployed  the `haproxy+keepalived` container with this file mounted as docker. (with --restart=always)
The `haproxy+keepalived` container internally monitors the process for each module.

I install the control planes after installing the architecture described above.

I don't know much about capv, but I think it's possible to automate it using capv.

yastij:

**What this PR does / why we need it**: This PR adds support for VIP load balancing through [kube-vip](https://github.com/plunder-app/kube-vip), this is done by mutating the pod located under `/etc/kubernetes/manifests/kube-vip.yaml` to add the right environment variables 

**Which issue(s) this PR fixes** : Fixes #819

/assign @randomvariable 

**Release note**:

```release-note
add kube-vip support on CAPV for control plane endpoints
```


yastij:
closing this as we can just get away with templating, now that kube-vip was reworked

PR #964, opened 2020-07-10
add support for the kube-vip flavor

yastij:

**What this PR does / why we need it**: This PR adds support for a kube-vip flavor, two env variables are introduced:

`CONTROL_PLANE_ENDPOINT_IP`: the VIP we want to use
`VIP_NETWORK_INTERFACE`: the network interface we want to bind the VIP to (e.g. `eth0`)

**Which issue(s) this PR fixes**: Fixes #

**Release note**:

```release-note
add support for a kube-vip flavor
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,7 +269,9 @@ manifests:  $(STAGE)-version-check $(STAGE)-flavors $(MANIFEST_DIR) $(BUILD_DIR)
 
 .PHONY: flavors
 flavors: $(FLAVOR_DIR)
-	go run ./packaging/flavorgen -f multi-host > $(FLAVOR_DIR)/cluster-template.yaml
+	go run ./packaging/flavorgen -f multi-host > $(FLAVOR_DIR)/cluster-template-haproxy.yaml
+	go run ./packaging/flavorgen -f vip > $(FLAVOR_DIR)/cluster-template.yaml
+
 
 .PHONY: release-flavors ## Create release flavor manifests
 release-flavors: release-version-check
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ e2e: e2e-image
 e2e: $(GINKGO) $(KUSTOMIZE) $(KIND) $(GOVC) ## Run e2e tests
 	$(MAKE) release-manifests
 	@mkdir -p $(E2E_TEMPLATE_DIR)
-	cp $(RELEASE_DIR)/cluster-template.yaml $(E2E_TEMPLATE_DIR)
+	cp $(RELEASE_DIR)/cluster-template-haproxy.yaml $(E2E_TEMPLATE_DIR)
 	@echo PATH=$(PATH)
 	@echo
 	@echo Contents of $(TOOLS_BIN_DIR):
--- a/packaging/flavorgen/flavors/generators.go
+++ b/packaging/flavorgen/flavors/generators.go
@@ -398,7 +398,7 @@ func kubeVIPPod() string {
 						},
 						{
 							Name:  "lb_name",
-							Value: "kcpEdnpoint",
+							Value: "kcpEndpoint",
 						},
 						{
 							Name:  "lb_bindtovip",
@@ -407,10 +407,6 @@ func kubeVIPPod() string {
 						{
 							Name:  "vip_interface",
 							Value: vipNetworkInterfaceVar,
-						},
-						{
-							Name:  "lb_type",
-							Value: "tcp",
 						},
 						{
 							Name:  "lb_type",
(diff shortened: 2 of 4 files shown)


randomvariable:
PR needs some revisions to handle envsubst.
Closing and will resubmit.
/close

yastij:
Rebased to fix CI

/assign @ncdc @detiber

packaging/flavorgen/flavors/generators.go
```diff
@@ -396,6 +505,13 @@ func newHAProxyLoadBalancer() infrav1.HAProxyLoadBalancer {
 }
 
 func newKubeadmControlplane(replicas int, infraTemplate infrav1.VSphereMachineTemplate) controlplanev1.KubeadmControlPlane {
+	files := []bootstrapv1.File{
```
ncdc, line 508: This should only be here if using kube-vip. It is currently also being generated for the HAProxy flavor.

yastij:
@ncdc - pushed

Makefile
```diff
@@ -116,7 +116,7 @@ e2e: e2e-image
 e2e: $(GINKGO) $(KUSTOMIZE) $(KIND) $(GOVC) ## Run e2e tests
 	$(MAKE) release-manifests
 	@mkdir -p $(E2E_TEMPLATE_DIR)
-	cp $(RELEASE_DIR)/cluster-template.yaml $(E2E_TEMPLATE_DIR)
+	cp $(RELEASE_DIR)/cluster-template-haproxy.yaml $(E2E_TEMPLATE_DIR)
```
ncdc, line 119: When would you want to switch e2e to use kube-vip?

packaging/flavorgen/flavors/generators.go
```diff
@@ -331,6 +341,105 @@ func defaultPreKubeadmCommands() []string {
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kube-vip",
+			Namespace: "kube-system",
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Name:  "kube-vip",
+					Image: "plndr/kube-vip:0.1.6",
```
ncdc, line 359: Either here or a follow-up: let's allow the image to be customizable + defaulted if the user doesn't specify an alternative.

packaging/flavorgen/flavors/generators.go
```diff
@@ -331,6 +341,105 @@ func defaultPreKubeadmCommands() []string {
...
+							Name:  "lb_backendport",
+							Value: "6443",
+						},
+						{
+							Name:  "vip_addpeerstolb",
+							Value: "true",
+						},
+						{
+							Name:  "lb_name",
+							Value: "kcpEdnpoint",
```
ncdc, line 401:
```suggestion
							Value: "kcpEndpoint",
```

packaging/flavorgen/flavors/generators.go
```diff
@@ -331,6 +341,105 @@ func defaultPreKubeadmCommands() []string {
...
+							Value: vipNetworkInterfaceVar,
+						},
+						{
+							Name:  "lb_type",
+							Value: "tcp",
+						},
+						{
+							Name:  "lb_type",
+							Value: "tcp",
+						},
```
ncdc, line 418: Dupe of previous var

packaging/flavorgen/flavors/generators.go
```diff
@@ -331,6 +341,105 @@ func defaultPreKubeadmCommands() []string {
...
+				},
+			},
+			HostNetwork: true,
+			Volumes: []v1.Volume{
+				{
+					Name: "kubeconfig",
+					VolumeSource: v1.VolumeSource{
+						HostPath: &v1.HostPathVolumeSource{
+							Path: "/etc/kubernetes/admin.conf",
+							Type: &hostPathType,
```
ncdc, line 429: Let's use `pointer.StringPtr()`

Makefile
yastij: Ideally once DHCP is supported in kube-vip, or when doing some testing around VIPs from other segments in our infra.

packaging/flavorgen/flavors/generators.go
yastij: we don't enable custom images in our upstream templates (e.g. CCM and CSI), usually users have template overrides with their own images

packaging/flavorgen/flavors/generators.go
yastij: `pointer.StringPtr()` would return a `*string`, here we need a `*HostPathType, I'm not sure we can cast this ? did you have something specific in mind ?

ncdc: Sorry, assumed it was a *string.

packaging/flavorgen/flavors/generators.go
ncdc: 👍

PR #971, opened 2020-07-21
replace featuregate flags to fix the e2e

yastij:

**What this PR does / why we need it**: This PR fixes the CI by replacing the featureflag with explicit values, this is just a short-term fix until #922 merges

Fixes #

/assign @ncdc 

**Release note**:

```release-note

```


ncdc:
You switched networks?

yastij:
@ncdc - yes, seems like the one we're using was heavily used, so switched to a dedicated one

PR #978, opened 2020-07-31
wait to have IP addresses reported before making a VSphereVM as ready

### LINKED ISSUE #938: VSphereVM should set ready state after network is reconciled
/kind bug

created a cluster with a control plane node, and observed the VSphereVM being set ready before the network is reconciled.

VSphereVM being set ready when network is reconciled, thus providing a consistent signal with VSphereMachines and the HAProxyLoadBalancer that instead are setting ready only after network is reconciled.

yastij:

**What this PR does / why we need it**: This PR waits for the `VSphereVM` to report Addresses before marking it as Ready

**Which issue(s) this PR fixes**  Fixes #938 

/assign @vincepri 

**Release note**:

```release-note
wait to report Addresses on the VSphereVM before marking it as ready
```

### DIFF
--- a/controllers/vspherevm_controller.go
+++ b/controllers/vspherevm_controller.go
@@ -348,6 +348,11 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
 	// Update the VSphereVM's network status.
 	r.reconcileNetwork(ctx, vm)
 
+	// we didn't get any addresses, requeue
+	if len(ctx.VSphereVM.Status.Addresses) == 0 {
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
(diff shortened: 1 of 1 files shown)


controllers/vspherevm_controller.go
```diff
@@ -348,6 +348,11 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
 	// Update the VSphereVM's network status.
 	r.reconcileNetwork(ctx, vm)
```
gab-satchi, line 349: It looks like this call [returns an error](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/controllers/vspheremachine_controller.go#L518-L521) when address isn't found. can probably just use that?

PR #986, opened 2020-08-06
Uses CRS to deploy CSI for default flavour

### LINKED ISSUE #718: deploy CSI with ClusterResourceSet
/kind feature

This is to discuss whether or not CSI should be installed as an add-on, and not reconciled as part of CAPV when deploying workload clusters.

cc @voor @mylesagray

gab-satchi:
This is a continuation of the work done by @Boutheina-Dab on #972 
Changes since that PR: 
- ClusterResourceSetBinding removed. Should be created automatically
- Switched from modifying HAProxy template to the default (kubevip)
- Reverted changes in e2e templates (will do a follow up commit for that shortly)
- Service Account and Cloud Config are injected via secrets. The rest of the required resources are done using ConfigMaps

Holding for e2e test changes.
/hold 

Fixes #718 

**Release note**:

```release-note
Default CAPV template will deploy CSI via ClusterResourceSet. Needs the ClusterResourceSet feature gate enabled
```


packaging/flavorgen/flavors/generators.go
```diff
@@ -178,6 +173,7 @@ func newCluster(vsphereCluster infrav1.VSphereCluster, controlPlane *controlplan
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      clusterNameVar,
 			Namespace: namespaceVar,
+			Labels:    clusterLabels(),
```
gab-satchi, line 176: Adding this label here so the ClusterResourceSet can use it to find and target this cluster

gab-satchi:
I got a local e2e test to pass with the new templates. It would be nice to set up a PVC test that would ensure everything in this PR works as expected but I'm not sure how to go about getting a dedicated IP for the `CONTROL_PLANE_ENDPOINT_IP`. 

/hold cancel

yastij:
@gab-satchi - I did test this, it's working as expected

PR #987, opened 2020-08-07
hardcode network interface to eth0 in the cluster-template.yaml

yastij:

**What this PR does / why we need it**: This PR fixes a bug in our templating, where if you supplied a custom network interface name to kube-vip, it wouldn't start (we weren't setting `DeviceName` into the `VSphereMachineTemplate`).
After giving more thoughts for a quickstart template we don't need to template that kind of things.

Fixes #

/assign @gab-satchi 

**Release note**:

```release-note

```

### DIFF
--- a/packaging/flavorgen/flavors/generators.go
+++ b/packaging/flavorgen/flavors/generators.go
@@ -53,7 +53,6 @@ const (
 	vSphereTemplateVar           = "${ VSPHERE_TEMPLATE }"
 	workerMachineCountVar        = "${ WORKER_MACHINE_COUNT }"
 	controlPlaneEndpointVar      = "${ CONTROL_PLANE_ENDPOINT_IP }"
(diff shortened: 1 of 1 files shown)


packaging/flavorgen/flavors/generators.go
```diff
@@ -386,7 +385,7 @@ func kubeVIPPod() string {
 						},
 						{
 							Name:  "vip_interface",
-							Value: vipNetworkInterfaceVar,
+							Value: "eth0",
```
vincepri, line 388: Want to add a comment why this is hardcoded now?

gab-satchi:
/lgtm

one less thing to specify now

packaging/flavorgen/flavors/generators.go
yastij: this shouldn't have been templated in the first place I think

yastij:
added a comment @vincepri

PR #1003, opened 2020-08-19
let deletion go through when the folder of a VSphereVM doesn't exist

### LINKED ISSUE #1000: CAPV doesn't delete the resources when the VM Folder doesn't exist
/kind bug

When you provision a cluster with a wrong folder name, you won't be able to delete the resources on the management cluster unless you create that folder, then trigger a delete

CAPV should be able to delete the resources when the supplied folder doesn't exist

**Environment:**

- Cluster-api-provider-vsphere version: v0.7.0

yastij:

**What this PR does / why we need it**: This PR fixes a bug where the deletion flow of a `VSphereVM` is blocked if its folder doesn't exist.

**Which issue(s) this PR fixes**: Fixes #1000 

/assign @vincepri @fabriziopandini 

**Release note**:

```release-note
let deletion go through when the folder of a VSphereVM doesn't exist
```

### DIFF
--- a/pkg/services/govmomi/service.go
+++ b/pkg/services/govmomi/service.go
@@ -159,7 +159,7 @@ func (vms *VMService) DestroyVM(ctx *context.VMContext) (infrav1.VirtualMachine,
 	if err != nil {
 		// If the VM's MoRef could not be found then the VM no longer exists. This
 		// is the desired state.
-		if isNotFound(err) {
+		if isNotFound(err) || isFolderNotFound(err) {
 			vm.State = infrav1.VirtualMachineStateNotFound
 			return vm, nil
 		}
--- a/pkg/services/govmomi/errors.go
+++ b/pkg/services/govmomi/errors.go
@@ -44,6 +44,15 @@ func isNotFound(err error) bool {
 	}
 }
 
+func isFolderNotFound(err error) bool {
+	switch err.(type) {
+	case *find.NotFoundError:
+		return true
+	default:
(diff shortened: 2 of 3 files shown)


pkg/services/govmomi/service.go
```diff
@@ -159,7 +159,7 @@ func (vms *VMService) DestroyVM(ctx *context.VMContext) (infrav1.VirtualMachine,
 	if err != nil {
 		// If the VM's MoRef could not be found then the VM no longer exists. This
 		// is the desired state.
-		if isNotFound(err) {
+		if isNotFound(err) || isFolderNotFound(err) {
```
vincepri, line 162: Do you want to try using https://golang.org/pkg/errors/#Is ?

vincepri: Or https://godoc.org/github.com/pkg/errors#Is given that's the package we actually use

yastij: @vincepri - I like this. can you raise an issue for this, I'd like to refactor the error handling to use it ?

randomvariable:
can we follow up and document this behaviour, as i'm sure it's going to be one of those things people might end up doing in vcenter and then find surprising. 
/approve

leave @vincepri to do lgtm wrt to the errors question.

pkg/services/govmomi/service.go
vincepri: #1004

PR #1014, opened 2020-08-24
Add ability to assign tags to virtual machines

alexander-demichev:
Add the ability to assign tags to virtual machines, similar to what is done in AWS. In order to do so, we need to have `Tags` field in `VSphereMachineTemplate` and create REST client. This change assumes that tags are already created by the user.

**Release note**:

-->
```release-note
Add the ability to assign tags to virtual machines. 
```


pkg/services/govmomi/create_test.go
```diff
@@ -38,6 +40,7 @@ func TestCreate(t *testing.T) {
 		t.Fatal(err)
 	}
 	model.Service.TLS = new(tls.Config)
+	model.Service.RegisterEndpoints = true
```
alexander-demichev, line 43: This line and `_ "github.com/vmware/govmomi/vapi/simulator"` are needed to make simulator register rest session endpoint or it returns an error - `unable to create rest client: POST https://127.0.0.1:63223/rest/com/vmware/cis/session: 404 Not Found`

yastij:
@alexander-demichev - we're planning to cut v0.7.1 today, let's review this, merge it and cut a release once it's ready

randomvariable:
Need to add the conversion webhooks for the tests to pass.

EleanorRigby:
@alexander-demichev :  I think these tags have to pre-exist in vCenter. Is there a way to dynamically create/delete them with VMs?

ncdc:
@alexander-demichev would you be able to elaborate on your use case here? I recognize that other infra providers such as CAPA allow this, but you may want to read https://www.vmware.com/content/dam/digitalmarketing/vmware/en/pdf/techpaper/performance/tagging-vsphere67-perf.pdf about tag performance in larger configurations.

yastij (commented):
I didn't review the tests yet, I'll do another pass monday

api/v1alpha3/vspheremachine_types.go
```diff
@@ -37,6 +37,10 @@ type VSphereMachineSpec struct {
 	// vsphere://12345678-1234-1234-1234-123456789abc
 	// +optional
 	ProviderID *string `json:"providerID,omitempty"`
+
+	// Tags is an optional set of tags to add to an instance.
+	// +optional
+	Tags []string `json:"tags,omitempty"`
```
yastij, line 43: I'd put these in the `VirtualMachineCloneSpec`

api/v1alpha3/vspherevm_types.go
```diff
@@ -46,6 +46,10 @@ type VSphereVMSpec struct {
 	// this CRD as unstructured data.
 	// +optional
 	BiosUUID string `json:"biosUUID,omitempty"`
+
+	// Tags is an optional set of tags to add to an instance.
+	// +optional
+	Tags []string `json:"tags,omitempty"`
```
yastij, line 52: once it's in the the `VirtualMachineCloneSpec` this shouldn't be needed

yastij (commented):
Another pass for the API

yastij:
@alexander-demichev - the scalability concerns pointed by @ncdc are standing. what is the use case we want to cover and at  what scale ?

alexander-demichev:
@yastij @ncdc I didn't know about performance issues with tags. My use case is similar to what tags do for other providers.

vrabbi:
i also have a few use cases for this. having the tags would greatly enhance the ability for extensibility using the VMware Ecosystem tools. the use cases i have seen for this so far are:
1. on-boarding clusters to vRealize Automation using the tag based on-boarding mechanism
2. using VMware Event Broker Appliance to run custom tasks when a specific event occurs. this use case is very helpful as there is no specific event in vSphere for a CAPV cluster being deployed as it is simply provisioning VMs from the vSphere perspective. by adding the tags a custom action could be built in VEBA which could run any custom logic when a node is created in a CAPV clusters. this could include DRS rules (anti-affinity or affinity), creating DNS records for the nodes, adding nodes to a CMDB , creating custom groups in VROPs for monitoring a cluster etc.

regarding the performance issues, while this is something to take into consideration, 25k tag assignments with good performance as mentioned in the article seems quite reasonable. many systems add tags in vSphere and the performance issue i dont think should be a blocker here either. just as maybe the user doesnt have enough resources isnt a blocker for automated deployments. 
having CAPV add tags on its own would be an issue but if its opt-in and user defined i think it makes a lot of sense to add this functionality

embano1:
Not sure about the initial use case here from reading the PR, but I'd suggest to use an async/throttled out-of-band (ie decoupled) mechanism based on vSphere events for this if possible, eg using the event broker or similar. It helps with separation of concerns (access permissions to VC), scalability concerns, network issues and making the controller logic less vulnerable to reconciliation/blocking issues due to heavy VC RPCs. We've seen users successfully adopting this model for tag management (beyond K8s controllers) and it works nicely  (orthogonal) with the async reconciliation pattern in K8s controllers.

yastij:
@vrabbi - I agree that we should have this. @alexander-demichev can you address the comments and rebase ?

alexander-demichev:
@yastij yes, I'll try to find time this week

api/v1alpha3/vspheremachine_types.go
yastij: can we make the tag category explicit also ? (this would enable key:value use cases)

alexander-demichev: @yastij Do you have an example of how to attach a tag from a specific category?

pkg/services/govmomi/service.go
```diff
@@ -273,6 +279,18 @@ func (vms *VMService) reconcilePowerState(ctx *virtualMachineContext) (bool, err
 	}
 }
 
+func (vms *VMService) reconcileTags(ctx *virtualMachineContext) error {
+	tagManager := tags.NewManager(ctx.Session.RestClient)
+
+	for _, tagName := range ctx.VSphereVM.Spec.Tags {
+		if err := tagManager.AttachTag(ctx, tagName, ctx.Ref); err != nil {
```
embano1, line 286: I recommend some form of Client-side rate limiter for these REST calls, see this note:

> As an additional consideration, to avoid denial-of-service attacks to the tagging service, there is a limit of
360 tagging-related operations per second. To detect if you are hitting this limit, look for messages like “Request rejected due to high request rate. Try again later.” in the vapi-endpoint log file (/var/log/vmware/vapi/endpoint/endpoint.log). If your creation code is exceeding this limit, we recommend inserting a pause of 2 ms or more between tag-creation calls.

Also, I'll try to implement the batch API for more efficient (time, resources) tagging in govmomi soon.

api/v1alpha3/types.go
```diff
@@ -133,6 +133,10 @@ type VirtualMachineCloneSpec struct {
 	// Defaults to empty map
 	// +optional
 	CustomVMXKeys map[string]string `json:"customVMXKeys,omitempty"`
+
+	// Tags is an optional set of tags to add to an instance.
+	// +optional
+	Tags []string `json:"tags,omitempty"`
```
embano1, line 139: Should we highlight that this is tagID or do we plan to support tag resolution and caching?

embano1:
@alexander-demichev FYI https://github.com/vmware/govmomi/pull/2211
Also note the current bug in https://github.com/vmware/govmomi/issues/2210 related to `attach-tag-to-multiple-objects` if you need that as well.

srm09:
Joining this PR conversation very late, is there a current blocker on this one? @alexander-demichev Are you waiting on any inputs/comments?

mazzy89:
This would come in very handy at the time when the Load Balancer provided by NSX-T is leveraged. In fact, you can create pools by targeting VMs using tags assigned to them. 

Interested to this here too.

TerryHowe:
This work here seems like it would be worth reviving or is there another effort?

srm09:
@TerryHowe you could bring this up during the CAPV office hours. We can revive the proposal and update the implementation as well.

gab-satchi:
Summarizing what's left to be done for the PR:
- rebase now that main branch has switched over to v1alpha4. The API changes will need to be modified to drop v1alpha2
- switch to the batch tagging API. also implies we use tagIDs in the spec and not the tag name. 
- rate limiting for the tag attach call. This may not be strictly necessary if we switch to the batch tagging

@alexander-demichev apologies for the long delay with this PR. Is this still something you're interested in contributing?

config/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachines.yaml
```diff
@@ -514,6 +514,11 @@ spec:
                   a linked clone. This field is ignored if LinkedClone is not enabled.
                   Defaults to the source's current snapshot.
                 type: string
+              tags:
+                description: Tags is an optional set of tags to add to an instance.
```
vrabbi, line 518: should be "Tags is an optional set of tag IDs to add to an instance."

config/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachinetemplates.yaml
```diff
@@ -574,6 +574,12 @@ spec:
                           to create a linked clone. This field is ignored if LinkedClone
                           is not enabled. Defaults to the source's current snapshot.
                         type: string
+                      tags:
+                        description: Tags is an optional set of tags to add to an
```
vrabbi, line 578: Should be "Tags is an optional set of tag IDs to add to an instance"

config/crd/bases/infrastructure.cluster.x-k8s.io_vspherevms.yaml
```diff
@@ -272,6 +272,11 @@ spec:
                   a linked clone. This field is ignored if LinkedClone is not enabled.
                   Defaults to the source's current snapshot.
                 type: string
+              tags:
+                description: Tags is an optional set of tags to add to an instance.
```
vrabbi, line 276: This should be "Tags is an optional set of tag IDs to add to an instance."

vrabbi (commented):
it should be explicit we are dealing with Tag IDs and not Tag Names

config/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachines.yaml
TerryHowe:
```suggestion
                description: Tags is an optional set of tag IDs to add to an instance.
```

alexander-demichev:
Hi all, I updated the PR:
1. bumped govmomi dependency so we can use batch tagging
2. changed API field name to TagIDs

I'm having an issue with generating conversions, it's failing with the following error:
`
E0818 13:26:54.674110   59681 conversion.go:755] Warning: could not find nor generate a final Conversion function for sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha4.VirtualMachineCloneSpec -> sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha3.VirtualMachineCloneSpec
E0818 13:26:54.674407   59681 conversion.go:756]   the following fields need manual conversion:
E0818 13:26:54.674417   59681 conversion.go:758]       - TagIDs
`

pkg/services/govmomi/service.go
```diff
@@ -140,6 +140,10 @@ func (vms *VMService) ReconcileVM(ctx *context.VMContext) (vm infrav1.VirtualMac
 		return vm, err
 	}
 
+	if err := vms.reconcileTags(vmCtx); err != nil {
```
embano1, line 143: I think error handling could be slightly improved here instead of reconciling (potentially forever) on any error, including unrecoverable errors.

To quote from the docs:

> AttachMultipleTagsToObject attaches multiple tag IDs to a managed object. This operation is idempotent. If a tag is already attached to the object, then the individual operation is a no-op and no error will be thrown. This operation is not atomic. If the underlying call fails with one or more tags not successfully attached to the managed object reference it might leave the managed object reference in a partially tagged state and needs to be resolved by the caller. In this case BatchErrors is returned and can be used to analyse failure reasons on each failed tag. Specified tagIDs must use URN-notation instead of display names or a generic error will be returned and no tagging operation will be performed. If the managed object reference does not exist a generic 403 Forbidden error will be returned. This operation was added in vSphere API 6.5.

We are working on improved error type handling (for assertions) but the above should give you an impression of different error scenarios.

alexander-demichev: Does it mean that there are no typed errors at the moment?

embano1: Only partially and the situation is improving as we export more errors. Please see the tests for the method which shows common error scenarios and the errors.

alexander-demichev:
@yastij Hi, I rebased the PR and added a condition

PR #1018, opened 2020-08-26
Add support for upgrading virtual hardware

### LINKED ISSUE #998: CAPV OVAs at HW image 15?
/kind bug

Before we make the change in `image-builder`, I wanted to confirm something first. CAPV always deploys vSphere CSI, correct? If it does, the documented required version for vSphere CSI is [hardware version 15](https://vsphere-csi-driver.sigs.k8s.io/driver-deployment/prerequisites.html), but `image-builder` is currently setting the hardware version to 13.

Are there any use cases where we don't want the hardware image version to be 15? I can't think of any if CAPV is *always* deploying vSphere CSI. I can picture scenarios were vSphere CSI isn't used (say, deploying a cluster on vSphere but using another storage solution), but in that case, I think it is likely that there will be some customization happening at both the CAPV manifest layer, and potentially the OVAs coming from `image-builder`. The hardware version is easily cusomizable in `image-builder`, but the default behavior should most likely support the documented needs of CNS, which is deployed by default with CAPV. Any objections around this?

EleanorRigby:

Upgrades the virtual hardware version of the machine to the latest supported on ESXi host it is cloned.
 - Assuming the cluster is composed of homogeneous ESXi versioned servers, it should not break any v-motions initiated by DRS.

We need this because it frees capv from image-builder dependency to bump virtual hardware whenever needed for components like CSI.

It is gated with a Boolean to not upgrade when requirements are such. Otherwise, defaults to true and upgrades hardware version.

Fixes #998 

I would like to add this as a one time op attached to the clone part of VM. Reconciling this fact is wasteful as it cannot be downgraded once it is upgraded. Better place/way to put this in?

**Release note**:
```release-note
Adds support to upgrade to latest virtual hardware supported on host it is cloned.
```


yastij:
@ncdc @codenrhoden - This would make the infrastructure mutable to some extent, I'm not sure we'll want to do this. If users want to upgrade the hardware version, they'll just need to rollout a new machine.

ncdc:
Given we haven't agreed to proceed with the power states proposal in Cluster API (https://github.com/kubernetes-sigs/cluster-api/issues/1810), I don't think we can proceed with this PR, as it requires the user to use a mechanism outside of Cluster API (namely, asking vCenter to power off the machine) to control the state of the VM.

From the special notes above:
> I would like to add this as a one time op attached to the clone part of VM. Reconciling this fact is wasteful as it cannot be downgraded once it is upgraded. Better place/way to put this in?

Unfortunately based on how things are currently coded, there is not an easy way to make this a one-time operation associated with the clone operation.

EleanorRigby:
> Given we haven't agreed to proceed with the power states proposal in Cluster API ([kubernetes-sigs/cluster-api#1810](https://github.com/kubernetes-sigs/cluster-api/issues/1810)), I don't think we can proceed with this PR, as it requires the user to use a mechanism outside of Cluster API (namely, asking vCenter to power off the machine) to control the state of the VM.

@ncdc The hardware upgrade does not power off VM but it happens before VM is powered on. It just checks the power state so as not to issue a wasteful vSphere call to upgrade(which cannot happen when powered off). 
It does not in any way issue any power on/off request. In the above proposal, when instance state is stopped/shut-down, there will be a lot of reconcilers like power/hardware/network/clone-reconcile that will anyways need to be stopped from affecting manually stopped state of the machine.


>This would make the infrastructure mutable to some extent, I'm not sure we'll want to do this. If users want to upgrade the hardware version, they'll just need to roll out a new machine.
@yastij : I do not think it will make it mutable. It is a change to template VM image, same as adding a network device or configuring DRS settings. I believe it is equivalent to `upgrading` hardware of template from UI and then using `clusterctl` to clone. The only thing different is that now upgrade will happen after cloning via code.

FWIW, I have come to know from different k8s distribution experience that a lot of enterprises or users would want the latest hardware version. At that point, VC7 supports version version `17` that has some cool [features](https://www.teimouri.net/whats-new-in-hardware-version-17/) . This change will seamlessly support all VC versions for CAPV without having to build separate stemcell

ncdc:
@EleanorRigby yes, I understand the upgrade happens before power-on - I wasn't sure if your goal is

1. Upgrade the HW version of an existing machine
1. Upgrade the HW version of a new machine, prior to powering it on (one-time operation)
1. Both/either

If it's 1, then we'd need power states in the API.

If it's only 2, then it _could_ make sense to ask/expect the user creating the OVAs to produce new ones with the newer hardware version.

Are there any reasons someone would **not** want to automatically upgrade their newly-cloned VM to the latest hardware version?

EleanorRigby:
@ncdc  So goal is `2` : to be able to upgrade for a new Machine. 

###### Virtual Hardware Version : to upgrade or not to upgrade , that's the question
from what I heard and worked on in past it is generally advised to upgrade virtual hardware as it means new motherboard. But here is an excerpt from an internal chat about this

```
I am particularly interested in understanding the impact of changing virtual hardware (upgrade) on OS and apps, specifically on apps in the containerized world (k8s). Would you know more about this? or forward me to someone who can advise more?

@XXXX  6 months ago
It’s kind of a general guidance. Imagine taking a motherboard out of the computer after shutting down the server and then booting it back up with a new motherboard… Not always a good idea as the operating system can freak out and not understand the underlying hardware changes. This happened in the past with some windows servers… I believe it was 2016 but I could be wrong. As far as containers or containerize workloads… I’m not very sure on how they would react to the same “hardware rug” being pulled out from underneath their feet. If that makes sense
```

Are there some changes you suggest I can do to make this PR better and less code/reconcile-cost intensive? 

I believe automation will anyway be better than the operator's manual effort. Moreover, if capv ever moves toward Continous Stemcell delivery through published content libraries it might be good to not maintain 3-4 different virtual hardware versioned stemcells for same k8s version and OS

ncdc:
@frapposelli what are your thoughts?

frapposelli:
In general I'm against this change, the virtual hardware is fundamentally tied to the OS install, and validating it should be a concern of `image-builder` and not of Cluster API.

While it's generally safe to upgrade from one version to another, there have been many instances where a vHW upgrade breaks the OS install because of a change in the underlying hardware and steps in the OS have the be performed prior to upgrading vHW. This is particularly true for Windows and Linux is also not immune.

EleanorRigby:
Closing PR as hardware issue will be addressed through separate VM image.

PR #1019, opened 2020-08-28
Add conditions for the reconcile delete workflow

fabriziopandini:
Adding conditions for the reconcile delete workflow

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/991

**Release note**:
```release-note
VSphereCluster, VSphereMachine, VSPhereVM now update conditions during the delete workflow
```

### DIFF
--- a/controllers/vspherecluster_controller.go
+++ b/controllers/vspherecluster_controller.go
@@ -30,6 +30,11 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha3"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/record"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/cloudprovider"
+	infrautilv1 "sigs.k8s.io/cluster-api-provider-vsphere/pkg/util"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
 	clusterutilv1 "sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/conditions"
@@ -43,12 +48,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
-
-	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha3"
-	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
-	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/record"
-	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/cloudprovider"
-	infrautilv1 "sigs.k8s.io/cluster-api-provider-vsphere/pkg/util"
 )
 
 var (
@@ -181,18 +180,6 @@ func (r clusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr er
 	// Always issue a patch when exiting this function so changes to the
 	// resource are patched back to the API server.
 	defer func() {
-		// always update the readyCondition.
-		conditions.SetSummary(clusterContext.VSphereCluster,
-			conditions.WithConditions(
-				infrav1.LoadBalancerAvailableCondition,
-				infrav1.CCMAvailableCondition,
-				infrav1.CSIAvailableCondition,
-			),
-			conditions.WithStepCounterIfOnly(
-				infrav1.LoadBalancerAvailableCondition,
-			),
-		)
-
 		if err := clusterContext.Patch(); err != nil {
 			if reterr == nil {
 				reterr = err
@@ -213,6 +200,10 @@ func (r clusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr er
 func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconcile.Result, error) {
 	ctx.Logger.Info("Reconciling VSphereCluster delete")
 
+	// ccm and csi needs the control plane endpoint (which is removed when the VSphereCluster is)
+	conditions.MarkFalse(ctx.VSphereCluster, infrav1.CCMAvailableCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
+	conditions.MarkFalse(ctx.VSphereCluster, infrav1.CSIAvailableCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
+
 	vsphereMachines, err := infrautilv1.GetVSphereMachinesInCluster(ctx, ctx.Client, ctx.VSphereCluster.Namespace, ctx.VSphereCluster.Name)
 	if err != nil {
 		return reconcile.Result{}, errors.Wrapf(err,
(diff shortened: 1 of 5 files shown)


fabriziopandini:
/hold 
for CAPI version v0.3.9

fabriziopandini:
/hold cancel
Rebased on top of CAPI v0.3.9

controllers/vspherecluster_controller.go
```diff
@@ -230,6 +217,7 @@ func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconci
 		return reconcile.Result{}, err
 	}
 	if len(haproxyLoadbalancers.Items) > 0 {
+		conditions.MarkFalse(ctx.VSphereCluster, infrav1.LoadBalancerAvailableCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
```
gab-satchi, line 220: Should we add the following after line 225? To flip the condition reason from Deleting to Deleted once the loadbalancers have been successfully removed

```
if conditions.Has(ctx.VSphereCluster, infrav1.LoadBalancerAvailableCondition) {
	conditions.MarkFalse(ctx.VSphereCluster, infrav1.LoadBalancerAvailableCondition, clusterv1.DeletedReason, clusterv1.ConditionSeverityInfo, "")
}
```

controllers/vspherecluster_controller.go
```diff
@@ -181,18 +180,6 @@ func (r clusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr er
 	// Always issue a patch when exiting this function so changes to the
 	// resource are patched back to the API server.
 	defer func() {
-		// always update the readyCondition.
```
gab-satchi, line 184: Don't think moving this into `func Patch()` changed the functionality but curious about the reason for this change. It felt more transparent to have the spec/status changes located in the controller and have patch just attempt to patch the object in the current state.

controllers/vspherecluster_controller.go
```diff
@@ -282,6 +270,13 @@ func (r clusterReconciler) reconcileNormal(ctx *context.ClusterContext) (reconci
 		return reconcile.Result{}, nil
 	}
 
+	// If the cluster is deleted, that's mean that the workload cluster is being deleted and so the CCM/CSI instances
+	if !ctx.Cluster.DeletionTimestamp.IsZero() {
+		conditions.MarkFalse(ctx.VSphereCluster, infrav1.CCMAvailableCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
```
gab-satchi, line 275: It looks like this block would only get exercised if the cluster is deleted while in the middle of a normal reconcile. Otherwise it would go through `reconcileDelete`.  Should this be part of reconcileDelete as well?

gab-satchi (commented):
I feel conditions for the delete flow is only truly useful when things go wrong or get stuck. In that situation, it would be nice to see some of the conditions with Reason `Deleted` and one or more in `Deleting` or even `DeletionFailed`. It looks like the controller just counts the vsphereMachines and waits so a DeleteionFailed isn't possible there but we're actively deleting the LoadBalancer so that could probably update to DeleteionFailed on error.

controllers/vspherevm_controller.go
```diff
@@ -286,8 +286,10 @@ func (r vmReconciler) reconcileDelete(ctx *context.VMContext) (reconcile.Result,
 	// TODO(akutz) Implement selection of VM service based on vSphere version
 	var vmService services.VirtualMachineService = &govmomi.VMService{}
 
+	conditions.MarkFalse(ctx.VSphereVM, infrav1.VMProvisionedCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
 	vm, err := vmService.DestroyVM(ctx)
 	if err != nil {
+		conditions.MarkFalse(ctx.VSphereVM, infrav1.VMProvisionedCondition, "DeletionFailed", clusterv1.ConditionSeverityInfo, err.Error())
```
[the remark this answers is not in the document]
fabriziopandini: Unfortunately, the `clusterv1.DeletionFailedReason` const didn't make v0.3.9 due to some e2e failures 😓. 
However, I will fix this as soon as CAPI dep will be updated to v0.3.10

controllers/vspherecluster_controller.go
fabriziopandini: Sure!
When a condition changes, we are required to update the summary condition as well, otherwise `ready` gets out of sync from the underlying conditions.
Given that in many controllers, we are starting to issue `Patch` in the middle of the reconciliation loop, I'm systematically moving "update the readyCondition" into `func Patch()` to ensure that we are respecting the requirement above.

controllers/vspherecluster_controller.go
fabriziopandini: Done
However, in a normal delete workflow immediately after the Load balancer is deleted the finalizer gets removed, so the change won't be visible to the users.

controllers/vspherecluster_controller.go
fabriziopandini: Done!

fabriziopandini:
@gab-satchi thanks for the feedback!
everything should be addressed and/or answer provided

controllers/vspherecluster_controller.go
gab-satchi: That makes sense. Thanks!

controllers/vspherecluster_controller.go
```diff
@@ -213,6 +200,12 @@ func (r clusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr er
 func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconcile.Result, error) {
 	ctx.Logger.Info("Reconciling VSphereCluster delete")
 
+	// If the cluster is deleted, that's mean that the workload cluster is being deleted and so the CCM/CSI instances
+	if !ctx.Cluster.DeletionTimestamp.IsZero() {
```
gab-satchi, line 204: Nit: no longer need the IF condition

controllers/vspherecluster_controller.go
gab-satchi: I was thinking of a case where the vsphereMachines aren't getting deleted properly but the loadbalancer was successfully deleted. Unsure how probable / common that failure is though.

controllers/vspherecluster_controller.go
```diff
@@ -230,11 +223,14 @@ func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconci
...
 	}
 	if len(haproxyLoadbalancers.Items) > 0 {
+		conditions.MarkFalse(ctx.VSphereCluster, infrav1.LoadBalancerAvailableCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
 		for _, lb := range haproxyLoadbalancers.Items {
 			if err := r.Client.Delete(ctx, lb.DeepCopy()); err != nil && !apierrors.IsNotFound(err) {
+				conditions.MarkFalse(ctx.VSphereCluster, infrav1.LoadBalancerAvailableCondition, "DeletionFailed", clusterv1.ConditionSeverityWarning, "")
 				return reconcile.Result{}, err
 			}
 		}
+		conditions.MarkFalse(ctx.VSphereCluster, infrav1.LoadBalancerAvailableCondition, clusterv1.DeletedReason, clusterv1.ConditionSeverityInfo, "")
```
gab-satchi, line 233: apologies, got my line numbers mixed up in original comment. I was suggesting to add this after the `if len(haproxyLoadbalancers.Items) > 0 {` block. As it stands, we requeue after a 10 second wait but may be prematurely setting the condition to Deleted.

controllers/vspherecluster_controller.go
fabriziopandini: I see a potential bug in the logic here, because we should not start deleting the load balancer if there are still machines, otherwise the cluster is busted (and if I'm not wrong cluster controller in CAPI has this control)

controllers/vspherecluster_controller.go
fabriziopandini: What if the vSphereCluster gets deleted but the Cluster not? is that a use case we want to support?
In that case CCM/CSI technically could continue to live in the workload cluster, even if there is nothing reconciling them...

controllers/vspherecluster_controller.go
fabriziopandini: no problems! done!

controllers/vspherecluster_controller.go
yastij: I agree with @fabriziopandini - this might turn into a potential bug when we start deleting in-cluster objects (e.g. `Services`, `PVCs`...), but for now we don't require `kube-apiserver` access to complete the deletion.

@fabriziopandini - can you file an issue for this ?

yastij: I agree with @fabriziopandini - this might turn into a potential bug when we start deleting in-cluster objects (e.g. `Services`, `PVCs`...), but for now we don't require `kube-apiserver` access to complete the deletion.

@fabriziopandini - can you file an issue for this ?

controllers/vspherecluster_controller.go
yastij: we should remove the if, as the ccm and csi needs the control plane endpoint (which is removed when the `VSphereCluster` is)

controllers/vspherecluster_controller.go
fabriziopandini: done, see https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1033

fabriziopandini:
@gab-satchi @yastij all comments should be address, waiting for a go/no-go before squashing commits

gab-satchi:
thanks @fabriziopandini lgtm

yastij (approved):
/approve

gab-satchi:
good to squash commits @fabriziopandini

yastij (commented):
/lgtm

PR #983, opened 2020-08-06
[e2e] Install CNI using ClusterResourceSet and bump up CAPI to v0.3.9

sedefsavas:
What this PR does / why we need it:
This PR installs e2e CNIs using ClusterResourceSet.

This PR should hold until CAPI v0.3.9 release or whenever these changes are merged: kubernetes-sigs/cluster-api#3332

/hold

PR #1025, opened 2020-09-08
Use envsubst syntax

fabriziopandini:
This PR adapt templates/flavor generator to the syntax of envsubst (without spaces inside ${}).
This makes it possible to use templates both with clusterctl - which tolerated spaces but not requires them - and the plain envsubst.

**Release note**:
```release-note
Change templates to use the envsubst syntax for variables (without spaces inside ${})
```

### DIFF
--- a/packaging/flavorgen/flavors/generators.go
+++ b/packaging/flavorgen/flavors/generators.go
@@ -33,29 +33,29 @@ import (
 )
 
 const (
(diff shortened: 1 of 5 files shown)


yastij (approved):
/approve
/lgtm

fabriziopandini:
rebased and applied changes to new templates introduced by https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/983

PR #1029, opened 2020-09-09
[e2e] In e2e cluster template, modify cluster CIDR to match with CNI CIDR

sedefsavas:
This PR changes the pod CIDR in the cluster template used in e2e tests.

test/e2e/data/infrastructure-vsphere/cluster-template.yaml is generated on the fly during `make e2e` based on the latest manifests. It would still be a good idea to keep a copy of it under test/e2e to see what template we are using for the tests.


gab-satchi:
I feel we should git ignore the generated cluster-template file. When debugging or simply running e2e, it will still be there for the user to see the  template getting applied but by ignoring it we don't track it twice in source control. And someone unaware of how e2e tests are run, won't waste time manually modifying both templates when making changes to cluster-template. WDYT?

yastij:
+1 on what @gab-satchi said

yastij (approved):
/approve
/assign @gab-satchi

test/e2e/data/infrastructure-vsphere/cluster-template.yaml
```diff
@@ -1,66 +1,185 @@
+apiVersion: v1
```
gab-satchi, line 1: I believe you need to `git rm` this file in order for the gitignore to work

sedefsavas: I removed it now but I feel like it would be good to keep it for visibility.
test/e2e/config/vsphere-ci.yaml will have reference to a template, that is not there.
`- sourcePath: "../../../test/e2e/data/infrastructure-vsphere/cluster-template.yaml"`

Similar to CAPZ is keeping the generated test templates: https://github.com/kubernetes-sigs/cluster-api-provider-azure/tree/master/templates/test.

gab-satchi: yeah it seems both dev and ci config yamls have references to the cluster-template. It's not ideal but as long as the `make e2e` target is used, the file should get generated.

PR #1032, opened 2020-09-21
bump kube-vip to 0.1.8 and adjust leader election params

yastij:

**What this PR does / why we need it**: This PR bumps kube-vip to `0.1.8` and sets the leader election params to the default values

**Release note**:

```release-note
bump kube-vip to v0.1.8 and adjust leader election params
```

### DIFF
--- a/packaging/flavorgen/flavors/generators.go
+++ b/packaging/flavorgen/flavors/generators.go
@@ -361,7 +361,7 @@ func kubeVIPPod() string {
 			Containers: []v1.Container{
 				{
 					Name:  "kube-vip",
-					Image: "plndr/kube-vip:0.1.7",
+					Image: "plndr/kube-vip:0.1.8",
 					Args: []string{
 						"start",
 					},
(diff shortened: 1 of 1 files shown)


packaging/flavorgen/flavors/generators.go
```diff
@@ -398,6 +398,18 @@ func kubeVIPPod() string {
 							Name:  "vip_interface",
 							Value: "eth0",
 						},
+						{
```
gab-satchi, line 401: > sets the leader election params to the default values

I see different defaults [here](https://github.com/plunder-app/kube-vip/pull/73/files#diff-280a01902e7f52d4ee3966850378b8a7R42-R44). were the defaults in this PR meant to be different from the ones in kube-vip?

thebsdbox: Those above are the defaults for `kube-vip` the defaults for k\k are longer, which are what are being used here.

thebsdbox (approved):
Looks ok to me. Would be nice to see if there was a way of having T-shirt sizes for cluster size failover. But it's a convo for another day.

PR #1040, opened 2020-09-29
Refactor the error handling

### LINKED ISSUE #1004: Use pkg/errors `Is` to determine error types
Brought up originally in #1003, we have a bunch of private functions that check if an error is of a certain type. These functions can become tedious and cumbersome over time to maintain. The `pkg/errors` package has an `errors.Is` function exposed that can be used for this purpose, we should cleanup our internal codebase.

/kind cleanup

evalsocket:
Brought up originally in #1003, we have a bunch of private functions that check if an error is of a certain type. These functions can become tedious and cumbersome overtime to maintain. 

This pr replace our custom error function with pkg/errors
  
**Which issue(s) this PR fixes** :
Fixes #1004 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/services/govmomi/service.go
+++ b/pkg/services/govmomi/service.go
@@ -72,12 +72,12 @@ func (vms *VMService) ReconcileVM(ctx *context.VMContext) (vm infrav1.VirtualMac
 	// Before going further, we need the VM's managed object reference.
 	vmRef, err := findVM(ctx)
 	if err != nil {
-		if !errors.As(err, find.NotFoundError{}) {
+		if !errors.As(err, errNotFound{}) {
 			return vm, err
 		}
 
 		// If the machine was not found by BIOS UUID it means that it got deleted from vcenter directly
-		if errors.As(err, find.NotFoundError{}) {
+		if !errors.As(err, find.NotFoundError{}) {
 			ctx.VSphereVM.Status.FailureReason = capierrors.MachineStatusErrorPtr(capierrors.UpdateMachineError)
 			ctx.VSphereVM.Status.FailureMessage = pointer.StringPtr(fmt.Sprintf("Unable to find VM by BIOS UUID %s. The vm was removed from infra", ctx.VSphereVM.Spec.BiosUUID))
 			return vm, err
--- a/pkg/services/govmomi/service.go
+++ b/pkg/services/govmomi/service.go
@@ -19,10 +19,10 @@ package govmomi
 import (
 	"encoding/base64"
 	"fmt"
-	"github.com/vmware/govmomi/find"
 
 	"github.com/pkg/errors"
 
+	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/property"
 	"github.com/vmware/govmomi/vim25/mo"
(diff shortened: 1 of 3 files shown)


pkg/services/govmomi/service.go
```diff
@@ -71,15 +72,17 @@ func (vms *VMService) ReconcileVM(ctx *context.VMContext) (vm infrav1.VirtualMac
...
+		if !errors.As(err,find.NotFoundError{}) {
 			return vm, err
 		}
 
 		// If the machine was not found by BIOS UUID it means that it got deleted from vcenter directly
-		if wasNotFoundByBIOSUUID(err) {
-			ctx.VSphereVM.Status.FailureReason = capierrors.MachineStatusErrorPtr(capierrors.UpdateMachineError)
-			ctx.VSphereVM.Status.FailureMessage = pointer.StringPtr(fmt.Sprintf("Unable to find VM by BIOS UUID %s. The vm was removed from infra", ctx.VSphereVM.Spec.BiosUUID))
-			return vm, err
+		if errors.As(err,find.NotFoundError{}) {
```
gab-satchi, line 80: don't need this extra IF as if it gets past line 75, this condition would be true

pkg/services/govmomi/service.go
```diff
@@ -71,15 +72,17 @@ func (vms *VMService) ReconcileVM(ctx *context.VMContext) (vm infrav1.VirtualMac
 	// Before going further, we need the VM's managed object reference.
 	vmRef, err := findVM(ctx)
 	if err != nil {
-		if !isNotFound(err) {
+		if !errors.As(err,find.NotFoundError{}) {
```
gab-satchi, line 75: We create a `errNotFound` when the err from the govomi lookup [comes back nil](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/pkg/services/govmomi/util.go#L61-L67) but the object reference is not set. With the original implementation, that would fall through this condition. With the new changes, it looks like it would return early as errNotFound won't have a `find.NotFoundError` in its chain.

pkg/services/govmomi/service.go
```diff
@@ -159,7 +162,7 @@ func (vms *VMService) DestroyVM(ctx *context.VMContext) (infrav1.VirtualMachine,
 	if err != nil {
 		// If the VM's MoRef could not be found then the VM no longer exists. This
 		// is the desired state.
-		if isNotFound(err) || isFolderNotFound(err) {
+		if errors.As(err,find.NotFoundError{}) || errors.As(err,find.NotFoundError{}) {
```
gab-satchi, line 165: Should this be `if errors.As(err, errNotFound{}) || errors.As(err,find.NotFoundError{}) {` ?

gab-satchi (commented):
Thanks for the changes. It looks like the linter is failing as well. You can check locally with a `make lint`

yastij:
@evalsocket - seems like there's a panic, see: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/1040/pull-cluster-api-provider-vsphere-e2e/1311676884823576576/artifacts/clusters/bootstrap/controllers/capv-controller-manager/capv-controller-manager-74597b647c-vwmrw/manager.log

srm09:
/remove-lifecycle stale
@yastij Is this still relevant?

yastij:
@srm09  Yes it is

PR #1046, opened 2020-10-19
re-order the deletion process to delete the haproxylb after the machines

### LINKED ISSUE #1033: Delete load balances should not start if there are still VSphereMachines
/kind bug

as per https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1019#discussion_r491992765

In reconcile delete, we are checking if there are pending VSphereMachines only _after_ deleting the load balancer, but in that case, the user won't be able to operate those machine due to the missing load balances.

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/f3012ec2b567a73f74ded82f20cf8f27a881f23b/controllers/vspherecluster_controller.go#L243-L246

Instead, we should have this test up in the same func so we don't start deleting the load balancer if there are still VSphereMachines (except that machine backing the load balancer itself)

yastij:

**What this PR does / why we need it**: This PR re-order the deletion process to delete the `HAProxyLoadBalancer` after the `vsphereMachines` are deleted

**Which issue(s) this PR fixes**: Fixes #1033

/assign @fabriziopandini 

**Release note**:

```release-note
re-order the deletion process to delete the haproxylb after the machines
```

### DIFF
--- a/controllers/vspherecluster_controller.go
+++ b/controllers/vspherecluster_controller.go
@@ -220,6 +220,12 @@ func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconci
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
+	if len(vsphereMachines) > 0 {
+		ctx.Logger.Info("Waiting for VSphereMachines to be deleted", "count", len(vsphereMachines))
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
 	if len(haproxyLoadbalancers.Items) > 0 {
 		conditions.MarkFalse(ctx.VSphereCluster, infrav1.LoadBalancerAvailableCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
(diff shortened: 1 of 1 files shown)


controllers/vspherecluster_controller.go
```diff
@@ -220,6 +220,12 @@ func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconci
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
+	if len(vsphereMachines) > 0 {
```
fabriziopandini, line 224: @yastij 
Background question. Is the VM backing the HAProxy included in `vsphereMachines`? if so this will hang indefinitely 😅

yastij: @fabriziopandini - no it's not included, `VSphereMachines` represent only Cluster API infra machines, anything that needs a VM generates a `VSphereVM`, so this check shouldn't block

fabriziopandini: thanks for the clarification!

yastij (commented):
/approve

PR #1059, opened 2020-12-03
move CPI over to CRS and refactor flavorgen

yastij:
Signed-off-by: Yassine TIJANI <ytijani@vmware.com>

**What this PR does / why we need it**: This PR moves our templates to deploying the CPI through CRS and refactors flavorgen for better re-usability across its packages

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
deploy CPI through CRS instead of using the built-in APIs
```

PR #1045, opened 2020-10-13
add the v1alpha4 API changes for CAPV

yastij:
Signed-off-by: Yassine TIJANI <ytijani@vmware.com>

**What this PR does / why we need it**: This PR adds the initial API changes that are going to be needed for v1alpha4, the PR isn't exhaustive, and we can update it as we start to add/remove fields/features

**Which issue(s) this PR fixes**: Fixes #

**Special notes for your reviewer**:

/assign @fabriziopandini @vincepri 

**Release note**:

```release-note

```

PR #1060, opened 2020-12-04
WIP: deploy CPI in same minor version as Kubernetes

gwang550:
This PR aims to solve the problem of deploying the CPI version to the minor version of K8s
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/packaging/flavorgen/flavors/generators.go
+++ b/packaging/flavorgen/flavors/generators.go
@@ -35,7 +35,6 @@ import (
 const (
 	clusterNameVar               = "${CLUSTER_NAME}"
 	controlPlaneMachineCountVar  = "${CONTROL_PLANE_MACHINE_COUNT}"
-	defaultCloudProviderImage    = "gcr.io/cloud-provider-vsphere/cpi/release/manager:v1.2.1"
 	defaultClusterCIDR           = "192.168.0.0/16"
 	defaultDiskGiB               = 25
 	defaultMemoryMiB             = 8192
@@ -104,6 +103,9 @@ var (
 		regexVar(vSphereTemplateVar),
 		regexVar(vSphereHaproxyTemplateVar),
 	}
+	k8sVersionToCPIVersion = map[string]string{
+		"v1.18.6":"gcr.io/cloud-provider-vsphere/cpi/release/manager:v1.2.1",
+	}
 )
 
 func regexVar(str string) string {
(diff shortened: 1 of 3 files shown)


packaging/flavorgen/flavors/generators.go
```diff
@@ -104,6 +103,9 @@ var (
 		regexVar(vSphereTemplateVar),
 		regexVar(vSphereHaproxyTemplateVar),
 	}
+	k8sVersionToCPIVersion = map[string]string{
+		"v1.18.6":"gcr.io/cloud-provider-vsphere/cpi/release/manager:v1.2.1",
```
yastij, line 107: as you see in #1059 we're moving away from baking the CPI into CAPV

@andrewsykim @dvonthenen - would it make sense to move/re-tag the CPI with minor kubernetes version ? if not, at worst consumers may maintain that mapping

andrewsykim: > would it make sense to move/re-tag the CPI with minor kubernetes version ? 

Yes, that is what this PR is trying to do. Going forward we are always tracking the Kubernetes minor versions, so `v1.18.6` needs to map to v1.18.0 (when released). However, the patch versions may not match so we still need to map a given Kubernetes minor version with the latest CPI patch version. See [this KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cloud-provider/1771-versioning-policy-for-external-cloud-providers) for more details. 

> if not, at worst consumers may maintain that mapping

I'm not sure we want to put this burden on users to be honest. I think the CRS approach is reasonable, as long as we do the versioning mapping there. @yastij would it be reasonable for @gwang550 to have a follow-up PR to update the generated CRS code when yours is merged? When are we expecting that to merge and are we going to remove the baked in deployment right after? If we think the baked in deployment is staying for a while it's valuable to make the necessary changes here, if we are replacing it entirely with CRS, let's have a follow-up PR for the CRS changes.

yastij: #1045 outlines the timeline for CPI and CSI

@andrewsykim - I expect the PR to merge today, I just need to wrap up a few things. Note that consumers that don't generate templates programmatically through something similar to https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/tree/master/packaging/flavorgen might not benefit from @gwang550's PR

andrewsykim: Not this specific PR, but we need to add logic in the template generate to pick the right version

gwang550: @yastij Let me if there is something that I need to pick up following your PR. In the meantime, I am going to look into how to corporate the logic into your PR

PR #1064, opened 2020-12-14
support storage policies in CAPV

### LINKED ISSUE #1043: Specify storage policy for VM
/kind feature

The vSphereMachineTemplate resource should allow a user to specify the storage policy for the VM. 

This can be accomplished today by specifying a vm template that uses the desired storage policy.  However, some users may want more flexibility.

**Environment:**

yastij:

**What this PR does / why we need it**: This PR introduces the support of storage policies within CAPV. The current plan is to enforce with webhooks that datastore shouldn't be set at the same time as storage policy name

**Which issue(s) this PR fixes**: Fixes #1043 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/services/govmomi/vcenter/clone.go
+++ b/pkg/services/govmomi/vcenter/clone.go
@@ -163,7 +163,6 @@ func Clone(ctx *context.VMContext, bootstrapData []byte) error {
 			MemoryMB:          memMiB,
 		},
 		Location: types.VirtualMachineRelocateSpec{
-			//	Datastore:    types.NewReference(datastore.Reference()),
 			DiskMoveType: string(diskMoveType),
 			Folder:       types.NewReference(folder.Reference()),
 			Pool:         types.NewReference(pool.Reference()),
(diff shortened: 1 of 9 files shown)


yastij (commented):
/approve
/assign @srm09

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -107,11 +108,6 @@ func Clone(ctx *context.VMContext, bootstrapData []byte) error {
 		return errors.Wrapf(err, "unable to get folder for %q", ctx)
 	}
 
-	datastore, err := ctx.Session.Finder.DatastoreOrDefault(ctx, ctx.VSphereVM.Spec.Datastore)
-	if err != nil {
-		return errors.Wrapf(err, "unable to get datastore for %q", ctx)
-	}
-
```
srm09, line 114: This would cause the datastore to not be set if there is no `StoragePolicyName` specified. Is that intentional?

PR #1072, opened 2021-01-06
[WIP] rate limit tasks to avoid overloading vcenter

### LINKED ISSUE #1070: rate-limit vcenter tasks for vsphereVMs
/kind feature

**Describe the solution you'd like**: given that today CAPI doesn't scale/create machines gradually, CAPV should rate-limit the creation of tasks, to avoid overloading vcenter. This is especially true for a large number of workers (e.g. 1k node clusters).

**Environment:**

- Cluster-api-provider-vsphere version:  0.7.1

yastij:

**What this PR does / why we need it**: This PR adds rate limiting to the tasks being queued, this avoids overloading vcenter when creating large scale clusters

**Which issue(s) this PR fixes**: Fixes #1070 

**Release note**:

```release-note
rate-limit task queueing for large clusters
```

### DIFF
--- a/pkg/services/govmomi/service.go
+++ b/pkg/services/govmomi/service.go
@@ -37,11 +37,18 @@ import (
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/extra"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/net"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/task"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/util"
 )
 
+const (
+	maxTaskCount = 80
+)
+
 // VMService provdes API to interact with the VMs using govmomi
-type VMService struct{}
+type VMService struct {
+	TaskCounter *task.Counter
+}
 
 // ReconcileVM makes sure that the VM is in the desired state by:
 //   1. Creating the VM if it does not exist, then...
@@ -66,7 +73,7 @@ func (vms *VMService) ReconcileVM(ctx *context.VMContext) (vm infrav1.VirtualMac
 	// VSphereVM resource once its associated task completes. If
 	// there is no task for the VSphereVM resource then no reconcile
 	// event is triggered.
-	defer reconcileVSphereVMOnTaskCompletion(ctx)
+	defer reconcileVSphereVMOnTaskCompletion(ctx, vms.TaskCounter)
 
 	// Before going further, we need the VM's managed object reference.
 	vmRef, err := findVM(ctx)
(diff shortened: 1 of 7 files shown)


pkg/services/govmomi/service.go
```diff
@@ -37,11 +37,18 @@ import (
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/extra"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/net"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/task"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/util"
 )
 
+const (
+	maxTaskCount = 80
```
fabriziopandini, line 45: we should make this value configurable, but given that the rate-limiting for now is a not concern a concern for the majority of the users it is not required to included this in this PR (it a low priority/backlog task)

pkg/services/govmomi/service.go
```diff
@@ -37,11 +37,18 @@ import (
...
+)
+
 // VMService provdes API to interact with the VMs using govmomi
-type VMService struct{}
+type VMService struct {
+	TaskCounter *task.Counter
```
fabriziopandini, line 50: This is definitely a good start for introducing rate-limiting.

However, I think we should document the scope of this task counter, which, if I got this right, is responsible for rate-limiting everything managed by this controller.

In the future might be we should be required to improve this by making CAPV aware of the overall task pressure of the underlying vSphere infrastructure (a vSphere infrastructure can have pressure from something external from CAPV), otherwise, this mechanism could not limit enough depending on the conditions.

Vice versa, CAPV pressure could be spread on many vSphere infrastructures so this mechanism could be limiting too much depending on the conditions.

Documenting the current scope could help us in managing future improvements if required.

fabriziopandini (commented):
SGTM to me, only two comments/idea for future improvements, but nothing blocking this pr to move on

jzhoucliqr:
another point to reduce the vcenter load is about the session validation. Currently for all the operation capv need to first call vcenter to validate the session, which create lots of loads and cause vcenter to return 503. We've added a session keepalive support, to only send a ping pang message in a configured interval, instead of calling vcenter to validate the session for every api call. 
we can make the interval configurable and submit a PR if you guys think this could be useful:  
https://github.com/spectrocloud/cluster-api-provider-vsphere/pull/8/files

srm09:
@jzhoucliqr I like the idea around not validating the session for every vCenter operation. Instead keeping the session alive by sending a heartbeat at a configurable interval might be helpful in reducing the number of API requests to vCenter. I looked at the PR that you linked to in the fork, had a few suggestions on it:
1. Using the [session.KeepAliveHandler](https://pkg.go.dev/github.com/vmware/govmomi/session@v0.23.1#KeepAliveHandler) as the handler function for keepAlive requests. We could use this handler to re-login in case of failures, instead of just invalidating the cache.
2. Making the keepAlive behavior configurable and exposing the configuration under a feature flag with sensible defaults, maybe just defaulting the behavior as false to start with.
3. Instead of having two different methods `GetOrCreateKeepAliveSession` and `GetOrCreateNonKeepAliveSession` with a lotta duplication, adding a condition around the creation of the soap client would be a bit more cleaner.

Let me know if these suggestions make sense.

jzhoucliqr:
@srm09 Thanks for the suggestions, make sense to me. The change in the fork was just a quick dirty fix.

srm09:
@jzhoucliqr Did you get time to open a PR with the changes from the spectrocloud fork?

jzhoucliqr:
> @jzhoucliqr Did you get time to open a PR with the changes from the spectrocloud fork? If you are tied with things, then I can help with those changes.

@srm09 Will open PR for this soon!

PR #1073, opened 2021-01-11
Adds escaping logic for vCenter credentials

### LINKED ISSUE #1071: username and passwords should be escaped
/kind bug

**What steps did you take and what happened:** currently credentials aren't escaped, this should be fixed when a user provides `DOMAIN\USER` username for example

**Environment:**

- Cluster-api-provider-vsphere version: 0.7.1

srm09:
This patch escapes the vCenter credentials while reading them from the credential file. This makes sure that samAccountUsername formats work as well.

Fixes #1071

n/a

**Release note**:
```release-note
Adds escaping logic for vCenter credentials
```

### DIFF
--- a/config/default/credentials.yaml
+++ b/config/default/credentials.yaml
@@ -6,5 +6,5 @@ metadata:
 type: Opaque
 stringData:
   credentials.yaml: |-
-    username: "${VSPHERE_USERNAME}"
-    password: "${VSPHERE_PASSWORD}"
+    username: '${VSPHERE_USERNAME}'
+    password: '${VSPHERE_PASSWORD}'
--- a/pkg/manager/options_test.go
+++ b/pkg/manager/options_test.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


pkg/manager/options.go
```diff
@@ -138,8 +139,8 @@ func (o *Options) defaults() {
 
 	if o.Username == "" || o.Password == "" {
 		credentials := o.getCredentials()
-		o.Username = credentials["username"]
-		o.Password = credentials["password"]
+		o.Username = escapeCredentials(credentials["username"])
```
yastij, line 142: actually, while playing with this PR I remembered that single quoting here https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/config/default/credentials.yaml#L9 would do the trick. so that might be simpler than what we were thinking

randomvariable: Yes, this approach seems a bit brittle. We have funny regex escaping for the ini files because INI files are not defined in a standard and the Go library that was being used had a bespoke interpretation. Otherwise we should rely on the YAML specification or the appropriate library escaping, otherwise we'll be forever chasing edge cases.

yastij: fun thing: printing/logging actually escapes before printing, so it can be difficult to know what's ends up in memory

srm09: >fun thing: printing/logging actually escapes before printing, so it can be difficult to know what's ends up in memory

@yastij This is what I was talking about on Slack yesterday. I will update the PR with the single quotes and modify the test.

yastij: ack, thank you @srm09 !

yastij (approved):
/approve
/lgtm

PR #1077, opened 2021-01-13
update README with new OVA links

codenrhoden:
Latest OVAs have been uploaded to GCS. Update README to show them. Removes link to 1.16 since it is EOL.

Since there are currently 4 K8s versions that are active, keeps 1.17 and adds 1.20.

Fixes #

At some point we should start linking Ubuntu 20.04 instead of 18.04.

**Release note**:

```release-note

```

/assign @yastij


yastij (approved):
/approve
/lgtm

MaxRink:
> At some point we should start linking Ubuntu 20.04 instead of 18.04.

that switch might be a good time to make the switch to EFI.
I know @randomvariable has done some work in that regard already.

PR #1115, opened 2021-02-18
Adds new reason for VMProvisionedCondition

### LINKED ISSUE #1116: VM Provisioning halts on incorrect conditions
/kind bug

Creation of a VSphereVM with multiple virtual network devices waits for static IP address allocation even though when one of the devices has DHCP set as true.

Steps:
1. Create a VSphereVM with multiple virtual network devices
2. Ensure one of the virtual network specs has `DHCP4` set to `true`
3. Control plane or worker VMs are never provisioned.
4. Combing through the logs shows the following:
`vspherevm_controller.go:321] capv-controller-manager/vspherevm-controller/default/multinw-192-gxnpn "msg"="vm is waiting for static ip to be available"`

VSphereVMs should be created successfully.

FYI [here is the proposal](https://docs.google.com/document/d/10I-HH67_xBk73pazhNmbJjOvtPgq9Uk6ZrD3vj06srs/edit#heading=h.3xhbc92zb407) that introduced these changes in the first place:

**Environment:**

- Cluster-api-provider-vsphere version: `0.7.4`
- Kubernetes version: (use `kubectl version`): `v1.19.1`
- OS (e.g. from `/etc/os-release`):  `ubuntu`

srm09:
This patch introduces a new reason to indicate that the `VMProvisionedCondition` is halted because of static IP Address allocation.

Fixes #1116 

n/a

**Release note**:
```release-note
Adds a new reason for VMProvisionedCondition
```

### DIFF
--- a/controllers/vspherevm_controller.go
+++ b/controllers/vspherevm_controller.go
@@ -214,7 +214,7 @@ func (r vmReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error) 
 		// newer than the local resource.
 		// nolint:errcheck
 		wait.PollImmediateInfinite(time.Second*1, func() (bool, error) {
-			// remoteObj refererences the same VSphereVM resource as it exists
+			// remoteObj references the same VSphereVM resource as it exists
 			// on the API server post the patch operation above. In a perfect world,
 			// the Status for localObj and remoteObj should be the same.
 			remoteObj := &infrav1.VSphereVM{}
@@ -318,6 +318,7 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
 	var vmService services.VirtualMachineService = &govmomi.VMService{}
 
 	if r.isWaitingForStaticIPAllocation(ctx) {
+		conditions.MarkFalse(ctx.VSphereVM, infrav1.VMProvisionedCondition, infrav1.WaitingForStaticIPAllocationReason, clusterv1.ConditionSeverityInfo, "")
 		ctx.Logger.Info("vm is waiting for static ip to be available")
 		return reconcile.Result{}, nil
 	}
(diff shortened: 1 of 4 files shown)


yastij:
cc @jzhoucliqr PTAL

controllers/vspherevm_controller.go
```diff
@@ -363,16 +364,20 @@ func (r vmReconciler) reconcileNormal(ctx *context.VMContext) (reconcile.Result,
 	return reconcile.Result{}, nil
 }
 
+// isWaitingForStaticIPAllocation checks whether the VM should wait for a static IP
+// to be allocated.
+// It checks the state of both DHCP4 and DHCP6 for all the network devices and if
+// any static IP addresses are specified.
```
jzhoucliqr, line 370: Am just wondering, when one of the NIC needs static IP, If the vm got created without the static IP, how will the static IP to be set into the VM?  Do we expect some additional agent running within the VM to configure the additional NIC?

srm09: Hey @jzhoucliqr , I looked at the proposal [here](https://docs.google.com/document/d/10I-HH67_xBk73pazhNmbJjOvtPgq9Uk6ZrD3vj06srs/edit#heading=h.iame4w8fq7my) and the idea was that we would be using an external IPAM for IP allocation and this address allocation logic is not to be included in the CAPV codebase. Hence made the changes in this way. Does that answer your question?

jzhoucliqr: Thanks @srm09 . The allocation is done by external IPAM, but to configure the IP within the VM is done by cloud-init, and we need to pass the IP info into the VM using guestinfo.metadata. So that means we'll need to wait for all IPs to be allocated then clone the VM, and set the correct static IP info.

yastij: just took a look at this, @jzhoucliqr is correct we need to have the static IPs in place before creating the VM. I think we can close this PR

yastij:
opening back, @srm09 can you update the PR to only keep the new reason when waiting for a static IP ?

controllers/vspherevm_controller.go
srm09: Thanks @yastij and @jzhoucliqr for providing additional context

srm09:
Since this is a new reason for the Condition, I am gonna change the base to `master` and cherry-pick it back to `release-0.7` branch if needed

yastij (approved):
/approve

@srm09 - can you cherry-pick to `release-0.7`

PR #1118, opened 2021-02-23
Utilize datastores found via storage policy

gab-satchi:
With these changes, we look for compatible datastores to use when a storagepolicy is defined.

- If storage policy and datastore is defined, it validates the datastore is contained in the compatible list of datastores for the storage policy
- Errors if storage policy returns an empty list of compatible datastore
- Falls back to using the default datastore if neither datastore or storage policy is defined.

**Release note**:
```release-note
NONE
```


yastij (approved):
/approve
/lgtm

PR #1133, opened 2021-03-05
fix panic and randomize chosen datastore when using storage policies

yastij:

**What this PR does / why we need it**: This PR fixes a panic where datastoreRef was nil that was introduced in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1118, it also randomize the picking of the compatible datastore, to avoid blocking when the first compatible datastore is full

**Release note**:

```release-note

```


yastij (commented):
/approve

gab-satchi:
thanks for the fix. 

/lgtm

PR #1141, opened 2021-03-10
Adds storage policy to e2e tests

gab-satchi:
Adds e2e test that: 
- creates tag, category and storage policy
- creates machine template that utilizes the storage policy

**Release note**:
```release-note
None
```


test/e2e/data/infrastructure-vsphere/cluster-template-storage-policy.yaml
```diff
@@ -0,0 +1,209 @@
+apiVersion: v1
```
yastij, line 1: is there a specific reason to have this file instead of using `packaging/flavorgen` ?

yastij (commented):
will do a more complete pass tomorrow

test/e2e/data/infrastructure-vsphere/cluster-template-storage-policy.yaml
gab-satchi: Not really, was a pattern when working with CAPA e2e. Just needed a 'flavour' without the VsphereMachineTemplate. Should I just use `MultiNodeTemplateWithHAProxy` and then take out machine template?

yastij: is there a reason why we don't need the `VSphereMachineTemplate` ?

gab-satchi: We definitely need it but the storage-policy name is determined at run time so the vsphereMachineTemplate is applied by the test once the storage-policy is set up.

yastij: Ideally we should just consume a storage policy that is already created in our environment, this way we can re-use the same one across jobs, without having to generate it at runtime and deal with the naming issues.

gab-satchi: ya could do that. Went with creating and cleaning up per test so the user would have less setup to do. And the validation becomes simpler when it's a new storage policy rather than an existing one (can just make sure at least 1 vm is using it rather than checking against the vm-id is in the list of objects using the policy).

With using an existing sp, we'd still need a new flavour that specifies the storage-policy in the machine template. Or do you think I should just modify the quick start to use storage policy as well? We should cover both scenarios with tests imo.

test/e2e/capv_quick_start_test.go
```diff
@@ -37,3 +52,187 @@ var _ = Describe("Cluster Creation using Cluster API quick-start test", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+		}
+		Expect(len(res)).To(BeNumerically(">", 0))
+
+		// Delete storage policy, tag and category
+		if pbmClient != nil {
+			_, _ = pbmClient.DeleteProfile(ctx, []types.PbmProfileId{*spID})
+		}
+		_ = tagsManager.DeleteTag(ctx, &tag)
+		_ = tagsManager.DeleteCategory(ctx, &cat)
```
srm09, line 138: Call them with defer to make sure they get run at the end of the method. Since there is a return after a failure.

Also, even if the `pbmClient` fails to exist, we should enforce deletion of the profile.

```golang 
..
defer tagsManager.DeleteTag(ctx, &tag)
defer tagsManager.DeleteCategory(ctx, &cat)
```

test/e2e/data/infrastructure-vsphere/cluster-template-storage-policy.yaml
srm09: I prefer the programmatic way of doing the setup for creating a storage policy, which makes sure that there is no extra setup that needs to be done before running the test.

srm09 (commented):
minor nits around resource cleanup for tests

test/e2e/data/infrastructure-vsphere/cluster-template-storage-policy.yaml
yastij: I'd rather add changes to `flavorgen` to generate both (sp-only, and sp with datastore) and use an existing storage policy. which would allow to get rid of the file + test against an existing storage policy

gab-satchi: removed file and added storage policy to haproxy flavour.

test/e2e/capv_quick_start_test.go
```diff
@@ -37,3 +50,101 @@ var _ = Describe("Cluster Creation using Cluster API quick-start test", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+		}
+		Expect(len(res)).To(BeNumerically(">", 0))
+
+		vms := getVSphereVMsForCluster(clusterName, namespace.Name)
+		Expect(len(vms.Items)).To(BeNumerically(">", 0))
+
+		datacenter, err := vsphereFinder.DatacenterOrDefault(ctx, vsphereDatacenter)
+		Expect(err).ShouldNot(HaveOccurred())
+		By("verifying storage policy is used by VMs")
```
gab-satchi, line 103: the query for objects using the storage policy returns server UID along with object reference ID. The vsphereVM spec has the BIOS UID which is different so forced to fetch the vm again and compare object references. let me know if there's a simpler way I missed

gab-satchi:
need to rebase on top of our main branch 

/hold

yastij (approved):
/approve
/lgtm

PR #1138, opened 2021-03-10
Updates references to v1alpha4 types

srm09:
**What this PR does / why we need it**:
The CAPI module in use now points to the latest on the `master` branch of CAPI. This patch updates the references from `v1alpha3` to `v1alpha4` types.
This patch has the following changes:
- Use the latest `master` from the CAPI repo.
- Updates controller-runtime to version `v0.8.2`
- Updates golang version to `v1.15.8`
- Updates the CAPV codebase to reference v1alpha4 types.
- Uses the CAPI nightly builds for E2E tests with v1alpha4 types

**Which issue(s) this PR fixes**:
n/a

**Special notes for your reviewer**:
n/a

**Release note**:
```release-note
Introduces support for v1alpha4 types
```

PR #1144, opened 2021-03-11
Adds support for controller tests

### LINKED ISSUE #1142: Add setup for controller tests
/kind feature

This issue tracks adding the setup code to add Controller tests leveraging the `envtest` model from CAPI and controller-runtime.

n/a

**Environment:**
n/a

srm09:
This patch introduces the setup code for introducing controller tests in the CAPV codebase.

Fixes #1142 

Depends on #1138

**Release note**:

```release-note
NONE
```


hack/fetch_ext_bins.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
```
gab-satchi, line 1: I see CAPA doing the same thing. Is this the only way to setup the test binaries? We also have the `/hack/tools` directory we use to go build binaries into

yastij (approved):
/approve

lgtm pending @gab-satchi's comment

hack/fetch_ext_bins.sh
srm09: The only concern here would be how resource intensive it is to `go build` these binaries locally. And if it takes more time than downloading the binaries, then sticking to this option would make more sense.
I can start an exploratory issue and continue some work there and let  this one merge for now. So that we can start accepting new PRs for more controller tests.

gab-satchi: thanks Sagar. Was more curious than anything.

PR #1147, opened 2021-03-15
Reconcile storage policy after cloning the VM

### LINKED ISSUE #1146: Storage policy applied before cloning causes clone to fail across datastores
/kind bug

Use a template that doesn't have a storage policy to spin up worker nodes on a different datastore with a storage policy.  The current code sets the new storage policy before cloning, and we receive the following logs in our vCenter logs:
`Changing or applying VM Storage Policies with Data Service capabilities during clone operations is disallowed. VM Storage Policies with Data Service capabilities can be assigned to the provisioned VM after the clone operation has been completed and before the VM has been powered on.`  Even though [this article](https://kb.vmware.com/s/article/78488) states that this occurs when the datastores have different settings, we verified that this occurs even when the datastores have all the same configuration settings (the only differences between our datastores are name and size) and the storage policy is compliant with both the source and destination datastore.  We also verified that this still happens when the template lives on the same datastore that the worker node would live on, which removes datastore differences from the equation.

I expected the template to be cloned to the new datastore with the storage policy applied.  One solution here, which is currently used by the terraform vsphere provider, is to clone the template without a storage policy and then switch to the new policy.  This does make the clone take longer, but solves this problem.

**Environment:**

- Cluster-api-provider-vsphere version: v0.7.6
- Kubernetes version: (use `kubectl version`): v1.18.12
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04 machines
- vCenter: 6.7

gab-satchi:
This PR fixes an issue where host based storage policies couldn't be applied to VMs during the clone process. With these changes, the storage policy is reconciled separately after the clone VM, reconcile network and reconcile metadata completes. 

Fixes #1146 

The reconcile function returns early if the VM is powered on. This is to avoid the scenario where the storage policies have already been reconciled and the VM is in a running state but the user either attaches extra disks or uses CSI to provision new disks with a different storage policy. The capv controller should not overwrite the storage policy on any disks other than the first 'root' one. 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/services/govmomi/service.go
+++ b/pkg/services/govmomi/service.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 
 	"github.com/pkg/errors"
+	"github.com/vmware/govmomi/pbm"
+	pbmTypes "github.com/vmware/govmomi/pbm/types"
 
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/property"
@@ -123,6 +125,10 @@ func (vms *VMService) ReconcileVM(ctx *context.VMContext) (vm infrav1.VirtualMac
 	}
 
 	if ok, err := vms.reconcileMetadata(vmCtx); err != nil || !ok {
+		return vm, err
+	}
+
+	if err := vms.reconcileStoragePolicy(vmCtx); err != nil {
 		return vm, err
 	}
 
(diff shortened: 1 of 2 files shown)


srm09 (commented):
minor nits, other lgtm

yastij (commented):
/approve

@gab-satchi - this passed the storage policy test. this looks overall good

Issue #1123, opened 2021-03-03
Multi-tenant support with a single manager instance


gab-satchi:
/kind feature

**Describe the solution you'd like**
CAPV manager currently requires credentials to a vcenter through its runtime options to operate. This issue is to track support for multi-tenancy through a single instance of the CAPV manager. 

There are currently less than ideal ways to deploy to multiple vcenters:
- Utilize the `Server` field in vsphere machine clone spec. This allows for machines to be deployed to vcenters other than the one the manager is configured for, but it requires the alternate vcenters to have the same credentials as the primary one. 
- Deploy multiple instances of the CAPV manager configured for each vcenter and have them watch a dedicated namespace. Not currently supported and may never be fully supported. CAPI's [multi-tenancy contract](https://master.cluster-api.sigs.k8s.io/developer/architecture/controllers/multi-tenancy.html) implicitly calls for a single manager instance. 

**Anything else you would like to add:**
Each cluster will need to specify credentials to be used to deploy the cluster, if it's something other than what the CAPV manager is configured for. This could be done through a reference to a secret. If a cluster doesn't specify credentials, CAPV can fall back to using the credentials configured in the manager.

PR #1149, opened 2021-03-18
First draft of multi-tenancy proposal

gab-satchi:
/kind design
Proposing a way to allow CAPV to accommodate multiple tenants within a single instance

#1123 

Thanks to the AWS and Azure folks for pioneering the multi-tenancy efforts in their respective providers. @devigned @randomvariable @nader-ziada @sedefsavas

### DIFF
--- a/docs/proposal/20210311-single-controller-multitenancy.md
+++ b/docs/proposal/20210311-single-controller-multitenancy.md
@@ -1,0 +1,317 @@
+# Single Controller Multitenancy
+
+```text
+---
+title: Single Controller Multitenancy
+authors:
+  - "@gab-satchi"
+reviewers:
+  - "@sedefsavas"
+  - "@davigned"
+  - "@nader-ziada"
+  - "@yastij"
+  - "@fabriziopandini"
+creation-date: 2021-03-11
+last-updated: 2021-03-11
+status: implementable
+see-also:
+  - https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/docs/proposal/20200506-single-controller-multitenancy.md
+  - https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/docs/proposals/20200720-single-controller-multitenancy.md
+---
+```
+
+## Table of Contents
+
+* [Single Controller Multitenancy](#single-controller-multitenancy)
+  * [Glossary](#glossary)
+  * [Summary](#summary)
+  * [Motivation](#motivation)
+    * [Goals](#goals)
+  * [Proposal](#proposal)
+    * [User Stories](#user-stories)
+      * [Story 1 - Deploying to multiple vCenters](#story-1---deploying-to-multiple-vcenters)
+      * [Story 2 - Deploying multiple clusters from a single account](#story-2---deploying-multiple-clusters-from-a-single-account)
+      * [Story 3 - Legacy behaviour](#story-3---legacy-behaviour)
+    * [Requirements](#requirements)
+      * [Functional Requirements](#functional-requirements)
+      * [Non-Functional Requirements](#non-functional-requirements)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+      * [Current State](#current-state)
+      * [Proposed Changes](#proposed-changes)
+      * [Controller Changes](#controller-changes)
+      * [Clusterctl Changes](#clusterctl-changes)
+    * [Security Model](#security-model)
+      * [Roles](#roles)
+      * [RBAC](#rbac)
+        * [Write Permissions](#write-permissions)
+        * [Namespace Restrictions](#namespace-restrictions)
+        * [CAPV Controller Requirements](#capv-controller-requirements)
+    * [Risks and Mitigations](#risks-and-mitigations)
+      * [Caching](#caching)
+  * [Alternatives](#alternatives)
+    * [Using only secrets to specify vSphere accounts](#using-only-secrets-to-specify-vsphere-accounts)
+      * [Benefits](#benefits)
+      * [Mitigations for current proposal](#mitigations-for-current-proposal)
+  * [Upgrade Strategy](#upgrade-strategy)
+  * [Additional Details](#additional-details)
+    * [Test Plan](#test-plan)
+    * [Graduation Criteria](#graduation-criteria)
+  * [Implementation History](#implementation-history)
+
+## Glossary
+
+* CAPV - An abbreviation of Cluster API Provider vSphere
+
+## Summary
+
+The CAPV controller is capable of managing infrastructure resources on a vCenter using the credentials it was provided during initialization. The credentials are provided via environment variables to clusterctl that get saved onto a secret that's used by the CAPV deployment.
+
+Credentials provided are used for the entire lifetime of the CAPV deployment which means a CAPV cluster can become broken if the provisioning CAPV deployment were to be reconfigred for another set of credentials.
+
+This proposal outlines new capabilities for CAPV that can assume different credentials, at runtime, on a per-cluster basis. The proposed changes will maintain backwards compatibility and maintain the existing behaviour without any extra user configuration.  
+
+## Motivation
+
+Larger organizations often need to separate management and workload spaces by utilizing separate credentials for each. The tooling (CAPV in this case) may run in the mangement account with the intent of provisioning infrastructure in the workload accounts. For CAPV to be most useful within these organizations, it will need to support multi-acocunt models.
+
+vSphere can also be deployed in edge scenarios. With CAPV's current capabilities, a management cluster will need to be deployed in each edge environment. With the new capabilities, a single management cluster can manage multiple edge deployments.  
+
+### Goals
+
+1. To enable VSphereCluster resources reconciliation using per cluster vSphere credentials.
+1. To allow sets of clusters to use the same set of vSphere credentials.
+1. To maintain backwards compatibility for users who do not intend to use these capabilities and want to continue specifying credentials through the CAPV deployment.
+
+## Proposal
+
+### User Stories
+
+#### Story 1 - Deploying to multiple vCenters
+
+Alex is an engineer in an organization that enforces strict vSphere account and environment architectures. They use a management vCenter and account for the management cluster where CAPV is running. Alex has a workload account in a dedicated vCenter.
+Alex can provision a new cluster in the workload account by creating Cluster API resources in the management cluster. The CAPV controller will utilize the workload credentials to provision the cluster in Alex's environment.
+
+#### Story 2 - Deploying multiple clusters from a single account
+
+Stacy works at an organization where the cloud admin provides them with a namespace to use on a management cluster. Stacy can deploy workload clusters without having to know or specify the account details. If Stacy tried to deploy into a namespace that isn't authorized to use the account, the cluster will fail to deploy. Stacy can still create clusters into their dev environments by providing the account details in a Secret.
+
+#### Story 3 - Legacy behaviour
+
+Erin is an engineer in a smaller less strict organization. They use a single vCenter and keep their management cluster up-to-date. Erin can create new vSphere clusters while omitting the `vSphereCluster.Spec.IdentityRef` field. The CAPV controller will use the credentials it was given at initialization to create new vSphere clusters.
+
+### Requirements
+
+#### Functional Requirements
+
+* FR1: CAPV MUST support credentials provided through a `VSphereClusterIdentity` and referenced by `VSphereCluster.IdentityRef` field (Stories 1,2)
+* FR2: CAPV MUST support credentials provided through a `Secret` and referenced by `VSphereCluster.IdentityRef` field (Story 1)
+* FR3: CAPV MUST support static credentials (Story 3)
+* FR4: CAPV MUST support clusterctl move scenarios
+* FR5: CAPV MUST prevent privilege escalation allowing users to create clusters in accounts they should not be able to (Story 2)
+
+#### Non-Functional Requirements
+
+* NFR1: Unit tests MUST exist for cluster and machine controllers that utilize the credentials
+* NFR2: e2e tests MUST exist for multi account scenarios
+
+### Implementation Details/Notes/Constraints
+
+#### Current State
+
+CAPV currently uses a session manager to create and cache sessions. Sessions are created (or retrieved from the cache) during cluster and vSphere VM reconcile loops.
+vSphere cluster uses the session as a sanity check to ensure connectivity. vSphere VM uses the session for VM lifecycle tasks on vcenter and the session is stored in a VMContext.
+
+```go
+type VMContext struct {
+    *ControllerContext
+    VSphereVM   *v1alpha3.VSphereVM
+    PatchHelper *patch.Helper
+    Logger      logr.Logger
+    Session     *session.Session
+}
+```
+
+#### Proposed Changes
+
+The proposed changes below allow users to specify the account to use for clusters during runtime. The credentials can be provided by two approaches:
+
+* Referencing a cluster scoped `VSphereClusterIdentity` to be used for credentials.
+* Referencing a `Secret` that contains the credentials in the same namespace as the cluster.
+
+Changed Resources
+
+* `VSphereCluster`
+
+New Resources
+
+* A cluster-scoped `VSphereClusterIdentity` represents the vCenter account to use for reconciliation. This type should also contain references to namespaces that are allowed to use the account.
+
+Changes to VSphereCluster
+A new field is added to the VSphereClusterSpec to reference the `VSphereClusterIdentity`. We intend to use a `VSphereIdentityReference` type, similar to that of `corev1.TypedLocalObjectReference` in order to ensure the only objects that can be referenced are either in the same namespace or are scoped globally.
+
+```go
+// VSphereClusterIdentity is the account to be used for vcenter actions
+type VSphereClusterIdentity struct {
+    metav1.TypeMeta `json:`",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    Spec VSphereClusterIdentitySpec `json:"spec,omitempty"`
+    Status VSphereClusterIdentityStatus `json:"status, omitempty"`
+}
+
+type VSphereClusterIdentitySpec struct {
+    // secretsRef references a secret in the CAPV controller namespace with the credentials to use
+    secretRef `json:"secretRef,omitempty"`
+
+    // AllowedNamespaces is used to identify which namespaces are allowed to use this account.
+    // Namespaces can be selected either by using a label selector.
+    // If this object is nil, no namespaces will be allowed.
+    // If this object is empty {}, clusters can use this identity from any namespace.
+    //
+    // +optional
+    AllowedNamespaces *AllowedNamespaces `json:"allowedNamespaces"`
+}
+
+type AllowedNamespaces struct {
+    // Selector is a standard Kubernetes LabelSelector.
+    // This is a standard Kubernetes LabelSelector, a label query over a set of resources.
+    // The result of matchLabels and matchExpressions are ANDed.
+    // +optional
+    Selector metav1.LabelSelector `json:"selector"`
+}
+
+type VSphereIdentityReference struct {
+    // Kind of the identity.
+    // +kubebuilder:validation:Enum=VSphereClusterIdentity,Secret
+    Kind string
+
+    // Name of the identity
+    // +kubebuilder:validation:MinLength=1
+    Name string
+}
+
+type VSphereClusterSpec struct {
+    ...
+    // +optional
+    IdentityRef *VSphereIdentityReference `json:"IdentityRef,omitempty"`
+}
+```
+
+#### Controller Changes
+
+* If IdentityRef is specified, and it references a `VSphereClusterIdentity`, the CRD is fetched and used to create a session
+* The secret referenced within `VSphereClusterIdentity` must be in the controller namespace
+* If IdentityRef is specified, and it references a `Secret`, the `Secret` is used to create a session. The secret must be in the same namespace as the `VSphereCluster`
+* The controller will add the VSphereCluster to OwnerReferences on a Secret that contains the account credentials
+* If IdentityRef is not specified, the controller will fallback to using credentials provided in the controller deployment
+* The session will be cached using the existing logic where the key is server + username + datacenter
+* The `IdentityRef` on a `VSphereCluster` will be mutable. Setting the `IdentityRef` to nil will cause the controller to fallback to the static credentials
+* The controller will fail to delete a `VSphereClusterIdentity` with an error if the account is used by `VSphereClusters`
+
+#### Clusterctl Changes
+
+Today, clusterctl move operates by tracking objectreferences within the same namespace, since we are now proposing to use cluster-scoped resources, we will need to add requisite support to clusterctl's object graph to track cluster-scoped resources that are used by the source cluster, and ensure they are moved. We will naively not delete cluster-scoped resources during a move, as they maybe referenced across namespaces. If a cluster uses a `Secret` for account credentials, the OwnerReference will get set by the controller and the `Secret` will be moved to the target cluster.
+
+### Security Model
+
+The intended RBAC model mirrors that for Service APIs:
+
+#### Roles
+
+For the purposes of this security model, 3 common roles have been identified:
+
+* **Infrastructure Provider**: The infrastructure provider (infra) is responsible for the overall environment that
+  the cluster(s) are operating in or the PaaS provider in a company.
+
+* **Management Cluster Operator**: The cluster operator (ops) is responsible for
+  administration of the Cluster API management cluster. They manage policies, network access,
+  application permissions.
+
+* **Workload Cluster Operator**: The workload cluster operator (dev) is responsible for
+  management of the cluster relevant to their particular applications .
+
+There are two primary components to the Service APIs security model: RBAC and namespace restrictions.
+
+#### RBAC
+
+RBAC (role-based access control) is the standard used for Kubernetes
+authorization. This allows users to configure who can perform actions on
+resources in specific scopes. RBAC can be used to enable each of the roles
+defined above. In most cases, it will be desirable to have all resources be
+readable by most roles, so instead we'll focus on write access for this model.
+
+##### Write Permissions
+
+|                              | VSphereClusterIdentity | Secret | Cluster |
+| ---------------------------- | ---------------------- | ------ | ------- |
+| Infrastructure Provider      | Yes                    | Yes    | Yes     |
+| Management Cluster Operators | Yes                    | Yes    | Yes     |
+| Workload Cluster Operator    | No                     | Yes    | Yes     |
+
+#### Namespace Restrictions
+
+* To prevent workload cluster operators from using cluster-scoped `VSphereClusterIdentity` that they should not be using, the `allowedNamespaces` will be used to dictate which namespaces are allowed to use the `VSphereClusterIdentity`.
+An empty slice indicates that the `VSphereClusterIdentity` can be used by any `VSphereCluster`
+
+#### CAPV Controller Requirements
+
+The CAPV controller will need to:
+
+* Populate conditions when cluster is misconfigured. `VSphereClusterIdentity` is not found or isn't compatible with the cluster due to namespace restrictions.
+* Not implement an invalid configuration. If a cluster is attempting to use the `VSphereClusterIdentity` from an invalid namespace, ignore it and indicate the issue through conditions.
+* Respond to changes in a `VSphereClusterIdentity` spec.
+
+### Risks and Mitigations
+
+#### Caching
+
+With multi-tenant support, a single CAPV instance may reconcile multiple clusters across many accounts.
+There's existing caching in the session manager that should be sufficient to handle the extra sessions that will get created.
+
+## Alternatives
+
+### Using only secrets to specify vSphere accounts
+
+The VSphereCluster will reference the Secret via an ObjectReference
+
+#### Benefits
+
+* Re-using secrets ensures encryption by default and provides a clear UX signal to end users that the data is meant to be secure
+* Keeps clusterctl move straightforward with the 1:1 cluster -> credential relationship
+
+#### Mitigations for current proposal
+
+* There are use cases where users would like to reuse the same account for multiple namespaces. See [Story 2](#story-2---deploying-multiple-clusters-from-a-single-account)
+
+## Upgrade Strategy
+
+The data changes are additive and optional. VSphereClusters that aren't configured with a `VSphereClusterIdentity` or `Secret` will default to the credentials initialized in the CAPV deployment.
+
+## Additional Details
+
+### Test Plan
+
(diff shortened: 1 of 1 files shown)


docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,255 @@
...
+    - [Graduation Criteria](#graduation-criteria)
+  - [Implementation History](#implementation-history)
+
+## Glossary
+- CAPV - An abbreviation of Cluster API Provider vSphere
+
+## Summary
+The CAPV controller is capable of managing infrastructure resources on a vCenter using the credentials it was provided during initialization. The credentials are provided via environment variables that get saved onto a secret that's used by the CAPV deployment. 
+
+Credentials provided are used for the entire lifetime of the CAPV deployment which means a CAPV cluster can become broken if the provisioning CAPV deployment were to be reconfigred for another set of credentials.
```
sedefsavas, line 63: How multitenancy can fix CAPV deployment credential issue? 
CAPA will start with credentials retrieved locally, then other roles could be used for creating workload clusters. 
When the local credentials (that CAPV uses) change, is the plan to update the secret?

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,255 @@
...
+    Password string `json:"password,omitempty"`
+
+    // AllowedNamespaces is an array of namespaces that VSphereClusters can
+    // use this account from.
+    //
+    // An empty list (default) indicates that VSphereClusters can use this
+    // account from any namespace. This field is intentionally not a
+    // pointer because the nil behavior (no namespaces) is undesirable here.
+    // +optional
+    AllowedNamespaces []string `json:"allowedNamespaces"`
```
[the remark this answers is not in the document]
nader-ziada: @sedefsavas we had some issues with implementing the label selector and that's why we only did the []string, but I do agree its good a consistent behaviour for the default across providers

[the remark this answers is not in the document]
gab-satchi: I like both the differences with CAPA. Was originally thinking label selector was overkill but I can see how that's useful with targeting multiple namespaces without having to list each of them out. Will update, thanks Sedef

docs/proposal/20210311-single-controller-multitenancy.md
gab-satchi: I don't think MT can fix any existing clusters from breaking when the deployment credentials are changed. This was more highlighting a cluster's dependance on the deployment credentials and if MT is fully utilized, it can move that dependency to a `VSphereAccount`

docs/proposal/20210311-single-controller-multitenancy.md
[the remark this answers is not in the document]
devigned: +1 to consistent behavior across providers.

What is the behavior when `NamespaceList` is nil and a selector is set? Is the result always the union of `NamespaceList` + `Selector`?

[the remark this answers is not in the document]
gab-satchi: yeah in CAPA's implementation, it is a union of the two. That'll offer the most flexibility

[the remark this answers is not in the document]
sedefsavas: empty `allowedNamespaces` --> allows all namespaces 
nil `allowedNamespaces` --> blocks all namespaces 

nil `NamespaceList` --> blocks all
nil `Selector` --> blocks all


> What is the behavior when NamespaceList is nil and a selector is set? Is the result always the union of NamespaceList + Selector?
Yes, if at least one of them matches a namespace, it is allowed.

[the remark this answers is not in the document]
sedefsavas: I drafted some examples for how we use allowedNamespaces in CAPA:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2319/files

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,295 @@
...
+
+    Spec VSphereAccountSpec `json:"spec,omitempty"`
+    Status VSphereAccountStatus `json:"status, omitempty"`
+}
+
+type VSphereAccountSpec struct {
+    // Server is the address of the vSphere endpoint.
+    Server string `json:"server,omitempty"`
+
+    Username string `json:"username,omitempty"`
```
yastij, line 159: remove `omitempty` ?

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,295 @@
...
+
+    Password string `json:"password,omitempty"`
```
yastij, line 161: same

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,295 @@
...
+
+    // AllowedNamespaces is used to identify which namespaces are allowed to use this account.
+    // Namespaces can be selected either by using an array of namespaces or with a label selector.
+    // If this object is nil, no namespaces will be allowed
```
yastij, line 165: are other providers failing closed too ?

yastij (commented):
did a first pass on the API

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,295 @@
+# Single Controller Multitenancy
+
+```text
+---
+title: Single Controller Multitenancy
+authors:
+  - "@gab-satchi"
+reviewers:
```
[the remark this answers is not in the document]
gab-satchi: ACK

docs/proposal/20210311-single-controller-multitenancy.md
gab-satchi: AWS is but Azure uses a string slice and if it's empty, will default to all namespaces are allowed.

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,295 @@
...
+
+The proposed changes below allow users to specify the account to use for clusters during runtime.
+
+Changed Resources
+
+- `VSphereCluster`
+
+New Resources
+
+- cluster-scoped `VSphereAccount` represents the vCenter account to use for reconciliation. This type should also contain a string list of namespaces in which it is allowed to be used.
```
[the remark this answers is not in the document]
gab-satchi: It's called out a bit in `Clusterctl Changes`. I don't think there's anything concrete in clusterctl regarding that yet. Some discussion captured here: https://github.com/kubernetes-sigs/cluster-api/issues/3354

We'd likely do something where we copy accounts (while avoiding duplicates) and almost never delete accounts as it can be used by other clusters or intended to be used for future clusters.

docs/proposal/20210311-single-controller-multitenancy.md
nader-ziada: I'm keep track of this to discuss with capz community

yastij: @randomvariable @sedefsavas -  IIRC CAPA also has namespaced resources for account management right ? so that might be a stop-gap until clusterctl supports the cluster-scoped ones

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,296 @@
...
+    // AllowedNamespaces is used to identify which namespaces are allowed to use this account.
+    // Namespaces can be selected either by using an array of namespaces or with a label selector.
+    // If this object is nil, no namespaces will be allowed
+    //
+    // +optional
+    AllowedNamespaces *AllowedNamespaces `json:"allowedNamespaces"`
+}
+
+type AllowedNamespaces struct {
+    // If both NamespaceList and Selector are nil/empty, no namespaces will be allowed to use the account (default behaviour)
```
fabriziopandini, line 173: What if both are specified? How do they combine (or or AND)?

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,296 @@
...
+type VSphereClusterSpec struct {
+    ...
+    // +optional
+    AccountRef *corev1.LocalObjectReference `json:"accountRef,omitempty"`
+}
+```
+
+#### Controller Changes
+
+- 1If AccountRef is specified, the CRD is fetched and used to create a session
```
fabriziopandini, line 194:
```suggestion
- If AccountRef is specified, the CRD is fetched and used to create a session
```

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,296 @@
...
+- The session will be cached using the existing logic where the key is server + username + datacenter
+- The controller should reconcile the VSphereAccount resource and ensure connectivity to the account
+- The controller will set the VsphereAccount as one of the OwnerReferences of the VsphereCluster if the VsphereAccount is in the same namespace as the VsphereCluster
```
fabriziopandini, line 197: I'm a little bit confused by this, my understanding till now was that VsphereAccount is a cluster scoped resource...

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,296 @@
...
+
+#### Clusterctl Changes
+
+Today, clusterctl move operates by tracking objectreferences within the same namespace, since we are now proposing to use cluster-scoped resources, we will need to add requisite support to clusterctl's object graph to track ownerReferences pointing at cluster-scoped resources, and ensure they are moved. We will naively not delete cluster-scoped resources during a move, as they maybe referenced across namespaces.
```
fabriziopandini, line 201: > clusterctl's object graph to track ownerReferences pointing at cluster-scoped resources

As far as I understood, the only link to VsphereAccount is AccountRef, and this makes things a little bit more complicated because the object tree is defined outside the ownerRefereces. 
We can define a solution to address this, but I really would like to come out with a design that works for all the providers, not only for CAPV
 
Also, what is the expected behavior if the cluster scope resource already exists in the target cluster? (override or ignore)

docs/proposal/20210311-single-controller-multitenancy.md
gab-satchi: It will be the union so OR. Will update to make that clearer

docs/proposal/20210311-single-controller-multitenancy.md
gab-satchi: It should be cluster scoped. CAPA called out cluster scoped vs. namespaced in the proposal but the implementation (so far) uses cluster scoped identities and doesn't set the owner reference. CAPZ seems to use namespaced crds for the AzureIdentity but also copies the identity per cluster.

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,298 @@
...
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    Spec VSphereAccountSpec `json:"spec,omitempty"`
+    Status VSphereAccountStatus `json:"status, omitempty"`
+}
+
+type VSphereAccountSpec struct {
+    Username string `json:"username"`
+
+    Password string `json:"password"`
```
jzhoucliqr, line 163: shall we put this username/password inside a secret?

docs/proposal/20210311-single-controller-multitenancy.md
randomvariable: Yes, the idea that you can enable both the cloud admin persona to set up cluster scoped resources for use by different teams, and then the devops and developer personas can also plug in their own dev environments using the namespace scoped resources.

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,298 @@
...
+New Resources
+
+- cluster-scoped `VSphereAccount` represents the vCenter account to use for reconciliation. This type should also contain a string list of namespaces in which it is allowed to be used.
+
+Changes to VSphereCluster
+A new field is added to the VSphereClusterSpec to reference the `VSphereAccount`. We intend to use `corev1.LocalObjectReference` in order to ensure the only objects that can be referenced are either in the same namespace or are scoped globally.
+
+```go
+// VSphereAccount is the account to be used for vcenter actions
+type VSphereAccount struct {
```
randomvariable, line 152:
```suggestion
type VSphereClusterAccount struct {
```

Cluster scoped resources should have "Cluster" in the name.

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,298 @@
...
+Changed Resources
+
+- `VSphereCluster`
+
+New Resources
+
+- cluster-scoped `VSphereAccount` represents the vCenter account to use for reconciliation. This type should also contain a string list of namespaces in which it is allowed to be used.
+
+Changes to VSphereCluster
+A new field is added to the VSphereClusterSpec to reference the `VSphereAccount`. We intend to use `corev1.LocalObjectReference` in order to ensure the only objects that can be referenced are either in the same namespace or are scoped globally.
```
randomvariable, line 148: We've created our own PrincipalReference type in CAPA rather than reuse TypedLocalObjectReference.

docs/proposal/20210311-single-controller-multitenancy.md
randomvariable: Yes, the actual credential should still be a SecretRef

docs/proposal/20210311-single-controller-multitenancy.md
randomvariable: >  the more it makes sense to have a 1:1 relationship with a cluster and account
I think this will defeat the purpose somewhat, as it limits the personas who can create clusters.

docs/proposal/20210311-single-controller-multitenancy.md
yastij: @gab-satchi - this should be a stop-gap until we know how to move cluster-scoped resources

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,296 @@
...
+}
+
+type VSphereClusterSpec struct {
+    ...
+    // +optional
+    AccountRef *corev1.LocalObjectReference `json:"accountRef,omitempty"`
+}
+```
+
+#### Controller Changes
```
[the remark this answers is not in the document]
gab-satchi: updated Controller Changes to include suggestions.

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+New Resources
+
+- cluster-scoped `VsphereClusterAccount` represents the vCenter account to use for reconciliation. This type should also contain references to namespaces that are allowed to use the account.
+
+Changes to VsphereCluster
+A new field is added to the VsphereClusterSpec to reference the `VsphereClusterAccount`. We intend to use `corev1.TypedLocalObjectReference` in order to ensure the only objects that can be referenced are either in the same namespace or are scoped globally.
+
+```go
+// VsphereClusterAccount is the account to be used for vcenter actions
+type VsphereClusterAccount struct {
```
fabriziopandini, line 164: I know that naming is being discussed with other providers, so feel to ignore…
Given that this type is referenced by `identityRef`, should this be `VsphereClusterIdentity` (or simply `VSphereIdentity`)?

docs/proposal/20210311-single-controller-multitenancy.md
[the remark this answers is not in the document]
randomvariable: I'm tempted to say we pull out the list option for v1alpha4. I will take the conversation to CAPI, but https://github.com/kubernetes/kubernetes/pull/101342 is going to move NamespaceDefaultLabelName to GA and has been defaulted to on from 1.21 .

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+type VsphereClusterSpec struct {
+    ...
+    // +optional
+    IdentityRef *corev1.TypedLocalObjectReference `json:"IdentityRef,omitempty"`
+}
+```
+
+#### Controller Changes
+
+- If IdentityRef is specified, and it references a `VsphereClusterAccount`, the CRD is fetched and used to create a session
```
yastij, line 207: @randomvariable @fabriziopandini  -  it might worthwhile to inject this reference if nothing is set **and** there are `VsphereClusterAccount` that are allowed for that namespace. This avoids having to communicate the kind and Name to folks that want to provision clusters and that don't care much about creds ?

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+- If IdentityRef is specified, and it references a `Secret`, the `Secret` is used to create a session
+- The controller will add the VsphereCluster to OwnerReferences on a Secret that contains the account credentials
+- If IdentityRef is not specified, the controller will fallback to using credentials provided in the controller deployment
+- The session will be cached using the existing logic where the key is server + username + datacenter
+- The `IdentityRef` on a `VsphereCluster` will be mutable. Setting the `IdentityRef` to nil will cause the controller to fallback to the static credentials
+- The controller will fail to delete a `VsphereClusterAccount` with an error if the account is used by `VsphereClusters`
+
+#### Clusterctl Changes
```
fabriziopandini, line 215: Might be I'm wrong, but I see a potential problem for move in the fact that `VsphereClusterAccount` refers to a secret, and this secret can be in any namespace

- Say we are moving clusters in namespace A
- With the suggested changes, move will include also `VsphereClusterAccount` objects
- Say that a VsphereClusterAccount store its credential in the namespace B 

As clusterctl is designed today (https://cluster-api.sigs.k8s.io/clusterctl/provider-contract.html#move), secret in namespace B won't be moved thus leading to an inconsistent `VsphereClusterAccount` in the target cluster...

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+
+Today, clusterctl move operates by tracking objectreferences within the same namespace, since we are now proposing to use cluster-scoped resources, we will need to add requisite support to clusterctl's object graph to track cluster-scoped resources that are used by the source cluster, and ensure they are moved. We will naively not delete cluster-scoped resources during a move, as they maybe referenced across namespaces. If a cluster uses a `Secret` for account credentials, the OwnerReference will get set by the controller and the `Secret` will be moved to the target cluster. Any `VsphereClusterAccount` that's required by the resources being moved, will need to be manually created on the target cluster prior to the move operation.
```
fabriziopandini, line 217: > clusterctl's object graph to track cluster-scoped resources that are used by the source cluster,

This is kind of complex, given that the only relation between cluster and `VsphereClusterAccount` is defined in the identityRef field, and thus unknown to the clusterctl move discovery algorithm (that should be generic/provider agnostic).

There are two possible workarounds: 
- Move all the `VsphereClusterAccount` (by adding the force move label on the VsphereClusterAccount  CRD)
- Move only specific `VsphereClusterAccount` (by adding the force move label only on those objects)

docs/proposal/20210311-single-controller-multitenancy.md
fabriziopandini: I find the idea interesting, because it makes it easier to decouple roles and responsibilities
What there are more than one Account eligible for a cluster?

docs/proposal/20210311-single-controller-multitenancy.md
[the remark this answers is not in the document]
randomvariable: PTAL https://github.com/kubernetes-sigs/cluster-api/pull/4514

docs/proposal/20210311-single-controller-multitenancy.md
randomvariable, line 207:
```suggestion
- If IdentityRef is specified, and it references a `VSphereClusterAccount`, the CRD is fetched and used to create a session
```

randomvariable: We have a mechanism in CAPA for that called AWSClusterControllerIdentity that respects the existing behaviour as well as EC2 instance profiles for cross-account role assumption which is a key requirement for AWS. See https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/proposal/20200506-single-controller-multitenancy.md#upgrade-strategy

Azure doesn't have cross-account role assumption so this wasn't an important use case for them, so was not implemented there.

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+    // +optional
+    IdentityRef *corev1.TypedLocalObjectReference `json:"IdentityRef,omitempty"`
+}
+```
+
+#### Controller Changes
+
+- If IdentityRef is specified, and it references a `VsphereClusterAccount`, the CRD is fetched and used to create a session
+- If IdentityRef is specified, and it references a `Secret`, the `Secret` is used to create a session
+- The controller will add the VsphereCluster to OwnerReferences on a Secret that contains the account credentials
```
randomvariable, line 209:
```suggestion
- The controller will add the VSphereCluster to OwnerReferences on a Secret that contains the account credentials
```

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+- If IdentityRef is not specified, the controller will fallback to using credentials provided in the controller deployment
+- The session will be cached using the existing logic where the key is server + username + datacenter
+- The `IdentityRef` on a `VsphereCluster` will be mutable. Setting the `IdentityRef` to nil will cause the controller to fallback to the static credentials
```
randomvariable, line 212:
```suggestion
- The `IdentityRef` on a `VSphereCluster` will be mutable. Setting the `IdentityRef` to nil will cause the controller to fallback to the static credentials
```

randomvariable: Consider https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/proposal/20200506-single-controller-multitenancy.md#upgrade-strategy for fallback behaviour.

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+- The controller will fail to delete a `VsphereClusterAccount` with an error if the account is used by `VsphereClusters`
```
randomvariable, line 213:
```suggestion
- The controller will fail to delete a `VSphereClusterAccount` with an error if the account is used by `VSphereClusters`
```

docs/proposal/20210311-single-controller-multitenancy.md
randomvariable: Cluster scoped resources should not support arbitrary namespaces for secrets IMO, but only the controller's own namespace, that is what I've put in https://github.com/kubernetes-sigs/cluster-api/pull/4514/files

docs/proposal/20210311-single-controller-multitenancy.md
randomvariable, line 217:
```suggestion
Today, clusterctl move operates by tracking objectreferences within the same namespace, since we are now proposing to use cluster-scoped resources, we will need to add requisite support to clusterctl's object graph to track cluster-scoped resources that are used by the source cluster, and ensure they are moved. We will naively not delete cluster-scoped resources during a move, as they maybe referenced across namespaces. If a cluster uses a `Secret` for account credentials, the OwnerReference will get set by the controller and the `Secret` will be moved to the target cluster. Any `VSphereClusterAccount` that's required by the resources being moved, will need to be manually created on the target cluster prior to the move operation.
```

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+
+|                              | VsphereClusterAccount | Secret | Cluster |
+| ---------------------------- | --------------------- | ------ | ------- |
+| Infrastructure Provider      | Yes                   | Yes    | Yes     |
+| Management Cluster Operators | Yes                   | Yes    | Yes     |
+| Workload Cluster Operator    | No                    | Yes    | Yes     |
+
+#### Namespace Restrictions
+
+- To prevent workload cluster operators from using cluster-scoped `VsphereClusterAccounts` that they should not be using, the `allowedNamespaces` will be used to dictate which namespaces are allowed to use the `VsphereClusterAccount`.
```
randomvariable, line 257:
```suggestion
- To prevent workload cluster operators from using cluster-scoped `VSphereClusterAccounts` that they should not be using, the `allowedNamespaces` will be used to dictate which namespaces are allowed to use the `VSphereClusterAccount`.
```

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+An empty slice indicates that the `VsphereClusterAccount` can be used by any `VsphereCluster`
+
+#### CAPV Controller Requirements
+
+The CAPV controller will need to:
+
+- Populate conditions when cluster is misconfigured. `VsphereClusterAccount` is not found or isn't compatible with the cluster due to namespace restrictions.
```
randomvariable, line 264:
```suggestion
- Populate conditions when cluster is misconfigured. `VSphereClusterAccount` is not found or isn't compatible with the cluster due to namespace restrictions.
```

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+- Not implement an invalid configuration. If a cluster is attempting to use the `VsphereClusterAccount` from an invalid namespace, ignore it and indicate the issue through conditions.
+- Respond to changes in a `VsphereClusterAccount` spec.
```
randomvariable, line 266:
```suggestion
- Respond to changes in a `VSphereClusterAccount` spec.
```

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,321 @@
...
+The VsphereCluster will reference the Secret via an ObjectReference
+
+#### Benefits
+
+- Re-using secrets ensures encryption by default and provides a clear UX signal to end users that the data is meant to be secure
+- Keeps clusterctl move straightforward with the 1:1 cluster -> credential relationship
+
+#### Mitigations for current proposal
+
+- There are usecases where users would like to reuse the same account for multiple namespaces. See [Story 2](#story-2---deploying-multiple-clusters-from-a-single-account)
```
randomvariable, line 288:
```suggestion
- There are use cases where users would like to reuse the same account for multiple namespaces. See [Story 2](#story-2---deploying-multiple-clusters-from-a-single-account)
```

yastij (approved):
/approve

this should be ready to go, I'll leave lgtm to @fabriziopandini re: move choices

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,317 @@
...
+    // AllowedNamespaces is used to identify which namespaces are allowed to use this account.
+    // Namespaces can be selected either by using a label selector.
+    // If this object is nil, no namespaces will be allowed.
+    // If this object is empty {}, clusters can use this identity from any namespace.
+    //
+    // +optional
+    AllowedNamespaces *AllowedNamespaces `json:"allowedNamespaces"`
+}
+
+type AllowedNamespaces struct {
```
fabriziopandini, line 175: according the general guidelines being proposed in CAPI, NamespaceList is missing

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,317 @@
...
+    // +optional
+    IdentityRef *VSphereIdentityReference `json:"IdentityRef,omitempty"`
+}
+```
+
+#### Controller Changes
+
+* If IdentityRef is specified, and it references a `VSphereClusterIdentity`, the CRD is fetched and used to create a session
+* The secret referenced within `VSphereClusterIdentity` must be in the controller namespace
+* If IdentityRef is specified, and it references a `Secret`, the `Secret` is used to create a session. The secret must be in the same namespace as the `VSphereCluster`
```
fabriziopandini, line 204: > The secret must be in the same namespace as the `VSphereCluster`

Am I wrong or the secrets for Global resources should be in controller namespace?

docs/proposal/20210311-single-controller-multitenancy.md
```diff
@@ -0,0 +1,317 @@
...
+* The session will be cached using the existing logic where the key is server + username + datacenter
+* The `IdentityRef` on a `VSphereCluster` will be mutable. Setting the `IdentityRef` to nil will cause the controller to fallback to the static credentials
+* The controller will fail to delete a `VSphereClusterIdentity` with an error if the account is used by `VSphereClusters`
+
+#### Clusterctl Changes
+
+Today, clusterctl move operates by tracking objectreferences within the same namespace, since we are now proposing to use cluster-scoped resources, we will need to add requisite support to clusterctl's object graph to track cluster-scoped resources that are used by the source cluster, and ensure they are moved. We will naively not delete cluster-scoped resources during a move, as they maybe referenced across namespaces. If a cluster uses a `Secret` for account credentials, the OwnerReference will get set by the controller and the `Secret` will be moved to the target cluster. Any `VSphereClusterIdentity` that's required by the resources being moved, will need to be manually created on the target cluster prior to the move operation.
```
fabriziopandini, line 213: > Any `VSphereClusterIdentity` that's required by the resources being moved, will need to be manually created on the target cluster prior to the move operation.

Is this still a requirement? AFAIK there is no need for manually moving objects (except you explicitly wont to do so)

docs/proposal/20210311-single-controller-multitenancy.md
yastij: this is for the local credentials (i.e. BYO creds within the namespace where you provision your cluster). But I agree for the global ones

docs/proposal/20210311-single-controller-multitenancy.md
yastij: agreed

docs/proposal/20210311-single-controller-multitenancy.md
yastij: +1

gab-satchi: The PR calls out to deprecate `NamespaceList` for v1alpha4

`Providers SHOULD mark NamespaceList as deprecated from v1alpha4`

docs/proposal/20210311-single-controller-multitenancy.md
gab-satchi: yeah when the secret is addressed through the global `VSphereClusterIdentity` then it's expected to be in the controller namespace. CAPV is also implementing a direct reference to a secret from a `VSphereCluster` and for that we expect it to be in the same namespace as the cluster.

docs/proposal/20210311-single-controller-multitenancy.md
gab-satchi: thanks will update.

PR #1151, opened 2021-03-19
fix panic and randomize chosen datastore when using storage policiess

gab-satchi:

Cherrypicking #1133 

/assign @yastij


yastij (approved):
/lgtm
/approve

PR #1153, opened 2021-03-22
Make conversion-gen output location explicit

### LINKED ISSUE #1152: Generated code may go to `$GOPATH/src` instead of in repo
/kind bug

Found in https://github.com/kubernetes-sigs/cluster-api/issues/4185

We [conversion-gen](https://pkg.go.dev/k8s.io/code-generator/cmd/conversion-gen) to generate some of the code. This command has an optional argument of `--output-base` that controls where the generated files are placed.

The default for this arg can vary. If `GOPATH` is set, then it will place files under `$GOPATH/src`. But if not set, it defaults to `./`, which is what we want to have the generated files within the repo.

The dependency on an environment variable is non-obvious and can cause stale files to be used if the user doesn't know about it. To make sure files are always placed where we need them, we should be explicit and set `--output-base=./`.

Files generated/over-written within the repo. In environment with GOPATH set, files were generated under `$GOPATH/src/api`.

stmcginnis:

The conversion-gen command has an --output-base argument to control
where generated files are placed. The default value for this argument
can vary depending on whether or not $GOPATH is set or not. This results
in our generated files being created in the wrong place for some people
in a subtle and non-obvious way.

For those running `make generate` with GOPATH set, the files are created
in `$GOPATH/src/[file path]`. For those without GOPATH set, files are
created where we expect them to be within the repo at `./[file path]`.

To make sure files are always generated to the location where we expect
them to be, this updates our calls to conversion-gen to always be
explicit and set `--output-base=./`.

Fixes #1152

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,11 @@ KIND := $(TOOLS_BIN_DIR)/kind
 KUSTOMIZE := $(TOOLS_BIN_DIR)/kustomize
 TOOLING_BINARIES := $(CONTROLLER_GEN) $(CONVERSION_GEN) $(GINKGO) $(GOLANGCI_LINT) $(GOVC) $(KIND) $(KUSTOMIZE)
 
+# Set --output-base for conversion-gen if we are not within GOPATH
+ifneq ($(abspath $(ROOT_DIR)),$(shell go env GOPATH)/src/sigs.k8s.io/cluster-api-provider-vsphere)
+	OUTPUT_BASE := --output-base=$(ROOT_DIR)
+endif
+
 # Allow overriding manifest generation destination directory
 MANIFEST_ROOT ?= ./config
 CRD_ROOT ?= $(MANIFEST_ROOT)/crd/bases
@@ -203,7 +208,7 @@ generate-go: $(CONTROLLER_GEN) $(CONVERSION_GEN) ## Runs Go related generate tar
 
 	$(CONVERSION_GEN) \
 		--input-dirs=./api/v1alpha3 \
-		--output-file-base=zz_generated.conversion \
+		--output-file-base=zz_generated.conversion $(OUTPUT_BASE) \
 		--go-header-file=./hack/boilerplate/boilerplate.generatego.txt
 
 .PHONY: generate-manifests


stmcginnis:
I've found issues with this implementation. There's a better approach - update coming shortly.

Makefile
```diff
@@ -56,6 +56,11 @@ KIND := $(TOOLS_BIN_DIR)/kind
 KUSTOMIZE := $(TOOLS_BIN_DIR)/kustomize
 TOOLING_BINARIES := $(CONTROLLER_GEN) $(CONVERSION_GEN) $(GINKGO) $(GOLANGCI_LINT) $(GOVC) $(KIND) $(KUSTOMIZE)
 
+# Set --output-base for conversion-gen if we are not within GOPATH
+ifneq ($(abspath $(ROOT_DIR)),$(shell go env GOPATH)/src/sigs.k8s.io/cluster-api)
```
gab-satchi, line 60: should the path be `cluster-api-provider-vsphere`?

stmcginnis: Yes indeed, it should. Copy/paste strikes again!

Thanks @gab-satchi, update coming.

Makefile
```diff
@@ -56,6 +56,11 @@ KIND := $(TOOLS_BIN_DIR)/kind
 KUSTOMIZE := $(TOOLS_BIN_DIR)/kustomize
 TOOLING_BINARIES := $(CONTROLLER_GEN) $(CONVERSION_GEN) $(GINKGO) $(GOLANGCI_LINT) $(GOVC) $(KIND) $(KUSTOMIZE)
 
+# Set --output-base for conversion-gen if we are not within GOPATH
+ifneq ($(abspath $(ROOT_DIR)),$(shell go env GOPATH)/src/sigs.k8s.io/cluster-api-provider-vsphere)
```
yastij, line 60: can we check If this approach is consistent across cluster-api + other providers ?

stmcginnis: Yeah, good question @yastij. This is currently being [implemented in cluster-api](https://github.com/kubernetes-sigs/cluster-api/pull/4349), so we may want to wait until that lands just to see if anything needs to change there.

But this is based on an existing approach from CAPZ here: https://github.com/kubernetes-sigs/cluster-api-provider-azure/commit/03a76e30059c9d467ee470600d3471f8bf6ba4fa

It is also handled in k/k, but in that case there is a small script (with a very large library of functions hidden behind it) that handles this: https://github.com/kubernetes/kubernetes/blob/master/hack/run-in-gopath.sh

Ultimately, I think conversion-gen code should be updated to handle running outside of GOPATH. It was designed when that was a safe assumption to make, but now it needs to be accounted for everywhere it is being used.

stmcginnis: PR has now merged to cluster-api, so this would match that and CAPZ. Now that cluster-api has merged, I plan on addressing the same issue in the other upstream cluster-api-provider-* repos.

yastij (approved):
/approve
/lgtm

PR #1181, opened 2021-05-11
Add multi-tenancy support

### LINKED ISSUE #1123: Multi-tenant support with a single manager instance

gab-satchi:
- VSphereCluster fetches Secret (from cluster namespace) or VSphereIdentity that links to a Secret in the controller namespace when verifying connectivity
- VsphereVM utilizes identity when IdentityRef is specified in cluster
- validates secret is stored in the correct namespace for the identity types
- adds namespaces allowed logic
- adds authentication related conditions to VSphereVM
- adds vSphere secret to generated cluster templates
- updates secrets with finalizers and ownerref when secret is meant to be used as the Identity

Fixes #1123 

**Release note**:
```release-note
- Multi-tenancy support. Clusters can either reference a Secret with credentials or a VSphereClusterIdentity
```


go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	k8s.io/apiextensions-apiserver v0.21.0-beta.1
 	k8s.io/apimachinery v0.21.0-beta.1
 	k8s.io/client-go v0.21.0-beta.1
+	k8s.io/cluster-bootstrap v0.21.0-beta.1
```
yastij, line 22: hmmm I'd expect something like api-machinery to have utility func to get data, might be worthwhile comparing with other providers

controllers/vspherevm_controller.go
```diff
@@ -135,14 +137,28 @@ func (r vmReconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Res
 		return reconcile.Result{}, err
 	}
 
-	// Get or create an authenticated session to the vSphere endpoint.
-	authSession, err := session.GetOrCreate(r.Context,
-		vsphereVM.Spec.Server, vsphereVM.Spec.Datacenter,
-		r.ControllerManagerContext.Username, r.ControllerManagerContext.Password, vsphereVM.Spec.Thumbprint)
+	// Get cluster object and then get VSphereCluster object
+	cluster, err := clusterutilv1.GetClusterFromMetadata(r.ControllerContext, r.Client, vsphereVM.ObjectMeta)
```
yastij, line 141: this logic should be part of `retrieveVcenterSession`

controllers/vsphereclusteridentity_controller.go
```diff
@@ -0,0 +1,179 @@
...
+		return reconcile.Result{}, errors.Wrapf(
+			err,
+			"failed to init patch helper for %s %s/%s",
+			identity.GroupVersionKind(),
+			identity.Namespace,
+			identity.Name)
+	}
+
+	defer func() {
+		conditions.SetSummary(identity, conditions.WithConditions(infrav1.CredentialsAvailableCondidtion))
```
yastij, line 105: for consistency we might want to have an `identity_context.go`, see `pkg/context`

controllers/vsphereclusteridentity_controller.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	if !clusterutilv1.IsOwnedByObject(secret, identity) {
+		if len(secret.OwnerReferences) > 0 {
+			conditions.MarkFalse(identity, infrav1.CredentialsAvailableCondidtion, infrav1.SecretAlreadyInUseReason, clusterv1.ConditionSeverityError, "secret being used by another Cluster/VSphereIdentity")
+			identity.Status.Ready = false
+			return reconcile.Result{}, errors.New("secret being used by another Cluster/VSphereIdentity")
+		}
+
+		identitySecretMU.Lock()
```
yastij, line 136: do we need a lock here ? I wouldn't expect concurrent reconciles of the same `VSphereClusterIdentity` to happen at the same time in the controller. If it's to deal with other identities adding the owner refs to this secret that should be handled by optimistic concurrency at the apiserver level + your check above

yastij (commented):
/approve

generally looks good, a few comments

yastij:
@gab-satchi - you might want to rebase this

controllers/vspherevm_controller.go
```diff
@@ -135,14 +137,28 @@ func (r vmReconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Res
...
+		r.Logger.Info("VsphereVM is missing cluster label or cluster does not exist")
+		return reconcile.Result{}, nil
+	}
+
+	key := client.ObjectKey{
+		Namespace: cluster.Namespace,
+		Name:      cluster.Spec.InfrastructureRef.Name,
+	}
+	vsphereCluster := &infrav1.VSphereCluster{}
+	err = r.Client.Get(r, key, vsphereCluster)
```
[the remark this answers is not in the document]
gab-satchi: now falls back to controller credentials if either Cluster or VSphereCluster is missing (or both)

go.mod
gab-satchi: CAPA seems to do it themselves. It's a small enough util so just added it to the identity package as well

controllers/vsphereclusteridentity_controller.go
gab-satchi: It crossed my mind during implementation but the logic is fully contained in the controller right now and there aren't any other resources constantly used by the identity. I think it makes more sense if/when the identity logic grows

controllers/vsphereclusteridentity_controller.go
gab-satchi: It was to deal with other identities adding themselves as the owner. I've removed it. An identical flow exists with clusters that use secrets directly. That's also currently utilizing a lock. Should that also be removed?

api/v1alpha3/vspherecluster_types.go
```diff
@@ -62,6 +62,11 @@ type VSphereClusterSpec struct {
 	// DEPRECATED: will be removed in v1alpha4
 	// +optional
 	LoadBalancerRef *corev1.ObjectReference `json:"loadBalancerRef,omitempty"`
+
+	// IdentityRef is a reference to either a Secret or VSphereClusterIdentity that contains
+	// the identity to use when reconciling the cluster.
+	// +optional
+	IdentityRef *VSphereIdentityReference `json:"identityRef,omitempty"`
```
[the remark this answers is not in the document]
gab-satchi: moved v1alpha4 changes to a separate commit, so the first commit can be cherry-picked onto `release-0.7`

controllers/vspherevm_controller_test.go
```diff
@@ -38,6 +38,25 @@ func TestReconcileNormal_WaitingForIPAddrAllocation(t *testing.T) {
 	s := model.Service.NewServer()
 	defer s.Close()
 
+	vsphereCluster := &infrav1.VSphereCluster{
```
yastij, line 41: can we add a test to ensure we're picking up the right credentials (i.e. the ones in the `identityRef` for the VMs)

controllers/vspherecluster_controller.go
```diff
@@ -330,7 +359,60 @@ func cloudProviderConfigurationAvailable(ctx *context.ClusterContext) bool {
 	return !reflect.DeepEqual(ctx.VSphereCluster.Spec.CloudProviderConfiguration, infrav1.CPIConfig{})
 }
 
+func (r clusterReconciler) reconcileIdentitySecret(ctx *context.ClusterContext) error {
+	vsphereCluster := ctx.VSphereCluster
```
yastij, line 363: can we add a test to cover secret reconciliation ? (happy path + maybe a case secret where the secret is already owned)

controllers/vsphereclusteridentity_controller.go
yastij: can you add a TODO to consider it in the future ?

controllers/vspherevm_controller_test.go
gab-satchi: added. The existing tests set the credentials through the controller manager context. the new test skips that and only sets it through the cluster

controllers/vsphereclusteridentity_controller.go
yastij: yes

PR #1187, opened 2021-05-27
Multi tenancy support for v1alpha3

gab-satchi:
- VSphereCluster fetches Secret (from cluster namespace) or VSphereIdentity that links to a Secret in the controller namespace when verifying connectivity
- VsphereVM utilizes identity when IdentityRef is specified in cluster
- validates secret is stored in the correct namespace for the identity types
- adds namespaces allowed logic
- adds authentication related conditions to VSphereVM
- adds vSphere secret to generated cluster templates
- updates secrets with finalizers and ownerref when secret is meant to be used as the Identity

```release-note
- Multi-tenancy support. Clusters can either reference a Secret with credentials or a VSphereClusterIdentity
```


controllers/vspherecluster_controller.go
```diff
@@ -240,6 +244,26 @@ func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconci
...
+		secret := &apiv1.Secret{}
+		secretKey := client.ObjectKey{
+			Namespace: ctx.VSphereCluster.Namespace,
+			Name:      ctx.VSphereCluster.Spec.IdentityRef.Name,
+		}
+		err := ctx.Client.Get(ctx, secretKey, secret)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+		if !apierrors.IsNotFound(err) {
```
yastij, line 258: we should be deleting the secret

controllers/vspherecluster_controller.go
```diff
@@ -240,6 +244,26 @@ func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconci
...
+			r.Logger.Info(fmt.Sprintf("Removing finalizer form Secret %s/%s", secret.Namespace, secret.Name))
+			ctrlutil.RemoveFinalizer(secret, infrav1.SecretIdentitySetFinalizer)
```
yastij, line 260: do we really need a finalizer ? i.e. are there things we want to perform before deleting the secret ?

gab-satchi: spoke about the finalizer offline. just to prevent accidental deletion of a secret that's still in use by a cluster.

api/v1alpha3/vsphereclusteridentity_types.go
```diff
@@ -0,0 +1,103 @@
...
+func (c *VSphereClusterIdentity) GetConditions() clusterv1.Conditions {
+	return c.Status.Conditions
+}
+
+func (c *VSphereClusterIdentity) SetConditions(conditions clusterv1.Conditions) {
+	c.Status.Conditions = conditions
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=vsphereclusteridentities,scope=Cluster,categories=cluster-api
```
yastij, line 81: missing `// +kubebuilder:storageversion` tag

controllers/vspherecluster_controller.go
```diff
@@ -334,7 +366,60 @@ func cloudProviderConfigurationAvailable(ctx *context.ClusterContext) bool {
...
+			Name:      vsphereCluster.Spec.IdentityRef.Name,
+		}
+		err := ctx.Client.Get(ctx, secretKey, secret)
+		if err != nil {
+			return err
+		}
+
+		// check if cluster is already an owner
+		if !clusterutilv1.IsOwnedByObject(secret, vsphereCluster) {
+			secretMU.Lock()
```
yastij, line 384: we don't need the mutex here too, we operate against a local copy, and optimistic concurrency would trigger a conflict

controllers/vspherecluster_controller.go
```diff
@@ -56,9 +57,12 @@ var (
 	clusterControlledType     = &infrav1.VSphereCluster{}
 	clusterControlledTypeName = reflect.TypeOf(clusterControlledType).Elem().Name()
 	clusterControlledTypeGVK  = infrav1.GroupVersion.WithKind(clusterControlledTypeName)
+	secretMU                  sync.Mutex
```
yastij, line 60: we don't need a mutex for this

gab-satchi:
1/2 tests passes so going to re-run

/test pull-cluster-api-provider-vsphere-e2e

api/v1alpha3/vsphereclusteridentity_types.go
gab-satchi: thanks for the catch

controllers/vspherecluster_controller.go
```diff
@@ -240,6 +243,29 @@ func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconci
...
 
+	// Remove finalizer on Identity Secret
+	if identity.IsSecretIdentity(ctx.VSphereCluster) {
+		secret := &apiv1.Secret{}
+		secretKey := client.ObjectKey{
+			Namespace: ctx.VSphereCluster.Namespace,
+			Name:      ctx.VSphereCluster.Spec.IdentityRef.Name,
+		}
+		err := ctx.Client.Get(ctx, secretKey, secret)
+		if err != nil && !apierrors.IsNotFound(err) {
```
yastij, line 254: I'd suggest something like:

```go
    if err != nil 
       if !apierrors.IsNotFound(err) {			
       }
    }
```

controllers/vspherecluster_controller.go
```diff
@@ -240,6 +243,29 @@ func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconci
...
+			return reconcile.Result{}, err
+		}
+		if !apierrors.IsNotFound(err) {
```
yastij, line 257: remove this If, we want this to happen anyway, as it'd be the case where the secret was found

yastij (approved):
/approve

PR #1188, opened 2021-05-31
remove the vspherecluster finalizer when the secret does not exist

yastij:

**What this PR does / why we need it**: This PR fixes a bug in multi-tenancy. When the secret is missing, the vspherecluster controller should continue and remove the finalizer from the deleting `VSphereCluster` to unblock deletion

**Release note**:

```release-note
remove the vspherecluster finalizer when the secret does not exist
```


yastij (commented):
/approve

gab-satchi:
thanks for catching it. 

/lgtm

PR #1192, opened 2021-06-02
add checks for network validation on update in vSphereMachine

MaxRink:

This adds checks so that networks in vSpherMachines are also validated on update and not just on create

none

**Release note**:

```release-note
Validate vSphereMachine Network assignment also upon update, not just on create
```

### DIFF
--- a/api/v1alpha4/vspheremachine_webhook.go
+++ b/api/v1alpha4/vspheremachine_webhook.go
@@ -59,6 +59,7 @@ func (m *VSphereMachine) ValidateCreate() error {
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (m *VSphereMachine) ValidateUpdate(old runtime.Object) error {
 	newVSphereMachine, err := runtime.DefaultUnstructuredConverter.ToUnstructured(m)
+
 	if err != nil {
 		return apierrors.NewInternalError(errors.Wrap(err, "failed to convert new VSphereMachine to unstructured object"))
 	}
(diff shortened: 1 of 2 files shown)


MaxRink:
should i also backport that to alpha3? 
I dont think that would make a breaking change, give that VMs that would pass before are broken anyway, as the network config will never be applied properly

api/v1alpha4/vspheremachine_webhook.go
```diff
@@ -59,6 +59,8 @@ func (m *VSphereMachine) ValidateCreate() error {
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (m *VSphereMachine) ValidateUpdate(old runtime.Object) error {
 	newVSphereMachine, err := runtime.DefaultUnstructuredConverter.ToUnstructured(m)
+	spec := m.Spec
```
gab-satchi, line 62: nit: move this closer to the for loop and maybe rename? Gets a bit confusing with multiple machine specs. There's already a oldVSphereMachineSpec so name would be difficult :p

gab-satchi (approved):
changes lgtm. Thanks! and yes we could backport to v1a3 as well.

api/v1alpha4/vspheremachine_webhook.go
MaxRink: would newMachineSpec be a more speaking name?
Ive kept it at spec as that name and logic got already used in the creation validation

gab-satchi: yah maybe just leave it as spec. moving it closer to the for loop should make it clearer

MaxRink: Ok, done.Ive also added a comment to clarify hat the block does

gab-satchi: thanks, squash the commits and i can lgtm

MaxRink:
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1194 is for alpha3

api/v1alpha4/vspheremachine_webhook.go
MaxRink: and squashed

yastij (approved):
/approve
/lgtm

PR #1195, opened 2021-06-04
v1a4 switches disk move type to the default: moveAllDiskBackingsAndDisallowSharing…

gab-satchi:
**Release note**:

```release-note
Fix bug with fullClone of a VM while expanding the disk size
```


yastij (approved):
/approve
/lgtm

@gab-satchi - do we have this in `release-0.7` ?

PR #1198, opened 2021-06-09
Implement externally managed annotation (also update cluster-api to v0.4.0)

asalkeld:
This PR implements https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20210203-externally-managed-cluster-infrastructure.md for the vsphere provider.

**Release note**:
```release-note
vsphere provider now supports the externally managed annotation as documented here https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20210203-externally-managed-cluster-infrastructure.md
```


srm09:
@asalkeld Can we hold this change for a few days until CAPI cuts a 0.4.0 tag and then update it once and for all? Unless this is a blocker.

asalkeld:
The e2e is currently failing with 
```
 error: error validating "STDIN": error validating data: [ValidationError(KubeadmControlPlane.spec): unknown field "machineTemplate" in io.x-k8s.cluster.controlplane.v1alpha4.KubeadmControlPlane.spec, ValidationError(KubeadmControlPlane.spec): missing required field "infrastructureTemplate" in io.x-k8s.cluster.controlplane.v1alpha4.KubeadmControlPlane.spec]; if you choose to ignore these errors, turn validation off with --validate=false
```
I believe because we are using too old image here: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/test/e2e/config/vsphere-ci.yaml#L15

since this change: https://github.com/kubernetes-sigs/cluster-api/commit/95f428eedf96aec466caede9ade343a95e03149f

@srm09 do you know how to update these values?

srm09:
@asalkeld At the time when I updated the values, I took the nightly build closest to the cluster-api commit that I had imported in the project. Since we are now importing the beta version, update the `vsphere-ci.yaml` and `vsphere-dev.aml` files with the v0.4.0-beta images.
```yaml
value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.4.0-beta.0/core-components.yaml"
```
And update the tags of the images to `v0.4.0-beta.0` in both files as well. 

These images are published with the CAPI release.

controllers/vspherecluster_controller.go
```diff
@@ -118,6 +137,7 @@ func AddClusterControllerToManager(ctx *context.ControllerManagerContext, mgr ma
 			&source.Channel{Source: ctx.GetGenericEventChannelFor(clusterControlledTypeGVK)},
 			&handler.EnqueueRequestForObject{},
 		).
+		WithEventFilter(predicates.ResourceIsNotExternallyManaged(reconciler.Logger)).
```
gab-satchi, line 140: shouldn't this be enough to ignore externally managed vsphereClusters? do we need the extra logic in the Cluster `Watches` above?

gab-satchi (commented):
changes lgtm. I think the predicates setup by the EventFilter will apply to VSphereClusters being queued up as a result of changes to a CAPI cluster but worth double checking.

gab-satchi:
The quickstart test is failing because v1alpha4 changes are now requiring an extra `IP_FAMILY` variable to be set. https://github.com/kubernetes-sigs/cluster-api/blob/v0.4.0-beta.0/test/e2e/quick_start.go#L76

controllers/vspherecluster_controller.go
asalkeld: Hi @gab-satchi  , this came up here https://kubernetes.slack.com/archives/C8TSNPY4T/p1620749340371300
so the issue is that we are watching a cluster then mapping it to a vsphereCluster,but the predicate is only applied on
the cluster, not the mapped vsphereCluster (and the cluster does not have the annotation).

gab-satchi: ah okay. thanks for the explanation!

go.mod
```diff
@@ -5,24 +5,32 @@ go 1.16
...
-	sigs.k8s.io/controller-runtime v0.9.0-alpha.1
-	sigs.k8s.io/kind v0.9.0
+	gopkg.in/warnings.v0 v0.1.2 // indirect
+	k8s.io/api v0.21.1
+	k8s.io/apiextensions-apiserver v0.21.1
+	k8s.io/apimachinery v0.21.1
+	k8s.io/client-go v0.21.1
+	k8s.io/klog/v2 v2.9.0
+	k8s.io/utils v0.0.0-20210527160623-6fdb442a123b
+	sigs.k8s.io/cluster-api v0.4.0-beta.0
```
gab-satchi, line 25: I think the CAPI v0.4.0 release is imminent (sometime this week). Should we wait for an official release and update to that?

gab-satchi (commented):
sorry for the back n forth with the capi version. I'm thinking it may be best to wait for the proper 0.4 release.

go.mod
yastij: +1

srm09:
/retitle Implement externally managed annotation (also update cluster-api to v0.4.0)

srm09 (approved):
/approve

go.mod
asalkeld: this is now done.

yastij (approved):
/lgtm
/approve

yastij:
@asalkeld - thanks, can you squash the commits ?

asalkeld:
> @asalkeld - thanks, can you squash the commits ?

Seriously? Squash the update and the externally managed annotation?

PR #1207, opened 2021-07-08
add VSphereClusterTemplate type

### LINKED ISSUE #1206: Add VSphereClusterTemplate type
/kind feature

Add a new CRD named VSphereClusterTemplate. This is needed for ClusterClass.
In the short term this type will not be used outside of ClusterClass.

ykakarap:
Add VSphereClusterTemplate type. This type is needed to work with [ClusterClass](https://github.com/kubernetes-sigs/cluster-api/pull/4678)
Fixes #1206 

**Release note**:

```release-note
Adds VSphereClusterTemplate type
```


fabriziopandini (commented):
@ykakarap thanks for this PR!

Could you kindly enable web hooks for the new type ensuring that template type applies the same defaulting & validation rules of the actual type, otherwise this could trigger continuous reconcile loop down the path when comparing the desired state (the VSphereTemplate object) with the actual state (the instance of VSphereCluster generated from the template)?

api/v1alpha4/vsphereclustertemplate_webhook.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if spec.Thumbprint != "" && spec.Insecure != nil && *spec.Insecure {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "Insecure"), spec.Insecure, "cannot be set to true at the same time as .spec.template.spec.Thumbprint"))
+	}
+
+	return aggregateObjErrors(ct.GroupVersionKind().GroupKind(), ct.Name, allErrs)
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (ct *VSphereClusterTemplate) ValidateUpdate(old runtime.Object) error {
```
fabriziopandini, line 49: Should we make this immutable, like the other templates?

fabriziopandini:
thanks for updating, changes lgtm to me
I leave the final word to CAPV maintainers

config/rbac/vsphereclustertemplate_editor_role.yaml
```diff
@@ -0,0 +1,24 @@
+# permissions for end users to edit vsphereclustertemplates.
```
srm09, line 1: This file is not required, could you please remove this.

config/rbac/vsphereclustertemplate_viewer_role.yaml
```diff
@@ -0,0 +1,20 @@
+# permissions for end users to view vsphereclustertemplates.
```
srm09, line 1: same as above.

config/samples/infrastructure_v1alpha4_vsphereclustertemplate.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
```
srm09, line 1: same as above.

api/v1alpha4/vspherecluster_webhook.go
```diff
@@ -51,3 +58,13 @@ func (c *VSphereCluster) ValidateUpdate(old runtime.Object) error {
...
 	return nil
 }
+
+func defaultVSphereCluterSpec(s *VSphereClusterSpec) {}
+
+func validateVSphereClusterSpec(s VSphereClusterSpec) field.ErrorList {
+	var allErrs field.ErrorList
+	if s.Thumbprint != "" && s.Insecure != nil && *s.Insecure {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "Insecure"), s.Insecure, "cannot be set to true at the same time as .spec.Thumbprint"))
+	}
```
srm09, line 68: This is an FYI, the struct variable `Insecure` is deprecated and is scheduled for removal in v1alpha4. So this will go away once the PR for removing deprecated types gets merged.

api/v1alpha4/vspherecluster_webhook.go
```diff
@@ -28,17 +29,23 @@ func (c *VSphereCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+// +kubebuilder:webhook:verbs=create;update,path=/default-infrastructure-cluster-x-k8s-io-v1alpha4-vspherecluster,mutating=true,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=vsphereclusters,versions=v1alpha4,name=default.vspherecluster.infrastructure.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
+// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1alpha4-vspherecluster,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=vsphereclusters,versions=v1alpha4,name=validation.vspherecluster.infrastructure.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
+
+var _ webhook.Defaulter = &VSphereCluster{}
+
+func (c *VSphereCluster) Default() {
+	defaultVSphereCluterSpec(&c.Spec)
+}
```
srm09, line 39: Is there a specific reason that the defaulting webhook is being included here? I see the implementation of the method is empty. If this is for the sake of future usage, let's introduce it only when there is some specific defaulting logic that needs to be implemented.

srm09 (commented):
The error in the tests is due to the new defaulting webhook being introduced for VSphereCluster type. If the webhook isn't necessary, we should remove it, which in turn would remove the error. Looking forward to hearing about the reasoning behind the webhook.

api/v1alpha4/vspherecluster_webhook.go
ykakarap: Yes, it is for future use cases. So that we make sure to introduce consistent defaulting for template and the cluster type when ever we do that.

srm09: My vote is to wait till the need arises for a separate webhook. Adding a note in the clustertemplate_webhook to make sure this needs to be replicated across the two types would be a way to move forward. 
Thoughts @gab-satchi @yastij ?

yastij: +1 on not adding dead code

ykakarap: Thanks for the feedback @srm09 @yastij. Will update the PR :)

config/rbac/vsphereclustertemplate_editor_role.yaml
srm09: The file `config/rbac/role.yaml` gets updated once we make changes to the controllers to allow them to read/update/delete/reconcile the template objects.

ykakarap:
@srm09 Addressed all review comments. PTAL.

ykakarap:
@srm09 Had to rebase because of conflicts. Can you please re-lgtm? 🙂

PR #1182, opened 2021-05-11
Multi AZ failure domain support

srm09:
**What this PR does / why we need it**:
This PR introduced support for Failure Domains in v1alpha4 API.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:
```release-note
Adds support for failure domains
```

PR #1208, opened 2021-07-08
Multi AZ failure domain support

srm09:
Adds support for failure domains

Fixes #

Backport for #1182 

**Release note**:
```release-note
Adds support for failure domains
```


pkg/services/govmomi/cluster/vmgroup.go
```diff
@@ -0,0 +1,77 @@
...
+package cluster
+
+import (
+	"context"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+func AddVMToGroup(ctx computeClusterContext, clusterName, vmGroupName, vm string) error {
```
yastij, line 26: why isn't the `clusterName` withing the `computeClusterContext`

yastij (commented):
commits 1 - 4

controllers/vspheredeploymentzone_controller_domain.go
```diff
@@ -0,0 +1,233 @@
...
+	if err := r.reconcileTopology(ctx); err != nil {
+		logger.Error(err, "topology is not configured correctly")
+		return errors.Wrap(err, "topology is not configured correctly")
+	}
+	return nil
+}
+
+func (r vsphereDeploymentZoneReconciler) reconcileInfraFailureDomain(ctx *context.VSphereDeploymentZoneContext, failureDomain infrav1.FailureDomain) error {
+	autoConfigure := false
+	if failureDomain.AutoConfigure != nil {
```
yastij, line 68: I assume we're defaulting this, so we don't need this check

controllers/vspheredeploymentzone_controller_domain.go
```diff
@@ -0,0 +1,233 @@
...
+
+TagVerifier:
+	for _, ref := range managedRefs {
+		attachedTags, err := ctx.AuthSession.TagManager.GetAttachedTags(ctx, ref)
+		if err != nil {
+			return err
+		}
+		for _, tag := range attachedTags {
+			if tag.Name == failureDomain.Name {
+				continue TagVerifier
```
yastij, line 163: why aren't we just breaking when we find the tag that matches ?

controllers/vspheredeploymentzone_controller_domain.go
```diff
@@ -0,0 +1,233 @@
...
+		err = errors.Wrapf(err, "failed to attach tag %s to region", failureDomain.Name)
+	}
+	return err
+}
+
+// managedRefFinder is the method to find the reference of the type specified in the Failure Domain
+type managedRefFinder func(goctx.Context) ([]object.Reference, error)
+
+// getManagedRefFinder returns the method to find the reference specified in the Failure Domain
+func getManagedRefFinder(failureDomain infrav1.FailureDomain, topology infrav1.Topology, finder *find.Finder) managedRefFinder {
```
yastij, line 203: this belongs into a govmomi package

controllers/vspheredeploymentzone_controller_domain.go
```diff
@@ -0,0 +1,233 @@
...
+
+	err = ctx.AuthSession.TagManager.AttachTag(ctx, failureDomain.Name, managedRef[0])
+	if err != nil {
+		err = errors.Wrapf(err, "failed to attach tag %s to region", failureDomain.Name)
+	}
+	return err
+}
+
+// managedRefFinder is the method to find the reference of the type specified in the Failure Domain
+type managedRefFinder func(goctx.Context) ([]object.Reference, error)
```
yastij, line 200: this belongs into a govmomi package

yastij (commented):
commit 5

api/v1alpha3/condition_consts.go
```diff
@@ -47,6 +47,20 @@ const (
...
+
+	// FailureDomainsSkippedReason (Severity=Info) documents that some of the failure domain statuses
+	// associated to the VSphereCluster are reported as not ready.
+	FailureDomainsSkippedReason = "FailureDomainsSkipped"
+
+	// FailureDomainsNotReadyReason (Severity=Info) documents that some of the failure domains
+	// associated to the VSphereCluster are not reporting the Ready status.
+	// Instead of reporting a false ready status, these failure domains are still under the process of reconciling
+	// and hence not yet reporting their status.
+	FailureDomainsNotReadyReason = "FailureDomainsNotReady"
```
gab-satchi, line 63: `FailureDomainsNotReady` doesn't really convey that we're waiting on them to report status. It almost sounds like they reported a Ready: false. Maybe something like `WaitingForFailureDomainStatus`?

gab-satchi (commented):
commit 6: just minor nits

api/v1alpha3/condition_consts.go
```diff
@@ -124,3 +126,48 @@ const (
...
+	// the VSphereDeploymentZone is misconfigured.
+	RegionMisconfiguredReason = "FailureDomainRegionMisconfigured"
+
+	// ZoneMisconfiguredReason (Severity=Error) documents that the zone for the Failure Domain associated to
+	// the VSphereDeploymentZone is misconfigured.
+	ZoneMisconfiguredReason = "FailureDomainZoneMisconfigured"
+
+	// ComputeClusterMisconfiguredReason (Severity=Error) documents that the Compute Cluster details for the Failure Domain
+	// associated to the VSphereDeploymentZone is misconfigured.
+	ComputeClusterMisconfiguredReason = "ComputeClusterMisconfigured"
```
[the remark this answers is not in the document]
srm09: Its a mix of both, its used if the cluster is not found or if the resource Pool does not belong to the compute cluster. Does it makes sense to break it down into 2 different reasons instead?

controllers/vspheredeploymentzone_controller_domain.go
srm09: The implicit assumption that I made here is that all the hosts in a host group should be tagged and hence I continue to loop through the rest of the references.
I thought about this but defensively assumed that we would want all hosts in the host group to be tagged. From the user's perspective, if you are using an explicit host group in the failure domain, all the hosts in that group are eligible for VM placement of VMs. And hence, it made sense for all of them to have the tag. 

But if that assumption is too rigid, we could assume that the validity of the failure domain depends on having at least one tagged host in the group. But that causes us to lose redundancy since all of the VM placement might happen on a single host.

controllers/vspheredeploymentzone_controller_domain.go
srm09: I specifically did not move it to the govmomi package since a lot of inputs are the CAPV API types. It isn't the case with the current setup of the `govmomi` package, but decided against it. 
On the other hand, I know this method has a lotta vSphere internals which do not need to reside in the controllers package. I will move it.

api/v1alpha3/condition_consts.go
```diff
@@ -124,3 +126,48 @@ const (
 	// SecretAlreadyInUseReason is used when another VSphereClusterIdentity is using the secret
 	SecretAlreadyInUseReason = "SecretInUse"
 )
+
+const (
+	// PlacementConstraintConfigurationCondition documents whether the placement constraint is configured correctly or not.
+	PlacementConstraintConfigurationCondition clusterv1.ConditionType = "PlacementConstraintConfiguration"
```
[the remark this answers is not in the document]
srm09: Yes, I did not like any of the names I came up with yesterday. Will mostly go with Gab's suggestion.

controllers/vspheremachine_controller.go
```diff
@@ -486,10 +493,80 @@ func (r machineReconciler) reconcileNormalPre7(ctx *context.MachineContext, vsph
...
+func overrideNetworkDeviceSpecs(deviceSpecs []infrav1.NetworkDeviceSpec, networks []string) []infrav1.NetworkDeviceSpec {
+	index, length := 0, len(networks)
+
+	devices := make([]infrav1.NetworkDeviceSpec, 0, integer.IntMax(length, len(deviceSpecs)))
+	// override the networks on the VM spec with placement constraint network definitions
+	for i := range deviceSpecs {
+		vmNetworkDeviceSpec := deviceSpecs[i]
+		if i < length {
+			index++
+			vmNetworkDeviceSpec.NetworkName = networks[i]
```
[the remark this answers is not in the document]
srm09: I added tests for all the 3 scenarios, >/</=
But I will recheck the logic to be sure.

controllers/vspheredeploymentzone_controller_domain.go
srm09: Relaxing the constraints to allow tags to be present on at least of the hosts in the host group. wdyt?

controllers/vspheremachine_controller.go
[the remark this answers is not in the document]
srm09: Line 550 makes sure that the network override is present in the failure domain topology spec.

pkg/services/govmomi/cluster/rule.go
```diff
@@ -0,0 +1,84 @@
...
+		return false
+	}
+	return *v.Mandatory
+}
+
+func (v vmHostAffinityRule) Disabled() bool {
+	if v.Enabled == nil {
+		return true
+	}
+	return negate(*v.Enabled)
```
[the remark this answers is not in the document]
srm09: jus wanted to clearly specify the intent and for beautification. :)

pkg/services/govmomi/cluster/service.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/types"
+
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/session"
+)
+
+type computeClusterContext interface {
+	context.Context
+
+	GetSession() *session.Session
```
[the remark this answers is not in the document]
srm09: The idea behind adding a new context here was to limit the scope of the context that is being passed to the package. In the controller where the methods that use this context are called from, they pass the `VSphereDeploymentZoneContext` which contains a lot more information that is necessary by this package.

pkg/services/govmomi/cluster/service.go
```diff
@@ -0,0 +1,68 @@
...
+
+	GetLogger() logr.Logger
```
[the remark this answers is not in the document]
srm09: answered above.

pkg/services/govmomi/cluster/rule.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func negate(input bool) bool {
+	return !input
+}
+
+func VerifyAffinityRule(ctx computeClusterContext, clusterName, hostGroupName, vmGroupName string) (Rule, error) {
```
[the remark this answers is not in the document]
srm09: Am not sure what does this one mean?

controllers/vspheredeploymentzone_controller_domain.go
yastij: hmmm no you're right, let's check all the hosts.

api/v1alpha3/condition_consts.go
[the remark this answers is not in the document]
yastij: yeah, I'd rely on `ResourcePoolNotFound` and `ComputeClusterNotFound`

pkg/services/govmomi/cluster/service.go
```diff
@@ -0,0 +1,68 @@
...
+	return refs, nil
+}
+
+func reconfigure(ctx context.Context, ccr *object.ClusterComputeResource, spec types.BaseComputeResourceConfigSpec) error {
+	reconfigureTask, err := ccr.Reconfigure(ctx, spec, true)
+	if err != nil {
+		return err
+	}
+
+	err = reconfigureTask.Wait(ctx)
```
yastij, line 63: we should probably revisit this to stop blocking CAPV's reconciliation, see the `defer reconcileVSphereVMOnTaskCompletion(ctx)` on `ReconcileVM`

controllers/vspheremachine_controller_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+		key = client.ObjectKey{Namespace: testNs.Name, Name: infraMachine.Name}
+	})
+
+	AfterEach(func() {
+		Expect(testEnv.Cleanup(ctx, capiCluster, testNs)).To(Succeed())
+	})
+
+	// TODO(srm09): Move this to a different test, since requires a more complicated setup
+	/*Context("In case of errors on the VSphereVM", func() {
```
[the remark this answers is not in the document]
srm09: removed this, since it is an unrelated test for VSphereVM. Will add it in a separate PR.

controllers/vspherevm_controller.go
```diff
@@ -138,6 +139,33 @@ func (r vmReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error)
...
+	if err := r.Client.Get(r, req.NamespacedName, vsphereMachine); err != nil {
+		if apierrors.IsNotFound(err) {
+			r.Logger.Info("Owner VSphereMachine not found, won't reconcile", "key", req.NamespacedName)
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	var vsphereFailureDomain infrav1.VSphereFailureDomain
+	if failureDomain := vsphereMachine.Spec.FailureDomain; failureDomain != nil {
```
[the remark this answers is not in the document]
srm09: Regardless of the VSphereVM, the failureDomain should be set on the VSphereMachine. Will include that in the vspheremachine controller logic as well.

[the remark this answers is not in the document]
srm09: And rely on the CAPI machine's failure domain in the VSphereVM controller as well.

pkg/services/govmomi/cluster/service.go
srm09: This reconfigure task is not a long-running task, does it still make sense to defer the reconciliation.

controllers/vspherevm_controller.go
```diff
@@ -138,6 +139,43 @@ func (r vmReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error)
 		return reconcile.Result{}, err
 	}
 
+	// Fetch the owner VSphereMachine.
+	vsphereMachine := &infrav1.VSphereMachine{}
+	hasMachineOwner := clusterutilv1.HasOwnerRef(vsphereVM.OwnerReferences, metav1.OwnerReference{
```
yastij, line 144: This breaks `HAProxyLoadBalancer` in v1alpha3 setups. the haproxy controller generates a `VSphereVM`, so a `VSphereVM` can have another ownerRef than a `VSphereMachine`

controllers/vspherevm_controller.go
```diff
@@ -156,13 +157,54 @@ func (r vmReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error)
 	}
 	conditions.MarkTrue(vsphereVM, infrav1.VCenterAvailableCondition)
 
+	var vsphereFailureDomain *infrav1.VSphereFailureDomain
```
yastij, line 160: what happens when we call `reconcileVM` on an HAProxy VM ? are we checking ctx.VSphereFailureDomain is not nil ?

srm09: Yes the `reconcileVMGroupInfo` has a nil check before using it.

yastij (approved):
/approve

Issue #1211, opened 2021-07-13
After clusterctl move, vsphere worker machines take ~10-12 minutes to go to "Running" phase


mrajashree:
/kind bug

**What steps did you take and what happened:**
Created a self-managed capv cluster following steps in the quickstart book
1. Created a Kind bootstrap cluster
2. Installed CAPI+CAPV providers through clusterctl
3. Created the target workload cluster
4. Installed CAPI+CAPV providers through clusterctl on this workload cluster
5. Performed clusterctl move to the workload cluster.

**What did you expect to happen:**
The worker machines' status field has the nodeRef value set after move, because of which it can be considered ready, but its phase is `Provisioned` and not `Ready` for about 10-12 minutes after the move.

**Anything else you would like to add:**
These are the Machine.Status fields of a worker machine post move, the difference between the `NodeHealthy` and `Ready` conditions' timestamps is ~13 min, which is the time taken for Machine to go to "Running" phase.
```
status:
  addresses:
  - address: 198.x.x.x
    type: ExternalIP
  bootstrapReady: true
  conditions:
  - lastTransitionTime: "2021-07-13T23:13:47Z"
    status: "True"
    type: Ready
  - lastTransitionTime: "2021-07-13T22:55:52Z"
    status: "True"
    type: BootstrapReady
  - lastTransitionTime: "2021-07-13T23:13:47Z"
    status: "True"
    type: InfrastructureReady
  - lastTransitionTime: "2021-07-13T22:55:52Z"
    status: "True"
    type: NodeHealthy
  infrastructureReady: true
  lastUpdated: "2021-07-13T23:13:47Z"
  nodeRef:
    apiVersion: v1
    kind: Node
    name: abcd
    uid: 2dc6356a-362b-432e-b573-026d2193f17a
  observedGeneration: 1
  phase: Running
```
And after move before those 13 minutes finish, the `Ready` condition is set to `False` with `WaitingForClusterInfrastructure` and none of the addresses are set.

**Environment**
- Cluster-api-provider-vsphere version: v0.7.8
- Kubernetes version: (use `kubectl version`): 1.19
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

PR #1212, opened 2021-07-13
Add watch on cluster.status.infrastructureReady to reconcile vsphereMachine when cluster infrastructure is ready

mrajashree:
After `clusterctl move` for self-managed CAPV clusters, worker machines take around 10-15 minutes to go to "Running" phase. 
It could be due to the following reasons:
1. During `clusterctl move`, the Status fields are removed. So none of the CAPI objects retain the status fields, including the Cluster object. Once the Cluster.Spec.Paused field is unset after move, controllers begin to reconcile and the status fields get set again.
2. The [capv vsphereMachine controller](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/controllers/vspheremachine_controller.go#L316) while reconciling the VsphereMachine object checks for the `Cluster.Status.InfrastructureReady` field before proceeding, which will be `false` immediately after the move until the cluster controller has reconciled & updated it.
3. The vsphereMachine controller gets the Cluster object at the start of every Reconcile, and it's possible that the CAPI cluster controller updates the Cluster object after it has already been picked up by the vsphereMachine controller. It's also possible that CAPV could be hitting [this bug](https://github.com/kubernetes-sigs/controller-runtime/issues/1464) where controller-runtime GET is reading stale data. And from the comments on that bug the recommended way is to keep retrying.
4. CAPV syncs with CAPI objects eventually after the sync period which is by default set to 10 minutes. Because of which, after around 10 minutes the vsphereMachine controller gets the latest copy of the Cluster object and reconciles VsphereMachine successfully because by then the Cluster.Status.InfrastructureReady field has been set to true.
5. So the solution is either reducing the sync period time so that CAPV gets the latest Cluster object sooner or requeuing VsphereMachine until Cluster.Status.InfrastructureReady is true. It is recommended to requeue the object instead of reducing the sync period in this case as per [these comments in the controller-runtime code](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.9.1/pkg/manager/manager.go#L108-L133).
4. This PR modifies the watch on Cluster object, to receive update events when cluster.status.infrastructureReady field becomes true. This way it enqueues the vsphereMachine for reconcilation once infrastructure is ready post clusterctl move

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1211

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/vspheremachine_controller.go
+++ b/controllers/vspheremachine_controller.go
@@ -34,14 +34,13 @@ import (
 	"sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
+	"sigs.k8s.io/cluster-api/util/predicates"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
@@ -143,19 +142,7 @@ func AddMachineControllerToManager(ctx *context.ControllerManagerContext, mgr ma
 		err = c.Watch(
 			&source.Kind{Type: &clusterv1.Cluster{}},
 			handler.EnqueueRequestsFromMapFunc(r.clusterToVSphereMachines),
-			predicate.Funcs{
-				UpdateFunc: func(e event.UpdateEvent) bool {
-					oldCluster := e.ObjectOld.(*clusterv1.Cluster)
-					newCluster := e.ObjectNew.(*clusterv1.Cluster)
-					return oldCluster.Spec.Paused && !newCluster.Spec.Paused
-				},
-				CreateFunc: func(e event.CreateEvent) bool {
-					if _, ok := e.Object.GetAnnotations()[clusterv1.PausedAnnotation]; !ok {
-						return false
-					}
-					return true
-				},
-			})
+			predicates.ClusterUnpausedAndInfrastructureReady(r.Logger))
 		if err != nil {
 			return err
 		}


neolit123:
deferring to the CAPV maintainers:
/assign yastij randomvariable

controllers/vspheremachine_controller.go
```diff
@@ -317,9 +317,9 @@ func (r machineReconciler) reconcileNormal(ctx *context.MachineContext) (reconci
 	ctrlutil.AddFinalizer(ctx.VSphereMachine, infrav1.MachineFinalizer)
 
 	if !ctx.Cluster.Status.InfrastructureReady {
-		ctx.Logger.Info("Cluster infrastructure is not ready yet")
+		ctx.Logger.Info("Cluster infrastructure is not ready yet, requeuing")
 		conditions.MarkFalse(ctx.VSphereMachine, infrav1.VMProvisionedCondition, infrav1.WaitingForClusterInfrastructureReason, clusterv1.ConditionSeverityInfo, "")
-		return reconcile.Result{}, nil
+		return reconcile.Result{RequeueAfter: 30*time.Second}, nil
```
vincepri, line 322: Should we have a watch here instead of requeueing?

mrajashree: @vincepri oh you mean just like there's a watch for Cluster for the VSphereCluster object, we should add one mapping Cluster to VSphereMachine?

vincepri: Yeah if we're waiting for it, we probably should add watch instead of using RequeueAfter

mrajashree: oh looks like there is a watch for Cluster, just need to update the predicate to check for this

srm09:
Retriggering, seems to be an unrelated failure

controllers/vspheremachine_controller.go
mrajashree: I tested changing the update predicate for the watch on Cluster but I see the same, machines reconcile after a while. I'll see if I missed something and test again

mrajashree: The watch on cluster object's infrastructureReady field change wasn't working because it wasn't enqueuing the vsphereMachine object. Opened a PR for that https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1306
With the change in clusterToVsphereMachine mapper, updating the watch on cluster object to enqueue vsphereMachine when cluster.status.infrastructureReady becomes true works. I'll rebase this PR when the other one is merged

mrajashree:
Updated the PR to replace predicates on cluster watch with [ClusterUnpausedAndInfrastructureReady](https://github.com/kubernetes-sigs/cluster-api/blob/main/util/predicates/cluster_predicates.go#L179)
This covers create events when Cluster.Spec.Paused is false and Cluster.Status.InfrastructureReady is true
And covers update events when either Cluster.Spec.Paused transitions to false or Cluster.Status.InfrastructureReady transitions to true

PR #1095, opened 2021-01-27
Removes fields marked as deprecated

srm09:
**What this PR does / why we need it**:
Removing deprecated types for v1a4 types

**Which issue(s) this PR fixes**:
Fixes #1038

**Special notes for your reviewer**:
n/a

**Release note**:
```release-note
Remove deprecated fields from v1alpha4 API types
```

PR #1214, opened 2021-07-14
Fix VSphereCluster VCenterAvailable condition

### LINKED ISSUE #1209: Make VCenterAvailable status/condition more meaningful to enable better troubleshooting of issues related to connectivity or misconfiguration
When bootstrapping a cluster using capv version `0.7.8` the `VCenterAvailable` condition stays `False` with  reason `VCenterUnreachable` and message `Secret "tkg-mgmt-vc" not found` long after VMs have been created. 

```
Every 2.0s: kubectl get vspherecluster mycluster -o json | jq .status                                                                                                                                                                                                                                            
{
  "conditions": [
    {
      "lastTransitionTime": "2021-07-09T19:43:31Z",
      "status": "True",
      "type": "Ready"
    },
    {
      "lastTransitionTime": "2021-07-09T19:43:31Z",
      "status": "True",
      "type": "LoadBalancerAvailable"
    },
    {
      "lastTransitionTime": "2021-07-09T19:43:30Z",
      "message": "Secret \"mycluster-vc\" not found",
      "reason": "VCenterUnreachable",
      "severity": "Error",
      "status": "False",
      "type": "VCenterAvailable"
    }
  ],
  "ready": true
}
```

From what I've observed, this condition stays in this state until the cluster is created successfully.

I've started to read through [the reconcile func](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/504148a6290b45dc8fcbb86f1781ccf6f22c58fd/controllers/vspherecluster_controller.go#L277) more closely to see if I can understand what is intended and why this condition is not cleared as I would expect.

Currently, this makes it hard to distinguish cases where there _really is_ a connectivity issue to the vCenter. We'd like to be able to do so more reliably in order to provide faster feedback to users when there is a misconfiguration or connectivity issue from bootstrap cluster to the management cluster. In particular, when bootstrapping IPv6 clusters, we've identified a couple cases where connectivity issues are likely to occur:

1. When using a kind cluster for bootstrapping and NAT is not configured for IPv6 (which Docker doesn't do automatically).
2. When a hostname is used for the vCenter, but no IPv6 DNS server has been configured for Docker or CoreDNS that can resolve the hostname

That said, we think this would be useful more generally in other scenarios as well.

/kind feature


**Environment:**

- Cluster-api-provider-vsphere version: 0.7.8

tylerschultz:

We now check the connectivity regardless of whether
CloudProviderConfiguration is set on the spec.

That property has been deprecated and it is valuable to check this
connectivity early and surface this on the VSphereCluster status rather
than just on VSphereVMs.

This also fixes a bug where the status could be set to "False" if the
secret was not found and then never properly set to "True" once the secret
becomes present.

Fixes #1209

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/vspherecluster_controller_test.go
+++ b/controllers/vspherecluster_controller_test.go
@@ -24,6 +24,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/vmware/govmomi/simulator"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -37,6 +38,8 @@ import (
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha4"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context/fake"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/identity"
+	"sigs.k8s.io/cluster-api-provider-vsphere/test/helpers"
 )
 
 const (
@@ -49,11 +52,20 @@ var _ = Describe("ClusterReconciler", func() {
 
 	Context("Reconcile an VSphereCluster", func() {
 		It("should create a cluster", func() {
+			fakeVCenter := startVcenter()
+			vcURL := fakeVCenter.ServerURL()
+			defer fakeVCenter.Destroy()
+
 			// Create the secret containing the credentials
+			password, _ := vcURL.User.Password()
 			secret := &corev1.Secret{
 				ObjectMeta: metav1.ObjectMeta{
 					GenerateName: "secret-",
 					Namespace:    "default",
+				},
+				Data: map[string][]byte{
+					identity.UsernameKey: []byte(vcURL.User.Username()),
+					identity.PasswordKey: []byte(password),
 				},
 			}
 			Expect(testEnv.Create(ctx, secret)).To(Succeed())
@@ -69,6 +81,7 @@ var _ = Describe("ClusterReconciler", func() {
 						Kind: infrav1.SecretKind,
 						Name: secret.Name,
 					},
+					Server: fmt.Sprintf("%s://%s", vcURL.Scheme, vcURL.Host),
 				},
 			}
 			Expect(testEnv.Create(ctx, instance)).To(Succeed())
@@ -128,6 +141,14 @@ var _ = Describe("ClusterReconciler", func() {
 					return false
 				}
 				return len(secret.OwnerReferences) > 0
+			}, timeout).Should(BeTrue())
+
+			By("setting the VSphereCluster's VCenterAvailableCondition to true")
+			Eventually(func() bool {
+				if err := testEnv.Get(ctx, key, instance); err != nil {
+					return false
+				}
+				return conditions.IsTrue(instance, infrav1.VCenterAvailableCondition)
 			}, timeout).Should(BeTrue())
 		})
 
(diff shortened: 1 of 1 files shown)


controllers/vspherecluster_controller.go
```diff
@@ -301,14 +301,12 @@ func (r clusterReconciler) reconcileNormal(ctx *context.ClusterContext) (reconci
 		return reconcile.Result{}, err
 	}
 
-	if cloudProviderConfigurationAvailable(ctx) {
-		if err := r.reconcileVCenterConnectivity(ctx); err != nil {
-			conditions.MarkFalse(ctx.VSphereCluster, infrav1.VCenterAvailableCondition, infrav1.VCenterUnreachableReason, clusterv1.ConditionSeverityError, err.Error())
-			return reconcile.Result{}, errors.Wrapf(err,
-				"unexpected error while probing vcenter for %s", ctx)
-		}
-		conditions.MarkTrue(ctx.VSphereCluster, infrav1.VCenterAvailableCondition)
```
srm09, line 310: This will break since the method `reconcileVCenterConnectivity` depends on the Datacenter value from the cloud provider configuration. Also, this change is included in another PR raised to remove the deprecated types in v1alpha4. I would suggest let us wait for that to merge since it removes all the deprecated types together, and alters the controller logic accordingly.

controllers/vspherecluster_controller_test.go
```diff
@@ -122,6 +135,19 @@ var _ = Describe("ClusterReconciler", func() {
...
+			By("setting the VSphereCluster's VCenterAvailableCondition to true")
+			Eventually(func() bool {
+				if err := testEnv.Get(ctx, key, instance); err != nil {
+					return false
+				}
+				for _, condition := range instance.GetConditions() {
+					if condition.Type == infrav1.VCenterAvailableCondition {
+						return condition.Status == corev1.ConditionTrue
+					}
+				}
```
srm09, line 148: There is a helper function we borrow from the CAPI codebase `sigs.k8s.io/cluster-api/util/conditions` to get a type of condition from an object, `Has` & `IsTrue`. I would suggest using the same.

srm09 (commented):
tldr; PR #1095 fixes this and removes the deprecated types too.

controllers/vspherecluster_controller.go
tylerschultz: We also saw that `reconcileVCenterConnectivity` refers to the cloud provider's Datacenter. We were under the impression that the Datacenter would be empty string in the case the cloud provider configuration was not present. The reachability and connectivity to the VCenter can still be verified in that case.

controllers/vspherecluster_controller_test.go
tylerschultz: Updated to use `IsTrue`. It will return false if the condition is not present, so there should be no need for `Has`.

controllers/vspherecluster_controller.go
mcwumbly: Just skimmed #1095 and agree it would fix #1209, so happy to see it resolved that way if it gets merged first. Certainly don't want to get in the way of merging that.

If valuable, we could slim this PR down to just add  the test coverage included here. 

We were similarly under the impression initially that the `cloudProviderConfigurationAvailable` guard was needed given that `reconcileVCenterConnectivity` accesses `Datacenter`, but on closer inspection, decided it was safe to remove that guard as @tylerschultz mentioned for the reasons he stated above.

Given that, we are also considering opening a followup PR against the `release-0.7` with this more targeted fix if that would be welcome.

mcwumbly: > Given that, we are also considering opening a followup PR against the release-0.7 with this more targeted fix if that would be welcome.

Here's that other PR: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1216

srm09:
@tylerschultz Could you update the PR with just the test changes, and we will work on getting the deprecated types PR merged soon?

tylerschultz:
The PR has been updated to only contain the test changes.

mcwumbly:
I think we expect this to fail now until https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1095 is merged, so we'll hold off on retesting until then...

mcwumbly:
@srm09 this has been rebased now and checks are passing now. can you take another look?

PR #1090, opened 2021-01-21
Updates image to use nonroot user for CAPV

srm09:
**What this PR does / why we need it**:
Uses nonroot user for CAPV

**Which issue(s) this PR fixes**:
Fixes #1089

**Special notes for your reviewer**:
Porting a similar PR from CAPI

**Release note**:
```release-note
Uses nonroot user for CAPV for successful deployment when PSPs are enabled 
```

PR #1216, opened 2021-07-17
Fix VSphereCluster VCenterAvailable condition

### LINKED ISSUE #1209: Make VCenterAvailable status/condition more meaningful to enable better troubleshooting of issues related to connectivity or misconfiguration

mcwumbly:
We now check the connectivity regardless of whether CloudProviderConfiguration is set on the spec.

That property has been deprecated and it is valuable to check this connectivity early and surface this on the VSphereCluster status rather than just on VSphereVMs.

This also fixes a bug where the status could be set to "False" if the secret was not found and then never properly set to "True" once the secret becomes present.

This fixes #1209 for the `release-0.7` branch. 

This fix was originally proposed against the `master` branch via PR #1214. However, it was noted in that PR that  #1095 already includes a fix for this issue on master so that is likely to be merged instead, at which point #1214 may be closed or just slimmed down to add test coverage.

While `reconcileVCenterConnectivity` accesses the Datacenter value on the deprecated, (but still present) CloudProviderConfiguration spec property, we believe it is always initialized to an empty string and is so is safe to call unconditionally as [it calls](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/7e1a78f2035d64321604a45195572fdc5f72f637/pkg/session/session.go#L96) the [`DatacenterOrDefault` func](https://github.com/vmware/govmomi/blob/cfbe1d59b449be9b5378251bc6e52c65bc5e1cbd/find/finder.go#L394) which handles being called with an empty string.

**Release note**:

```release-note
Fix VSphereCluster VCenterAvailable condition
```

cc @tylerschultz

### DIFF
--- a/controllers/vspherecluster_controller.go
+++ b/controllers/vspherecluster_controller.go
@@ -301,14 +301,12 @@ func (r clusterReconciler) reconcileNormal(ctx *context.ClusterContext) (reconci
 		return reconcile.Result{}, err
 	}
 
-	if cloudProviderConfigurationAvailable(ctx) {
-		if err := r.reconcileVCenterConnectivity(ctx); err != nil {
-			conditions.MarkFalse(ctx.VSphereCluster, infrav1.VCenterAvailableCondition, infrav1.VCenterUnreachableReason, clusterv1.ConditionSeverityError, err.Error())
-			return reconcile.Result{}, errors.Wrapf(err,
-				"unexpected error while probing vcenter for %s", ctx)
-		}
-		conditions.MarkTrue(ctx.VSphereCluster, infrav1.VCenterAvailableCondition)
+	if err := r.reconcileVCenterConnectivity(ctx); err != nil {
+		conditions.MarkFalse(ctx.VSphereCluster, infrav1.VCenterAvailableCondition, infrav1.VCenterUnreachableReason, clusterv1.ConditionSeverityError, err.Error())
+		return reconcile.Result{}, errors.Wrapf(err,
+			"unexpected error while probing vcenter for %s", ctx)
 	}
+	conditions.MarkTrue(ctx.VSphereCluster, infrav1.VCenterAvailableCondition)
 
 	// Reconcile the VSphereCluster's load balancer.
 	if ok, err := r.reconcileLoadBalancer(ctx); !ok {
--- a/controllers/vspherecluster_controller_test.go
+++ b/controllers/vspherecluster_controller_test.go
@@ -24,6 +24,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/vmware/govmomi/simulator"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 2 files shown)


gab-satchi (approved):
changes lgtm. I think we can merge this and trim down the v1alpha4 PR to just the tests once the cleanup PR (#1090) is merged.

mcwumbly:
I'll look closer at these test failures before kicking of retries.

tylerschultz:
This test failed for us locally and then passed again in a later run. I
believe this failure to be a flake.

srm09:
@gab-satchi This will break the e2e tests once we move to using the kube-vip flavorgen in the CI.
@mcwumbly This PR should also include a check for the presence of the `Datacenter` in the `VSphereCluster.Spec.CloudProviderConfiguration.Workspace` spec. If not present, send an empty string instead.

tylerschultz:
Hi @srm09, 

Can you please elaborate on how these changes will break e2e's when kube-vip flavorgen is introduced? Is there something we can do to mitigate future problems?

Regarding the concern over guarding the Datacenter field value, this should not be necessary. There are no pointers in this chain of structs `VSphereCluster.Spec.CloudProviderConfiguration.Workspace.Datacenter`, so there should be no nils to worry about. The default value for a string var is empty string. The code will work fine without guards, and empty string will be sent if the value is not set. Is there something I'm missing?

cc @gab-satchi @mcwumbly

srm09:
> y. There are no pointers in this chain of structs `VSphereCluster.Spec.CloudProviderConfiguration.Workspace.Datacenter`, so there should be no nils to worry about. The default value for a string var is empty string. The code will work fine without guards, and empty string will be sent if the value is not set. Is there something I'm missing?

Yea I misunderstood that part, and an empty string will be sent by default. That should work out fine since in that case, it would use the default datacenter.
And I was wrongly fixating on the CloudProviderConfiguration not being present will cause an issue once we move over to using kube-vip + CRS for the CPI config. But that won't be the case, so it all should work fine. Sorry for the false alarm.

PR #1237, opened 2021-08-26
Proposal for GPU and PCI passthrough support

geetikabatra:
This commit adds a proposal which enables GPU support
in CAPV using.

Signed-off-by: Geetika Batra <geetikab@vmware.com>

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/docs/proposal/20210823-gpu-support.md
+++ b/docs/proposal/20210823-gpu-support.md
@@ -1,0 +1,189 @@
+# GPU and PCI passthrough support in CAPV
+
+```text
+---
+title: GPU and PCI passthrough support in CAPV
+authors:
+  - "@geetikabatra"
+reviewers:
+  - "@vijaykumar"
+  - "@ankitaswamy"
+  - "@sonasingh46"
+  - "@pshail"
+  
+creation-date: 2021-08-23
+last-updated: 2021-08-25
+status: implementable
+```
+
+## Table of Contents
+
+- [GPU and PCI passthrough support in CAPV](#gpu-and-pci-passthrough-support-in-capv)
+  - [Table of Contents](#table-of-contents)
+  - [Glossary](#glossary)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+  - [Proposal](#proposal)
+    - [User Stories](#user-stories)
+      - [Story 1 - using vGPU](#story-1---using-vgpu)
+      - [Story 2 - GPU Direct implementation](#story-2---gpu-direct-implementation)
+      - [Story 3 - PCI passthrough for single node Customer](#story-3---pci-passthrough-for-single-node-customer)
+    - [Requirements](#requirements)
+      - [Functional Requirements](#functional-requirements)
+      - [Non-Functional Requirements](#non-functional-requirements)
+    - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+      - [Proposed Changes](#proposed-changes)
+      - [Current Scope of vendor](#current-scope-of-vendor)
+      - [Additional Webhooks](#additional-webhooks)
+      - [Controller Changes](#controller-changes)
+  - [Additional Details](#additional-details)
+    - [Test Plan](#test-plan)
+      - [Subtasks](#subtasks)
+    - [Graduation Criteria](#graduation-criteria)
+  - [Implementation History](#implementation-history)
+
+## Glossary
+
+- CAPV - An abbreviation of Cluster API Provider vSphere
+
+## Summary
+
+CAPV currently does not support provisioning of GPU-based worker nodes and hence limits the deployment of workloads that needs GPU. This proposal outlines ways to enable GPU support in CAPV. This means that the k8s cluster created via CAPV can have GPU enabled nodes that can be finally consumed by the workloads(pods) that request GPU. The proposed changes will maintain backward compatibility and maintain the existing behaviour without any extra user configuration.
+
+## Motivation
+
+CAPV does not currently support GPU-accelerated workloads which have a limited addressable market. Competitively, all of the hyper-scale cloud providers (AWS, Azure, & GCP) offer GPU-accelerated virtualization and Kubernetes platforms. Many of them additionally offer robust vertically integrated AI/ML solutions building on open source technology. GPU support is a long-requested feature. With this capability, users can deploy GPU accelerated workloads on the Kubernetes cluster provisioned via CAPV
+
+### Goals
+
+1. To allow vGPU support for multi-node clusters.
+2. To allow GPU Direct support for multi-node clusters.
+3. To allow PCI passthrough support for single-node clusters.
+
+## Proposal
+
+### User Stories
+
+#### Story 1 - using vGPU
+
+Stacy works at an IT department of an organization where AI/ML models are run every hour and the entire staff wants to leverage GPU-supported nodes. VGPU's are able to provide close to bare-metal performance. Stacy needs a lot of sharing GPUs.
+vGPU provides a full-fledged cluster from which a GPU can be requested.
+
+#### Story 2 - GPU Direct implementation
+
+Tony is an Engineer at a big organization with multiple centers across the Globe. Tony needs a big GPU cluster to address the needs of his organization. So they intend to use GPU Direct. GPU direct is technically a pool of GPUs connected by a network card in a kind of peer-to-peer network. This is the latest technology and gives the advantage of shared GPUs from a pool. GPU  direct is the technology that helps in linking multiple GPUs. Tony wants to leverage this technology so that it is easier for him to manage resource allocation.
+
+#### Story 3 - PCI passthrough for single node Customer
+
+Alex is an Engineer at a retail organization that requires a single GPU node. They use one node with GPU attached and want to keep things simple. Alex can simply add this GPU-connected machine to the cluster and that should do the job. While selecting nodes, Alex can use appropriate labels to run his AI/ML workload on this particular node. PCI passthrough will provide direct GPU support. The challenge that Alex can face is that using passthrough Alex wouldn't be able to migrate nodes.
+
+### Requirements
+
+#### Functional Requirements
+
+- FR1: CAPV MUST support vGPUs.
+- FR2: CAPV MUST support GPUDirect.
+- FR3: CAPV MUST support PCI Passthrough.
+
+#### Non-Functional Requirements
+
+- NFR1: Unit tests MUST exist for all 3 supports
+- NFR2: e2e tests MUST exist for 3 supports
+
+### Implementation Details/Notes/Constraints
+
+- Using the Vsphere failure domain to dictate where the sphere VM gets created. It needs to have access to GPU; otherwise, cloning will fail. The users' responsibility is to ensure that whatever failure domain they are trying to put the VM on(GPU VM). It must have access to the required physical device.
+
+#### Proposed Changes
+
+```go
+// VirtualMachineCloneSpec is information used to clone a virtual machine.
+type VirtualMachineCloneSpec struct {
+  // PciDevices is the list of pci devices used by the virtual machine.
+  // +optional
+  PciDevices []PCIDeviceSpec `json:"pcidevices"`
+  // VgpuDevices is the list of vgpu devices used by the virtual machine.
+  // +optional
+  VgpuDevices []VGPUDeviceSpec `json:"vgpudevices"`
+}
+//VGPUDeviceSpec defines virtual machine's VGPU configuration
+type VGPUDeviceSpec struct {
+ // ProfileName is the Profile Name of a virtual machine's VGPU, in string.
+ // Defaults to the eponymous property value in the template from which the
+ // virtual machine is cloned.
+ // +optional
+ ProfileName string `json:"profileName,omitempty"`
+}
+//PCIDeviceSpec defines virtual machine's PCI configuration
+type PCIDeviceSpec struct {
+ // DeviceID is the device ID of a virtual machine's PCI, in integer.
+ // Defaults to the eponymous property value in the template from which the
+ // virtual machine is cloned.
+ // +optional
+ DeviceID int32 `json:"deviceId,omitempty"`
+ // VendorId is the vendor ID of a virtual machine's PCI, in integer.
+ // Defaults to the eponymous property value in the template from which the
+ // virtual machine is cloned.
+ // +optional
+ VendorID int32 `json:"vendorId,omitempty"`
+}
+```
+
+#### Current Scope of vendor
+
+- Nvidia
+
+#### Additional Webhooks
+
+- Check if `DeviceID` is present or not in `VSphereMachine template`.
+
+#### Controller Changes
+
+- If vGPU is present, the CR populates the ProfileName, which adds the vGPU to the list of devices.
+- If GPU is present via PCI passthrough, `device ID`, and `vendor ID`. The CR populates the variables respective to VGPU, i.e., GPUProfileName.
+- Similarly, the controller also initializes variables related to Direct devices(for PCI passthrough and GPUDirect). i.e., `GPUDeviceId` and `GPUVendorId`.
+- Controller will call a function `getGpuSpecs` device, which will use the information stated above to determine if that specific GPU is present or not. If it is, then VM with that GPU is spawned.
+- VSphereVMStatus object will return the Device info to the user stating whether the VM is in ready state or in failed state.
+  
+## Additional Details
+
+### Test Plan
+
+- Unit tests for cluster controller to test behaviour when vGPU is requested.
+- Unit tests for cluster controller to test behaviour if vGPU works as expected.
+- Unit tests for cluster controller to test if multiple vGPUs are requested, do all of them behave as expected.
+- Unit tests for cluster controller to test GPU's requested from GPU direct work as expected.
+- Unit tests for cluster controller to test PCI passthrough works as expected.
+- Unit tests for a single node PCI passthrough.
+- E2E tests for GPU enabled workload clusters
+
+#### Subtasks
+
+- Enabling GPU based infrastructure, will be hosted behind a VPN.
+- Setting up webhooks to trigger the GPU based E2E tests.
+- Actual test scenarios
+  - E2E test to create a GPU-enabled cluster with one control plane node and one worker node.
+    - Wait for a node to have a "nvidia.com/gpu" allocatable resource.
+    - E2E test to run a GPU-based calculation.
+    - Run a CUDA vector calculation job
+
+### Graduation Criteria
+
+Alpha
+
+- Support VGPU
+- Support GPU Direct
+- Support PCI PAssthrough
+
+Beta
+
+N/A
+
+Stable
+
+- Two releases since beta.
+
+## Implementation History
+
+- 08/23/2021: Initial Proposal


docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,200 @@
...
+
+#### Story 2 - GPU Direct implementation
+
+Tony is an Engineer at a big organization with multiple centers across the Globe. Tony needs a big GPU cluster to address the needs of his 
+organization. So they intend to use GPU Direc. GPU direct is  technically a pool of GPU’s connected by a network card in a kind of  peer to peer network. This is the latest technology and gives advantage of shared GPU’s from a pool. GPU  direct is the technology which helps in linking multiple GPU’s. Tony wants to leverage this technology so that it is easier for him to manage resource allocation.
+
+
+#### Story 3 - PCI passthrough for single node Customer
+
+Alex is an Engineer at retail organization that requires single GPU node. They use one node with GPU attached and want to keep things simple. Alex can simply add this GPU connected machine to the cluster and that should do the job. While selecting nodes, Alex can use appropriate labels to run his AI/ML workload on this particular node. PCI passthrough will provide direct GPU support. Challenges that Alex can face is that using passthrough Alex wouln't be able to migrate nodes.
```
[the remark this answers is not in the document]
geetikabatra: @MaxRink I don't get your question here. Non GPU will be counted among regular nodes in my opinion.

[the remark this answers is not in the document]
MaxRink: I mean passthrough for PCI-E devices that are not GPUs, as PCI-E passthrough shouldnt really care if its a GPU, a NIC or even an HBA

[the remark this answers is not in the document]
geetikabatra: Yeah, If we put it that way, yes. It also applies for non GPU devices.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,143 @@
...
+### Implementation Details/Notes/Constraints
+
+#### Controller Changes
+
+```go
+// vGPUProfileName is the profileName of the vGPU, in string.
+// Defaults to the eponymous property value in the template from which the
+// virtual machine is cloned.
+// +optional
+VGPUProfileName string `json:"vGPUProfileName,omitempty"`
```
gab-satchi, line 100: Missing some info on how this string will get integrated into the existing API types. Also, in order to support PCI passthrough, we'd need more information. From the [docs](https://vdc-download.vmware.com/vmwb-repository/dcr-public/b50dcbbf-051d-4204-a3e7-e1b618c1e384/538cf2ec-b34f-4bae-a332-3820ef9e7773/vim.vm.device.VirtualPCIPassthrough.AllowedDevice.html) the `deviceId` and `vendorId` are required

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,164 @@
...
+```go
+// vGPUProfileName is the Profile Name of a virtual machine's vGPU, in string.
+// Defaults to the eponymous property value in the template from which the
+// virtual machine is cloned.
+// +optional
+VGPUProfileName string `json:"vGPUProfileName,omitempty"`
+// vGPUDeviceId is the device ID of a virtual machine's vGPU, in integer.
+// Defaults to the eponymous property value in the template from which the
+// virtual machine is cloned.
+// +optional
```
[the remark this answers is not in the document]
pshail: The optional fields of revisionId, subDeviceId, subVendorId should also be included

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,164 @@
...
+    the template from which the virtual machine is cloned.
+  format: int32
+  type: integer
+```
+
+## Additional Details
+
+### Test Plan
+
+* Unit tests for cluster controller to test behavior vGPU is requested.
```
pshail, line 137:
```suggestion
* Unit tests for cluster controller to test behavior when vGPU is requested.
```

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,164 @@
...
+
+```text
+---
+title: GPU support CAPV
+authors:
+  - "@geetikabatra"
+reviewers:
+  - "@vijaykumar"
+  - "@ankitaswamy"
+  - "@ashutoshkumar"
```
sonasingh46, line 11: If you want to put it, my github ID is : sonasingh46
Hence,
```
@ashutoshkumar ---> @sonasingh46
```

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,164 @@
...
+## Glossary
+
+* CAPV - An abbreviation of Cluster API Provider vSphere
+
+## Summary
+
+CAPV currently does not support GPU-based infrastructure.
+
+This proposal outlines ways to provision GPU support in CAPV. The proposed changes will maintain backward compatibility and maintain the existing behavior without any extra user configuration.
+
```
sonasingh46, line 49: n_p:
To make these lines more clear, I was thinking to put some lines explaining what does GPU support in CAPV means?

e.g. 

```
CAPV currently does not support provisioning of GPU-based worker nodes and hence limits the deployment of workloads that needs GPU 

This proposal outlines ways to enable GPU support in CAPV. This means that the k8s cluster created via CAPV can have GPU enabled nodes that can be finally consumed by the workloads(pods) that request GPU.   The proposed changes will maintain backward compatibility and maintain the existing behaviour without any extra user configuration.

"As an end user, I want to deploy GPU accelerated workloads on the Kubernetes cluster created via CAPV

```

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,164 @@
...
+#### Story 2 - GPU Direct implementation
+
+Tony is an Engineer at a big organization with multiple centers across the Globe. Tony needs a big GPU cluster to address the needs of his organization. So they intend to use GPU Direct. GPU direct is technically a pool of GPUs connected by a network card in a kind of peer-to-peer network. This is the latest technology and gives the advantage of shared GPUs from a pool. GPU  direct is the technology that helps in linking multiple GPUs. Tony wants to leverage this technology so that it is easier for him to manage resource allocation.
+
+#### Story 3 - PCI passthrough for single node Customer
+
+Alex is an Engineer at a retail organization that requires a single GPU node. They use one node with GPU attached and want to keep things simple. Alex can simply add this GPU-connected machine to the cluster and that should do the job. While selecting nodes, Alex can use appropriate labels to run his AI/ML workload on this particular node. PCI passthrough will provide direct GPU support. The challenge that Alex can face is that using passthrough Alex wouldn't be able to migrate nodes.
```
sonasingh46, line 76: A non-goal item but pasting here for reference. 
https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/#node-labeller

sonasingh46: >The challenge that Alex can face is that using passthrough Alex wouldn't be able to migrate nodes.
Trying to understand this part. What is meant by node migration here. 

I am thinking the following: 
If a node for example, `node-1` has GPU available via PCI passthrough and for some reason this node cannot be used, can't we bring up a new node `node-1'` and make a GPU available to it and then de-comission the previous `node-1`. 
The workloads should actually get migrated automatically to the newer nodes as we will have the same  GPU label

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,164 @@
...
+4. To allow PCI passthrough support for single-node clusters.
+
+## Proposal
+
+### User Stories
+
+#### Story 1 - using vGPU
+
+Stacy works at an IT department of an organization where AI/ML models are run every hour and the entire staff wants to leverage GPU-supported nodes. VGPU's are able to provide close to bare-metal performance. Stacy needs a lot of sharing GPUs.
+vGPU provides a full-fleshed cluster from which a virtual GPU can be requested.
```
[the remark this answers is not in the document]
sonasingh46: Requesting a vGPU does not necessarily mean that it can be shared across, right? 
One can still request a vGPU and it may not be shareable.  
For example, in the below doc ( specific to Nvidia ), I can see that we can request vGPU in multiple modes and one of the mode can allow vGPU to be shared. 
https://docs.nvidia.com/grid/latest/grid-vgpu-user-guide/index.html

The question I am trying to ask is that, when we say that we have a vGPU available, does it inherently mean that it is shareable? 
Are we trying to say that from CAPV end user prespective, when a user request for a vGPU that always means it is shareable ( we make sure from out implementation that we provision vGPU in a shareable mode ), which sounds fine to me.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,164 @@
...
+
+#### Story 2 - GPU Direct implementation
+
+Tony is an Engineer at a big organization with multiple centers across the Globe. Tony needs a big GPU cluster to address the needs of his organization. So they intend to use GPU Direct. GPU direct is technically a pool of GPUs connected by a network card in a kind of peer-to-peer network. This is the latest technology and gives the advantage of shared GPUs from a pool. GPU  direct is the technology that helps in linking multiple GPUs. Tony wants to leverage this technology so that it is easier for him to manage resource allocation.
```
sonasingh46, line 72: vGPU and GPU direct seems to be available by Nvidia. Is this vendor dependent? Are there other providers also providing this tech/feature? 
These questions may help design the CR spec schema in my opinion.

docs/proposal/20210823-gpu-support.md
sonasingh46: As we are talking about vGPU, GPU direct also, these needs to covered in the schema and possibly in a generic way. What are your thoughts? 
According to the [docs](https://blogs.vmware.com/apps/2018/09/using-gpus-with-virtual-machines-on-vsphere-part-3-installing-the-nvidia-grid-technology.html): 
```
A vGPU profile allows you to assign a GPU solely to one virtual machine’s use or to be used in a shared mode with others.
```
That means `VGPUProfileName ` does not fit with other mode i.e. `pci-passthrough` or `gpu-direct`

Ankitasw: @sonasingh46 i think we need to define different schema for different techniques, as defined in vm-operator https://github.com/vmware-tanzu/vm-operator/blob/main/config/crd/bases/vmoperator.vmware.com_virtualmachineclasses.yaml#L31-L38
Not sure though, this would be applicable only when each mode has different fields.

sonasingh46: >i think we need to define different schema for different techniques
Sure.  
I meant is those schemas needs to be present in the CR spec. ( The bigger CR schema ) 
Something like this https://github.com/vmware-tanzu/vm-operator-api/blob/master/api/v1alpha1/virtualmachineclass_types.go#L55

Also, while we are designing the schema, it definitely helps to draw inspiration from the vm-operator api types and it is possible that we might come up with slight different approach for the schema for GPU after discussions. 
Open to discussions.

docs/proposal/20210823-gpu-support.md
geetikabatra: @sonasingh46. I think it is self-explanatory.  But, if you strongly feel I should rephrase it. I would certainly do that.

docs/proposal/20210823-gpu-support.md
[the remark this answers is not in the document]
geetikabatra: Inherently it's not shareable. Sharing is enabled inside of physical hardware using one of the above techniques you pointed out, which is the user's own choice. We assume it to be just a piece of GPU(shareable or non-shareable).  Sharing should be automatically handled by the vGPU instance(Need to double-check that).

docs/proposal/20210823-gpu-support.md
geetikabatra: This is indeed vendor-dependent.  This proposal is keeping in mind Nvidia GPUs and it's specific operator.  Intel has an entirely different architecture and method. Currently, we are targeting Nvidia.

docs/proposal/20210823-gpu-support.md
sonasingh46: I was only confused by `provision GPU support in CAPV`. It is just a nit pick from me. Let us see what others have to say. Should be fine.

docs/proposal/20210823-gpu-support.md
geetikabatra: I have removed 'v' from the variable names as this was not Vgpu specific.

docs/proposal/20210823-gpu-support.md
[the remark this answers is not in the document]
geetikabatra: The keyword optional here implies that the node might or might not have GPU attached.
Also, the implementation ignores the fields `revisionId, subDeviceId, subVendorId`. This is debatable though. I am open to discussion regarding this.

geetikabatra:
@gab-satchi I have added the steps of implementation as part of. controller changes as we discussed last week. Please take a look once. Thanks in advance.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,163 @@
...
+## Motivation
+
+CAPV does not currently support GPU-accelerated workloads which have a limited addressable market. Competitively, all of the hyper-scale cloud providers (AWS, Azure, & GCP) offer GPU-accelerated virtualization and Kubernetes platforms. Many of them additionally offer robust vertically integrated AI/ML solutions building on open source technology. GPU support is a long-requested feature. With this capability, users can leverage GPU support.
+
+### Goals
+
+1. To enable GPU Support on Vsphere Kubernetes Cluster.
+2. To allow vGPU support for multi-node clusters.
+3. To allow GPU Direct support for multi-node clusters.
+4. To allow PCI passthrough support for single-node clusters.
```
gab-satchi, line 61: Does PCI passthrough necessarily limit things to a single node cluster? I also don't think goal 1 is needed as 2-4 covers that.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Proposal
+
+### User Stories
+
+#### Story 1 - using vGPU
+
+Stacy works at an IT department of an organization where AI/ML models are run every hour and the entire staff wants to leverage GPU-supported nodes. VGPU's are able to provide close to bare-metal performance. Stacy needs a lot of sharing GPUs.
+vGPU provides a full-fledged cluster from which a virtual GPU can be requested.
```
gab-satchi, line 70: `vGPU provides a full-fledged cluster from which a virtual GPU can be requested`

I don't really follow this. Isn't vGPU = virtual GPU?

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,163 @@
...
+}
+```
+
+#### Controller Changes
+
+* If VGPU or PCI device is present, the CRD is fetched and used to identify the expected devices.
+* The controller initiallizes variables respective to VGPU i.e `GPUProfileName`.
+* Similarly, controller also initializes variables related to Direct devices(for PCI passthrough and GPUDirect) ie. `GPUDeviceId` and `GPUVendorId`.
+* Using the above variables, respective objects of backing info of type `vimTypes.VirtualPCIPassthroughVmiopBackingInfo` with `{Vgpu: <GPUProfileName>}` for VGPU's and `vimTypes.VirtualPCIPassthroughAllowedDevice` for GPUDirect should be created by the controller.
+* Controller will call a function called as createPCI device which will use the information in above step to create objects of type `vimTypes.VirtualPCIPassthrough`
```
gab-satchi, line 124: should the VirtualMachineClone spec contain an array of PCI devices?

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,163 @@
...
+  // Defaults to the eponymous property value in the template from which the
+  // virtual machine is cloned.
+  // +optional
+  GPUVendorId int32 `json:"GPUVendorId,omitempty"`
+}
+```
+
+#### Controller Changes
+
+* If VGPU or PCI device is present, the CRD is fetched and used to identify the expected devices.
```
gab-satchi, line 120: which CRD will get fetched here?

docs/proposal/20210823-gpu-support.md
geetikabatra: Not sure about it. But there was a discussion going on that we need a Mellanox Network Operator for multi-node.

docs/proposal/20210823-gpu-support.md
geetikabatra: Fixing it.

docs/proposal/20210823-gpu-support.md
geetikabatra: I have added the fields into a GPU struct and loaded it inside of VirtualMachineClone as a slice.

geetikabatra:
@yastij, @gab-satchi, and @srm09. Please take a look.

docs/proposal/20210823-gpu-support.md
geetikabatra: I have rephrased this  since no new CRD would be fetched and existing CRD's would be used.

MaxRink:
> I think we should separate out PCI passthrough and vGPU support into two distinct APIs, and place everything related to PCI passthrough in a PCI passthrough spec instead of forcing people to use GPUs for that.

Yes, i also view that as a good idea, given that PCI Passthrough behaves quite different.
I'm also not sure how well the currently proposed approach of needing to specify Vendor and PCIE ID will work, given that PCIE ID depends on Slot placement, which can vary wildly between server models.
We might need some decoupling / autodiscover logic here, otherwise the UX for MDs with PCI-E Devices might be suboptimal.

Ankitasw:
> We might need some decoupling / autodiscover logic here, otherwise the UX for MDs with PCI-E Devices might be suboptimal.
Yes, that's a valid point, I think we need to look at this part during our implementation if we need some autodiscover logic (which at this point I am not sure).
@yastij @gab-satchi do you have any comment for @MaxRink

MaxRink:
> > We might need some decoupling / autodiscover logic here, otherwise the UX for MDs with PCI-E Devices might be suboptimal.
> 
> Yes, that's a valid point, I think we need to look at this part during our implementation if we need some autodiscover logic(which at this point I am not sure). @yastij @gab-satchi do you have any comment for @MaxRink

The other option im currently thinking about would be something following the pattern used by CSI.
Defining a deviceClass ( for example a mellanox NIC, RTX 3090 or whatever PCIe devie you want) and then just referencing that deviceClass instead of an individual Device. 
That might be the easiest from an UX perspective as mapping VMs to physical servers will become quite a constraint to take care of.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+}
+
+//PCIDeviceSpec defines virtual machine's PCI configuration
+type PCIDeviceSpec struct {
+
+ // DeviceId is the device ID of a virtual machine's PCI, in integer.
+ // Defaults to the eponymous property value in the template from which the
+ // virtual machine is cloned.
+ // +optional
+ DeviceId int32 `json:"deviceId,omitempty"`
```
yastij, line 139: nit: `DeviceID`

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+ // VendorId is the vendor ID of a virtual machine's PCI, in integer.
+ // Defaults to the eponymous property value in the template from which the
+ // virtual machine is cloned.
+ // +optional
+ VendorId int32 `json:"vendorId,omitempty"`
```
yastij, line 145: nit: `VendorID`

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+```go
+
+// VirtualMachineCloneSpec is information used to clone a virtual machine.
+type VirtualMachineCloneSpec struct {
+  // Vgpu is the Vgpu configuration for this machine's VM.
+ // +optional
+ Vgpu []VGPUSpec `json:"vgpu,omitempty"`
+ // Pci is the Pci configuration for this machine's VM.
+ // +optional
+ Pci []PCISpec `json:"pci,omitempty"`
```
yastij, line 108: why don't we instead have something like this?
```golang
PCIDevices []PCIDeviceSpec `json:"pciDevices,omitempty"`
```

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+}
+
+//PCISpec defines virtual machine's PCI configuration
+type PCISpec struct {
```
yastij, line 112: I don't think we need this struct here

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### Proposed Changes
+
+```go
+
+// VirtualMachineCloneSpec is information used to clone a virtual machine.
+type VirtualMachineCloneSpec struct {
+  // Vgpu is the Vgpu configuration for this machine's VM.
+ // +optional
+ Vgpu []VGPUSpec `json:"vgpu,omitempty"`
```
yastij, line 105: why don't we instead have something like this?
```golang
VGPUDevices []VGPUDeviceSpec `json:"vgpuDevices,omitempty"` 
```

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+}
+
+//PCISpec defines virtual machine's PCI configuration
+type PCISpec struct {
+ // Devices is the list of pci devices used by the virtual machine.
+ PciDevices []PCIDeviceSpec `json:"pcidevices"`
+}
+
+//VGPUSpec defines virtual machine's VGPU configuration
+type VGPUSpec struct {
```
yastij, line 118: I don't think we need this struct

yastij (changes requested):
did an API review, a few comments to fix before merging. I also Agree that splitting vgpu and pci made things clearer. Thanks !

docs/proposal/20210823-gpu-support.md
srm09: Since this implementation is vendor dependent, we should add a section around saying this implementation is specific to Nvidia.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+ // VendorId is the vendor ID of a virtual machine's PCI, in integer.
+ // Defaults to the eponymous property value in the template from which the
+ // virtual machine is cloned.
+ // +optional
+ VendorId int32 `json:"vendorId,omitempty"`
+}
+
+```
+
+#### Controller Changes
```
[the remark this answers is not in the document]
srm09: From my POV, it would make more sense to the reader to list out the changes required for each GPU support type.
For example:
for vGPU, populate the ProfileName and add to the list of devices.
for PCI passthrough, populate the device and vendor IDs and add to the list of devices.
and so on.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+- If VGPU is present, VGPU's `ProfileName` would be fetched from `VirtualMachineCloneSpec` and device object (`createPCIPassThroughDevice` from govmomi library) is formed using this VGPU `ProfileName`. This device object is added to
+- the slice of devices and then used by `vspheremachine_controller` to schedule nodes.
```
srm09, line 153: The VM placement is dictated by the failureDomain present on the VirtualMachineSpec (maybe placement on a particular host or group op hosts), so what happens if the GPUProfile is not available on that host.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+- Using the above variables, respective objects of backing info of type `vimTypes.VirtualPCIPassthroughVmiopBackingInfo` with `{Vgpu: <GPUProfileName>}` for VGPU's and `vimTypes.VirtualPCIPassthroughAllowedDevice` for GPUDirect should be created by the controller.
+- Controller will call a function called as createPCI device which will use the information in above step to create objects of type `vimTypes.VirtualPCIPassthrough`
+- The create PCI device function will also be responsible for comparing the current devices with the expected devices.
+- The create method will return a slice `deviceChanges` of type `vimTypes.BaseVirtualDeviceConfigSpec`.
+- If there are no current devices and new devices are added. The slice `devicechanges` would reflect only add operations which would be equivalent to the number of devices added.
+- If the current devices are same as proposed devices. The length of the device changes remains 0.
+- Adding Vgpu devices and PCI devices with differnet backing info would return one remove and one add operation for each device.
+So the length of device changes would be 2*devices with different backing info.
```
srm09, line 164: Since the VSphereMachineTemplate and VSphereMachine objects are immutable, how would this case arise? Why would vGPU or PCI devices with different backing info be needed?

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+## Additional Details
+
+### Test Plan
+
+- Unit tests for cluster controller to test behavior when vGPU is requested.
+- Unit tests for cluster controller to test behavior if vGPU works as expected.
+- Unit tests for cluster controller to test if multiple vGPU's are requested, do all of them behave as expected.
+- Unit tests for cluster controller to test GPU's requested from GPU direct work as expected.
+- Unit tests for cluster controller to test PCI passthrough works as expected.
+- E2E test with a GPU Direct cluster running in different availability zones.
```
srm09, line 175: Upstream CI does not have GPU support, so running E2E tests would not be possible.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+- Unit tests for a single node PCI passthrough.
+- E2E test with the PCI passthrough.
```
srm09, line 177: same as above

srm09 (commented):
One of the missing things here is, changes to the `VSphereVMStatus` object to bubble the Device info to the user.
Another thing to take into consideration is if any new Conditions are needed to surface error scenarios or maybe just new reasons which can be set for existing conditions if any error happens during VM creation.

docs/proposal/20210823-gpu-support.md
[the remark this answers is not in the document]
srm09: @yastij Does it make sense to separate out the PCI Passthrough support to a separate proposal? This could also enable the workflow of provisioning VMs physical NICs attached to the VMs created by CAPV.

And if it is too much work to separate the proposal out, I +1 Naadir's comments on retitling the proposal since PCI passthrough support is not strictly GPU related.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+}
+
+```
+
+#### Controller Changes
+
+- If VGPU is present, VGPU's `ProfileName` would be fetched from `VirtualMachineCloneSpec` and device object (`createPCIPassThroughDevice` from govmomi library) is formed using this VGPU `ProfileName`. This device object is added to
+- the slice of devices and then used by `vspheremachine_controller` to schedule nodes.
+- If PCI device is present, PCI device's `DeviceId` and `VendorId` would be fetched from `VirtualMachineCloneSpec` and device object(`createPCIPassThroughDevice` from govmomi library) is formed using this information and then used by the `vspheremachine_controller` to schedule nodes.
+- The controller initiallizes variables respective to VGPU i.e `GPUProfileName`.
```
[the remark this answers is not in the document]
geetikabatra: It basically is another way of saying that `GPUProfile name` is now set and can be used.

docs/proposal/20210823-gpu-support.md
geetikabatra: Need a broader discussion on this.

docs/proposal/20210823-gpu-support.md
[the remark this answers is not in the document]
geetikabatra: @srm09 I added a detailed overview of the work as Gab had asked me to do so a couple of months back. I can remove it if you want.

docs/proposal/20210823-gpu-support.md
geetikabatra: Just curious, how does Network Devices work. What I see is, they call a `Finder.Network` in `govmomi` and just return that list of devices. In the case of GPU's there is no such method as `finder.PCIDevices`. All the workflow above will not be required if we have a list of GPU devices. 

P.S If above stands true, then I guess, I have to implement a PCI finder in govmomi.

docs/proposal/20210823-gpu-support.md
geetikabatra: Need a broader discussion on this.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,197 @@
...
+ DeviceId int32 `json:"deviceId,omitempty"`
+
+ // VendorId is the vendor ID of a virtual machine's PCI, in integer.
+ // Defaults to the eponymous property value in the template from which the
+ // virtual machine is cloned.
+ // +optional
+ VendorId int32 `json:"vendorId,omitempty"`
+}
+
+```
```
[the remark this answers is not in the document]
geetikabatra: Done!

docs/proposal/20210823-gpu-support.md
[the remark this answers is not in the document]
geetikabatra: /rephrase

docs/proposal/20210823-gpu-support.md
geetikabatra: If we are using the Vsphere failure domain to dictate where the sphere VM gets created. It needs to have access to GPU, otherwise cloning will fail. It is users' responsibility to make sure that whatever failure domain they are trying to put the VM on(GPU VM). It must have the access to the required physical device.

docs/proposal/20210823-gpu-support.md
geetikabatra: //need to file an issue in govmomi

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,170 @@
...
+ // virtual machine is cloned.
+ // +optional
+ GPUVendorId int32 `json:"GPUVendorId,omitempty"`
+}
+
+```
+
+#### Controller Changes
+
+- If vGPU is present, the CR populates the ProfileName, which adds the vGPU to the list of devices.
```
geetikabatra, line 136: @srm09 PTAL

geetikabatra:
@srm09 PTAL

docs/proposal/20210823-gpu-support.md
srm09: Lets create an issue to track the E2E testing for GPU separately. Whatever mechanism we end up taking to run the tests on GPU specialized hardware hosted behind a VPN, we should use the issue to describe the solution and update the progress.

docs/proposal/20210823-gpu-support.md
srm09: This seems to be taken care of by https://github.com/vmware/govmomi/pull/2772. Resolving this

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,177 @@
+# GPU support in CAPV
```
srm09, line 1: We had discussed updating the title of the proposal to explicitly call out the PCI Passthrough support, could you include that here?

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,177 @@
...
+
+```text
+---
+title: GPU support in CAPV
```
srm09, line 5: Same as above

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,177 @@
...
+    - [Graduation Criteria](#graduation-criteria)
+  - [Implementation History](#implementation-history)
+
+## Glossary
+
+- CAPV - An abbreviation of Cluster API Provider vSphere
+
+## Summary
+
+CAPV currently does not support GPU-based infrastructure.
```
srm09, line 49: nit: remove this line, the below paragraph talks about the same thing.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,177 @@
...
+
+CAPV currently does not support provisioning of GPU-based worker nodes and hence limits the deployment of workloads that needs GPU. This proposal outlines ways to enable GPU support in CAPV. This means that the k8s cluster created via CAPV can have GPU enabled nodes that can be finally consumed by the workloads(pods) that request GPU. The proposed changes will maintain backward compatibility and maintain the existing behaviour without any extra user configuration. "As an end user, I want to deploy GPU accelerated workloads on the Kubernetes cluster created via CAPV
```
srm09, line 51:
```suggestion
CAPV currently does not support provisioning of GPU-based worker nodes and hence limits the deployment of workloads that needs GPU. This proposal outlines ways to enable GPU support in CAPV. This means that the k8s cluster created via CAPV can have GPU enabled nodes that can be finally consumed by the workloads(pods) that request GPU. The proposed changes will maintain backward compatibility and maintain the existing behaviour without any extra user configuration.
```

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Motivation
+
+CAPV does not currently support GPU-accelerated workloads which have a limited addressable market. Competitively, all of the hyper-scale cloud providers (AWS, Azure, & GCP) offer GPU-accelerated virtualization and Kubernetes platforms. Many of them additionally offer robust vertically integrated AI/ML solutions building on open source technology. GPU support is a long-requested feature. With this capability, users can leverage GPU support.
```
srm09, line 55:
```suggestion
CAPV does not currently support GPU-accelerated workloads which have a limited addressable market. Competitively, all of the hyper-scale cloud providers (AWS, Azure, & GCP) offer GPU-accelerated virtualization and Kubernetes platforms. Many of them additionally offer robust vertically integrated AI/ML solutions building on open source technology. GPU support is a long-requested feature. With this capability, users can deploy GPU accelerated workloads on the Kubernetes cluster provisioned via CAPV
```

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,177 @@
...
+
+- FR1: CAPV MUST support vGPUs.
+- FR2: CAPV MUST support GPUDirect.
+- FR3: CAPV MUST support PCI Passthrough.
+
+#### Non-Functional Requirements
+
+- NFR1: Unit tests MUST exist for all 3 supports
+
+### Implementation Details/Notes/Constraints
```
srm09, line 92: Add a section for **Constraints** and call out the discussion in the [comment](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1237#discussion_r813153403) around placement of VMs when using Failure Domains.

docs/proposal/20210823-gpu-support.md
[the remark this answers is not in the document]
srm09: @geetikabatra do not see the validations listed out in the proposal

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,177 @@
...
+
+#### Proposed Changes
+
+```go
+
+// VirtualMachineCloneSpec is information used to clone a virtual machine.
+type VirtualMachineCloneSpec struct {
+  // GPU is the GPU configuration for this machine's VM.
```
srm09, line 100: nit
```suggestion
  // GPU is the GPU configuration for the virtual machine.
```

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,177 @@
...
+- Unit tests for cluster controller to test GPU's requested from GPU direct work as expected.
+- Unit tests for cluster controller to test PCI passthrough works as expected.
+- Unit tests for a single node PCI passthrough.
+- Enabling GPU based infrastructure.
+- Setting up webhooks to trigger the GPU based E2E tests.
+- E2E test to create a GPU-enabled cluster with one control plane node and one worker node.
+  - Subtasks:
+    - Wait for a node to have a "nvidia.com/gpu" allocatable resource.
+- E2E test to run a GPU-based calculation.
+  - Run a CUDA vector calculation job
```
srm09, line 157:
```suggestion
- E2E tests for GPU enabled workload clusters
   #### Subtasks:
   - Enabling GPU based infrastructure, will be hosted behind a VPN.
   - Setting up webhooks to trigger the GPU based E2E tests.
   - Actual test scenarios
      - E2E test to create a GPU-enabled cluster with one control plane node and one worker node.
      - Wait for a node to have a "nvidia.com/gpu" allocatable resource.
      - E2E test to run a GPU-based calculation.
      - Run a CUDA vector calculation job
```

srm09 (changes requested):
It is missing the updates to the `VSphereVMStatus` to provide the Device info to the user which was called out in [this](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1237#pullrequestreview-872529095) comment. 
I do not see the validation webhook changes in the final proposal, was [this change](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1237#discussion_r811872980) overridden?

Could you please add the two things mentioned above and also take a look at the open conversations and resolve them if appropriate? 
Thanks for all the work on the proposal, it is almost there! 👍🏾

docs/proposal/20210823-gpu-support.md
geetikabatra: Done!

docs/proposal/20210823-gpu-support.md
geetikabatra: Noted this one as a work item. Marking this resolved.

docs/proposal/20210823-gpu-support.md
```diff
@@ -0,0 +1,199 @@
...
+ // +optional
+ VendorID int32 `json:"vendorId,omitempty"`
+}
+```
+
+#### Current Scope of vendor
+
+- Nvidia
+
+#### Additional Webhooks
```
geetikabatra, line 147: @srm09 This section says all about validations.

docs/proposal/20210823-gpu-support.md
geetikabatra: Had updated an outdated version in my previous commit. Updated the new one.

docs/proposal/20210823-gpu-support.md
srm09: +1 to what Yassine has said here ^

docs/proposal/20210823-gpu-support.md
srm09: The VGPUSpec adds nothing apart from a wrapper around the VGPUDeviceSpec. Makes sense to remove this too. The NetworkSpec comprises of a bunch of other fields which are a part of the Spec, hence having the wrapper struct makes more sense in that case. And this approach applies to all the other structs above.

srm09 (commented):
One final set of changes. @yastij can you take a look at it too?

docs/proposal/20210823-gpu-support.md
geetikabatra: Done!

docs/proposal/20210823-gpu-support.md
geetikabatra: Done!

docs/proposal/20210823-gpu-support.md
geetikabatra: Removed!

docs/proposal/20210823-gpu-support.md
geetikabatra: Done!

geetikabatra:
@yastij and @srm09 This can be merged now!

geetikabatra:
> One of the missing things here is, changes to the `VSphereVMStatus` object to bubble the Device info to the user. Another thing to take into consideration is if any new Conditions are needed to surface error scenarios or maybe just new reasons which can be set for existing conditions if any error happens during VM creation.

This is addressed in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1237/files#diff-da5c64bcd81497eebce388a7c1e0250d75fd759fb2131341546691e3d7475e41R147

PR #1239, opened 2021-08-26
Add feature gate machinery

geetikabatra:
Currently CAPV doesn't have cpability of adding features via
feature gates. This change would add feature gate machinery.

Signed-off-by: Geetika Batra <geetikab@vmware.com>

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -1,0 +1,42 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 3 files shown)


feature/feature.go
```diff
@@ -0,0 +1,42 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
gab-satchi, line 2: nit: should be 2021

feature/gates.go
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
gab-satchi, line 2: nit: should be 2021

yastij (commented):
/approve

I'll leave lgtm to @gab-satchi

feature/feature.go
geetikabatra: Done.

PR #1274, opened 2021-10-20
Assign datacenter only if specified

### LINKED ISSUE #1266: VSphereCluster reconciliation fails for VC with multiple Datacenters
/kind bug

Using a VC with multiple datacenters configured causes the VSphereCluster reconciliation to error out with the following error:
```
1007 20:30:16.008439       1 controller.go:302] controller/vspherecluster "msg"="Reconciler error" "error"="unexpected error while probing vcenter for infrastructure.cluster.x-k8s.io/v1alpha4, Kind=VSphereCluster vsphere-tsttmdc-bn/tardis-1: unable to find datacenter \"\": default datacenter resolves to multiple instances, please specify" "name"="foocluster" "namespace"="foo-ns" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="VSphereCluster" 
```

During the VSphereCluster reconciliation loop, the presence/connectivity of the provided vCenter endpoint is verified. As part of the verification loop, CAPV creates a session object which gets reused for future usage. As part of the v1alpha4 move, we deprecated the Cloud Provider configuration that was a part of the VSphereCluster Spec field. The cloud provider configuration had a **mandatory** datacenter field that was used during session creation. In the absence of the datacenter value, the session creation logic looks for the default datacenter which cannot be correctly resolved when there is more than one DC instance. This causes the finder logic to barf the above error and return the reconcile loop with an error.

Cluster reconciliation should continue successfully without the above error message.

A possible fix would be to amend the vCenter connectivity logic to not rely on the presence of the Datacenter field. Instead of creating the session object and storing it in the cache, we can just test whether the vCenter endpoint & credentials are valid and exit early. 

**Environment:**

- Cluster-api-provider-vsphere version:  v0.8.1
- Kubernetes version: (use `kubectl version`): n/a
- OS (e.g. from `/etc/os-release`): n/a

farodin91:

Fixes #1266

**Release note**:

```release-note
Assign datacenter only if specified
```

### DIFF
--- a/controllers/vspheredeploymentzone_controller_domain_test.go
+++ b/controllers/vspheredeploymentzone_controller_domain_test.go
@@ -33,6 +33,10 @@ import (
 	"sigs.k8s.io/cluster-api-provider-vsphere/test/helpers"
 )
 
+const (
+	defaultDatacenter string = "DC0"
+)
+
 func TestVsphereDeploymentZoneReconciler_Reconcile_VerifyFailureDomain(t *testing.T) {
 	t.Run("for Compute Cluster Zone Failure Domain", ForComputeClusterZone)
 	t.Run("for Host Group Zone Failure Domain", ForHostGroupZone)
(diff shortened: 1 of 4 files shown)


farodin91:
@gab-satchi `make test` did pass locally now on my machine.

PR #1284, opened 2021-10-28
🐛 Adds deletion logic for orphaned VSphereMachines

srm09:
Deletion of orphaned VSphereMachine (VSphereMachine object without Machine object owner references) during VSphereCluster object deletion.

Fixes #

This is a special scenario wherein the Machine objects are deleted before they can set owner references on the VSphereMachine object. This leads to the object being orphaned and the VSphereCluster object deletion is stuck forever.  

**Release note**:

```release-note

```


yastij:
rebased the PR

PR #1303, opened 2021-11-09
Update golangci-lint to 1.43.0

stmcginnis:

This updates the version of golangcli-lint we are using to 1.43.0. This
version picks up linter releases that address issues when running under
go 1.17. This prepares us for eventually raising the required golang
version for the project and makes sure developers that have newer
versions installed can run local tests without issues.

The newer versoin of golangci-lint brings with it additional checks that
were not enforced before. Due to that, in addition to raising the
golangci-lint version this also fixes a lot of (most minor) things that
the newer version uncovered.

This touches a *very large number* of files. In almost all cases though,
the changes in these files are very minor cosmetic updates.

There are several linters that are disabled. Most of these would
require even more modifications to the existing codebase with little
benefit to correctness. We can remove them from the exclusion list
at a future date if we find one that would be useful to enforce, then
have a more focused PR to address its issues.

Can be tested locally by running `make lint-go-full` along with the other CI tests.

**Release note**:

```release-note
NONE
```


.golangci.yml
```diff
@@ -1,17 +1,37 @@
 linters:
   enable-all: true
   disable:
+    - contextcheck
```
sbueringer, line 4: @stmcginnis @srm09 WDYT could we maybe change this to specify the list of enabled linters instead?
I think it would make it way easier to see which linters are enabled (and also potentially to compare it with other providers to pick up new linters)

stmcginnis: We could, but my fear would be we would miss some useful linters that we just aren't aware of yet. Enabling all of them and then disabling explicitly helps, in my opinion, by making it explicit what we are deciding we do not want to lint on.

If/when we upgrade the version of golangci-lint that we're using, it has the risk that it pulls in more linters that we were not enforcing before. But that can be useful to realize there is maybe something we should consider that would improve our code quality. Personally, I would rather go through that and make the explicit decision that we are not going to enforce something, than to risk not knowing about issues because we're ignoring everything by default.

Happy to change this though if the general consensus is to do it the other way.

sbueringer: Yup agree, your perspective sounds very reasonable. It's essentially a trade-off between "seeing" what's enabled vs missing out on new linters with newer golangci-lint versions. I'm fine with both, just thought I'd ask :)

sbueringer: @vincepri We definitely don't have to do it the same way in all provider / core CAPI, but I wonder what's your take on it?

sbueringer: I've opened a Slack thread: https://kubernetes.slack.com/archives/C8TSNPY4T/p1642768320122200

.golangci.yml
```diff
@@ -1,17 +1,37 @@
...
+    - goerr113
     - gomnd
+    - gomoddirectives
+    - ireturn
+    - lll
+    - nilnil
+    - nlreturn
+    - paralleltest
+    - tagliatelle
+    - thelper
```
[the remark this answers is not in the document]
stmcginnis: Sure, I'll enable that and post an update!

.golangci.yml
scdubey: I had a discussion with @srm09, for CAPV it would be good idea to align with CAPI. So we would have to change it to enable list in that case.

stmcginnis:
/retest

Random job failures:
> Failed to open /logs/process-log.txt: open /logs/process-log.txt: no such file or directory

srm09:
@stmcginnis The upstream CI is now fixed, so could you make the linter changes to disable all and only enable specific linters, probably following the list that CAPI supports and over time curate it.
If there are any issues which require code-refactoring, explicitly list those out and I will create separate GH issues which could be worked upon separately.

Thanks for your patience, this has been open a while. We will get it merged soon enough.

srm09:
Also, do rebase the PR with the latest from the default branch to get the upstream CI fixes.

stmcginnis:
No worries, thanks @srm09! I'll get things updated shortly.

srm09:
Thanks for the PR.
/lgtm
/approve

cc: @scdubey

PR #1287, opened 2021-10-29
Adds supervisor types

gab-satchi:
Changes for types:
- move api folder to apis, set multigroup: true and supervisor based APIs under `/apis/vmware`
- creates a new `infrastructure-components-supervisor` which contains the new CRDs. Original `infrastructure-components` remains unchanged

**What this PR does / why we need it**:
Adds the supervisor types specified in #1265 

/hold 

**Release note**:
```release-note
Adds Supervisor types for VSphereCluster, VSphereMachineTemplate and VSphereMachine
```

PR #1309, opened 2021-11-16
Adds controller logic to support Supervisor types

gab-satchi:
Based on top of #1287 

Changes for controllers:
* main.go loads different set of controllers depending on which CRDs are loaded. branching logic looks for the VSphereCluster
* vspherecluster_controller loads a different reconciler depending on which object is loaded
* adds a new vspherecluster context for supervisor based cluster
* adds new services for network, controlplane endpoint and resource policy reconciliation for a supervisor based cluster
* adds new utility functions for new API types
* ~adds required Tanzu API types under external. These types will likely move into a central repo in the future to avoid the duplication~
Testing:
* unit tests are ported over.
* integration tests are WIP
* e2e tests currently do not exercise supervisor based clusters.

**Release note**:
```release-note
Adds support for Supervisor based VSphereCluster and VSphereMachine
```


pkg/services/network/network.go
```diff
@@ -0,0 +1,334 @@
...
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/util"
+)
+
+const (
+	NSXTTypeNetwork = "nsx-t"
+	// This constant is also defined in VM Operator
+	NSXTVNetSelectorKey = "ncp.vmware.com/virtual-network-name"
+
+	CAPWDefaultNetworkLabel    = "capw.vmware.com/is-default-network" //nolint:gosec
```
yastij, line 45: this is probably to support backward compatibility for existing setups, but let's discuss this a bit more. regardless we likely need a new label named `capv.vmware.com/is-default-network`

pkg/services/network/network.go
```diff
@@ -0,0 +1,334 @@
...
+
+	networkList := &netopv1.NetworkList{}
+	err := np.client.List(ctx, networkList, client.InNamespace(ctx.Cluster.Namespace), client.MatchingLabels(labels))
+	if err != nil {
+		return nil, err
+	}
+
+	switch len(networkList.Items) {
+	case 0:
+		return nil, fmt.Errorf("no default CAPW Network found")
```
yastij, line 127: nit: CAPV

pkg/services/network/network.go
```diff
@@ -0,0 +1,334 @@
...
+	case 1:
+		return &networkList.Items[0], nil
+	default:
+		return nil, fmt.Errorf("more than one default CAPW Network found: %d", len(networkList.Items))
```
yastij, line 131: same

pkg/services/vmoperator/control_plane_endpoint.go
```diff
@@ -0,0 +1,226 @@
...
+	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/vmware/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context/vmware"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services"
+)
+
+const (
+	defaultAPIBindPort                   = 6443
+	controlPlaneServiceAPIServerPortName = "apiserver"
+
+	clusterSelectorKey = "capw.vmware.com/cluster.name"
```
yastij, line 40: we need to update this to `capv.vmware.com/cluster.name`

pkg/services/vmoperator/control_plane_endpoint.go
```diff
@@ -0,0 +1,226 @@
...
+	nodeSelectorKey    = "capw.vmware.com/cluster.role"
```
yastij, line 41: same

pkg/services/vmoperator/control_plane_endpoint.go
```diff
@@ -0,0 +1,226 @@
...
+
+func controlPlaneVMServiceName(ctx *vmware.ClusterContext) string {
+	return fmt.Sprintf("%s-control-plane-service", ctx.Cluster.Name)
+}
+
+// ClusterRoleVMLabels returns labels applied to a VirtualMachine in the cluster. The Control Plane
+// VM Service uses these labels to select VMs, as does the Cloud Provider.
+func clusterRoleVMLabels(ctx *vmware.ClusterContext, controlPlane bool) map[string]string {
+	result := map[string]string{
+		clusterSelectorKey: ctx.Cluster.Name,
```
yastij, line 109: labels updated below should reflect then here and then in the `VirtualMachineService`'s selector + `VirtualMachine` labels, since this is called when calling `getVMLabels()`

yastij (commented):
first pass on labels and semantics

yastij (commented):
vspherecluster part1

main.go
```diff
@@ -154,59 +157,31 @@ func main() {
...
-		}
-		if err := controllers.AddVsphereClusterIdentityControllerToManager(ctx, mgr); err != nil {
-			return err
-		}
-		if err := controllers.AddVSphereDeploymentZoneControllerToManager(ctx, mgr); err != nil {
+		supervisorCluster := &vmwarev1b1.VSphereCluster{}
+		gvr = vmwarev1b1.GroupVersion.WithResource(reflect.TypeOf(supervisorCluster).Elem().Name())
+		_, err = mgr.GetRESTMapper().KindFor(gvr)
+		switch {
+		case err == nil:
```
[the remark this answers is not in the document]
gab-satchi: We can. The block would look like this: 
```
if err != nil {
	if meta.IsNoMatchError(err) {
		setupLog.Info(fmt.Sprintf("CRD for %s not loaded, skipping.", gvr.String()))
	} else {
		return err
	}
} else {
	if err := setupVAPIControllers(ctx, mgr); err != nil {
		return err
	}
}
```

I just thought the switch would look less complex

controllers/supervisorvspherecluster_reconciler.go
```diff
@@ -0,0 +1,399 @@
...
+			return reconcile.Result{}, errors.Wrapf(err, "unexpected error while reconciling control plane endpoint for %s", ctx.VSphereCluster.Name)
+		}
+	}
+
+	ctx.VSphereCluster.Status.Ready = true
+	ctx.Logger.V(2).Info("Reconcile completed, vsphereCluster is infrastructure-ready")
+	return reconcile.Result{}, nil
+}
+
+func (r *supervisorClusterReconciler) reconcileControlPlaneEndpoint(ctx *vmware.ClusterContext) (bool, error) {
```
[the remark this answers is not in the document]
gab-satchi: I'm also not a fan of the `VIMMachineService` vs `VmopMachineService`. Are there better names/structure to use for those?

main.go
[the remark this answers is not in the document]
yastij: tbh, I'm not for or against if or a switch, I'm more concerned about checking `err == nil` when it's more idiomatic to check for `err != nil` in go

controllers/supervisorvspherecluster_reconciler.go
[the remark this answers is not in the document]
yastij: `VIMMachineService` and `VmopMachineService` seem to be fine

controllers/vspheremachine_controller.go
```diff
@@ -105,41 +114,159 @@ func AddMachineControllerToManager(ctx *context.ControllerManagerContext, mgr ma
...
+				Machine:           machine,
+				Logger:            r.Logger.WithName(req.Namespace).WithName(req.Name),
+			},
+			ClusterContext: clusterContext,
+			PatchHelper:    patchHelper,
+			VSphereCluster: vsphereCluster,
+			VSphereMachine: vsphereMachine,
+		}
+		defer func() {
+			r.Recorder.EmitEvent(vsphereMachine, "Reconcile", reterr, true)
```
yastij, line 241: I think that this can be common to both code paths

controllers/vspheremachine_controller.go
```diff
@@ -105,41 +114,159 @@ func AddMachineControllerToManager(ctx *context.ControllerManagerContext, mgr ma
...
+			if err := machineContext.Patch(); err != nil {
+				machineContext.Logger.Error(err, "patch failed", "machine", machineContext.String())
+				if reterr == nil {
+					reterr = err
+				}
+			}
+		}()
+
+		// Handle deleted machines
+		if !vsphereMachine.ObjectMeta.DeletionTimestamp.IsZero() {
```
yastij, line 251: you don't need the vsphereMachine object, you use `meta.Accessor(obj)` to just get the meta and make it common across both code paths

controllers/vspheremachine_controller.go
```diff
@@ -105,41 +114,159 @@ func AddMachineControllerToManager(ctx *context.ControllerManagerContext, mgr ma
...
+			return r.reconcileDelete(machineContext)
+		}
+
+		if vsphereClusterErr != nil {
```
yastij, line 255: why are we checking the error here and not early on?

controllers/vspheremachine_controller.go
```diff
@@ -105,41 +114,159 @@ func AddMachineControllerToManager(ctx *context.ControllerManagerContext, mgr ma
...
+			r.Logger.Info("unable to retrieve VSphereCluster", "error", vsphereClusterErr)
+			return reconcile.Result{}, nil
+		}
+
+		if machine == nil {
```
yastij, line 260: why aren't we moving this to the top and make it common between code paths?

controllers/vspheremachine_controller.go
```diff
@@ -49,30 +45,50 @@ import (
...
+// +kubebuilder:rbac:groups=vmware.infrastructure.cluster.x-k8s.io,resources=vspheremachines,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=vmware.infrastructure.cluster.x-k8s.io,resources=vspheremachines/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines;machines/status,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=events,verbs=get;list;watch;create;update;patch
 
 // AddMachineControllerToManager adds the machine controller to the provided
 // manager.
-func AddMachineControllerToManager(ctx *context.ControllerManagerContext, mgr manager.Manager) error {
+func AddMachineControllerToManager(ctx *context.ControllerManagerContext, mgr manager.Manager, controlledType client.Object) error {
+	var supervisorBased bool
```
yastij, line 69: can we hide behind a util func?

yastij (commented):
pass on the vspheremachine controller structure

controllers/vspheremachine_controller.go
gab-satchi: Supervisor clusters don't wait for the machines to get removed before allowing the delete on themselves like VSphereClusters. So we only error out if it's not a delete on the machine i.e: reconcileNormal since that will need a cluster present.

controllers/vspheremachine_controller.go
gab-satchi: same reason as the delete above^ Supervisor machines can be in an orphaned state and delete should go through. I do think it'll be useful for VSphereMachines to behave like  that as well. To make it impossible for vspheremachines to become orphaned and stuck on delete.

controllers/vspheremachine_controller.go
gab-satchi: moved into a common defer function. Couldn't keep it in a common place for both paths as defers get evaluated immediately, just not executed till function return. So had to initialize the machineContext first. If we can converge the prerequisites for reconcileDelete/Normal then we can have a common path.

gab-satchi:
Thanks for the review Yassine. I think if we can standardize how the delete flow is implemented for both the VSphereMachines, we can further move things into a common path. 

With CAPV VSphereMachines, we require all the capi resources to exist before we attempt a delete. 
With Supervisor VSphereMachine, it still allows deletes to go through.

yastij (approved):
/approve
/lgtm

Great work there @gab-satchi !

Issue #1328, opened 2021-12-14
Mysterious certs issue when using HTTP_PROXY


jayunit100:
/kind bug

**What steps did you take and what happened:**

I defined HTTP_PROXY variables, like so:
```
      env:
        - name: HTTP_PROXY
          value: http://192.168.116.1:80
        - name: HTTPS_PROXY
          value: http://192.168.116.1:80
        - name: NO_PROXY
          value: 100.64.0.1,100.96.0.9/24,192.168.111.89,192.168.111.1/24,192.168.116.1/24,192.168.117.1/24,192.168.118.1/24,192.168.119.1/24,192.168.120.1/24,172.18.0.2/24
```

And the code here, seems to fail (i.e. `roblem creating controller manager`)...

```
setupLog.Info("creating controller manager", "version", version.Get().String())
	managerOpts.AddToManager = addToManager
	mgr, err := manager.New(managerOpts)
	if err != nil {
		setupLog.Error(err, "problem creating controller manager")
		os.Exit(1)
	}
```

However oddly, the reason this failure happens is: 

```
entrypoint "msg"="problem running controller manager" "error"="open /tmp/k8s-webhook-server/serving-certs/tls.cert"
```

Who makes this cert, and, why does it *not* exist in cases where the HTTP_PROXY is missing?

**What did you expect to happen:**

HTTP_PROXY and HTTPS_PROXY wouldnt cause an error related to certificates being available... 

**Anything else you would like to add:**

I see this bug in a TKG cluster, however, it appears to be related to the way proxies and certs are mounted.  

**Environment:**

- Cluster-api-provider-vsphere version: master branch
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

jayunit100:
maybe relateed to https://github.com/kubernetes-sigs/cluster-api/blob/main/bootstrap/kubeadm/main.go#L125

josh-ferrell:
This should be the capv-serving-cert certificate that cert-manager creates. Just on a hunch, you may want to try adding ".svc" and ".svc.cluster.local" to NO_PROXY

PR #1330, opened 2021-12-14
Add a logging statement so that in networking blips the dissapearance…

jayunit100:
Fixes part of seeing #1328 when it happens


pkg/manager/manager.go
```diff
@@ -72,6 +73,7 @@ func New(opts Options) (Manager, error) {
 	}
 
 	// Build the controller manager.
+	klog.Infof("Making new manager from options: %v", opts)
```
randomvariable, line 76: This needs to be redacted. We absolutely cannot print credentials to the log.

randomvariable (changes requested):

pkg/manager/manager.go
jayunit100: yup, fixed

PR #1390, opened 2022-02-01
Adds a provision to skip E2E tests

srm09:
Use the `GINKGO_SKIP` args to add `-skip` flag (to the `ginkgo run` command) for each space separated arguments passed to the variable. This alongwith GINKGO_FOCUS var enables us to run/skip specific e2e tests
This has been borrowed from the CAPI E2E test setup which eliminates the
need to create multiple sh files to run specific tests.

Fixes n/a

Once the PR is merged, the follow up task would be to create test-infra jobs to run `conformance` and `clusterctl-upgrade` tests in their own jobs.

**Release note**:
```release-note
NONE
```

**Example**
To skip say the PR Blocking test (the PR Blocking tests should have the label [PR-Blocking] in its `Describe` directive), we can use:
```bash
# This will skip all tests with [PR-Blocking] when running all the tests
GINKGO_SKIP="\[PR-Blocking\]" go test ./...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,12 @@ test-integration: $(GINKGO) $(KUSTOMIZE) $(KIND)
 	time $(GINKGO) -v ./test/integration -- --config="$(INTEGRATION_CONF_FILE)" --artifacts-folder="$(ARTIFACTS_PATH)"
 
 GINKGO_FOCUS ?=
+GINKGO_SKIP ?=
+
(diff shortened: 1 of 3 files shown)


geetikabatra:
@srm09 Thanks for the detailed explanation 😃

yastij (commented):
/lgtm

PR #1387, opened 2022-01-31
🐛 Add missing API type conversions

aartij17:
**What this PR does / why we need it**:
- adds missing conversions for API types - `VSphereClusterIdentity`, `VSphereClusterTemplates`
- add a new make target to run `conversion verification` tests

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1363 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
NONE
```

Issue #1363, opened 2022-01-20
Conversion is not completely implemented for all API types


sbueringer:
/kind bug

Hey folks,
in core CAPI we have a tool called conversion-verifier which loops over all types and ensures the controller-runtime conversion interfaces are implemented correctly, i.e. it checks the types have the Hub / ConvertTo / ConvertFrom functions.

The results for CAPV are:
* type "infrastructure.cluster.x-k8s.io/VSphereClusterIdentity" in package "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha4" marked as storage version but it's not convertible, missing Hub() method
* type "infrastructure.cluster.x-k8s.io/VSphereClusterIdentity" in package "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha3" it's not convertible, missing ConvertFrom() and ConvertTo() methods
* type "infrastructure.cluster.x-k8s.io/VSphereFailureDomain" in package "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha4" marked as storage version but it's not convertible, missing Hub() method
* type "infrastructure.cluster.x-k8s.io/VSphereFailureDomain" in package "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha3" it's not convertible, missing ConvertFrom() and ConvertTo() methods
* type "infrastructure.cluster.x-k8s.io/VSphereClusterIdentityList" in package "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha4" marked as storage version but it's not convertible, missing Hub() method
* type "infrastructure.cluster.x-k8s.io/VSphereClusterIdentityList" in package "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha3" it's not convertible, missing ConvertFrom() and ConvertTo() methods
* type "infrastructure.cluster.x-k8s.io/VSphereDeploymentZoneList" in package "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha4" marked as storage version but it's not convertible, missing Hub() method
* type "infrastructure.cluster.x-k8s.io/VSphereDeploymentZoneList" in package "sigs.k8s.io/cluster-api-provider-vsphere/api/v1alpha3" it's not convertible, missing ConvertFrom() and ConvertTo() methods
[...]

PR #1393, opened 2022-02-01
✨ add script to run `verify-conversions` make target

aartij17:
#1387  introduced a new make target `verify-conversions`. This PR adds a script which will be run as a pre-submit job to verify that the conversion is implemented for all API types. 

Test-infra issue - #https://github.com/kubernetes/test-infra/issues/25075  
Once this PR is merged, we could probably combine all the `verify` targets under a single make target. Similar to what @sbueringer suggested #https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1363#issuecomment-1021933590

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -306,12 +306,50 @@ manifests:  $(STAGE)-version-check $(STAGE)-flavors $(MANIFEST_DIR) $(BUILD_DIR)
 	"$(KUSTOMIZE)" build $(BUILD_DIR)/config/default > $(MANIFEST_DIR)/infrastructure-components.yaml
 	"$(KUSTOMIZE)" build $(BUILD_DIR)/config/supervisor > $(MANIFEST_DIR)/infrastructure-components-supervisor.yaml
 
+## --------------------------------------
+## Verification
+## --------------------------------------
+
+.PHONY: verify
+verify: ## Runs all the verify targets
+	$(MAKE) verify-boilerplate
+	$(MAKE) verify-crds
+	$(MAKE) verify-gen
+	$(MAKE) verify-modules
+	$(MAKE) verify-conversions
+
+.PHONY: verify-boilerplate
+verify-boilerplate: ## Verifies all sources have appropriate boilerplate
+	./hack/verify-boilerplate.sh
+
+.PHONY: verify-crds
+verify-crds: ## Verifies the committed CRDs are up-to-date
+	./hack/verify-crds.sh
+
+.PHONY: verify-gen
+verify-gen: generate  ## Verfiy go generated files are up to date
+	@if !(git diff --quiet HEAD); then \
+		git diff; \
+		echo "generated files are out of date, run make generate"; exit 1; \
+	fi
+
+.PHONY: verify-modules
+verify-modules: modules  ## Verify go modules are up to date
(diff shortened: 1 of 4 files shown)


hack/verify-api-conversion.sh
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+cd "${REPO_ROOT}" || exit 1
```
sbueringer, line 22: I wonder if we should instead introduce a scripts/ci-verify.sh which calls `make verify`

`make verify` would then additionally also call `$(make) verify-conversions)`

(Mostly to align with other providers and to have a single verify job instead of one per validation)

aartij17: Yes, I just saw your other comment on the test-infra PR too! 
I think it makes sense to unify all the `verify` make targets :) Thanks @sbueringer!

Makefile
```diff
@@ -309,12 +309,47 @@ manifests:  $(STAGE)-version-check $(STAGE)-flavors $(MANIFEST_DIR) $(BUILD_DIR)
...
 	"$(KUSTOMIZE)" build $(BUILD_DIR)/config/supervisor > $(MANIFEST_DIR)/infrastructure-components-supervisor.yaml
 
+## --------------------------------------
+## Verification
+## --------------------------------------
+
+.PHONY: verify
+verify: ## Runs all the verify targets
+	$(MAKE) verify-boilerplate
+	$(MAKE) verify-crds
```
srm09, line 319: Missing 2 targets here.

geetikabatra:
@aartij17 You might want to link this PR to the issue https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1393

hack/verify-api-conversion.sh
sbueringer: Just to mention it here to. Had a short discussion yesterday about the jobs in the CAPV office hours and we would keep the separate jobs. Goal is to make it easier to see on PRs which verification fails.

Additionally we want to have a verify target which can be used for convenience locally to run all verifications at once.

Makefile
```diff
@@ -306,12 +306,50 @@ manifests:  $(STAGE)-version-check $(STAGE)-flavors $(MANIFEST_DIR) $(BUILD_DIR)
...
+	$(MAKE) verify-gen
+	$(MAKE) verify-modules
+	$(MAKE) verify-conversions
+
+.PHONY: verify-boilerplate
+verify-boilerplate: ## Verifies all sources have appropriate boilerplate
+	./hack/verify-boilerplate.sh
+
+.PHONY: verify-crds
+verify-crds: ## Verifies the committed CRDs are up-to-date
```
sbueringer, line 326: @aartij17 @srm09 @scdubey
Just an observation for a potential follow-up PR. 

I think verify-crds.sh is redundant. The `verify-gen` target is also running `generate-manifests` and then checking if there is a diff.

aartij17: Sure, thanks for pointing it out @sbueringer! 😄 I will create an issue to remove it.

PR #1403, opened 2022-02-03
🐛 use correct CAPI version in the require clause in go.mod

aartij17:
The `e2e` tests create the resource objects and use regex to fetch the latest CAPI version. This led to the tests picking the older version of CAPI (in spite of newer version mentioned in the replace clause). This PR updates the version in the require clause of `go.mod`. 

**Release note**:

```release-note
NONE
```


aartij17:
Creating another PR for this issue.

PR #1409, opened 2022-02-03
Fixes CAPI CRD path lookup for envtests

### LINKED ISSUE #1410: Envtest CAPI CRD path lookup is broken
/kind bug

Currently, before running the integration tests, we setup the local API server provisioned by envtest with CAPI CRDs. The path for the CRDs is plucked from the `go.mod` file. The lookup for this currently only takes into account the `require` [section](https://go.dev/ref/mod#go-mod-file-require) of the file.
This would overlook any subsequent upgrades that might happen to the CAPI module. Instead institute a dynamic lookup for the CAPI version to use to populate the CRDs.

The latest version of the module is used to fetch the CRDs

From the `go.mod` file no matter what CAPI version is used using the `replace` [directive](https://go.dev/ref/mod#go-mod-file-replace), only the version from the `require` [section](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/go.mod#L33) is picked up.
A dynamic lookup which always picks up the latest and greatest version should be used.

**Environment:**

- Cluster-api-provider-vsphere version: `main` branch

/assign

aartij17:
Fixes CAPI CRD path lookup for envtests

fixes #1410 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/vmware-tanzu/vm-operator/external/tanzu-topology v0.0.0-20211209213435-0f4ab286f64f
 	github.com/vmware/govmomi v0.27.1
 	golang.org/x/crypto v0.0.0-20210817164053-32db794688a5
+	golang.org/x/mod v0.4.2
 	golang.org/x/oauth2 v0.0.0-20210819190943-2bc19b11175f
 	gopkg.in/gcfg.v1 v1.2.3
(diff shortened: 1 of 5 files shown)


srm09:
/retitle Fixes CAPI CRD path lookup for envtests
/hold

aartij17:
This was a super clean solution! Thank you :) 
/lgtm

PR #1412, opened 2022-02-04
Bump golangci-lint to v1.44.0 and fix errors

johananl:

This PR bumps golangci-lint to v1.44.0.

None

**Release note**:

```release-note
NONE
```


srm09:
This is something to verify since when I was trying out this lint version change, I used the go mod edit -replace command to update the version.
This command was adding a statement to replace the go.mod version, but still the version being built was 1.43.0 instead of 1.44.0
Did you go down that route and face the same problem?

johananl:
> This is something to verify since when I was trying out this lint version change, I used the go mod edit -replace command to update the version. This command was adding a statement to replace the go.mod version, but still the version being built was 1.43.0 instead of 1.44.0 Did you go down that route and face the same problem?

Nope, I just updated `hack/tools/go.mod` and ran `make lint`. What did I miss? 😕

johananl:
@srm09 OK, I ran `go mod edit -replace github.com/golangci/golangci-lint=github.com/golangci/golangci-lint@v1.44.0` and indeed the built golangci-lint version is `v1.43.0`. However, I am not sure why this command should be the way to update the linter. Why not do the following, then run `go mod tidy`?

```diff
diff --git a/hack/tools/go.mod b/hack/tools/go.mod
index 2a555638..24e60203 100644
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -3,7 +3,7 @@ module sigs.k8s.io/cluster-api-provider-vsphere/hack/tools
 go 1.15

 require (
-       github.com/golangci/golangci-lint v1.43.0
+       github.com/golangci/golangci-lint v1.44.0
        github.com/onsi/ginkgo v1.16.5
        github.com/vmware/govmomi v0.23.1
        k8s.io/code-generator v0.22.2
```

srm09:
> @srm09 OK, I ran `go mod edit -replace github.com/golangci/golangci-lint=github.com/golangci/golangci-lint@v1.44.0` and indeed the built golangci-lint version is `v1.43.0`. However, I am not sure why this command should be the way to update the linter. Why not do the following, then run `go mod tidy`?

`go mod tidy` will primarily only add any missing module requirements. An upgrade isn't technically a missing requirement, so am not sure if `tidy` will help in upgrading the version.

I faced an issue while editing the `go.mod` file directly, wherein syncing dependencies would result in a cryptic error around being unable to pull some versions of certain modules with an unusable version number. To avoid the same pitfall, I was hoping that we use the `go mod edit` commands to alter the file. But this `replace` is wonky for the `golangci-lint` module as it reports the version from the golangci-lint binary built locally as `1.43.0`. Am unable to find an answer for that one too.

But running the linter with this updated replace directive in the `go.mod` does reveal the linter issues that you fixed in this PR. So I think even though the version is being incorrectly reported (with `golangci-lint version`), it is still using the newer version to run the linter.

johananl:
P.S. I am still quite certain `go mod edit -replace` isn't applicable to our use case here: What it does is add a `replace` directive to `go.mod` whereas what we want is to modify the existing `require` directive to point at the newer version.

srm09:
> P.S. I am still quite certain `go mod edit -replace` isn't applicable to our use case here: What it does is add a `replace` directive to `go.mod` whereas what we want is to modify the existing `require` directive to point at the newer version.
> 

The `replace` directive replaces the specific version of a module with a different version, which also does what we want to do. But I read more around what the usage for the `-replace` flag is, and although it does the job for us, it is not the correct intent behind that flag. `go mod edit -require` is what should be used to upgrade dependencies.
TIL ^^

PR #1430, opened 2022-02-11
Removes a separate IP claim for upgrade tests

srm09:
Since the prow jobs now run the clusterctl upgrade tests as a separate job, we do not need two separate IP claims. 
Plus the IP claim freeing up logic does not get run if the job abruptly gets killed off, which leads two orphaned IP claims which can never be reused unless explicitly cleaned up.

n/a

Fix for smoother CI operations

**Release note**:
```release-note
NONE
```


srm09:
/close
Invalid assumption of how clusterctl upgrade tests work

PR #1446, opened 2022-02-17
Replace GC_KIND variable with SKIP_RESOURCE_CLEANUP

geetikabatra:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note

```

PR #1453, opened 2022-02-20
Add USE_EXISTING_CLUSTER variable for e2e tests

### LINKED ISSUE #1439: Expose e2e startup variables in the Makefile
Currently the e2e test suite that lives under `test/e2e` directory provides a couple of useful flags which make developing e2e tests easier:
- `e2e.skip-resource-cleanup` which can be set on test startup to avoid cleaning up resources after the test runs. This can be used to debug test failures 
- `e2e.use-existing-cluster` which can be leveraged to use an existing k8s cluster (kind, or otherwise) as the management cluster to deploy resources and run tests against.

Though these flags exist, there is no mechanism to set them while running the Makefile targets `make e2e`. It'd be very useful to customize the values of these flags without actually having to edit the Makefile.

This issue proposes adding 2 new variables in the `Makefile`, name `SKIP_RESOURCE_CLEANUP` with a default value of `false` and `USE_EXISTING_CLUSTER` with a default value of `false`. These variables can be used in the `e2e` Makefile target used to run e2e tests.

Default value of Makefile variables can be set by using the `?=` operator
To set the Makefile variables, either export them as environment variables or specify them as `FOO=bar make e2e` before running the command

/good-first-issue
/help
/area testing

ditsuke:
**What this PR does / why we need it**

Adds `USE_EXISTING_CLUSTER` as an environment variable to the Makefile. This variable is passed to the e2e test command where it enables using an existing cluster instead of creating a new one.:

Fixes #1439.

`nil`

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -40,6 +40,7 @@ The first step to running the e2e tests is setting up the required environment v
 
 | Flag                       | Description                                                           |Default Value
 `SKIP_RESOURCE_CLEANUP`      | This flags skips cleanup of the resources created during the tests as well as the kind/bootstrap cluster        | `false`
+`USE_EXISTING_CLUSTER`       | This flag enables the usage of an existing K8S cluster as the management cluster to run tests against.          | `false`
 
 ### Running the e2e tests
 
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,7 @@ VMOP_CRD_ROOT ?= $(MANIFEST_ROOT)/deployments/integration-tests/crds
 WEBHOOK_ROOT ?= $(MANIFEST_ROOT)/webhook
 RBAC_ROOT ?= $(MANIFEST_ROOT)/rbac
 SKIP_RESOURCE_CLEANUP ?= false
+USE_EXISTING_CLUSTER ?= false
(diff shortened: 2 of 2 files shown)


srm09:
@ditsuke Can you add a doc entry similar to the one in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1446
Otherwise /lgtm

ditsuke:
@srm09
I've rebased to fae6ef88467e608665e2902e2bb0aaeb4cee67ed and added the doc entry.

test/e2e/README.md
```diff
@@ -40,6 +40,7 @@ The first step to running the e2e tests is setting up the required environment v
 
 | Flag                       | Description                                                           |Default Value
 `SKIP_RESOURCE_CLEANUP`      | This flags skips cleanup of the resources created during the tests as well as the kind/bootstrap cluster        | `false`
+`USE_EXISTING_CLUSTER`       | This flag enables the usage of an existing K8S cluster as the management cluster to run tests against.
```
srm09, line 43: Missed adding the default value column.

ditsuke: Whoops! Added and squashed the commits.

PR #1456, opened 2022-02-23
Adds load restrictor flag to template generation

srm09:
This will allow kustomize to build the templates using files outside its own root directory and hence remove duplication of same files across directories

n/a

**Release note**:
```release-note
NONE
```


srm09:
/approve
This is just minor kustomization YAML changes for e2e tests.

PR #1459, opened 2022-02-24
🌱 make clusterclass test non `PR-Blocking`

aartij17:
As part of the ClusterClass flavorgen PR, I made the tests `PR-Blocking` in order to get immediate feedback from the prow jobs. But we don't need these tests to be PR-Blocking. This PR takes care of that.


srm09:
/lgtm
/approve
We will wait for this test to run a few times before blocking PRs on this.

PR #1485, opened 2022-03-17
Removes usage of deprecated ObjecteMeta field

### LINKED ISSUE #1481: CAPV should not use ObjectMeta.ClusterName
There is an upstream discussion about dropping ObjectMeta.ClusterName: kubernetes/kubernetes#108717. Related CAPI issue: kubernetes-sigs/cluster-api#6305

This field is not populated by Kubernetes and we should not use it as Cluster Name.

This issue is about doing some research where we use it in CAPV and fix it if it is used.

/kind bug
/assign

srm09:
This patch removes the reference to `ObjectMeta.ClusterName` field from the codebase.

Fixes #1481

Removes the field being used in the error message, no behavioral change

**Release note**:
```release-note
Removes usage of deprecated ObjecteMeta field
```

### DIFF
--- a/pkg/util/machines.go
+++ b/pkg/util/machines.go
@@ -214,8 +214,8 @@ func GetMachineMetadata(hostname string, vsphereVM infrav1.VSphereVM, networkSta
 	}); err != nil {
 		return nil, errors.Wrapf(
 			err,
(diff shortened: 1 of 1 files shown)


srm09:
/retest
Interestingly the e2e test passed, but the job reports a failure. Retriggering.

PR #1486, opened 2022-03-17
🐛 Delete legacy finalizer, fix setting ownerRef for `VSphereClusterIdentity `

### LINKED ISSUE #1482: Fix `v1alpha3` -> `v1beta1` upgrade workflow for `VSphereClusterIdentity`
/kind bug

Upgraded `v1alpha3` cluster to `v1beta1`. Noticed that the identity secret has different finalizers being set in both the versions - 
`v1alpha3` sets `identity/infrastructure.cluster.x-k8s.io` while `v1beta1` sets `vspherecluster/infrastructure.cluster.x-k8s.io`

This issue is to fix the following - 
1. Add a fix to remove the previously set finalizer. Without this, the cluster deletion is not successful
2. Check for only the group and kind, and not the apiVersion, when setting ownerReferences on the secret
3. Remove secret deletion from the vspherecluster reconciler deletion loop since the secret should automatically get deleted successfully once the finalizers are removed and the cluster is deleted

The workload cluster should get deleted successfully after upgrading the cluster from `v1alpha3` to `v1alpha4`

**Environment:**

- Cluster-api-provider-vsphere version: v1.1.0

aartij17:
This PR does this following - 
- Check if the older VSphereClusterIdentity is present and delete it during the reconcileDelete phase. This is to make sure that clusters upgraded from v1alpha3 to v1alpha4/v1beta1 can be deleted once the finalizers are removed
- Check only the group and kind(and not the version) while setting ownerRef on VSphereClusterIdentity
- Remove explicit secret deletion from VSphereClusterReconciler (this already happens during the delete reconcile loop of VSphereClusterIdentity)

Fixes #1482

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/vspherecluster_reconciler.go
+++ b/controllers/vspherecluster_reconciler.go
@@ -48,6 +48,10 @@ import (
 	infrautilv1 "sigs.k8s.io/cluster-api-provider-vsphere/pkg/util"
 )
 
+// legacyIdentityFinalizer is deprecated and should be used only while upgrading the cluster
+// from v1alpha3(v.0.7).
+const legacyIdentityFinalizer string = "identity/infrastructure.cluster.x-k8s.io"
+
 type clusterReconciler struct {
 	*context.ControllerContext
 }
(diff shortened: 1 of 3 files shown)


controllers/vspherecluster_reconciler.go
```diff
@@ -48,6 +48,8 @@ import (
 	infrautilv1 "sigs.k8s.io/cluster-api-provider-vsphere/pkg/util"
 )
 
+const legacyIdentityFinalizer string = "identity/infrastructure.cluster.x-k8s.io"
```
srm09, line 51: Do add an explicit comment marking it as Deprecated, kept only for upgrade issues.

controllers/vspherecluster_reconciler.go
```diff
@@ -256,11 +262,9 @@ func (r clusterReconciler) reconcileIdentitySecret(ctx *context.ClusterContext)
 		}
 
 		// check if cluster is already an owner
-		if !clusterutilv1.IsOwnedByObject(secret, vsphereCluster) {
-			if len(secret.GetOwnerReferences()) > 0 {
-				return fmt.Errorf("another cluster has set the OwnerRef for secret: %s/%s", secret.Namespace, secret.Name)
```
srm09, line 261: Well I think we need to have the check that this secret is owned by a particular vSphereCluster object only? If this secret owned by another VSphereCluster, then we should throw an error.

srm09 (changes requested):

controllers/vspherecluster_reconciler.go
```diff
@@ -260,7 +270,10 @@ func (r clusterReconciler) reconcileIdentitySecret(ctx *context.ClusterContext)
 			if len(secret.GetOwnerReferences()) > 0 {
 				return fmt.Errorf("another cluster has set the OwnerRef for secret: %s/%s", secret.Namespace, secret.Name)
 			}
-
+		}
+		// check only the group and the kind of the ownerRef
+		if !clusterutilv1.HasOwner(secret.OwnerReferences, clusterv1.GroupVersion.String(),
```
yastij, line 275: why do we have this here? Isn't the check above sufficient?

yastij (commented):
/approve

PR #1518, opened 2022-04-19
Start the Keep Alive handler for VSphere client used in e2e tests

srm09:
Some of the e2e test runs on testgrid keep failing due to failures in the test listed in `storage_policy_test.go` file. The reason for the failure is that the govmomi Client object we use to assert some conditions times out due to inactivity. As part of an earlier PR, a `Keep-Alive` handler was added to the client which did not seem to solve the issue. The reason was that the groutine for the handler needs an explicit call to `client.Login()` method to get activated. 

This patch slightly modifies the creation of the govmomi Client object and adds an explicit call to login after adding the keep alive handler and also logs out the session once the test is finished.

n/a

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/config/vsphere-ci.yaml
+++ b/test/e2e/config/vsphere-ci.yaml
@@ -187,6 +187,7 @@ intervals:
   default/wait-worker-nodes: ["10m", "10s"]
   default/wait-delete-cluster: ["5m", "10s"]
   default/wait-machine-upgrade: ["15m", "1m"]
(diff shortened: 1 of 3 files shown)


yastij (commented):
/lgtm
/hold

feel free to remove the hold when you want

srm09:
/unhold
Thanks for taking a look. Tested this one locally on multiple runs, and it worked.

PR #1568, opened 2022-07-06
Removing the ensure-go check from ci-test-coverage.sh

geetikabatra:
Currently there is a go version compatibility issue.
Test infra job is failing as go version in kubekins image is 1.17
while in dependencies like clusterapi@test is 1.16 and go mod
version in hack/tools is 1.15. This is causing gomodules to throw
error and hence failing the test coverage job.
Removing this line as kubekins by default has 1.17 go installed.
This script will run assuming that go will be present in the kubekins.

Signed-off-by: geetikab@vmware.com <geetikab@vmware.com>

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/hack/ci-test-coverage.sh
+++ b/hack/ci-test-coverage.sh
@@ -19,8 +19,6 @@ set -o nounset
 set -o pipefail
 
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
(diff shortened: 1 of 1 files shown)


srm09:
This is not the correct fix for this one, can you create an issue for this and we can discuss more on the issue.
/close

geetikabatra:
Sure!

PR #1570, opened 2022-07-07
OS field and Windows VM name shrink

### LINKED ISSUE #1292: Workaround the 15 character limit for AD in CAPV, temporarily (Cross linked issue to CAPI), until CAPI has a broader fix
/kind bug

we can't easily add machines for AD (active directory on windows nodes) bc of https://github.com/kubernetes-sigs/cluster-api/issues/2217 

Should we do a band aid similar to what is done in CAPZ? (ill try to dig up a link, but evidently CAPZ allows for smaller machine names by doing a provider specific fix for windows).
 
```
// Name returns the Azure Machine Pool Name.
func (m *MachinePoolScope) Name() string {
	// Windows Machine pools names cannot be longer than 9 chars
	if m.AzureMachinePool.Spec.Template.OSDisk.OSType == azure.WindowsOS && len(m.AzureMachinePool.Name) > 9 {
		return "win-" + m.AzureMachinePool.Name[len(m.AzureMachinePool.Name)-5:]
	}
	return m.AzureMachinePool.Name
}
``` (from machinepool.go)

### PR DESCRIPTION
**Pull request by knabben**
Windows VM name greater than 15 characters can cause issues when multiple machines join a domain and end up with the same NetBIOS name due the OS shrink of the name and same cluster prefix.

This PR adds a new field to identify the machine OS and mutate the VM name via Mutation Webhook when the OS=Windows. The same approach is being used on CAPZ.

Fixes #1292

This is a reopen of https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1052

**Release note**:

```release-note
Adding OS field on VirtualMachineCloneSpec and renaming of Windows VMs when the length is greater than 15 characters.
```

### DIFF
--- a/apis/v1beta1/vspherevm_webhook.go
+++ b/apis/v1beta1/vspherevm_webhook.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"strings"
 
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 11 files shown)


vrabbi (approved):
Looms good. Thanks @knabben!

apis/v1beta1/vspherevm_webhook.go
```diff
@@ -35,6 +36,15 @@ func (r *VSphereVM) SetupWebhookWithManager(mgr ctrl.Manager) error {
 }
 
 // +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1beta1-vspherevm,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=vspherevms,versions=v1beta1,name=validation.vspherevm.infrastructure.x-k8s.io,sideEffects=None,admissionReviewVersions=v1beta1
+// +kubebuilder:webhook:verbs=create;update,path=/mutate-infrastructure-cluster-x-k8s-io-v1beta1-vspherevm,mutating=true,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=vspherevms,versions=v1beta1,name=default.vspherevm.infrastructure.x-k8s.io,sideEffects=None,admissionReviewVersions=v1beta1
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type.
+func (r *VSphereVM) Default() {
+	// Windows hostnames must be < 16 characters in length
```
srm09, line 43: Shouldn't this also set the default OS value to Linux?

PR #1576, opened 2022-07-12
🌱 Uses cluster name for resource policy creation

srm09:
This PR updates the resource policy creation logic to use the name of the Cluster object, instead of the `VSphereCluster` object which is not deterministic in case of topology controller. Hence, using the Cluster object name makes it deterministic for the purpose of UX.

Fixes #

n/a

**Release note**:

```release-note
Uses cluster name for resource policy creation
```


yastij (approved):
/approve
/lgtm

PR #1581, opened 2022-07-22
Use CloudInit bootstrap w VM Operator

akutz:
This patch updates CAPV to use the VM Operator CloudInit bootstrap method instead of ExtraConfig. This change ensures there is no race between VM Tools and CloudInit since using ExtraConfig will cause Guest OS Customization
to configure the network, resulting in a reboot while CloudInit is running. Specifying CloudInit as the bootstrap provider ensures VM Operator uses CloudInit to bootstrap the network.

Please note this patch will only work in vSphere 7P05+.

NA

This is required for uTKG.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/services/vmoperator/vmopmachine.go
+++ b/pkg/services/vmoperator/vmopmachine.go
@@ -277,7 +277,7 @@ func (v VmopMachineService) reconcileVMOperatorVM(ctx *vmware.SupervisorMachineC
 		vmOperatorVM.Spec.ResourcePolicyName = ctx.VSphereCluster.Status.ResourcePolicyName
 		vmOperatorVM.Spec.VmMetadata = &vmoprv1.VirtualMachineMetadata{
 			ConfigMapName: vmwareutil.GetBootstrapConfigMapName(ctx.VSphereMachine.Name),
-			Transport:     "ExtraConfig",
+			Transport:     "CloudInit",
 		}
 
 		// VMOperator supports readiness probe and will add/remove endpoints to a
--- a/test/integration/integration_suite_test.go
+++ b/test/integration/integration_suite_test.go
@@ -727,8 +727,7 @@ func assertVirtualMachineState(machine *clusterv1.Machine, vm *vmoprv1.VirtualMa
 	Expect(vm.Spec.ImageName).ShouldNot(BeEmpty())
 	Expect(machine.Spec.Version).ShouldNot(BeNil(), "Error accessing nil Spec.Version for machine %s", machine.Name)
 	Expect(vm.Spec.VmMetadata).NotTo(BeNil())
-	// TODO: Aarti: not sure where to get this varaible from
-	//Expect(vm.Spec.VmMetadata.Transport).To(Equal(vmoprv1.VirtualMachineMetadataExtraConfigTransport))
+	Expect(vm.Spec.VmMetadata.Transport).To(Equal(vmoprv1.VirtualMachineMetadataCloudInitTransport))
 	Expect(vm.Spec.VmMetadata.ConfigMapName).ToNot(BeNil())
 }
 


test/integration/integration_suite_test.go
```diff
@@ -728,7 +728,7 @@ func assertVirtualMachineState(machine *clusterv1.Machine, vm *vmoprv1.VirtualMa
 	Expect(machine.Spec.Version).ShouldNot(BeNil(), "Error accessing nil Spec.Version for machine %s", machine.Name)
 	Expect(vm.Spec.VmMetadata).NotTo(BeNil())
 	// TODO: Aarti: not sure where to get this varaible from
-	//Expect(vm.Spec.VmMetadata.Transport).To(Equal(vmoprv1.VirtualMachineMetadataExtraConfigTransport))
+	//Expect(vm.Spec.VmMetadata.Transport).To(Equal(vmoprv1.VirtualMachineMetadataCloudInitTransport))
```
yastij, line 731: we should uncomment this `Expect` call and remove the `TODO` at the same time

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2021-2022 The Kubernetes Authors.
```
yastij, line 2: no need to update this

test/integration/integration_suite_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2021-2022 The Kubernetes Authors.
```
yastij, line 2: same

yastij (commented):
/approve

akutz:
@srm09 / @yastij / @sidharthsurana,

Do any of the three of you know whether the PR tests include validation against Supervisor and VM Operator integration?

pkg/services/vmoperator/vmopmachine.go
akutz: The file changed though, and dates are updated to reflect the most recent year, hence the range.

test/integration/integration_suite_test.go
akutz: See previous comment.

test/integration/integration_suite_test.go
akutz: Ack.

aruneshpa (approved):

test/integration/integration_suite_test.go
```diff
@@ -727,8 +727,7 @@ func assertVirtualMachineState(machine *clusterv1.Machine, vm *vmoprv1.VirtualMa
 	Expect(vm.Spec.ImageName).ShouldNot(BeEmpty())
 	Expect(machine.Spec.Version).ShouldNot(BeNil(), "Error accessing nil Spec.Version for machine %s", machine.Name)
 	Expect(vm.Spec.VmMetadata).NotTo(BeNil())
-	// TODO: Aarti: not sure where to get this varaible from
-	//Expect(vm.Spec.VmMetadata.Transport).To(Equal(vmoprv1.VirtualMachineMetadataExtraConfigTransport))
+	Expect(vm.Spec.VmMetadata.Transport).To(Equal(vmoprv1.VirtualMachineMetadataCloudInitTransport))
 	Expect(vm.Spec.VmMetadata.ConfigMapName).ToNot(BeNil())
```
bryanv, line 731: While you're here, the more appropriate matcher would be `BeEmpty()`

bryanv (approved):

test/integration/integration_suite_test.go
akutz: Absolutely, but I'm not going to respin the tests again. I just made the minimal changes, but you're not wrong.

srm09:
/lgtm

We do not have the setup to test supervisor types in the upstream CI.

PR #1585, opened 2022-07-26
add configuration that disables DHCP DNS servers

### LINKED ISSUE #1582: Support configuration to disable DHCP DNS servers
/kind feature

As a user, I would like to toggle a setting (`ignoreDHCPNameservers`) that will cause nodes to ignore the nameservers that DHCP assigns, so that I have full control over my DNS servers for my cluster.

This can be done with netplan already through the `dhcp{4,6}-overrides.use-dns` configuration option. However this currently only has an effect on the `networkd` backend. [netplan docs](https://netplan.io/reference#:~:text=The%20dhcp4%2Doverrides%20and%20dhcp6,override%20the%20default%20DHCP%20behavior.&text=Default%3A%20true%20.,effect%20on%20the%20networkd%20backend.)

We would like to add a field to the `NetworkDeviceSpec` on the `VsphereMachineTemplate`, something like:
```
// IgnoreDHCPNameservers is a flag that indicates whether or not to use the
// DNS servers received from the DHCP server.
// If true the DNS servers from the DHCP server will be ignored and only
// the statically configured nameservers will be added to the devices DNS
// servers.
// If false the DNS servers from DHCP and any statically configured
// nameservers will be added to the devices DNS servers.
// Please note that this only works with the `networkd` backend for netplan.
// +optional
IgnoreDHCPNameservers bool `json:"ignoreDHCPNameservers,omitempty"`
```

An example simplified config without all fields
```
---
kind: VSphereMachineTemplate
metadata:
  name: my-cluster
  namespace: my-namespace
spec:
  template:
    spec:
      network:
        devices:
          - networkName: "my-favorite-network"
            dhcp4: true
            dhcp6: true
            nameservers: [8.8.8.8, 8.8.4.4, 2001:4860:4860::8888]
            ignoreDHCPNameservers: true
```

An example metadata output
```
instance-id: "test-vm"
local-hostname: "test-vm"
wait-on-network:
  ipv4: true
  ipv6: true
network:
  version: 2
  ethernets:
    id0:
      match:
        macaddress: "00:00:00:00:00"
      set-name: "ens192"
      wakeonlan: true
      dhcp4: true
      dhcp6: true
      dhcp4-overrides:
        use-dns: false
      dhcp6-overrides:
        use-dns: false
      nameservers:
        addresses:
        - 8.8.8.8
        - 8.8.4.4
        - 2001:4860:4860::8888
```

The resulting netplan yaml
```
network:
  ethernets:
    id0:
      dhcp4: true
      dhcp4-overrides:
        use-dns: false
      dhcp6: true
      dhcp6-overrides:
        use-dns: false
      match:
        macaddress: xx:xx:xx:xx:xx:xx
      nameservers:
        addresses:
        - 8.8.8.8
        - 8.8.4.4
        - 2001:4860:4860::8888
      set-name: eth0
      wakeonlan: true
  version: 2
```

We'd love to get any input on the design and resolve any concerns there might be.

This is related to https://github.com/vmware-tanzu/tanzu-framework/issues/1103

We are happy to PR this change and we have one started so we could explore our design. We are struggling a bit with running the e2e tests on our own and could use any help there as well but that might be something better to drop into k8s slack.

adobley:
DHCP servers can provide DNS servers. This behavior is not always desired. With this setting a user can choose to have only the nameservers they set be used by the VMs. An image must know how to process this metadata field for it to have the desired effect. We have tested this with Ubuntu 20.04 images and netplan does respect this setting.

Fixes #1582

We have manually tested this with Ubuntu 20.04 and Photon-3, there will be some changes needed to the Photon-3 image to write the correct networkd config.

**Release note**:

```release-note
Configuration added to allow ignoring the nameservers that DHCP assigns
```


adobley:
Closing this PR in favor of a different approach in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1615

PR #1587, opened 2022-07-27
Only set PVC zone annotation when zonal storage classes are required

dougm:

The CSI zone annotation must be set when using a zonal storage class,
which is required when the cluster has multiple (3) zones.
Single zone clusters (legacy/default) do not support zonal storage and must not
have the zone annotation set.

Follow up to #1573


yastij (approved):
/approve
/lgtm

PR #1594, opened 2022-08-16
🌱 Updates VSphereCluster reconcile request enqueue mechanism

### LINKED ISSUE #1597: Change the reconcile request mechanism for VSphereCluster object
/kind feature

Like in many of the places in the codebase, there is an inherent assumption that the names of the `Cluster` and `VSphereCluster` objects are the same. That is not the case, since the advent of topology controller which uses a suffix to generate the name of the objects.
Due to this, the logic to enqueue requests for `VSphereCluster` objects when a `VSphereMachine` object changes needs to be updated.  

- [Logic](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/d2494c3c13fe1159a43e5a7f5dc9dff5de889c11/controllers/vmware/vspherecluster_reconciler.go#L346) for vmware.infrastructure.x-k8s.io.VSphereCluster object.

**Environment:**

srm09:
This patch fixes the logic of enqueuing requests to the `VSphereCluster` controller when `VSphereMachine` objects get changed. This is needed since the name of the `Cluster` object is different than the name of the `VSphereCluster` object in case of topology controller.

It also cleans up the unused `GetMachineInCluster()` function which can be replaced by `GetFilteredMachinesForCluster()` function from the CAPI `util` package.

Fixes #1597

_This patch alters the logic of the controller to not enqueue reconcile requests directly for the name of the provider service account objects and instead enqueue the request for the owner VSphereCluster objects._

**Release note**:
```release-note
Fixes reconcile request mechanism for VSphereCluster objects
```

### DIFF
--- a/pkg/util/machines.go
+++ b/pkg/util/machines.go
@@ -36,31 +36,6 @@ import (
 	vmwarev1b1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/vmware/v1beta1"
 )
 
-// GetMachinesInCluster gets a cluster's Machine resources.
-func GetMachinesInCluster(
-	ctx context.Context,
-	controllerClient client.Client,
-	namespace, clusterName string) ([]*clusterv1.Machine, error) {
-	labels := map[string]string{clusterv1.ClusterLabelName: clusterName}
-	machineList := &clusterv1.MachineList{}
-
-	if err := controllerClient.List(
-		ctx, machineList,
-		client.InNamespace(namespace),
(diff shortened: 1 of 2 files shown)


srm09:
Removing the `ProviderServiceAccount` controller changes to a different PR instead, since it might require a bigger change.

srm09:
/retitle 🌱 Updates VSphereCluster reconcile request enqueue mechanism

yastij (approved):
/lgtm
/approve

PR #1599, opened 2022-08-17
🌱 Changes enqueue logic for provider service account objects

### LINKED ISSUE #1598: Update Provider Service Account enqueue logic
/kind feature

Currently, for the provider service account objects we create a controller which watches the ProviderServiceAccount objects and also add
watches on the `ProviderServiceAccount` and `ServcieAccount` objects and enqueue the requests for the VSphereCluster objects associated to these objects.
The controller starts the reconciliation process by treating each of the request objects having the namespeace/name combo of a vSphereCluster object. The side-effect is that the ProviderServiceAccount objects also get enqueued in the reuqest queue, since the controller watches objects of those type. 
kube-builder does not allow the controllers to be built without having a base type that they are tracking. So, this causes the provider sec account objects to be enqueued as requests to the service account controller which show up as failures in the logs. These are harmless failures since there is a separate watch added to the `ProviderServiceAccount` objects to enqueue the associated `VSphereCluster` objects to the controller's queue.

Instead of this round-about way of enqueeuong requests, it is better to enqueue the ProviderServiceAccount objects to the controller's queue, and update the reconcile logic to handle such requests.

No behavioral changes are introduced in CAPV as a part of this change, hence no issues are bound to occur during upgrades.

**Environment:**

srm09:
This patch changes the reconcile logic so as to enqueue only `ProviderServiceAccount` objects and fetch the `VSphereCluster` objects associated to the these objects in the reconcile loop.

Fixes #1598

n/a

**Release note**:
```release-note
Changes enqueue logic for provider service account objects
```

### DIFF
--- a/controllers/serviceaccount_controller.go
+++ b/controllers/serviceaccount_controller.go
@@ -85,14 +85,10 @@ func AddServiceAccountProviderControllerToManager(ctx *context.ControllerManager
 	}
 
 	return ctrl.NewControllerManagedBy(mgr).For(controlledType).
-		// Watch a ProviderServiceAccount
-		Watches(
-			&source.Kind{Type: &vmwarev1.ProviderServiceAccount{}},
-			handler.EnqueueRequestsFromMapFunc(r.providerServiceAccountToVSphereCluster),
-		).
+		// Watch a ServiceAccount
 		Watches(
 			&source.Kind{Type: &corev1.ServiceAccount{}},
-			handler.EnqueueRequestsFromMapFunc(r.serviceAccountToVSphereCluster),
+			handler.EnqueueRequestsFromMapFunc(r.toProviderServiceAccount),
 		).
 		Complete(r)
 }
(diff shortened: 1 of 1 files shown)


yastij (commented):
/approve

controllers/serviceaccount_controller.go
```diff
@@ -85,14 +85,10 @@ func AddServiceAccountProviderControllerToManager(ctx *context.ControllerManager
 	}
 
 	return ctrl.NewControllerManagedBy(mgr).For(controlledType).
-		// Watch a ProviderServiceAccount
-		Watches(
```
yastij, line 89: we should still be watching for simple `ProviderServiceAccount` changes

srm09: The `.For()` adds a default watch not the controlled type which in this case is the `ProviderServiceAccount` object.

geetikabatra:
lgtm

PR #1602, opened 2022-08-23
🌱 Update CAPI to v1.2.1

### LINKED ISSUE #1572: Update CAPI version to v1.2.0
/kind feature

Once the new CAPI version is released, we should update the dependency in CAPV as well.

**Environment:**

- Cluster-api-provider-vsphere version: `main`

/hold
/lifecycle active
/assign

srm09:
This patch updates the CAPI version used by CAPV to `v1.2.1`. Following is the summary of all the changes:
1. Updates k8s dependencies to k8s 1.24
2. Update golang to v1.18
3. Updates CAPI references to v1.2.1
4. Changes logging to be in line with upstream k/k logging
5. Changes `VSphereMachineTemplate` to support SSA and dry runs.
6. Updates golang-ci lint to v1.47.3

Fixes #1572

The e2e test for SSA does not exercise the changes for SSA currently, since we are waiting for https://github.com/kubernetes-sigs/cluster-api/pull/7134 to be merged, added a TODO to describe the change.

**Release note**:
```release-note
Update CAPI dependency to v1.2.1
```


pkg/record/recorder.go
```diff
@@ -51,24 +51,28 @@ type recorder struct {
 	record.EventRecorder
 }
 
+func (r recorder) String(reason string) string {
+	return cases.Title(language.English, cases.NoLower).String(reason)
```
sbueringer, line 55:
```suggestion
	return cases.Title(language.Und).String(reason)
```
nit, we used that in core CAPI (just in case you want to use the same)

xref: https://en.wiktionary.org/wiki/und#:~:text=(international%20standards)%20ISO%20639%2D,code%20for%20an%20undetermined%20language.

main.go
```diff
@@ -45,10 +48,12 @@ import (
 )
 
 var (
-	setupLog = ctrllog.Log.WithName("entrypoint")
+	setupLog   = ctrllog.Log.WithName("entrypoint")
```
sbueringer, line 51: Just fyi we're using `ctrl.Log.WithName("setup")` with `ctrl "sigs.k8s.io/controller-runtime"` here instead. Not sure what difference it makes

main.go
```diff
@@ -59,14 +64,10 @@ var (
 	defaultKeepAliveDuration = constants.DefaultKeepAliveDuration
 )
 
-func main() {
-	rand.Seed(time.Now().UnixNano())
-
-	klog.InitFlags(nil)
-	ctrllog.SetLogger(klogr.New())
-	if err := flag.Set("v", "2"); err != nil {
```
sbueringer, line 67: Looks like CAPV was using --v=2 hard-coded before (?). Not sure how exactly that works and if `--v` provided by users was still used. (maybe it just set the default value given that flag parse happens below)

Dropping it should change the amounts of logs emitted by CAPV (per default (?)).

CAPI is running per default with --v=0 at the moment (even though Kuberentes recommends --v=2 for production).

Can't tell what the intention here is, just wanted to surface it.

main.go
```diff
@@ -129,39 +131,54 @@ func main() {
...
+	)
+
+	feature.MutableGates.AddFlag(fs)
+}
+
+func main() {
+	rand.Seed(time.Now().UnixNano())
+
+	InitFlags(pflag.CommandLine)
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
```
sbueringer, line 154:
```suggestion
        pflag.CommandLine.SetNormalizeFunc(cliflag.WordSepNormalizeFunc)
	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
```
You might want to add that as well. I think it was added in CAPI a while back

apis/v1beta1/vspheremachinetemplate_webhooks_test.go
```diff
@@ -54,7 +55,8 @@ func TestVSphereMachineTemplate_ValidateCreate(t *testing.T) {
 	}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			err := tc.vsphereMachine.ValidateCreate()
+			webhook := &VSphereMachineTemplateWebhook{}
+			err := webhook.ValidateCreate(context.TODO(), tc.vsphereMachine)
```
sbueringer, line 59:
```suggestion
			err := webhook.ValidateCreate(context.Background(), tc.vsphereMachine)
```
nit: context.TODO() is usually used in places where at some point a real ctx should be introduced (usually as soon as a ctx is passed into the current func).

Given that we absolutely don't care about it here I would use `context.Background()` in cases like this instead

P.S. context.TODO is meant in a sense of that it's a TODO to replace it at some point and folks can grep for context.TODO.

config/manager/manager.yaml
```diff
@@ -39,3 +39,5 @@ spec:
       tolerations:
         - effect: NoSchedule
           key: node-role.kubernetes.io/master
```
sbueringer, line 41: There are a few more places where `node-role.kubernetes.io/master` is used. Sometimes as nodeSelector sometimes as taint.

I think they should be adjusted as well, but I'm not sure

pkg/record/recorder.go
johananl: FYI, switching from `strings.Title` to `cases.Title` without `cases.NoLower` changes the resulting string. We might want to keep `cases.NoLower` here.

Context: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1608#discussion_r960609376

srm09: Was the reason why I introduced the lower case option when working on the PR.

main.go
srm09: Had omitted it since I wasn't sure what exactly was the use for the normalize func, and if that is something really required. Did some digging around, seems like it replaces the `_` in flags to `-`, so say the `--enable-keep-alive` flag can also be `--enable_keep_alive` flag.
No harm having it in the code. Thanks!

main.go
srm09: For the sake ok consistency, changed it :)

pkg/record/recorder.go
sbueringer: Thx! I'll test what Und does and then probably follow-up if the same change in core CAPI

main.go
srm09:
```bash
cluster-api-provider-vsphere on  update-capi-v1.2.0 [$!?] via 🏎💨 v1.18.5
❯ ./capv-manager
F0904 14:55:33.140969   66061 main.go:159] failed to set log level: no such flag -v

cluster-api-provider-vsphere on  update-capi-v1.2.0 [$!?] via 🏎💨 v1.18.5
❯ ./capv-manager -v 2
F0904 14:55:33.140969   66061 main.go:159] failed to set log level: no such flag -v
```
Have been seeing this error when setting the logging verbosity level in the code as well as a command line arg.

config/manager/manager.yaml
srm09: There is no node selector logic that CAPV uses to run the manager.

sbueringer: I just found code when grepping for it, e.g.:
* cloud-controller-manager.go
* csi.go
* calico.yaml

I can't tell if those places are relevant or not. Feel free to just resolve if that finding doesn't make sense :)

main.go
sbueringer: I think you have to use 2x `-`

```plain
go run main.go --v=10
```
Worked for me

config/manager/manager.yaml
randomvariable: Other than the calico.yaml, that was long replaced by applying CPI/CSI using CRS. The functions should probably be marked as deprecated.

hack/tools/go.mod
```diff
@@ -15,3 +15,236 @@ require (
 	sigs.k8s.io/kustomize/kustomize/v4 v4.4.0
 	sigs.k8s.io/testing_frameworks v0.1.2
 )
+
+require (
+	4d63.com/gochecknoglobals v0.1.0 // indirect
```
yastij, line 20: is this intake expected? all coming from golangci-lint?

packaging/flavorgen/flavors/generators.go
```diff
@@ -246,10 +246,9 @@ func defaultKubeadmInitSpec(files []bootstrapv1.File) bootstrapv1.KubeadmConfigS
 			},
 			ControllerManager: defaultControlPlaneComponent(),
 		},
-		Users:                    defaultUsers(),
-		PreKubeadmCommands:       defaultPreKubeadmCommands(),
-		UseExperimentalRetryJoin: true,
```
yastij, line 251: why did we drop `UseExperimentalRetryJoin` it's still needed in slow/flaky envs

yastij (commented):
/approve

few nits/questions

apis/v1alpha3/cloudprovider_types.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// cloudprovider_types contains API types for the vSphere cloud provider.
+// Package v1alpha3 cloudprovider_types contains API types for the vSphere cloud provider.
```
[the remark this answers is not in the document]
srm09: These fixes are from `make fix`, if you think, I can exclude the deprecated API types from the linter.

packaging/flavorgen/flavors/generators.go
srm09: This has been deprecated in kubeadmconfig types. Checkout https://github.com/kubernetes-sigs/cluster-api/blob/8b4214d72762394144b83dd6d14986ff7e274095/bootstrap/kubeadm/api/v1beta1/kubeadmconfig_types.go#L100-L101

hack/tools/go.mod
srm09: Most of them are because we updated the golang version to 1.18 which basically brings in the Module graph pruning changes (from go@1.17)

apis/v1alpha3/cloudprovider_types.go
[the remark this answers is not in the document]
srm09: Done, removed the changes from the deprecated API types.

PR #1615, opened 2022-09-13
✨ add ability to configure DHCP overrides

### LINKED ISSUE #1582: Support configuration to disable DHCP DNS servers

adobley:

This PR adds the ability to configure DHCP overrides for vSphere machines. This is important for users who want to tweak behaviors of their networks, such as using only the DNS servers that they configured.

Fixes #1582

Our end-to-end test isn't concerned with the functional change these settings make, but with checking that capv writes the metadata correctly. Many of these settings are highly environment specific, knowing IPs or other configurations that could be difficult to account for in CI/dev environments. The responsibility of translating the metadata to network configuration falls on the distribution which we did not want to test here.

**Release note**:

```release-note
Add ability to configure DHCP overrides
```


srm09:
@adobley This needs a rebase. 
/ok-to-test
@yastij @vrabbi this PR could benefit from a review.

srm09:
@adobley could you rebase this PR, I can get a review on this one today

srm09:
/test pull-cluster-api-provider-vsphere-full-e2e
To run the new e2e test introduced.

srm09:
Considering the special notes that you have added, I wonder if the end to end test does provide something concrete? Could we not write unit and/or controller tests to mimic this behavior?

test/e2e/dhcp_overrides_test.go
```diff
@@ -0,0 +1,121 @@
...
+	})
+})
+
+func getVMMetadata(vm types.VSphereVM) []byte {
+	finder := find.NewFinder(vsphereClient.Client, false)
+	dc, err := finder.Datacenter(ctx, vm.Spec.Datacenter)
+	Expect(err).NotTo(HaveOccurred())
+	finder.SetDatacenter(dc)
+
+	vmObj, err := finder.VirtualMachine(ctx, fmt.Sprintf("/%s/vm/%s/%s", vm.Spec.Datacenter, vm.Spec.Folder, vm.Name))
```
srm09, line 100:
```suggestion
	vmObj, err := finder.VirtualMachine(ctx, vm.Name)
```
should be enough to find the VM.

test/e2e/dhcp_overrides_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+type Ethernet struct {
+	DHCP4Overrides DHCPOverrides `yaml:"dhcp4-overrides"`
+}
+
+type DHCPOverrides struct {
+	SendHostname *bool `yaml:"send-hostname"`
+}
+
+var _ = Describe("DHCPOverrides configuration test", func() {
```
srm09, line 53: This is missing an AfterEach block for cleanup of the cluster. Checkout https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/8cbf21df6b1632b48cf81cdc732bf1e44dca51b8/test/e2e/anti_affinity_test.go#L39-L59

test/e2e/dhcp_overrides_test.go
```diff
@@ -0,0 +1,121 @@
...
+				err := yaml.Unmarshal(metadata, guestInfoMetadata)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(guestInfoMetadata.Network.Ethernets["id0"].DHCP4Overrides.SendHostname).NotTo(BeNil())
+				Expect(*guestInfoMetadata.Network.Ethernets["id0"].DHCP4Overrides.SendHostname).To(BeFalse())
+			}
+		})
+	})
+})
+
+func getVMMetadata(vm types.VSphereVM) []byte {
```
srm09, line 94: rewrite the signature to send an error if the metadata is not found and assert on the error on line 83, where the function is called.

test/e2e/dhcp_overrides_test.go
```diff
@@ -0,0 +1,121 @@
...
+	finder := find.NewFinder(vsphereClient.Client, false)
+	dc, err := finder.Datacenter(ctx, vm.Spec.Datacenter)
+	Expect(err).NotTo(HaveOccurred())
+	finder.SetDatacenter(dc)
```
srm09, line 98: There is a global finder initialized for the e2e tests, consider using that.

srm09 (changes requested):
Lotta changes in the e2e test, I have added some references to other tests you can use to model this e2e.

pkg/util/machines_test.go
```diff
@@ -382,6 +382,84 @@ network:
       wakeonlan: true
       dhcp4: true
       dhcp6: true
+`,
+		},
+		{
+			name: "dhcp4+dhcp6+dhcpOverrides",
```
srm09, line 388: add an explicit assertion for when the values are not set, just to verify the default behavior.

srm09:
Also if it is not too much, could you try adding a controller test as well.

test/e2e/dhcp_overrides_test.go
adobley: Updated, thanks!

test/e2e/dhcp_overrides_test.go
adobley: Updated, thanks!

pkg/util/machines_test.go
adobley: Updated, thanks!

yastij:
Mostly lgtm pending tests passing, same ask as @srm09 regarding controller tests. Thanks for working on this!

test/e2e/dhcp_overrides_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+			list := getVSphereVMsForCluster(clusterName, namespace.Name)
+			for _, vm := range list.Items {
+				metadata, err := getVMMetadata(vm)
+				Expect(err).NotTo(HaveOccurred())
+				guestInfoMetadata := &GuestInfoMetadata{}
+				err = yaml.Unmarshal(metadata, guestInfoMetadata)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(guestInfoMetadata.Network.Ethernets["id0"].DHCP4Overrides.SendHostname).NotTo(BeNil())
+				Expect(*guestInfoMetadata.Network.Ethernets["id0"].DHCP4Overrides.SendHostname).To(BeFalse())
```
srm09, line 98: From the `dhcp-overrides.yaml`, the override is being set only on the CP machine template. Update the kustomization to add it to both the templates, or a better test is to add different properties for the CP and worker node machine template.
When asserting them, the main difference would be that the CP VM would have the `cluster.x-k8s.io/control-plane: ""` label. So writing an assertion based off that would be the best thing to do.

adobley: Thanks! Updated!

srm09:
The [DHCP override e2e test](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/1615/pull-cluster-api-provider-vsphere-full-e2e/1592973058149191680#1:build-log.txt%3A804-864) passed in the recent run.

PR #1618, opened 2022-09-14
Maintain flavors in template directory

johananl:
**What this PR does / why we need it**: Maintain generated flavors in the repo. This allows generating workload clusters using the `--from` flag of `clusterctl generate` which is useful during development.

This practice is used in other providers such as CAPA and CAPZ. Furthermore, it is **expected** that all providers publish the generated templates in the repo. From the CAPI [docs](https://cluster-api.sigs.k8s.io/clusterctl/provider-contract.html#provider-repositories):

>Additionally, the provider repository SHOULD contain the following files:
>
>- Workload cluster templates

None

**Special notes for your reviewer**: Are we running `make verify-gen` somewhere in the CI? If not, I think we should because this would ensure contributors don't forget to run `make generate` before committing.

**Release note**:

```release-note
Maintain workload cluster templates in the CAPV repository.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,7 @@ BIN_DIR := $(ROOT_DIR)/bin
 TOOLS_DIR := $(ROOT_DIR)/hack/tools
 TOOLS_BIN_DIR := $(TOOLS_DIR)/bin
 export PATH := $(abspath $(TOOLS_BIN_DIR)):$(PATH)
+FLAVOR_DIR := $(ROOT_DIR)/templates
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -260,6 +261,7 @@ modules: ## Runs go mod to ensure proper vendoring
 generate: ## Generate code
 	$(MAKE) generate-go
 	$(MAKE) generate-manifests
+	$(MAKE) generate-flavors
```
yastij, line 264: +1

yastij (commented):
/approve
/assign @srm09 

assigning @srm09 for a final pass and lgtm

PR #1623, opened 2022-09-16
🌱 Exposes a new reason for VMProvisionedCondition

### LINKED ISSUE #1490: can we have more fine grained power state (i.e. a reason such as "powered on, waiting for IP") ?
/kind bug

- feel free to close this issue, it might not be actionable but i wanted to write it up... 
- were seeing VMs on some hardware coming online as IPv6 vms instead of IPv4 ones.
- in the same cluster, were seeing these stuck in PoweringOn state
```
  - lastTransitionTime: "2022-03-23T14:25:34Z"
    message: 1 of 2 completed
    reason: PoweringOn
    severity: Info
    status: "False"
    type: Ready
```

Is it possible that, if a VM cant get an IPv4 address, it might be misinterpretted as being in the `PoweringOn` state eternally ? If so, can we have an "intermediate" state of "PoweredOnWaitingIp" or something ? 

In the code for the `reconcilePowerState` , we say `powering on`...

```       
    case infrav1.VirtualMachinePowerStatePoweredOff:
                ctx.Logger.Info("powering on")
                task, err := ctx.Obj.PowerOn(ctx)
```
If the machine is in the `Off` state.  We then mark the machine as "false" bc its "PoweringOn".

However, I feel like this might be misleading, bc we are currently seeing "Powered On" machines that are flagged by CAPI as "ready=False" , "reason=PoweringOn"... 

Can we add more granularity to that log message, something like:

```
ctx.Logger.Info("Machine was powered on but has no IP... waiting for it to be fully networked...")
```

- More fine grained information about the state of powered off machines ... 
- Is it easy to add a more meaningfull value to the condition_consts.go 
- Have the reason for the reason, i.e., what do we really mean when we say it is in "PoweringOn" state?

srm09:
Adds a new reason on the VSphereVM object's condition to convey that the Machine is waiting on the IP address. This is set if the VSphereVM has no IP addresses reported in its state.

Fixes #1490

**Release note**:

```release-note
Exposes reason to convey VSphereVM waiting for IP address
```

### DIFF
--- a/apis/v1beta1/condition_consts.go
+++ b/apis/v1beta1/condition_consts.go
@@ -61,6 +61,12 @@ const (
 	// a static IP address.
 	WaitingForStaticIPAllocationReason = "WaitingForStaticIPAllocation"
 
+	// WaitingForIPAllocationReason (Severity=Info) documents a VSphereVM waiting for the allocation of
+	// an IP address.
+	// This is used when the dhcp4 or dhcp6 for a VSphereVM is set and the VSphereVM is waiting for the
+	// relevant IP address  to show up on the VM.
+	WaitingForIPAllocationReason = "WaitingForIPAllocation"
+
 	// CloningReason documents (Severity=Info) a VSphereMachine/VSphereVM currently executing the clone operation.
 	CloningReason = "Cloning"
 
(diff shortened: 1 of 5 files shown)


jayunit100:
Yesssssss thank you

PR #1632, opened 2022-09-29
📖 Proposal for IPAM support

flawedmatrix:
This PR adds a proposal for adding IPAM support in CAPV.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/docs/proposal/20220929-ipam-support.md
+++ b/docs/proposal/20220929-ipam-support.md
@@ -1,0 +1,446 @@
+﻿# CAPV IPAM Support
+
+```text
+---
+title: CAPV IPAM Support
+authors:
+  - "@christianang"
+  - "@flawedmatrix"
+  - "@adobley"
+  - "@tylerschultz"
+reviewers:
+  - "@yastij"
+  - "@srm09"
+  - "@vrabbi"
+  - "@schrej"
+creation-date: 2022-9-13
+last-updated: 2022-10-11
+status: Implementable
+---
+```
+
+## Table of Contents
+
+- [CAPV IPAM Support](#capv-ipam-support)
+  - [Table of Contents](#table-of-contents)
+  - [Glossary](#glossary)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals/Future Work](#non-goalsfuture-work)
+  - [Proposal](#proposal)
+    - [User Stories](#user-stories)
+      - [Story 1](#story-1)
+      - [Story 2](#story-2)
+      - [Story 3](#story-3)
+    - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+      - [API Changes](#api-changes)
+      - [VSphereVM Controller Changes](#vspherevm-controller-changes)
+      - [Owner Reference and Finalizer](#owner-reference-and-finalizer)
+      - [Flavors](#flavors)
+      - [Validation](#validation)
+      - [Nameservers](#nameservers)
+    - [Security Model](#security-model)
+    - [Risks and Mitigations](#risks-and-mitigations)
+  - [Alternatives](#alternatives)
+  - [Upgrade Strategy](#upgrade-strategy)
+  - [Additional Details](#additional-details)
+    - [Test Plan](#test-plan)
+    - [Graduation Criteria](#graduation-criteria)
+    - [Version Skew Strategy](#version-skew-strategy)
+  - [References](#references)
+  - [Implementation History](#implementation-history)
+
+## Glossary
+
+Refer to the [Cluster API IPAM Integration Glossary](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20220125-ipam-integration.md#glossary)
+
+## Summary
+
+This proposal adds CAPV support for the IPAM API contract described in the
+[IPAM Integration
+proposal](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20220125-ipam-integration.md).
+CAPV will be able to allocate/release IP Addresses from an IPAM provider's IP
+Pool to be used by `VSphereMachine`s.
+
+## Motivation
+
+The initial IPAM integration proposal explains the motivation for IPAM pretty well:
+
+> IP address management for machines is currently left to the infrastructure
+> providers for CAPI. Most on-premise providers (e.g. CAPV) allow the use of
+> either DHCP or statically configured IPs. Since Machines are created from
+> templates, static allocation would require a single template for each machine,
+> which prevents dynamic scaling of nodes without custom controllers that create
+> new templates.
+>
+> While DHCP is a viable solution for dynamic IP assignment which also enables
+> scaling, it can cause problems in conjunction with CAPI. Especially in smaller
+> networks rolling cluster upgrades can exhaust the network in use. When multiple
+> machines are replaced in quick succession, each of them will get a new DHCP
+> lease. Unless the lease time is very short, at least twice as many IPs as the
+> maximum number of nodes has to be allocated to a cluster.
+>
+> Metal3 has an ip-address-manager component that allows for in-cluster
+> management of IP addresses through custom resources, allowing to avoid DHCP.
+> CAPV allows to omit the address from the template while having DHCP disabled,
+> and will wait until it is set manually, allowing to implement custom
+> controllers to take care of IP assignments. At DTAG we've extended metal3's
+> ip-address-manager and wrote a custom controller for CAPV to integrate both
+> with Infoblox, our IPAM solution of choice. The CAPV project has shown
+> interest, and there has been a proposal to allow integrating metal3's
+> ip-address-manager with external systems.
+>
+> All on-premise providers have a need for IP Address Management, since they
+> can't leverage SDN features that are available in clouds such as AWS, Azure or
+> GCP. We therefore propose to add an API contract to CAPI, which allows
+> infrastructure providers to integrate with different IPAM systems. A similar
+> approach to Kubernetes' PersistentVolumes should be used where infrastructure
+> providers create Claims that reference a specific IP Pool, and IPAM providers
+> fulfill claims to Pools managed by them.
+
+### Goals
+
+- Implement support for the IPAM API contract in CAPV
+- Support both IPV4 and IPv6
+- Support dual-stack
+- Allow dynamically allocating and releasing addresses
+- Allow running multiple IPAM providers in parallel
+- Support running any IPAM provider
+
+### Non-Goals/Future Work
+
+- Configuring nameservers from Pool
+- Handling updates to the devices on `VSphereMachine` objects
+- Handling updates to the `addressesFromPools` list on individual devices
+- User direct manipulation of `IPAddressClaim`s
+- IP Address Management – this is the job of the IPAM provider
+- Support for specific IPs for specific machines - a machine will receive a
+  random IP in the pool. Implementation of this feature can be left to
+  future work.
+
+## Proposal
+
+Following the contract specified by the IPAM integration proposal we plan to:
+
+- Add the `addressesFromPools` field to the `spec.network.devices` section of
+  the `VSphereMachine` and `VSphereMachineTemplate` CRDs. The
+  `addressesFromPools` field will allow a user to specify that the network
+  device's IP address should be claimed from an IP pool.
+- If a `VSphereMachine` has been specified to claim an IP address from an
+  IP pool, then CAPV will create an `IPAddressClaim`, which the IPAM provider
+  will use to allocate an `IPAddress`. This `IPAddress` can then be used by the
+  `VSphereMachine`.
+
+### User Stories
+
+#### Story 1
+
+As a user, I want to be able to create a `VSphereMachine` where the IP address
+for the VM comes from an IPAM Provider instead of from DHCP.
+
+#### Story 2
+
+As a user I want to be able to allocate either IPv4 or IPv6 addresses for my
+`VSphereMachine`.
+
+#### Story 3
+
+As a user I want to be able to release my IP address allocation when the
+`VSphereMachine` is deleted for future `VSphereMachine`s to use.
+
+### Implementation Details/Notes/Constraints
+
+#### API Changes
+
+Adding `addressesFromPools` to the `VSphereMachineTemplate`
+
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
+kind: VSphereMachineTemplate
+metadata:
+  name: example
+  namespace: vsphere-site1
+spec:
+  template:
+    spec:
+      cloneMode: FullClone
+      numCPUs: 8
+      memoryMiB: 8192
+      diskGiB: 45
+      network:
+        devices:
+        - dhcp4: false
+          dhcp6: false
+          networkName: VM Network
+          addressesFromPools:
+            # reference to the pool
+            - apiGroup: ipam.cluster.x-k8s.io
+              kind: InClusterIPPool
+              name: testpool4
+            - apiGroup: ipam.cluster.x-k8s.io
+              kind: InClusterIPPool
+              name: testpool6
+```
+
+Add `AddressesFromPools` to the `NetworkDeviceSpec`:
+
+```go
+// NetworkDeviceSpec defines the network configuration for a virtual machine's
+// network device.
+type NetworkDeviceSpec struct {
+ // FromPools is a list of references to the pools from which IP addresses should be claimed.
+ AddressesFromPools []corev1.TypedLocalObjectReference `json:"addressesFromPools,omitempty"`
+}
+```
+
+#### VSphereVM Controller Changes
+
+1. If the `addressesFromPools` field is set on a device then this controller
+   will create a new `IPAddressClaim` for each pool configured on the device.
+   This `IPAddressClaim` will be created in the same namespace as the
+   `VSphereVM` / `VSphereMachine` / `VSphereMachineTemplate`. The `IPAddressClaim`
+   will be named according to the device index and the pool index in the
+   `VSphereVM` spec. The controller will not continue the VM creation process
+   until all `IPAddressClaim`s have been fulfilled.
+
+    ```yaml
+    apiVersion: ipam.cluster.x-k8s.io/v1alpha1
+    kind: IPAddressClaim
+    metadata:
+      name: "example-claim-0-0"
+      namespace: "vsphere-site1"
+      finalizers: ["ipam.cluster.x-k8s.io/ip-claim-protection"]
+      ownerReferences:
+      - apiVersion: "infrastructure.cluster.x-k8s.io/v1beta1"
+        kind: "VSphereVM"
+        name: "prod-fdftj"
+        uid: "097ccb1e-959f-4357-b0bf-cb3eccb2a185"
+    spec:
+      poolRef:
+        apiGroup: ipam.cluster.x-k8s.io
+        kind: InClusterIPPool
+        name: testpool4
+    status:
+      addressRef: null
+    ```
+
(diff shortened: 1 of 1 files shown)


docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,424 @@
...
+
+
+## Glossary
+
+Refer to the [Cluster API IPAM Integration Glossary](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20220125-ipam-integration.md#glossary)
+
+## Summary
+
+This proposal adds CAPV support for the IPAM API contract described in the IPAM
+Integration proposal. CAPV will be able to allocate/release IP Addresses from
```
srm09, line 62:
```suggestion
This proposal adds CAPV support for the IPAM API contract described in the [IPAM
Integration proposal](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20220125-ipam-integration.md). CAPV will be able to allocate/release IP Addresses from

```

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,424 @@
...
+> fulfill claims to Pools managed by them.
+
+### Goals
+
+* Implement support for the IPAM API contract in CAPV
+* Support both IPV4 and IPv6
+* Support dual-stack
+* Allows dynamically allocating and releasing addresses
+* Allows running multiple IPAM providers in parallel
+* Supports running any IPAM provider
```
srm09, line 109:
```suggestion
* Allow dynamically allocating and releasing addresses
* Allow running multiple IPAM providers in parallel
* Support running any IPAM provider
```

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,424 @@
...
+
+1. If the FromPools field is set on a device then this controller will create a
+   new IPAddressClaim for each pool configured on each device. The
+   IPAddressClaim will have an annotation containing the mac address of the
+   device, which will be used to correlate the IPAddress to the correct device.
+   The controller will not continue the VM creation process until all
+   IPAddressClaims have been fulfilled.
+
+        Note: The name of the IP Address Claim will have an appended randomly
+        generated hash for uniqueness[a][b][c][d][e][f][g].
```
srm09, line 204: Using the `generateName` field in the `Metadata` object will take care of the randomly generated naming scheme.

srm09 (commented):
minor nits, but lgtm.
Like we chatted in the office hours, we can retrofit the changes to the CI once we have more clarity on those.

srm09:
Run `make lint-markdown` to run the markdown verifier locally and fix the issues.

yastij (approved):
/approve

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,422 @@
...
+      memoryMiB: 8192
+      diskGiB: 45
+      network:
+        devices:
+        - dhcp4: false
+          dhcp6: false
+          networkName: VM Network
+          fromPools:
+            # reference to the pool
+            - group: ipam.cluster.x-k8s.io/v1alpha1
```
schrej, line 177: This should either be _only_ the group, aka. `ipam.cluster.x-k8s.io`, or be renamed to `apiVersion`.

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,422 @@
...
+ // FromPools is a list of references to the pools from which IP addresses should be claimed.
+ FromPools []corev1.TypedLocalObjectReference `json:"fromPools,omitempty"`
+}
+```
+
+#### VSphereVM Controller Changes
+
+1. If the FromPools field is set on a device then this controller will create a
+   new IPAddressClaim for each pool configured on each device. The
+   IPAddressClaim will have an annotation containing the mac address of the
```
schrej, line 200: Just curious how this works - do you create the machine directly, without booting it, and then wait for IPs?
In our hacked controller we're currently using the network name to match, but I guess you could technically have multiple interfaces to the same network.

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,422 @@
...
+   device, which will be used to correlate the IPAddress to the correct device.
+   The controller will not continue the VM creation process until all
+   IPAddressClaims have been fulfilled.
+
+   **Note**: The name of the IP Address Claim will have an appended randomly
+   generated hash for uniqueness by use of the `generatedName` field
```
schrej, line 206: So `<machine name>-<hash>`?

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,422 @@
...
+### Risks and Mitigations
+
+- Unresponsive IPAM providers can prevent successful creation of a cluster
+- Any issues with fulfilling the IPAddressClaim can prevent successful creation of a cluster
+  - This is partially by design i.e we don't want the cluster to be created if
+    it can't fulfill the ip address claim, but this is just added complexity
+    that a user has to be aware of.
+- Wrong IP address allocations (e.g. duplicates) can brick a cluster’s network
+- Incorrectly assigning IP address claims to the wrong device
+- How do we handle a user deleting a claim directly? This could create issues
```
schrej, line 356: It could be protected by a finalizer until the machine is deleted. If users delete it regardless they're on their own.

schrej (commented):
Just a few notes, otherwise lgtm.

docs/proposal/20220929-ipam-support.md
tylerschultz: Good catch, this is a mistake in the proposal. The field is not called group, it's called `apiGroup`. Update coming.

docs/proposal/20220929-ipam-support.md
tylerschultz: We've made a pass at an implementation that does not require using the mac address, nor any annotation. We're using the device index and pool index naming scheme as you suggested, and it seems to be working. The code does use the mac address internally to map pool specs to devices, but it is not exposed to the user in any way. Perhaps the mention of using the mac address should be removed from the proposal. 

The code creates the vm in a stopped state and the vm has the devices which can be retrieved from the vSphere api. The devices have mac addresses assigned. The pools are reconciled before the machine is powered on.

docs/proposal/20220929-ipam-support.md
tylerschultz: This is outdated too. The claims will have `<vm-name>-<device-dx>-<pool-idx>` names, per your suggestion.

docs/proposal/20220929-ipam-support.md
tylerschultz: We think this is covered in the section named Owner References. There is an existing finalizer on the VSphereVM. The vm object is not deleted until the vm is actually gone. We've added an owner ref (back to the vm) to the claim, and the claim will be cascade deleted when the vm object is deleted. CAPV controllers will not delete claims directly, and relies on the garbage collector. This seems safe to us, and a finalizer didn't seem necessary. If a user deletes the Claim manually, they'll have a bad day.

tylerschultz: We've been discussing this more. Because deleting a claim can cause an IP to be assigned to multiple devices, which would be a calamity, we think we should add the finalizer to the claim. We'll update the proposal. The finalizer would ensure the vm is no longer present in the vSphere api.

schrej:
/lgtm

/hold
so @srm09 can have another pass

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+// network device.
+type NetworkDeviceSpec struct {
+ // FromPools is a list of references to the pools from which IP addresses should be claimed.
+ FromPools []corev1.TypedLocalObjectReference `json:"fromPools,omitempty"`
+}
+```
+
+#### VSphereVM Controller Changes
+
+1. If the FromPools field is set on a device then this controller will create a
```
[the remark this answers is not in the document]
schrej: In this case "on the devices" means "in the device spec of the VSphereMachine object" I think.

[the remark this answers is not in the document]
tylerschultz: I agree this is confusing. The device exists on the VM object in the vSphere API before the vm is powered on. The device spec of the VSphereMachine Kubernetes api object describes configuration for that device. 

@geetikabatra I don't quite understand your second question? The IPAM provider will assign one IPAddress per IPAddressClaim. One IPAddressClaim will be created for each entry in the fromPools array. The IPAM provider is configured with the IPs that to be made assignable.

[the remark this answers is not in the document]
geetikabatra: Thanks, @schrej, and @tylerschultz. So, I am wondering does the IP Range in `fromPools` is coming from the actual device(NIC) in the device spec of the VSphereMachineObject? There could be many devices like this associated. Would `fromPools`barray change with each VSphereMacgine object?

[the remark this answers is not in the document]
tylerschultz: @geetikabatra There is no IP range described in `fromPools`. The `fromPools` section only describes the name, group and kind of a IPAM provider. The IPAM Provider, a separate deployment/controller (not described in depth here), is configured with the pool and ip range. Each VSphereMachine's fromPool entry would receive one IP from the pool's range.

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+> CAPV allows to ommit the address from the template while having DHCP disabled,
+> and will wait until it is set manually, allowing to implement custom
+> controllers to take care of IP assignments. At DTAG we've extended metal3's
+> ip-address-manager and wrote a custom controller for CAPV to integrate both
+> with Infoblox, our IPAM solution of choice. The CAPV project has shown
+> interest, and there has been a proposal to allow integrating metal3's
+> ip-address-manager with external systems.
+>
+> All on-premise providers have a need for IP Address Management, since they
+> can't leverage SDN features that are available in clouds such as AWS, Azure or
```
geetikabatra, line 95: Hi! Why can't we leverage SDN features of cloud providers?

docs/proposal/20220929-ipam-support.md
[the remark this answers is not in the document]
geetikabatra: Aah! Okay, I get it now. A suggestion, we should not name it to `fromPools`. It should be more intuitive like `IPProvider`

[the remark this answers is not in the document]
schrej: I don't think `IPProvider` would be intuitive. The idea of the IPAM contract is to have IP address pools defined in the cluster, which are managed by their corresponding IPAM provider. In theory you could of course have one massive Pool of addresses used by all clusters,  but it's also likely that each cluster uses their own pool.
Since you interact with Pools, and not the provider, when configuring where an Address comes from, `pool` should be included in the name.
Maybe something like `addressesFromPools` would be more clear.

FWIW: For the metal3 provider it's [going to be called](https://github.com/metal3-io/cluster-api-provider-metal3/pull/769) `fromPoolRef`, but there you can only reference a single pool as the configuration looks a bit different in general.

docs/proposal/20220929-ipam-support.md
schrej: This is a quote from the CAPI IPAM proposal, not written specifically for this one.

When you are on-premise, you can't use a cloud provider's SDN. It might be a viable use case to run vSphere on a cloud, and integrate it with the provider's SDN, but that's not the case for on-prem setups. The features I'm talking about here is the ability for the compute service of the cloud to request addresses from the networking service, and assign them statically. Even if you have SDN in your datacenter, you still need to integrate with it, which is something the IPAM contract allows you to do.

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+    lacks a namespace field on the reference.
+
+### Risks and Mitigations
+
+- Unresponsive IPAM providers can prevent successful creation of a cluster
+- Any issues with fulfilling the IPAddressClaim can prevent successful creation of a cluster
+  - This is partially by design i.e we don't want the cluster to be created if
+    it can't fulfill the ip address claim, but this is just added complexity
+    that a user has to be aware of.
+- Wrong IP address allocations (e.g. duplicates) can brick a cluster’s network
```
geetikabatra, line 352: A nit pick: brick->break. Or does this have a meaning I am not aware of?

docs/proposal/20220929-ipam-support.md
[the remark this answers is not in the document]
geetikabatra: `addressesFromPools` also sounds good to me.

docs/proposal/20220929-ipam-support.md
schrej: IT slang for a broken device. A device without its firmware/software and no way to recover it has no use, but you can still use it as a brick or paperweight.
It's used as a synonym for "break" here. Might be my fault if it's from the CAPI IPAM proposal.

docs/proposal/20220929-ipam-support.md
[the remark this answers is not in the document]
tylerschultz: Ok, sounds like a plan. We'll rename to `addressesFromPools`.

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+    spec:
+      poolRef:
+        apiGroup: ipam.cluster.x-k8s.io
+        kind: InClusterIPPool
+        name: testpool4
+    status:
+      addressRef: null
+    ```
+
+1. It will set the VSphereMachine Status to reflect that an IPAddressClaim(s)
```
srm09, line 222: This is more of an implementation detail but just calling it out. Since there might be cases of multiple entires in the `FromPools` field as it is a slice, we should create a single condition and call out how many of the IP addresses have been allocated, or how many of the IP addresses are remaining to be allocated. Similar to the way MachineDeployment or KCP has conditions around how many machines are ready/available.

srm09 (commented):
added some minor suggestions and a question or two.
Do squash the commits into a single one and this one is ready to merge.
/approve

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+   The controller will not continue the VM creation process until all
+   IPAddressClaims have been fulfilled.
+
+    ```yaml
+    apiVersion: ipam.cluster.x-k8s.io/v1alpha1
+    kind: IPAddressClaim
+    metadata:
+      name: "example-claim-0-0"
+      namespace: "vsphere-site1"
+      finalizers: ["ipam.cluster.x-k8s.io/ip-claim-protection"]
```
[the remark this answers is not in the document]
srm09: What was the consideration around setting a finalizer? From what I have seen in the CAP* controllers, the controllers that reconcile the objects are the one setting and removing the finalizers. And I was under the assumption, that the IPAM provider implementation would have controllers that watch and add the finalizers.

docs/proposal/20220929-ipam-support.md
tylerschultz: We've updated the proposal to include status on IPs waiting/allocated.

docs/proposal/20220929-ipam-support.md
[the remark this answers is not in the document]
schrej: The IPAM provider will also set a finalized to free up the IP after the claim is removed.

The finalizer set by the infra provider is to prevent deletion of the claim while the IP is still in use by the Machine.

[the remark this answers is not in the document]
tylerschultz: We've updated the 'Owner Reference and Finalizer' section. Let us know if there's still more to explain.

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+ FromPools []corev1.TypedLocalObjectReference `json:"addressesFromPools,omitempty"`
+}
+```
+
+#### VSphereVM Controller Changes
+
+1. If the FromPools field is set on a device then this controller will create a
+   new IPAddressClaim for each pool configured on each device. The
+   IPAddressClaim will be named according to the device index and the pool
+   index in the VSphereVM spec.
```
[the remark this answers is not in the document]
tylerschultz: This section has been updated with the suggestion.

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+#### Story 3
+
+As a user I want to be able to release my IP Address allocation when the
+VSphereMachine is deleted for future VSphereMachines to use.
+
+### Implementation Details/Notes/Constraints
+
+#### API Changes
+
+Adding `addressesFromPools`(type []corev1.TypedLocalObjectReference) to the VSphereMachineTemplate
```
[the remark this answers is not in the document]
tylerschultz: updated

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+  then CAPV will create an IPAddressClaim, which the IPAM provider will use to
+  allocate an IP Address. This IPAddress can then be used by the
+  VSphereMachine.
+
+### User Stories
+
+#### Story 1
+
+As a user, I want to be able to create a VSphereMachine where the IP Address
+for the machine comes from an IPAM Provider instead of DHCP.
```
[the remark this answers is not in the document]
tylerschultz: updated

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+Following the contract specified by the IPAM integration proposal we plan to:
+
+- Add the addressesFromPools field to the spec.network.devices section of the
+  VSphereMachine and VSphereMachineTemplate. The addressesFromPools field will allow a
+  user to specify that the network device IPAddress should be claimed from an
+  IPPool.
+- If a VSphereMachine has been specified to claim an IP Address from an IPPool,
+  then CAPV will create an IPAddressClaim, which the IPAM provider will use to
+  allocate an IP Address. This IPAddress can then be used by the
+  VSphereMachine.
```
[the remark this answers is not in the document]
tylerschultz: updated

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+> While DHCP is a viable solution for dynamic IP assignment which also enables
+> scaling, it can cause problems in conjunction with CAPI. Especially in smaller
+> networks rolling cluster upgrades can exhaust the network in use. When multiple
+> machines are replaced in quick succession, each of them will get a new DHCP
+> lease. Unless the lease time is very short, at least twice as many IPs as the
+> maximum number of nodes has to be allocated to a cluster.
+>
+> Metal3 has an ip-address-manager component that allows for in-cluster
+> management of IP addresses through custom resources, allowing to avoid DHCP.
+> CAPV allows to ommit the address from the template while having DHCP disabled,
```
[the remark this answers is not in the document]
tylerschultz: updated

docs/proposal/20220929-ipam-support.md
```diff
@@ -0,0 +1,423 @@
...
+  - "@adobley"
+  - "@tylerschultz"
+reviewers:
+  - "@yastij"
+  - "@srm09"
+  - "@vrabbi"
+  - "@schrej"
+creation-date: 2022-9-13
+last-updated: 2022-10-11
+status: In Review
```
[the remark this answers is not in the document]
tylerschultz: updated

srm09:
Just one last thing about the name of the finalizer to protect the IP address in use, since the finalizer is being applied by CAPV we should tweak the name to something on the lines of the existing finalizers, `vspherevm.infrastructure.cluster.x-k8s.io/ip-address-protection`. We can discuss more on the implementation PR, this is a nit.

Thanks for all the hard work here, this looks good to go.
/lgtm

srm09:
whoops missed this
/unhold

PR #1641, opened 2022-10-12
🐛  Adds default resource pool logic for cluster module creation

### LINKED ISSUE #1640: Cluster creation fails when no resource pool is specified
/kind bug

- Enable the node anti affinity feature for the CAPV deployment
- Create a cluster with the KCP and the MD objects using `VSphereMachineTemplate` objects without a resource pool, set `VSPHERE_RESOURCE_POOL` to `''` 
- Cluster creation fails since the `VSphereCluster` object never moves to `Ready` state.

- Cluster creation should succeed since resource pool is an optional field for `VSphereMachineTemplate` CR.

Similar to the clone operation, if no resource pool is specified, then the resource pool lookup should fallback to the default value.

**Environment:**

- Cluster-api-provider-vsphere version: v1.4.0

srm09:
This patch adds the logic to fall back to the default resource pool, if no resource pool is specified, during cluster module creation. It also checks whether the resource pool is owned by a compute cluster or a standalone host. In case of the owner being a standalone host, a warning is introduced on the VSphereCluster object and the cluster creation proceeds.

Fixes #1640

**Release note**:

```release-note
Adds default resource pool logic for cluster module creation
```

### DIFF
--- a/controllers/clustermodule_reconciler.go
+++ b/controllers/clustermodule_reconciler.go
@@ -17,9 +17,11 @@ limitations under the License.
 package controllers
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util"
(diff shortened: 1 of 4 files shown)


controllers/clustermodule_reconciler.go
```diff
@@ -253,3 +253,29 @@ func (r Reconciler) fetchMachineOwnerObjects(ctx *context.ClusterContext) (map[s
...
+
+func generateClusterModuleErrorMessage(errList []clusterModError) string {
+	sb := strings.Builder{}
+	sb.WriteString("failed to create cluster modules for: ")
+
+	for _, e := range errList {
+		sb.WriteString(fmt.Sprintf("%s %s, ", e.name, e.err.Error()))
+	}
+	msg := sb.String()
+	return msg[:len(msg)-2]
```
yastij, line 280: why is this len -2 ?

srm09: Remove the trailing `, ` after the final object name and error

PR #1644, opened 2022-10-13
🐛 Allow mutation of OS field and only when it's empty

yastij:

**What this PR does / why we need it**: This PR allows defaulting/mutation of the OS field, this is needed to fix errors when upgrading the controller

**Which issue(s) this PR fixes** : Fixes https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1643

**Release note**:

```release-note

```


srm09:
/retitle 🐛 Allow mutation of OS field and only when it's empty

srm09:
@yastij this might need a rebase

PR #1653, opened 2022-10-20
✨ Creates an explicit token secret for service account

### LINKED ISSUE #1651: TKC creation fails for 1.24 Supervisor cluster
/kind bug

Currently, the provider service account controller is responsible for creating the service account in the namespace specified in the spec of the `ProviderServiceAccount` object. It then waits for the token controller to populate the secret which is then copied over to the guest cluster.
With 1.24, the `LegacyServiceAccountTokenNoAutoGeneration` is enabled by default.  Due to this, Secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount. Check the release notes [here](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes)

For a supervisor cluster with k8s 1.24, creation of TKC fails since the CPI and CSI pods never move to Running state. 
```
vmware-system-cloud-provider   guest-cluster-cloud-provider-8547fc5667-wl2z4            0/1     ContainerCreating   0          4h30m
vmware-system-csi              vsphere-csi-controller-5d7c9896db-pj8qr                  0/6     Pending             0          4h30m
```
More investigation shows, that the pods are unable to mount the volume containing the secret for the service account.
```
Warning  FailedMount  41m (x17 over 4h30m)    kubelet  Unable to attach or mount volumes: unmounted volumes=[ccm-provider], unattached volumes=[kube-api-access-drl2m ccm-config ccm-owner-reference ccm-provider]: timed out waiting for the condition
  Warning  FailedMount  11m (x52 over 4h28m)    kubelet  Unable to attach or mount volumes: unmounted volumes=[ccm-provider], unattached volumes=[ccm-config ccm-owner-reference ccm-provider kube-api-access-drl2m]: timed out waiting for the condition
  Warning  FailedMount  102s (x141 over 4h32m)  kubelet  MountVolume.SetUp failed for volume "ccm-provider" : secret "cloud-provider-creds" not found
```

The TKC creation should succeed.

**Environment:**

- Cluster-api-provider-vsphere version: v1.3.1
- Kubernetes version: (use `kubectl version`): 1.24 

/assign

srm09:
This patch changes the default behavior of waiting for the generation of a default secret for the service account and propagate that over to the guest cluster.
Instead, alongwith the service account a secret of type "kubernetes.io/service-account-token" is created to host the token for the service account. The controller waits for the realization of this secret and then propagates this data over to the guest cluster.

This is required since 1.24 disables the auto generation of a default token secret for every service account created.

Fixes #1651 

**Release note**:
```release-note
Creates an explicit token secret for service account
```

### DIFF
--- a/controllers/serviceaccount_controller.go
+++ b/controllers/serviceaccount_controller.go
@@ -89,9 +89,10 @@ func AddServiceAccountProviderControllerToManager(ctx *context.ControllerManager
 			&source.Kind{Type: &vmwarev1.ProviderServiceAccount{}},
 			handler.EnqueueRequestsFromMapFunc(r.providerServiceAccountToVSphereCluster),
 		).
+		// Watches the secrets to re-enqueue once the service account token is set
 		Watches(
-			&source.Kind{Type: &corev1.ServiceAccount{}},
-			handler.EnqueueRequestsFromMapFunc(r.serviceAccountToVSphereCluster),
+			&source.Kind{Type: &corev1.Secret{}},
+			handler.EnqueueRequestsFromMapFunc(r.secretToVSphereCluster),
 		).
 		Complete(r)
 }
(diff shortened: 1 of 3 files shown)


controllers/serviceaccount_controller.go
```diff
@@ -89,9 +89,10 @@ func AddServiceAccountProviderControllerToManager(ctx *context.ControllerManager
 			&source.Kind{Type: &vmwarev1.ProviderServiceAccount{}},
 			handler.EnqueueRequestsFromMapFunc(r.providerServiceAccountToVSphereCluster),
 		).
+		// Watches the secrets to re-enqueue once the service account token is set
 		Watches(
-			&source.Kind{Type: &corev1.ServiceAccount{}},
-			handler.EnqueueRequestsFromMapFunc(r.serviceAccountToVSphereCluster),
+			&source.Kind{Type: &corev1.Secret{}},
```
yastij, line 94: why are we watching secrets instead of SAs? sure we're creating the secret now, but that's still done because there's a new service account being setup.

yastij (commented):
/approve

one comment though

controllers/serviceaccount_controller.go
srm09: Because the generated secret eventually gets populated with the token for the service account, which needs to be synced over to the guest cluster

yastij (commented):
/lgtm

PR #1666, opened 2022-10-25
✨ Add implementation of Node IPAM support

### LINKED ISSUE #1550: IPAM support for CAPV
Umbrella issue for all tasks required to implement IPAM support for CAPV
This issue will be updated over time

- [ ] IPAM Contract implementation - https://github.com/kubernetes-sigs/cluster-api/pull/6313

tylerschultz:

- Adds `addressesFromPools` to the VSphereVM network device spec
- Creates IPAddressClaims when AddressesFromPools is specified
- IP addresses created by IPAM providers are assigned to vSphere machines via the VM metadata
- IPv4 Gateways from IPAddresses (created by IPAM providers) are assigned to the Gateway4 field
- IPv6 Gateways from IPAddresses (created by IPAM providers) are assigned to the Gateway6 field
- All gateways in the same IP family assigned to a VSphereVM network device must be the same.
- This implementation expects that an appropriate gateway for the IP family exists for every IPAddress that comes from an IPAM provider.
- Adds the `IPAddressClaimed` Condition to provide visibility of the status of IP address acquisition
- Adds the finalizer to prevent IPAMClaims from being deleted early
- IPAddressClaims have an OwnerReference to VSphereVMs to ensure they are garbage-collected when the VM is deleted.

Co-authored-by: Aidan Obley <aobley@vmware.com>
Co-authored-by: Tyler Schultz <tschultz@vmware.com>
Co-authored-by: Christian Ang <angc@vmware.com>
Co-authored-by: Edwin Xie <exie@vmware.com>

Fixes #1550 

This PR implements this proposal: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1632

The e2e test will come in a separate PR. The IPAM feature requires an unreleased version of CAPI with a version of clusterctl with `--ipam-provider` flag and related IPAM CRDs. The main branch of CAPI also bumps to ginkgo v2. https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1667

**Release note**:

```release-note
Adds support for node IPAM
```


controllers/vspherevm_controller.go
```diff
@@ -306,6 +308,31 @@ func (r vmReconciler) reconcileDelete(ctx *context.VMContext) (reconcile.Result,
...
+			ctx.Logger.Info(fmt.Sprintf("removing finalizer for IPAddressClaim %s", ipAddrClaimName))
+			ipAddrClaimKey := apitypes.NamespacedName{
+				Namespace: ctx.VSphereVM.Namespace,
+				Name:      ipAddrClaimName,
+			}
+			if err := ctx.Client.Get(ctx, ipAddrClaimKey, ipAddrClaim); err != nil {
+				if apierrors.IsNotFound(err) {
+					continue
+				}
+				return reconcile.Result{}, errors.Wrapf(err, fmt.Sprintf("failed to find IPAddressClaim %s to remove the finalizer", ipAddrClaimName))
```
tylerschultz, line 326: This line should be called out. We see that other errors in the delete reconcile do not return errors, they log and eat the errors. We've chosen to return the error during finalizer removal errors.

Does anyone have context on why errors are not returned? Should this new code not return errors?

pkg/util/machines.go
```diff
@@ -114,12 +114,24 @@ func IsControlPlaneMachine(machine metav1.Object) bool {
 // GetMachineMetadata returns the cloud-init metadata as a base-64 encoded
 // string for a given VSphereMachine.
 func GetMachineMetadata(hostname string, vsphereVM infrav1.VSphereVM, networkStatuses ...infrav1.NetworkStatus) ([]byte, error) {
+	return GetMachineMetadataWithIPAMState(hostname, vsphereVM, map[string]infrav1.NetworkDeviceSpec{}, networkStatuses...)
+}
+
+// GetMachineMetadata returns the cloud-init metadata as a base-64 encoded
+// string for a given VSphereMachine.
+// IPAM state includes IP and Gateways that should be added to each device.
```
srm09, line 122: Update the comment and consider making it a private method if not used directly outside the package.

pkg/util/machines_test.go
```diff
@@ -644,14 +645,87 @@ network:
       wakeonlan: true
       dhcp4: false
       dhcp6: true
+`,
+		},
+		{
+			name: "ipam",
```
srm09, line 651: nit: describe the case better

apis/v1beta1/vspherevm_types.go
```diff
@@ -28,6 +28,10 @@ const (
 	// VMFinalizer allows the reconciler to clean up resources associated
 	// with a VSphereVM before removing it from the API Server.
 	VMFinalizer = "vspherevm.infrastructure.cluster.x-k8s.io"
+
+	// IPClaimFinalizer allows the reconciler to prevent deletion of an
+	// IPAddressClaim that is in use.
+	IPAddressClaimFinalizer = "ipam.cluster.x-k8s.io/ip-claim-protection"
```
srm09, line 34: What do you think about changing this reflect that it is being controller by CAPV, 
`vspherevm.infrastructure.cluster.x-k8s.io/ip-claim-protection` ?

srm09 (commented):
Did an initial pass, bunch of nits apart from setting the Conditions on the VSphereVM object. Need to fully review the `reconcileIPAddresses` function once, but the changes overall look good.

config/rbac/role.yaml
```diff
@@ -309,6 +309,24 @@ rules:
...
+- apiGroups:
+  - ipam.cluster.x-k8s.io
+  resources:
+  - ipaddressclaims
+  verbs:
+  - create
+  - get
+  - list
+  - update
+  - watch
```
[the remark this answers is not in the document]
tylerschultz: Delete does not appear to be necessary. We've seen the objects get deleted successfully. We'll make another pass to be sure.

pkg/services/govmomi/service.go
```diff
@@ -250,13 +263,176 @@ func (vms *VMService) reconcileNetworkStatus(ctx *virtualMachineContext) error {
...
+				ctx.Logger.Info(fmt.Sprintf("error fetching IPAddressClaim %s: %s", ipAddrClaimName, err.Error()))
+				return false, err
+			}
+
+			ipAddrName := ipAddrClaim.Status.AddressRef.Name
+			ctx.Logger.Info(fmt.Sprintf("fetched IPAddressClaim %s with AddressRef.Name %s", ipAddrClaimName, ipAddrName))
+			if ipAddrName == "" {
+				ctx.Logger.Info(fmt.Sprintf("name was empty on claim IPAddressClaim.Status.AddressRef.Name %s: %s", ipAddrClaimName, ipAddrName))
+				msg := "Waiting for IPAddressClaim to have an IPAddress bound"
+				markIPAddressClaimedConditionWaitingForClaimAddress(ctx.VSphereVM, msg)
```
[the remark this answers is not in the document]
tylerschultz: While possible, I don't think overwriting the condition is a likely scenario.

The VM is what is being reconciled, so the controller would see all of the desired claims in one pass. If any claims are missing, they'd be created and the status would be set as above.

The only way the status would be overwritten would be if the VM had failed reconciliation in a previous pass, and then  `addressesFromPools` section were to be amended. IIRC, the `VSphereVM`s are not mutated, and instead are recreated.

Per our offline conversation, we said we'd improve the condition status logic to include more information about desired count/actual count in a later PR.

controllers/vspherevm_controller.go
tylerschultz: @srm09 Do you have context on why during delete that errors are ignored and not returned?

pkg/services/govmomi/service.go
```diff
@@ -125,6 +130,14 @@ func (vms *VMService) ReconcileVM(ctx *context.VMContext) (vm infrav1.VirtualMac
 		return vm, err
 	}
 
+	if ok, err := vms.reconcileIPAddressClaims(vmCtx); err != nil || !ok {
```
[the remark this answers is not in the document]
tylerschultz: Correct. Ensure the claims exist. Don't wait for the IPAMProvider to create IPAddresses. The next function waits.

pkg/services/govmomi/service.go
```diff
@@ -250,13 +263,176 @@ func (vms *VMService) reconcileNetworkStatus(ctx *virtualMachineContext) error {
...
+func (vms *VMService) reconcileIPAddressClaims(ctx *virtualMachineContext) (bool, error) {
+	for devIdx, device := range ctx.VSphereVM.Spec.Network.Devices {
+		for poolRefIdx, poolRef := range device.AddressesFromPools {
+			// check if claim exists
+			ipAddrClaim := &ipamv1.IPAddressClaim{}
+			ipAddrClaimName := fmt.Sprintf("%s-%d-%d", ctx.VSphereVM.Name, devIdx, poolRefIdx)
+			ipAddrClaimKey := apitypes.NamespacedName{
+				Namespace: ctx.VSphereVM.Namespace,
+				Name:      ipAddrClaimName,
+			}
```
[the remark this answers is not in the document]
tylerschultz: Sure, I'll add some logging. 

I worry that might be noisy? Each reconcile loop would repeat the found status. The next block of code logs the creation of each claim.

pkg/services/govmomi/service.go
```diff
@@ -250,13 +263,176 @@ func (vms *VMService) reconcileNetworkStatus(ctx *virtualMachineContext) error {
...
+				Name:      ipAddrName,
+			}
+			if err = ctx.Client.Get(ctx, ipAddrKey, ipAddr); err != nil {
+				return false, err
+			}
+
+			toAdd := fmt.Sprintf("%s/%d", ipAddr.Spec.Address, ipAddr.Spec.Prefix)
+			parsedIP, _, err := net.ParseCIDR(toAdd)
+			if err != nil {
+				msg := fmt.Sprintf("IPAddress %s/%s has invalid ip address: %q",
```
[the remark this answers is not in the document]
tylerschultz: We had been logging errors, but removed them because we were seeing duplicate logs. 
We're seeing that errors returned during reconciliation are logged. Are there cases where the errors would not be logged?

pkg/util/machines.go
tylerschultz: We get apidiff failures when we reduce the visibility of functions in this package. Is that ok? I don't think the functions in this package are used widely. I'm not a big fan of the factoring. The functions are public so that other packages can use them, but they're really internal functions.

apis/v1beta1/vspherevm_types.go
tylerschultz: updated.

pkg/util/machines_test.go
tylerschultz: updated

pkg/manager/manager.go
```diff
@@ -31,6 +31,7 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
 	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
+	ipamv1a1 "sigs.k8s.io/cluster-api/exp/ipam/api/v1alpha1"
```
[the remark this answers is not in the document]
tylerschultz: updated

controllers/vspherevm_controller.go
randomvariable: It depends on whether or not the resource should be immediately requeued for retry. If so, return the error. If you only want to retry during a normal reconciliation interval, eat the error and return nil.

tylerschultz: In our case, we need to make sure the claim's finalizer is removed, or the IP will not be returned to the pool, so we think it's appropriate to return an error.

It is odd to us that when deleting the VM object that if something were to go wrong, it would eat the error. That is why we ask if there's some other reason we should eat the error too.

randomvariable: I don't think so. Probably historical mostly and trying to avoid early exponential backoff.

pkg/services/govmomi/service.go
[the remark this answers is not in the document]
srm09: Higher verbosity, ideally > 4 since CAPV by default uses 4 in its manifests.

pkg/util/machines.go
srm09: I meant update the first word in the comment to correctly reflect the name of the method
~~GetMachineMetadata~~ --> GetMachineMetadataWithIPAMState

tylerschultz: Ahhh, didn't see it. Fixed!

pkg/services/govmomi/service.go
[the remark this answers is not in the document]
tylerschultz: updated

pkg/services/govmomi/service.go
```diff
@@ -250,13 +263,176 @@ func (vms *VMService) reconcileNetworkStatus(ctx *virtualMachineContext) error {
...
+	ctx.IPAMState = map[string]infrav1.NetworkDeviceSpec{}
+	for devIdx, device := range ctx.VSphereVM.Spec.Network.Devices {
+		var ipAddrs []string
+		var gateway4 string
+		var gateway6 string
+
+		for poolRefIdx := range device.AddressesFromPools {
+			// check if claim exists
+			ipAddrClaim := &ipamv1.IPAddressClaim{}
+			ipAddrClaimName := fmt.Sprintf("%s-%d-%d", ctx.VSphereVM.Name, devIdx, poolRefIdx)
```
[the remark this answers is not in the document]
tylerschultz: updated

pkg/services/govmomi/service.go
```diff
@@ -250,13 +263,176 @@ func (vms *VMService) reconcileNetworkStatus(ctx *virtualMachineContext) error {
...
+			ipAddrClaimKey := apitypes.NamespacedName{
+				Namespace: ctx.VSphereVM.Namespace,
+				Name:      ipAddrClaimName,
+			}
+			var err error
+			ctx.Logger.Info(fmt.Sprintf("fetching IPAddressClaim %s with key %s", ipAddrClaimName, ipAddrClaimKey.String()))
+			if err = ctx.Client.Get(ctx, ipAddrClaimKey, ipAddrClaim); err != nil && !apierrors.IsNotFound(err) {
+				ctx.Logger.Info(fmt.Sprintf("error fetching IPAddressClaim %s: %s", ipAddrClaimName, err.Error()))
```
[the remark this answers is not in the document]
tylerschultz: updated

pkg/services/govmomi/service.go
```diff
@@ -250,13 +263,176 @@ func (vms *VMService) reconcileNetworkStatus(ctx *virtualMachineContext) error {
...
+				return false, err
+			}
+
+			ipAddrName := ipAddrClaim.Status.AddressRef.Name
+			ctx.Logger.Info(fmt.Sprintf("fetched IPAddressClaim %s with AddressRef.Name %s", ipAddrClaimName, ipAddrName))
+			if ipAddrName == "" {
+				ctx.Logger.Info(fmt.Sprintf("name was empty on claim IPAddressClaim.Status.AddressRef.Name %s: %s", ipAddrClaimName, ipAddrName))
```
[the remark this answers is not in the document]
tylerschultz: updated

pkg/services/govmomi/service.go
```diff
@@ -250,13 +263,176 @@ func (vms *VMService) reconcileNetworkStatus(ctx *virtualMachineContext) error {
...
+		for poolRefIdx := range device.AddressesFromPools {
+			// check if claim exists
+			ipAddrClaim := &ipamv1.IPAddressClaim{}
+			ipAddrClaimName := fmt.Sprintf("%s-%d-%d", ctx.VSphereVM.Name, devIdx, poolRefIdx)
+			ipAddrClaimKey := apitypes.NamespacedName{
+				Namespace: ctx.VSphereVM.Namespace,
+				Name:      ipAddrClaimName,
+			}
+			var err error
+			ctx.Logger.Info(fmt.Sprintf("fetching IPAddressClaim %s with key %s", ipAddrClaimName, ipAddrClaimKey.String()))
```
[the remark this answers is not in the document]
tylerschultz: updated

tylerschultz:
Apologies, the conversations were folded and I didn't see several items! All should be resolved now. Thanks!

controllers/vspherevm_controller.go
```diff
@@ -306,6 +308,31 @@ func (r vmReconciler) reconcileDelete(ctx *context.VMContext) (reconcile.Result,
 		r.Logger.V(6).Info("unable to delete node", "err", err)
 	}
 
+	// Remove finalizers from any ipam claims
```
srm09, line 311: If this controller is responsible for querying and removing the finalizer on the `IPAdressClaim`, then the kubebuilder brace directives should be a part of this file as well. Any particular reason why these get set on the `vspheremachine_controller.go` only.
My vote would be to duplicate for code readability if it is needed in both places. That way each controller clearly states all the rbac permissions it requires.

pkg/services/govmomi/service.go
```diff
@@ -250,13 +263,179 @@ func (vms *VMService) reconcileNetworkStatus(ctx *virtualMachineContext) error {
...
+				}
+				if err = ctx.Client.Create(ctx, claim); err != nil {
+					return false, err
+				}
+				msg := "Waiting for IPAddressClaim to have an IPAddress bound"
+				markIPAddressClaimedConditionWaitingForClaimAddress(ctx.VSphereVM, msg)
+			}
+		}
+	}
+
```
srm09, line 309: nit: remove empty line

pkg/services/govmomi/service.go
```diff
@@ -250,13 +263,179 @@ func (vms *VMService) reconcileNetworkStatus(ctx *virtualMachineContext) error {
...
+	return true, nil
+}
+
+func (vms *VMService) reconcileIPAddresses(ctx *virtualMachineContext) (bool, error) {
```
srm09, line 313: nit: add a comment stating the intent of the method

srm09 (commented):
lgtm apart from a few nits.
/approve
/hold

srm09:
@yastij @schrej Do take a pass at reviewing the PR and feel free to remove the hold once you feel it is good to go.

controllers/vspherevm_controller.go
tylerschultz: Good catch, this was a mistake. When we started this work we assumed we'd be modifying the `vspheremachine_controller.go`, but it turned out we needed to modify the `vspherevm_controller.go` instead. We've moved the kubebuilder brace directives to `vspherevm_controller.go`.

pkg/services/govmomi/service.go
tylerschultz: updated

pkg/services/govmomi/service.go
tylerschultz: updated

pkg/services/govmomi/service.go
```diff
@@ -580,3 +765,28 @@ func (vms *VMService) reconcileClusterModuleMembership(ctx *virtualMachineContex
...
+func markIPAddressClaimedConditionWaitingForClaimAddress(vm *infrav1.VSphereVM, msg string) {
+	conditions.MarkFalse(vm,
+		infrav1.IPAddressClaimedCondition,
+		infrav1.WaitingForIPAddressReason,
+		clusterv1.ConditionSeverityInfo,
+		msg)
+}
+
+func isIPv4(ip net.IP) bool {
+	return ip.To4() != nil
```
schrej, line 787: FYI: There is a new `netip` package since go1.18 with a nicer API: https://pkg.go.dev/net/netip#Addr.Is4

schrej (approved):
Very clean, well done!
Just two small things, otherwise
/lgtm

I didn't look at tests in detail though.

pkg/services/govmomi/service.go
tylerschultz: cool, we've switched over to `netip`

controllers/vspherevm_controller.go
```diff
@@ -306,6 +310,31 @@ func (r vmReconciler) reconcileDelete(ctx *context.VMContext) (reconcile.Result,
...
+	for devIdx, device := range ctx.VSphereVM.Spec.Network.Devices {
+		for poolRefIdx := range device.AddressesFromPools {
+			// check if claim exists
+			ipAddrClaim := &ipamv1.IPAddressClaim{}
+			ipAddrClaimName := fmt.Sprintf("%s-%d-%d", ctx.VSphereVM.Name, devIdx, poolRefIdx)
```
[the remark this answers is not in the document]
tylerschultz: We missed this one! Thanks

pkg/services/govmomi/service.go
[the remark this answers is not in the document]
tylerschultz: We considered putting it sooner, but it seems worthwhile to ensure the devices are in order before creating claims. The checking of network device status seems to run fairly quick - at least in our envs. 

We don't understand what the implications of checking hardware version are, we again thought we should wait for this 'pre-check'.

We like where it is, but we open to moving it to be sooner. Do you have a suggestion? We will be back again soon in this area of the code when we make updates to how IPAM condition status is reported. Perhaps we can revisit then?

pkg/util/machines.go
```diff
@@ -114,12 +114,24 @@ func IsControlPlaneMachine(machine metav1.Object) bool {
 // GetMachineMetadata returns the cloud-init metadata as a base-64 encoded
 // string for a given VSphereMachine.
 func GetMachineMetadata(hostname string, vsphereVM infrav1.VSphereVM, networkStatuses ...infrav1.NetworkStatus) ([]byte, error) {
+	return GetMachineMetadataWithIPAMState(hostname, vsphereVM, map[string]infrav1.NetworkDeviceSpec{}, networkStatuses...)
```
yastij, line 117: I don't think we need rename this. ultimately this function is supposed to return the metadata of a VM, we shouldn't leak the content into the name of the function

pkg/services/govmomi/service.go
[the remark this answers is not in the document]
yastij: +1 I don't think it's worth moving up for now, the sequencing we likely want:
- clone the VM with the devices
- verify devices and report the first wave of info in the status
- reconcile IPClaims
- verify the claims are fulfilled
- reconfigure the VM by setting the metadata

@tylerschultz - the only question I have is mostly around speed, did you notice longer cluster creation time or is it roughly stable?

yastij (commented):
/approve

yastij:
lgtm pending answers to the question above

pkg/services/govmomi/service.go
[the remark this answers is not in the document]
schrej: From a logical perspective that sequence makes sense of course. My argument would be to start with the IP Claim creation. Then clone the VM and verify devices. Then verify that claims are fulfilled.

This just gives the ipam providers a bit of time to potentially fulfil the claims in the meantime. If you create them and then check them right away it's quite likely that they aren't fulfilled yet, and reconciliation is aborted and needs to start over.

But I'm fine with both strategies, was just a suggestion.

pkg/util/machines.go
```diff
@@ -114,12 +114,24 @@ func IsControlPlaneMachine(machine metav1.Object) bool {
...
+// IPAM state includes IP and Gateways that should be added to each device.
+func GetMachineMetadataWithIPAMState(hostname string, vsphereVM infrav1.VSphereVM, ipamState map[string]infrav1.NetworkDeviceSpec, networkStatuses ...infrav1.NetworkStatus) ([]byte, error) {
 	// Create a copy of the devices and add their MAC addresses from a network status.
 	devices := make([]infrav1.NetworkDeviceSpec, integer.IntMax(len(vsphereVM.Spec.Network.Devices), len(networkStatuses)))
 
 	var waitForIPv4, waitForIPv6 bool
 	for i := range vsphereVM.Spec.Network.Devices {
 		vsphereVM.Spec.Network.Devices[i].DeepCopyInto(&devices[i])
+		if state, ok := ipamState[devices[i].MACAddr]; ok {
+			devices[i].IPAddrs = append(devices[i].IPAddrs, state.IPAddrs...)
```
yastij, line 131: this might not work with clusterClass, this is a case of co-authored slices. we need to be able a unique key so that server-side apply works. cc @fabriziopandini

yastij (commented):
/hold

pkg/util/machines.go
fabriziopandini: Confirmed.
if order to allow one contreller (api) to create the device entry, and another (capv) to set the ipAddr into device entries, the list of devices must be defined as +ListMap,

yastij: ok, just confirmed it seems like we're never retrofitting `.spec.devices.ipAddrs` so it shouldn't be an issue. 

@tylerschultz  can you confirm we're not patching `.spec.devices.ipAddrs` of the vsphereVM with the result of the fulfilled claim?

pkg/util/machines.go
tylerschultz: We made a second function to preserve the original function. `make apidiff` fails when we add a parameter to the original function. We can add the parameter to the original function name, but that means changing the 'public' api. 
Is it acceptable to submit this PR with failing apidiff?

pkg/util/machines.go
tylerschultz: We had originally updated `spec.devices.ipAddrs` on the VSphereVM, but we realized we were modifying the spec of the object so we backed it out. We changed it to use the in-memory map instead. We do not modify the `.spec.devices.ipAddrs`.

pkg/services/govmomi/service.go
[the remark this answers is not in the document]
tylerschultz: We've not noticed any delay. 

Our take is that in the grand scheme of standing up a cluster, the extra reconcile loop or 2 is trivial compared to the time it takes to boot a vm. OTOH, every loop counts, and perhaps we can optimize this.

We propose leaving where it is now and seeing how it goes.

pkg/util/machines.go
yastij: yeah we're using it mostly as a signal to know what we're breaking, we're not providing contract stability for functions

pkg/services/govmomi/service.go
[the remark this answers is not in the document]
yastij: @schrej - thinking more about, I think creating the IPClaims as part of the controller process before cloning makes sense. Ideally the VM Service layer should be mostly reading resources and translating them into vSphere constructs.

@tylerschultz do you mind filing an issue to address this later?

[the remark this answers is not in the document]
tylerschultz: will do

srm09:
/unhold
I will let @yastij add the final lgtm which should let this one merge.

tylerschultz:
Thanks all!

PR #1670, opened 2022-10-28
🌱 Skip creation of cluster modules in certain cases

### LINKED ISSUE #1647: Anti affinity feature assumes the infra version of cluster nodes
/kind bug

With the introduction of anti affinity, CAPV controller assumes that the infra template in use by the workload cluster provisioned via CAPV will only be `VSphereMachineTemplate`. This breaks an assumption that CAPI can provision a cluster with nodes belonging to multiple infra providers.

The cluster provisioning should progress and succeed even with the anti-affinity feature turned on.

When reconciling the cluster modules, the controller should only take into account objects referring `VSphereMachineTemplate` as their infra templates to use.

**Environment:**

- Cluster-api-provider-vsphere version: `v1.4.1` with anti affinity feature turned on.

srm09:
This patch adds a check for the following when creating cluster modules:
1. If the infra machine template kind used by MD/KCP object points to a different template kind (not VSphereMachineTemplate)
2. If the machine clone spec in the VSphereMachineTemplate points to a different vCenter URL than the VSphereCluster object. Basically, the case where some worker nodes for a cluster are deployed on a different vCenter instance.

Fixes #1647 and  1669

**Release note**:

```release-note
🌱 Skip creation of cluster modules in certain cases
```

### DIFF
--- a/controllers/clustermodule_reconciler.go
+++ b/controllers/clustermodule_reconciler.go
@@ -119,6 +119,10 @@ func (r Reconciler) Reconcile(ctx *context.ClusterContext) (reconcile.Result, er
 			modErrs = append(modErrs, clusterModError{obj.GetName(), err})
 			continue
 		}
+		// module creation was skipped
+		if moduleUUID == "" {
+			continue
+		}
 		clusterModuleSpecs = append(clusterModuleSpecs, infrav1.ClusterModule{
 			ControlPlane:     obj.IsControlPlane(),
 			TargetObjectName: obj.GetName(),
--- a/pkg/clustermodule/service.go
+++ b/pkg/clustermodule/service.go
@@ -19,6 +19,7 @@ package clustermodule
 import (
 	goctx "context"
 
+	"github.com/pkg/errors"
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/vim25/types"
 
@@ -27,6 +28,8 @@ import (
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/session"
 )
 
+const validMachineTemplate = "VSphereMachineTemplate"
+
 type service struct{}
 
 func NewService() Service {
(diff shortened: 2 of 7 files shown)


wjun (commented):
Could you explain if the second vCenter is used, will we create clustermodule for vms on the second vCenter? If not, does it mean there is no anti-affinity for the vms on the second vCenter?

wjun (approved):
LGTM.

controllers/clustermodule_reconciler.go
```diff
@@ -119,6 +119,10 @@ func (r Reconciler) Reconcile(ctx *context.ClusterContext) (reconcile.Result, er
 			modErrs = append(modErrs, clusterModError{obj.GetName(), err})
 			continue
 		}
+		// module creation was skipped
+		if moduleUUID == "" {
```
wjun, line 123: When will a moduleUUID be ""?

pkg/clustermodule/service.go
```diff
@@ -34,12 +35,28 @@ func NewService() Service {
...
-	logger := ctx.Logger.WithValues("object", wrapper.GetName())
-	template, err := fetchMachineTemplate(ctx, wrapper)
+	logger := ctx.Logger.WithValues("object", wrapper.GetName(), "namespace", wrapper.GetNamespace())
+
+	templateRef, err := fetchTemplateRef(ctx, ctx.Client, wrapper)
+	if err != nil {
+		logger.V(4).Error(err, "error fetching template for object")
+		return "", errors.Wrapf(err, "error fetching machine template for object %s/%s", wrapper.GetNamespace(), wrapper.GetName())
+	}
+	if templateRef.Kind != "VSphereMachineTemplate" {
```
wjun, line 45: Better to use a constant value instead of hard coded string "VSphereMachineTemplate"

sbueringer (commented):
Just a nit + the question that Jun brought up

yastij (approved):
/approve

lgtm pending the comments above.

sbueringer:
> Could you explain if the second vCenter is used, will we create clustermodule for vms on the second vCenter? If not, does it mean there is no anti-affinity for the vms on the second vCenter?

@srm09

srm09:
Missed this one, for the second vCenter we are not creating the cluster modules, hence the check for the vCenter URL. There is no anti affinity guarantee for VMs on the second vCenter. The stretched cluster setup is something that we do not fully support yet, so this is a known limitation.

PR #1673, opened 2022-11-01
Adds provision to set VMX Version

geetikabatra:
Currently, VMX version has to be set in the OVA.
This functionality provides a way to set VMX Version from users end, so that the created VM's default version is the set one and there is no need to set it as a metadata for the custom OVA.

Signed-off-by: geetikab@vmware.com <geetikab@vmware.com>

Fixes #

**Release note**:

```release-note

```


geetikabatra:
lgtm

apis/v1beta1/types.go
```diff
@@ -165,11 +165,16 @@ type VirtualMachineCloneSpec struct {
...
 	// OS is the Operating System of the virtual machine
 	// Defaults to Linux
 	// +optional
 	OS OS `json:"os,omitempty"`
+	// VMXVersion is the hardware version of the virtual machine.
+	// Defaults to the eponymous property value in the template from which the
+	// virtual machine is cloned.
+	// Check the compatibility with the ESXi version before setting the value.
+	// +optional
+	VMXVersion string `json:"vmxVersion,omitempty"`
```
yastij, line 177: we should probably rename to `hardwareVersion`

apis/v1beta1/vspheremachinetemplate_webhook.go
```diff
@@ -68,6 +69,12 @@ func (v *VSphereMachineTemplateWebhook) ValidateCreate(_ context.Context, raw ru
 			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "template", "spec", "network", "devices", "ipAddrs"), "cannot be set in templates"))
 		}
 	}
+	if spec.VMXVersion != "" {
+		r := regexp.MustCompile("^vmx-[1-9][0-9]?$")
```
yastij, line 73: we should drop the `?` as it'll allow `vmx-` as valid, which isn't

yastij (commented):
/approve

I'd like also to see an e2e test for this.

@geetikabatra - can you file an issue and assign yourself?

apis/v1beta1/vspheremachinetemplate_webhook.go
srm09: Added unit tests to check this particular scenario, it is working fine. The `[1-9]` part makes sure that any digit (from 1 to 9) is needed and the question mark is for the second set `[0-9]` which is not necessary in case like `vmx-9` or `vmx-4`.

apis/v1beta1/types.go
srm09: makes sense, done.

geetikabatra:
@yastij Sure!

PR #1675, opened 2022-11-01
🐛 Adds ESXi host info label validation and sanitization logic

### LINKED ISSUE #1674: Cluster creation fails when using invalid label values for ESXi host info
/kind bug

1. In the vCenter instance, configure the ESXi hosts to have an IPv6 address as the primary address.
2. Create a cluster with 1 control plane and 1 worker nodes
3. Cluster creation fails with the following errors in the logs:
```
E1031 17:21:43.140137       1 vspheremachine_controller.go:331] capv-controller-manager/vspheremachine-controller "msg"="failed to patch machine with host info label" "error"="Machine.cluster.x-k8s.io \"mgmt-vc-control-plane-27f4d\" is invalid: metadata.labels: Invalid value: \"2620:124:6020:c003:0:69ff:fe2b:3763\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')"  "machine "="mgmt-vc-control-plane-27f4d"
E1031 17:21:43.140980       1 controller.go:317] controller/vspheremachine "msg"="Reconciler error" "error"="Machine.cluster.x-k8s.io \"mgmt-vc-control-plane-27f4d\" is invalid: metadata.labels: Invalid value: \"2620:124:6020:c003:0:69ff:fe2b:3763\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')" "name"="mgmt-vc-control-plane-q5kzq" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="VSphereMachine"
``` 

Cluster creation proceeds without any errors.

**Environment:**

- Cluster-api-provider-vsphere version: v1.4.1

/assign

srm09:
This patch sanitizes the labels to be set on the Machine object to be conformant to the label value guidelines set in the k8s official docs. It also adds a validation to check if the sanitized label is conformant using the library functions from the apimachinery.

Fixes #1674

With the anti affinity feature, we also included the logic to propagate ESXi host info to the CAPI Machine object which then gets transferred over to the k8s nodes. Since, the ESXi host information is added as a label, it should follow the label value  conventions set by the k8s docs.

**Release note**:
```release-note
Adds ESXi host info label validation and sanitization logic
```

### DIFF
--- a/controllers/vspheremachine_controller.go
+++ b/controllers/vspheremachine_controller.go
@@ -29,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	apitypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	clusterutilv1 "sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/annotations"
@@ -58,6 +59,8 @@ import (
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/vmoperator"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/util"
 )
+
+const hostInfoErrStr = "host info cannot be used as a label value"
 
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=vspheremachines,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=vspheremachines/status,verbs=get;update;patch
@@ -345,6 +348,14 @@ func (r *machineReconciler) patchMachineLabelsWithHostInfo(ctx context.MachineCo
 		return err
 	}
 
+	info := util.SanitizeHostInfoLabel(hostInfo)
+	errs := validation.IsValidLabelValue(info)
+	if len(errs) > 0 {
+		err := errors.Errorf("%s: %s", hostInfoErrStr, strings.Join(errs, ","))
+		r.Logger.Error(err, hostInfoErrStr, "info", hostInfo)
+		return err
+	}
+
 	machine := ctx.GetMachine()
 	patchHelper, err := patch.NewHelper(machine, r.Client)
 	if err != nil {
@@ -352,7 +363,7 @@ func (r *machineReconciler) patchMachineLabelsWithHostInfo(ctx context.MachineCo
 	}
 
 	labels := machine.GetLabels()
-	labels[constants.ESXiHostInfoLabel] = hostInfo
+	labels[constants.ESXiHostInfoLabel] = info
 	machine.Labels = labels
 
 	return patchHelper.Patch(r, machine)
--- a/pkg/util/label.go
+++ b/pkg/util/label.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"fmt"
+	"net"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+// SanitizeHostInfoLabel ensures that the ESXi host information passed as a parameter confirms to
+// the label value constraints documented at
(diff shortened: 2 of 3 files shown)


pkg/util/label.go
```diff
@@ -0,0 +1,69 @@
...
+// https://k8s.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+//
+// The expected inputs for the object are IP addresses or FQDNs of the ESXi hosts.
+func SanitizeHostInfoLabel(info string) string {
+	updatedInfo := stripZoneInfo(info)
+	ip := net.ParseIP(updatedInfo)
+	if ip != nil {
+		if ipv4 := ip.To4(); ipv4 == nil {
+			// case when it is an IPv6 address
+			return fmt.Sprintf("%s.ipv6-literal", strings.ReplaceAll(updatedInfo, ":", "-"))
```
zjs, line 38: **nitpick**: Maybe include a comment explaining that this would be at most `8 * 4 (address) + 7 (dashess) + 12 (suffix) = 51` characters? (Mostly to make the thought process explicit in case there's some case in which it's not true.)

zjs (approved):
Looks good to me. Thank you for the quick fix!

controllers/vspheremachine_controller.go
```diff
@@ -345,14 +348,22 @@ func (r *machineReconciler) patchMachineLabelsWithHostInfo(ctx context.MachineCo
...
+
 	machine := ctx.GetMachine()
 	patchHelper, err := patch.NewHelper(machine, r.Client)
 	if err != nil {
 		return err
 	}
 
 	labels := machine.GetLabels()
-	labels[constants.ESXiHostInfoLabel] = hostInfo
+	labels[constants.ESXiHostInfoLabel] = info
```
adam-jian-zhang, line 366: Will truncate the host info label affect anti-affinity logic? For example, if after sanitize, the hostInfoLabel are the same for two esxi hosts, will it affect our anti-affinity logic?
Another question, if DRS kick in, the VMs is vMotioned to another host, will this label get updated accordingly?

pkg/util/label.go
```diff
@@ -0,0 +1,74 @@
...
+// https://k8s.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+//
+// The expected inputs for the object are IP addresses or FQDNs of the ESXi hosts.
+func SanitizeHostInfoLabel(info string) string {
+	updatedInfo := stripZoneInfo(info)
+	ip := net.ParseIP(updatedInfo)
+	if ip != nil {
+		if ipv4 := ip.To4(); ipv4 == nil {
+			// In case of an IPv6 address, replace `:` with the acceptable `-` character.
+			// The size for the string would never exceed 51 (8 * 4 (address) + 7 (dashes) + 12 (suffix) = 51) characters.
```
sbueringer, line 38:
```suggestion
			// The size for the string would never exceed 52 (8 * 4 (address) + 7 (dashes) + 13 (suffix) = 52) characters.
```
nit: I think you missed the dot of the suffix

pkg/util/label.go
```diff
@@ -0,0 +1,74 @@
...
+// TODO (srm09): consider removing it in the future.
+func stripZoneInfo(info string) string {
+	idx := strings.LastIndex(info, "%")
+	if idx == -1 {
+		return info
+	}
+	return info[:idx]
+}
+
+func sanitizeLabelLength(inputURL string) string {
```
sbueringer, line 58: nit: maybe add a comment to describe the sanitize logic
(I wonder if a func name like truncateLabel would fit better)

pkg/util/label_test.go
```diff
@@ -0,0 +1,101 @@
...
+			input:    "foo-1.bar.com",
+			expected: "foo-1.bar.com",
+		},
+		{
+			name:     "for DNS entry with 63 characters",
+			input:    "esx13-r09.p01.1d91f0ee4f14b7e83bd42.australiaeast.avs.belch.com",
+			expected: "esx13-r09.p01.1d91f0ee4f14b7e83bd42.australiaeast.avs.belch.com",
+		},
+		{
+			name:     "for DNS entry with > 63 characters",
```
sbueringer, line 79: Maybe add other test cases where:
* multiple DNS segments have to be dropped
* dropping a DNS segment results in exactly 63 characters

sbueringer (commented):
Just some nits.

I don't have enough context to assess the implications, only did a basic code review

controllers/vspheremachine_controller.go
srm09: Nope, the anti affinity logic is independent of the label. The logic is to get the host info from the VM and add it as a label to the Machine object. Eventually, on the next reconcile, the label on the Machine object will be updated to display the latest info.

pkg/util/label.go
srm09: Updated the function name, makes more sense that way.

pkg/util/label.go
srm09: Updated.

pkg/util/label_test.go
srm09: Done

PR #1681, opened 2022-11-04
:seedling: Run `make generate`

johananl:
**What this PR does / why we need it**: Some manifests are out of date following a kube-vip update.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: None

**Special notes for your reviewer**: Need to check if we should update the CI to ensure manifests are always up to date.

**Release note**:

```release-note
None
```

### DIFF
--- a/templates/cluster-template-topology.yaml
+++ b/templates/cluster-template-topology.yaml
@@ -48,7 +48,7 @@ spec:
               value: "10"
             - name: vip_retryperiod
               value: "2"
-            image: ghcr.io/kube-vip/kube-vip:v0.5.0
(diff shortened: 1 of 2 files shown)


johananl:
>It would be wise to keep a job that does a diff on the committed manifest v/s the generated the manifest and fails if there is any diff. This should be run as a pre-submit job to make sure this never happens again.

True. Here is the job 🙂 https://github.com/kubernetes/test-infra/pull/27898

PR #1691, opened 2022-11-10
🐛 Check for nil Cluster InfrastructureRef in ClusterModule reconcile

killianmuldoon:
Signed-off-by: killianmuldoon <kmuldoon@vmware.com>

Fix a nil-pointer crash when InfrastructureRef isn't set on the Cluster.

This is part of the ClusterModule reconciler.

### DIFF
--- a/controllers/clustermodule_reconciler.go
+++ b/controllers/clustermodule_reconciler.go
@@ -162,6 +162,11 @@ func (r Reconciler) toAffinityInput(obj client.Object) []reconcile.Request {
 		return nil
 	}
 
+	if cluster.Spec.InfrastructureRef == nil {
+		r.Logger.Error(err, "cluster infrastructureRef not set. Requeing",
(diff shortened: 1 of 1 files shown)


controllers/clustermodule_reconciler.go
```diff
@@ -162,6 +162,11 @@ func (r Reconciler) toAffinityInput(obj client.Object) []reconcile.Request {
 		return nil
 	}
 
+	if cluster.Spec.InfrastructureRef == nil {
```
killianmuldoon, line 165: A bunch of these tests are failing for me locally with or without these changes - not sure what the root cause is.

srm09: Pulled them locally, they have passed. Yassine triggered a retry, should work out.

yastij (approved):
/approve
/lgtm

srm09:
/retitle 🐛 Check for nil Cluster InfrastructureRef in ClusterModule reconcile

PR #1692, opened 2022-11-11
🐛 Removes hard dependency on cluster module info for VM deletion

srm09:
This patch removes the hard dependency of finding the cluster module information when deleting a VM. If the information is missing, due to broken owner reference links, then we should let the controller proceed with VM deletion.

Fixes #

**Release note**:

```release-note
Removes hard dependency on cluster module info for VM deletion
```

### DIFF
--- a/controllers/vspherevm_controller.go
+++ b/controllers/vspherevm_controller.go
@@ -211,11 +211,6 @@ func (r vmReconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Res
 		return reconcile.Result{}, nil
 	}
 
-	clusterModule, err := r.fetchClusterModuleInfo(vsphereCluster, machine)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-
 	var vsphereFailureDomain *infrav1.VSphereFailureDomain
 	if failureDomain := machine.Spec.FailureDomain; failureDomain != nil {
 		vsphereDeploymentZone := &infrav1.VSphereDeploymentZone{}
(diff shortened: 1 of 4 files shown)


controllers/vspherevm_controller.go
```diff
@@ -275,6 +269,15 @@ func (r vmReconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Res
 		}
 	}
 
+	clusterModuleInfo, err := r.fetchClusterModuleInfo(vsphereCluster, machine)
```
yastij, line 272: we shouldn't swallow the notfound error in this function, as it makes most the logic bellow unclear. Please file an issue for this.

controllers/vspherevm_controller.go
```diff
@@ -275,6 +269,15 @@ func (r vmReconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Res
...
+	// If cluster module information cannot be fetched for a VM being deleted,
+	// we should not block VM deletion since the cluster module is updated
+	// once the VM gets removed.
+	if err != nil && vsphereVM.ObjectMeta.DeletionTimestamp.IsZero() {
+		return reconcile.Result{}, err
+	}
+	vmContext.ClusterModuleInfo = clusterModuleInfo
```
yastij, line 279: If this is nil on deletion because of the `MachineSet` being gone, I assume we're not removing the VM from the clusterModule? If so does it have any impact when trying to delete the the whole cluster and hence the cluster modules?

controllers/vspherevm_controller.go
```diff
@@ -275,6 +269,15 @@ func (r vmReconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Res
 		}
 	}
 
+	clusterModuleInfo, err := r.fetchClusterModuleInfo(vsphereCluster, machine)
+	// If cluster module information cannot be fetched for a VM being deleted,
+	// we should not block VM deletion since the cluster module is updated
+	// once the VM gets removed.
+	if err != nil && vsphereVM.ObjectMeta.DeletionTimestamp.IsZero() {
```
yastij, line 276: this logic is a bit complex, please add a controller test specifically for it

yastij (commented):
/approve
/lgtm

PR #1703, opened 2022-11-30
✨ Adds dependabot configuration

srm09:
Adds dependabot configuration

Fixes n/a

n/a

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+# GitHub Actions
+- package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,27 @@
...
+      prefix: ":seedling:"
+  labels:
+    - "ok-to-test"
+# Go
+- package-ecosystem: "gomod"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+  ignore:
+  # Ignore k8s modules as they are upgraded manually
```
sbueringer, line 20: We have some more excludes in core CAPI: https://github.com/kubernetes-sigs/cluster-api/blob/main/.github/dependabot.yml

Just in case you want to exclude them as well

sbueringer: Ah sounds good

PR #1708, opened 2022-11-30
🌱 Checks for feature flag during VM reconciliation

### LINKED ISSUE #1702: CAPV looking for KubeadmControlPlane despite using non-kubeadm control plane provider
/kind bug

Deployed CAPV into a KIND cluster with `clusterctl` using a custom, in-house control-plane and bootstrap provider: 

`clusterctl init --infrastructure vsphere --bootstrap custom --control-plane custom`

Then we apply our manifests telling the `Cluster` CRD to utilize our custom control plane (for the sake of example, called `CustomControlPlane`):

```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  labels:
    cluster.x-k8s.io/cluster-name: capi-custom
  name: capi-custom
  namespace: default
spec:
  ...
  controlPlaneRef:
    apiVersion: controlplane.cluster.x-k8s.io/v1beta1
    kind: CustomControlPlane
    name: capi-custom
  infrastructureRef:
    kind: VSphereCluster
    name: capi-custom
```

CAPV starts, but then continually fails to reconcile, in pod logs we see CAPV is still try to match `KubeadmControlPlane`, despite not being denoted in the `Cluster` resource:

```
"controller-runtime/source: if kind is a CRD, it should be installed before calling Start" err="no matches for kind \"KubeadmControlPlane\" in version \"controlplane.cluster.x-k8s.io/v1beta1\"" kind="KubeadmControlPlane.controlplane.cluster.x-k8s.io"
```

If I were to then install the Kubeadm CRDs explicitly, CAPV will still fail, but now with:

```
controller.go:317] controller/vspherevm "msg"="Reconciler error" "error"="unable to find owner reference with APIVersion controlplane.cluster.x-k8s.io/v1beta1 and Kind KubeadmControlPlane" "name"="capi-custom-vc94v" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="VSphereVM"
```

CAPV to utilize the apiVersion/kind denoted in the `Cluster` object. In the above example: `CustomControlPlane`.

Ran a quick test through several versions, and it appears to start with releases >= `v1.3.2`.  Previous releases reconcile as expected with the custom control plane / bootstrap providers.

It appears that: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1638 added: `clustermodule_reconciler.go` which is looking for `KubeadmControlPlane` explicitly, regardless of what is set in the top-level `Cluster` object.

**Environment:**

- Cluster-api-provider-vsphere version: v1.5.0
- Kubernetes version: (use `kubectl version`):  v1.25.3
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04

srm09:
Adds the check for the state of the anti affinity feature flag when reconciling VM. When reconciling the VM, the cluster module info should be looked up only when the NodeAntiAffinity feature flag is turned on. This also solves the issue when the control plane provider is something other than the KubeadmControlPlane, in which case we do not want to allow checks for cluster module info.

Fixes #1702 

n/a

**Release note**:
```release-note
Checks for feature flag during VM reconciliation
```

### DIFF
--- a/controllers/vspherevm_controller.go
+++ b/controllers/vspherevm_controller.go
@@ -48,6 +48,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-vsphere/feature"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/clustermodule"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/identity"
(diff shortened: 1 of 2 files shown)


yastij (approved):
/approve
/lgtm

PR #1721, opened 2022-12-02
✨ ipam display count of address claims bound in condition

### LINKED ISSUE #1689: IPAM status conditions enhancements
/kind feature

The `IPAddressClaimed` status condition for IPAM IPAddressClaim and IPAddress reconciliation shows only the first encountered invalid IPAddress, if one is deemed to be invalid. Other IPAddressClaim/IPAddress pairs may be successfully reconciled, but the messages do not show this information.

The condition messages should be enhanced to show the number of pending/total addresses.

This controller was cited as an example of how it may look: https://github.com/kubernetes-sigs/cluster-api/blob/bda7e5cca7b57767437f3627e8cf2d3d389e173f/internal/controllers/machineset/machineset_controller.go

This is a follow on to this PR: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1666

**Environment:**

adobley:
while there are unbound ip address claims the condition will show the count of bound to total claims, giving some feedback on how much work is left

Fixes #1689 

**Release note**:

```release-note
when using AddressesFromPools the vSphereMachine status will reflect how many claims are awaiting addresses
```

### DIFF
--- a/pkg/services/govmomi/service.go
+++ b/pkg/services/govmomi/service.go
@@ -20,6 +20,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net/netip"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/vmware/govmomi/object"
@@ -52,6 +53,18 @@ import (
 
 // VMService provdes API to interact with the VMs using govmomi.
 type VMService struct{}
+
+// ipamDeviceConfig aids and holds state for the process of parsing IPAM
+// addresses for a given device.
+type ipamDeviceConfig struct {
+	DeviceIndex         int
+	IPAMAddresses       []*ipamv1.IPAddress
+	MACAddress          string
+	NetworkSpecGateway4 string
+	IPAMConfigGateway4  string
+	NetworkSpecGateway6 string
+	IPAMConfigGateway6  string
+}
 
 // ReconcileVM makes sure that the VM is in the desired state by:
 //  1. Creating the VM if it does not exist, then...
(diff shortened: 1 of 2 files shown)


pkg/services/govmomi/service.go
```diff
@@ -441,6 +425,61 @@ func (vms *VMService) reconcileIPAddresses(ctx *virtualMachineContext) (bool, er
...
+			ipAddrClaimName := IPAddressClaimName(ctx.VSphereVM.Name, devIdx, poolRefIdx)
+			ipAddrClaimKey := apitypes.NamespacedName{
+				Namespace: ctx.VSphereVM.Namespace,
+				Name:      ipAddrClaimName,
+			}
+			var err error
+			ctx.Logger.V(5).Info("fetching IPAddressClaim", "name", ipAddrClaimKey.String())
+			if err = ctx.Client.Get(ctx, ipAddrClaimKey, ipAddrClaim); err != nil && !apierrors.IsNotFound(err) {
+				ctx.Logger.Error(err, "error fetching IPAddressClaim", "name", ipAddrClaimName)
+				return nil, err
```
srm09, line 451: even with the error we should check for the state of the other IPAddressClaim objects and aggregate any errors at the end of the for loop, construct an error and return.

srm09 (changes requested):
some linter fixes to be made as well.
/ok-to-test

pkg/services/govmomi/service.go
tylerschultz: While I agree with the spirit of this suggestion, I'm not sure I agree with this one specifically.

The prior step in this reconciliation loop does a find or create for the claims. 

In the current step of the loop, the controller is waiting for the claim to have an address bound. The case where the claim is not found is handled. The only reason this if condition would be hit is if there is an api error. Continuing on and attempting to look up other claims will likely fail too. Gathering up more api errors seemingly provides diminished value.

Unless you think it is important, we'll skip on this suggestion, and instead look for other places where we can apply the suggested strategy. 

WDYT?

tylerschultz:
FYI - We think we've covered the feedback, but we haven't run this code and tried it out. We'll let you know when we think we're ready for another round of review.

tylerschultz:
Update: we've deployed the changes and we are happy with what we're seeing. 
- The condition status shows "0 of 1..." in a nice way when the vm is waiting for IPs.
- The status messages format nicely when multiple IPs cause errors in the VSphereVM yaml.

One area of possible concern would be that these changes include newlines when multiple errors are logged. 

This PR should be good for another round of review.

adobley:
We were seeing build errors fetching a version of vm-operator. We bumped that in another PR (#1732) and have rebased on that branch for now.

adobley:
Due to #1733 we have rebased onto #1734 until that gets merged so that the tests can run.

adobley:
> Due to #1733 we have rebased onto #1734 until that gets merged so that the tests can run.

Rebased back onto master now that #1734 is fixed. This should be good to go once everything passes.

pkg/services/govmomi/service.go
adobley: @srm09

adobley:
Rebased as it had been a few. I'll keep my eye on this but if everything passes we should be good to go again.

PR #1505, opened 2022-04-04
build: Support arm64 images

jimmidyson:
**What this PR does / why we need it**:
This PR uses buildx to create multi-platform images, both for amd64
and arm64. This enables creating bootstrap clusters on Macs with M1
chips, which require pulling arm64 based images.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
Support running CAPV controller manager on arm64
```

PR #1731, opened 2023-01-04
:seedling: chore: publishes arm64 images with release

faiq:
Uses docker build targets for release scripts to reduce repeating logic introduced in #1505. This also makes the release script publish arm64 images for users on m1 macs.

Fixes #

**Release note**:

```release-note
publishes arm64 images with release
```

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -102,16 +102,19 @@ function build_images() {
   esac
 
   # Manager image
-  make docker-buildx
   ARCH=("arm64" "amd64")
   for arch in "${ARCH[@]}"; do
     echo "building ${MANAGER_IMAGE_NAME}:${VERSION} for arch ${arch}"
-    ARCH=${arch} FULL_DOCKER_IMG="${MANAGER_IMAGE_NAME}:${VERSION}" make docker-build
+	  docker buildx inspect capv &>/dev/null || docker buildx create --name capv
+    docker buildx build --builder capv --platform linux/"${arch}" --output=type=docker --pull \
+      -f Dockerfile \
+      -t "${MANAGER_IMAGE_NAME}":"${VERSION}-${arch}" \
+      .
+    if [ "${LATEST}" ]; then
+      echo "tagging image ${MANAGER_IMAGE_NAME}:${VERSION}-${arch} as latest"
+      docker tag "${MANAGER_IMAGE_NAME}":"${VERSION}-${arch}" "${MANAGER_IMAGE_NAME}:latest-${arch}"
+    fi
   done
-  if [ "${LATEST}" ]; then
-    echo "tagging image ${MANAGER_IMAGE_NAME}:${VERSION} as latest"
-    docker tag "${MANAGER_IMAGE_NAME}":"${VERSION}" "${MANAGER_IMAGE_NAME}":latest
-  fi
 }
 
 function logout() {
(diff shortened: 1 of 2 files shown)


srm09:
@faiq I was looking at your changes, and this multi version image release can be done by a simple change in the `hack/release.sh` file which looks something like this below:

```
cluster-api-provider-vsphere on  main [$!?] via 🏎💨 v1.19
at 15:30:48 ➜ git diff
diff --git a/hack/release.sh b/hack/release.sh
index 3e10f165..7a0a0fee 100755
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -102,16 +102,20 @@ function build_images() {
   esac

   # Manager image
-  ARCH=$(go env GOARCH)
-  echo "building ${MANAGER_IMAGE_NAME}:${VERSION} for arch ${ARCH}"
-  docker buildx build --platform linux/"${ARCH}" --output=type=docker --pull \
-    -f Dockerfile \
-    -t "${MANAGER_IMAGE_NAME}":"${VERSION}" \
-    .
-  if [ "${LATEST}" ]; then
-    echo "tagging image ${MANAGER_IMAGE_NAME}:${VERSION} as latest"
-    docker tag "${MANAGER_IMAGE_NAME}":"${VERSION}" "${MANAGER_IMAGE_NAME}":latest
-  fi
+  ARCH=("amd64" "arm64")
+  for arch in "${ARCH[@]}"
+  do
+    local IMG_NAME=${MANAGER_IMAGE_NAME}-${arch}
+    echo "building ${IMG_NAME}:${VERSION} for arch ${arch}"
+    docker buildx build --platform linux/"${arch}" --output=type=docker --pull \
+      -f Dockerfile \
+      -t "${IMG_NAME}":"${VERSION}" \
+      .
+    if [ "${LATEST}" ]; then
+      echo "tagging image ${IMG_NAME}:${VERSION} as latest"
+      docker tag "${IMG_NAME}":"${VERSION}" "${IMG_NAME}":latest
+    fi
+  done
 }

 function logout() {
```

wdyt?

faiq:
thanks @srm09 i agree the changes could be made there, but I felt it was better to reuse the makefile target to follow the principle of DRY code. However since it does touch the makefile as well, I can understand your concerns. 

I'm fine either way, I just need arm64 images :) so if doing it one way will get them for me faster im happy to do it

srm09:
I am of the view that currently the release process does not include the Makefile in any way. We should continue this to unblock this and then work on the refactoring in a follow up issue/PR.

faiq:
sounds good. i'll make the changes

faiq:
@srm09 updated the PR

srm09 (changes requested):
I think the name of the image should be {image-name}-{arch}:{version} instead of the {image-name}:{version}-{arch} as mentioned in the PR.

faiq:
@srm09 updated, sorry I misread the image format, but i think everything should be fine now!

hack/release.sh
```diff
@@ -102,16 +102,19 @@ function build_images() {
...
-    echo "tagging image ${MANAGER_IMAGE_NAME}:${VERSION} as latest"
-    docker tag "${MANAGER_IMAGE_NAME}":"${VERSION}" "${MANAGER_IMAGE_NAME}":latest
-  fi
+  ARCH=("arm64" "amd64")
+  for arch in "${ARCH[@]}"; do
+    echo "building ${MANAGER_IMAGE_NAME}:${VERSION} for arch ${arch}"
+	  docker buildx inspect capv &>/dev/null || docker buildx create --name capv
+    docker buildx build --builder capv --platform linux/"${arch}" --output=type=docker --pull \
+      -f Dockerfile \
+      -t "${MANAGER_IMAGE_NAME}":"${VERSION}-${arch}" \
```
srm09, line 111:
```suggestion
      -t "${MANAGER_IMAGE_NAME}-${arch}:"${VERSION}" \
```

hack/release.sh
```diff
@@ -102,16 +102,19 @@ function build_images() {
...
+      .
+    if [ "${LATEST}" ]; then
+      echo "tagging image ${MANAGER_IMAGE_NAME}:${VERSION}-${arch} as latest"
+      docker tag "${MANAGER_IMAGE_NAME}":"${VERSION}-${arch}" "${MANAGER_IMAGE_NAME}:latest-${arch}"
```
srm09, line 115:
```suggestion
      echo "tagging image ${MANAGER_IMAGE_NAME}-${arch}:${VERSION} as latest"
      docker tag "${MANAGER_IMAGE_NAME}-${arch}":"${VERSION}" "${MANAGER_IMAGE_NAME}-${arch}:latest"
```

hack/release.sh
```diff
@@ -137,11 +140,26 @@ function push_images() {
   login
 
   # Manager image
-  echo "pushing ${MANAGER_IMAGE_NAME}:${VERSION}"
-  docker push "${MANAGER_IMAGE_NAME}":"${VERSION}"
+  ARCH=("arm64" "amd64")
+  for arch in "${ARCH[@]}"; do
+    echo "pushing ${MANAGER_IMAGE_NAME}:${VERSION}-${arch}"
```
srm09, line 145: Same suggestion as above.

hack/release.sh
```diff
@@ -137,11 +140,26 @@ function push_images() {
...
+    --amend "${MANAGER_IMAGE_NAME}:${VERSION}-amd64"
+  docker manifest push "${MANAGER_IMAGE_NAME}:${VERSION}"
   if [ "${LATEST}" ]; then
-    echo "also pushing ${MANAGER_IMAGE_NAME}:${VERSION} as latest"
-    docker push "${MANAGER_IMAGE_NAME}":latest
+    docker manifest create \
+      "${MANAGER_IMAGE_NAME}:latest" \
+      --amend "${MANAGER_IMAGE_NAME}:latest-arm64" \
+      --amend "${MANAGER_IMAGE_NAME}:latest-amd64"
+    docker manifest push "${MANAGER_IMAGE_NAME}:latest"
```
srm09, line 162: I am unfamiliar with this, could you elaborate on what this command is doing??

srm09 (changes requested):

hack/release.sh
faiq: yes absolutely, the manifest is saying that the the image at a particular tag has more than one architectures and it points to where those are.

in this case, we are saying the latest image has two architectures and each of those correspond to the existing image tags

faiq:
@srm09 can you please take a look again. the image names are updated with the format

hack/release.sh
faiq: Addressed with common variable

hack/release.sh
faiq: Addressed with common variable

srm09:
/lgtm
/approve

Thanks for submitting this.

PR #1758, opened 2023-01-30
:bug: Renaming the Windows VM name on creation

knabben:
Renaming the vsphereVM object in the Webhook will cause wrong patching in the controller later, fixing the bug and renaming it directly in the controller on creation time.

Fixes #

**Release note**:

```release-note
Fixes Windows vspherevm rename when VSphereMachine OS is Windows.
```

### DIFF
--- a/apis/v1beta1/vspherevm_webhook_test.go
+++ b/apis/v1beta1/vspherevm_webhook_test.go
@@ -45,10 +45,6 @@ func TestVSphereVM_Default(t *testing.T) {
 	g.Expect(LinuxVM.Spec.OS).To(Equal(Linux))
 	g.Expect(WindowsVM.Spec.OS).To(Equal(Windows))
 	g.Expect(NoOSVM.Spec.OS).To(Equal(Linux))
-
-	// WindowsVM gets name updated to be 15 characters. Linux remains unchanged
-	g.Expect(WindowsVM.Name).To(Equal("cluster-m-2qj6q"))
-	g.Expect(LinuxVM.Name).To(Equal("linux-control-plane-qkkbv"))
 }
 
 //nolint
@@ -81,7 +77,7 @@ func TestVSphereVM_ValidateCreate(t *testing.T) {
 			wantErr:   true,
 		},
 		{
-			name:      "name too long for Linux VM",
+			name:      "no error with name too long for Linux VM",
 			vSphereVM: createVSphereVM(linuxVMName, "foo.com", "", "", []string{"192.168.0.1/32", "192.168.0.3/32"}, nil, Linux),
 			wantErr:   false,
 		},
--- a/pkg/services/vimmachine.go
+++ b/pkg/services/vimmachine.go
@@ -19,6 +19,7 @@ package services
 import (
 	goctx "context"
 	"encoding/json"
+	"strings"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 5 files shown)


pkg/services/vimmachine.go
```diff
@@ -181,8 +182,8 @@ func (v *VimMachineService) GetHostInfo(c context.MachineContext) (string, error
 
 	vsphereVM := &infrav1.VSphereVM{}
 	if err := ctx.Client.Get(ctx, client.ObjectKey{
-		Namespace: ctx.Machine.Namespace,
-		Name:      ctx.Machine.Name,
+		Namespace: ctx.VSphereMachine.Namespace,
```
srm09, line 185: nit: Does not matter what you use, the namespace for the entire hierarchy of CAPI/CAPV objects will be the same.

pkg/services/vimmachine.go
```diff
@@ -450,6 +451,14 @@ func (v *VimMachineService) createOrPatchVSPhereVM(ctx *context.VIMMachineContex
...
 }
 
+// renameWindowsVM returns the renamed machine name with 15 characters length.
+func renameWindowsVM(ctx *context.VIMMachineContext, machineName string) string {
+	if ctx.VSphereMachine.Spec.OS == infrav1.Windows && len(machineName) > 15 {
+		return strings.TrimSuffix(machineName[0:9], "-") + "-" + machineName[len(machineName)-5:]
+	}
+	return machineName
+}
+
```
srm09, line 461: A better way to name the function would be something like `generateVMObjectName` and add a comment which calls out the specific logic for Windows based VSphereVM objects.

apis/v1beta1/vspherevm_webhook_test.go
```diff
@@ -46,9 +46,8 @@ func TestVSphereVM_Default(t *testing.T) {
 	g.Expect(WindowsVM.Spec.OS).To(Equal(Windows))
 	g.Expect(NoOSVM.Spec.OS).To(Equal(Linux))
 
-	// WindowsVM gets name updated to be 15 characters. Linux remains unchanged
-	g.Expect(WindowsVM.Name).To(Equal("cluster-m-2qj6q"))
-	g.Expect(LinuxVM.Name).To(Equal("linux-control-plane-qkkbv"))
+	g.Expect(WindowsVM.Name).To(Equal(windowsVMName))
+	g.Expect(LinuxVM.Name).To(Equal(linuxVMName))
```
srm09, line 50: Remove this altogether, these assertions add no value now.

srm09 (commented):
Minor nits and also add tests for the changes made to the functions in `vimmachine.go` files. Unit test the name gen function, and a good use case to introduce tests for the `createOrPatchVSPhereVM` function as well.

pkg/services/vimmachine.go
knabben: ok, can be a good idea to `vspheremachine` and maintain the standard across calls. wdyt?

pkg/services/vimmachine_test.go
```diff
@@ -237,3 +237,59 @@ var _ = Describe("VimMachineService_GetHostInfo", func() {
 	})
 
 })
+
+var _ = Describe("VimMachineService_createOrPatchVSPhereVM", func() {
```
knabben, line 241: if you have more use cases in mind, happy to increase the coverage in a follow up.

srm09: Sure this is a good enough place to get started

PR #1762, opened 2023-02-01
🌱 Drops version check when tracking owners

srm09:
During the VM reconciliation, the controllers track down the parent of the Machine objects using the owner reference hierarchy. The owner reference checks were comparing the Group as well as the API version to determins and fetch the owner.

Currently, the API version used for comparison was selected from the storageVersion of the CAPI objects on the API server. Due to this, any objects created via an earlier CAPI version, say v1alpha3, which sets the owner reference API version to group name + v1alpha3 were not being found since no such object would exist. A similar object with the updated API type is stored in the api server.

This drift happens since CAPI does not update the owner references of the objects when moving to newer API versions. There is an issue tracking this https://github.com/kubernetes-sigs/cluster-api/issues/7224 which will solve the stale owner ref API version problem. Until then we are dropping the version check and just relying on the group info and kind combination to fetch the parent object.

Fixes #

n/a

**Release note**:

```release-note
Drops version check when tracking owners
```

### DIFF
--- a/pkg/util/fetch_object.go
+++ b/pkg/util/fetch_object.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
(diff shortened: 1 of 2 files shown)


srm09:
Re-adding the lgtm label since it was lost during a force push.
/approve

PR #1775, opened 2023-02-12
🐛 Reorder PCI device attachment operation

### LINKED ISSUE #1749: Upgrading the VM hardware version is blocked by PCI device attachment
/kind bug

- Create a cluster with a `VSphereMachineTemplate` 
   - with PCI devices attached.
   - with virtual machine hardware version to the minimum required version (`vmx-17` and above)
- Machines/VMs using the above template cannot start up since the VM hardware version never gets upgraded.

- VMs hardware version is upgraded on the runtime and the PCI device attachment is successful

CAPV attaches the PCI devices to the VM as part of the Clone VM task before upgrading the hardware version of the VM which is possible only via the Upgrade VM Hardware task, which is queued later in the VM reconcile loop. So, the cloning never really succeeds coz the PCI device attachment on a VM with hardware version < vmx-17 is not supported.

**Environment:**

- Cluster-api-provider-vsphere version: `v1.5.1`
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04

/assign

srm09:
This patch reorders the process of attaching PCI devices to the VM. Previously, the PCI devices were attached during the VM cloning operation, but the PCI device attachment requires a minimum hardware version of vmx-17.
Instead of forcing the template to be hardcoded with a minimum hardware version, CAPV introduced the capability of upgrading the hardware version of the VM post clone operation succeeds. This would mean that the PCI devices need to be attached after the upgrade hardware VM task is finished and the hardware version of the VM has been successfully upgraded to some compatible version specified by the user.

Fixes #1749

The govmomi dependency has been updated to v0.30.2, which is the latest version released at this time, since v0.30.0 is necessary for unit test support for this reordering.

**Release note**:

```release-note
 Reorder PCI device attachment operation
```

### DIFF
--- a/apis/v1beta1/condition_consts.go
+++ b/apis/v1beta1/condition_consts.go
@@ -93,8 +93,19 @@ const (
 	// NOTE: This reason does not apply to VSphereVM (this state happens after the VSphereVM is in ready state).
 	WaitingForNetworkAddressesReason = "WaitingForNetworkAddresses"
 
-	// TagsAttachmentFailedReason (Severity=Error) documents a VSPhereMachine/VSphereVM tags attachment failure.
+	// TagsAttachmentFailedReason (Severity=Error) documents a VSphereMachine/VSphereVM tags attachment failure.
 	TagsAttachmentFailedReason = "TagsAttachmentFailed"
+
+	// PCIDevicesDetachedCondition documents the status of the attached PCI devices on the VSphereVM.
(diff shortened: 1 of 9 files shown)


geetikabatra:
Needs rebasing, otherwise looks good to me.

srm09:
Raising a manual cherry pick for this PR because of the merge conflicts above.

PR #1803, opened 2023-02-27
:sparkles: Make the cleaning of cluster's identity secrets configurable with default = true (as is)

jkremser:
https://kubernetes.slack.com/archives/CKFGK3SSD/p1677506848387019

It adds another command argument that can change the default behavior of secret deletion. By default it behaves as it was => deleting the secrets referenced from vspherecluster CRs, but if set to `false`, it won't be doing it.

**Release note**:

```release-note

```


srm09:
@jkremser Were you ok with the slack conversation we had around this one? Is this ok to close the PR?

chrischdi:
/close

Due to lack of feedback.

If you are still interested in this feature, feel free to reopen the PR.

PR #1804, opened 2023-02-27
:bug: Calculate default primary disk size from base template

dlipovetsky:
The primary disk size is [optional](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/48a9daca388238ea942c407b0752c4eeaeb09204/apis/v1beta1/types.go#L151), according to the API.
    
Prior to this commit, if the user did not set an explicit primary disk size, VM creation would fail with a validation error.
    
This PR sets the primary disk size to the base template primary disk size when the user does not set an explicit primary disk size.

Fixes #

**Release note**:

```release-note
Default the machine primary disk size to the base template disk size.
```

### DIFF
--- a/pkg/services/govmomi/vcenter/clone.go
+++ b/pkg/services/govmomi/vcenter/clone.go
@@ -350,12 +350,16 @@ func getDiskSpec(ctx *context.VMContext, devices object.VirtualDeviceList) ([]ty
 }
 
 func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (types.BaseVirtualDeviceConfigSpec, error) {
-	if disk.CapacityInKB > diskCloneCapacityKB {
+	switch {
+	case diskCloneCapacityKB == 0:
+		// No disk size specified for the clone. Default to the template disk capacity.
+	case diskCloneCapacityKB > 0 && diskCloneCapacityKB >= disk.CapacityInKB:
+		disk.CapacityInKB = diskCloneCapacityKB
+	case diskCloneCapacityKB > 0 && diskCloneCapacityKB < disk.CapacityInKB:
 		return nil, errors.Errorf(
 			"can't resize template disk down, initial capacity is larger: %dKiB > %dKiB",
 			disk.CapacityInKB, diskCloneCapacityKB)
 	}
-	disk.CapacityInKB = diskCloneCapacityKB
 
 	return &types.VirtualDeviceConfigSpec{
 		Operation: types.VirtualDeviceConfigSpecOperationEdit,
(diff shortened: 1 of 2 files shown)


dlipovetsky:
I think some jobs are failing due to test infrastructure.

- [pull-cluster-api-provider-vsphere-integration-test ](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/1804/pull-cluster-api-provider-vsphere-integration-test/1630321118915072000#1:build-log.txt%3A369)
```
rate limit for github api has been reached. Please wait one hour or get a personal API token and assign it to the GITHUB_TOKEN environment variable
```

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -350,12 +350,16 @@ func getDiskSpec(ctx *context.VMContext, devices object.VirtualDeviceList) ([]ty
...
+	switch {
+	case diskCloneCapacityKB == 0:
+		// Use to the template disk capacity.
+	case diskCloneCapacityKB > 0 && diskCloneCapacityKB >= disk.CapacityInKB:
+		disk.CapacityInKB = diskCloneCapacityKB
+	case diskCloneCapacityKB > 0 && diskCloneCapacityKB < disk.CapacityInKB:
 		return nil, errors.Errorf(
 			"can't resize template disk down, initial capacity is larger: %dKiB > %dKiB",
 			disk.CapacityInKB, diskCloneCapacityKB)
 	}
```
srm09, line 362:
```suggestion
	if diskCloneCapacityKB > 0 {
		if disk.CapacityInKB > diskCloneCapacityKB {
			return nil, errors.Errorf(
				"can't resize template disk down, initial capacity is larger: %dKiB > %dKiB",
				disk.CapacityInKB, diskCloneCapacityKB)
		}
		disk.CapacityInKB = diskCloneCapacityKB
	}
```
Does this logic make sense?

srm09:
/retest
Also, could you please squash the changes into a single commit.

pkg/services/govmomi/vcenter/clone.go
dlipovetsky: I tried this approach first, then tried the switch statement.

I found the switch statement to more clearly communicate that there are 3 possible cases (no size specified, valid size specified, invalid size specified).

Please let me know if you feel strongly that I should use the if, rather than the switch.

dlipovetsky:
Thanks for reviewing, @srm09!

dlipovetsky:
I fixed the comment in the switch statement, squashed commits, and force-pushed.

pkg/services/govmomi/vcenter/clone.go
srm09: My major concern was with the empty switch case but I see your point in the switch case laying out the 3 possible cases much more clearly.

PR #1809, opened 2023-03-01
🌱 Adds junit reporter to the ginkgo command

srm09:
Ginkgo 2.0 removed the support of custom reporters when running the test suite, and instead provided it as a command line argument on the command itself. This patch makes sure that the report of the test suite gets created in the artifacts folder.

n/a

Before Gingko 2.0 test grid reported the test runs as following
image

After the upgrade, the test grid report shows as below:
image

**Release note**:

```release-note
NONE
```


adam-jian-zhang (approved):
LGTM

srm09:
/approve
/lgtm

Adding a manual label to unblock this merge so that we can unblock the next minor release. Uploading the screenshots of the fixes prow job
image
image

PR #1839, opened 2023-03-27
:sparkles: content library support proposal

### LINKED ISSUE #1838: Content Library Support
/kind feature

[A clear and concise description of what you want to happen.]

We support clone from vSphere template in current CAPV implementation. In some cases, user might wish to use content library as the source of the OVA/templates for better managbility and consistency. For example if user have multiple vCenter environments, content library can help manage propagation of the VM templates, and ensure consistencies of the OVAs/templates.

**Environment:**

adam-jian-zhang:

Proposal for content library support.

Fixes #1838

**Release note**:

```release-note

Support content library as the source for VM clone.

```

### DIFF
--- a/docs/proposal/content-library-support.md
+++ b/docs/proposal/content-library-support.md
@@ -1,0 +1,134 @@
+# Content Library Support
+
+```text
+---
+title: Support for Content Library
+authors:
+  - "@adam-jian-zhang"
+reviewers:  
+  - "@srm09"
+  - "@yastij"
+creation-date: 2023-03-27
+last-updated: 2023-03-27
+status: proposed
+---
+```
+
+## Table of Contents
+
+* [Content Library Support](#Content-Library-Support)
+  * [Table of Contents](#table-of-contents)
+  * [Glossary](#glossary)
+  * [Summary](#summary)
+  * [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+  * [Proposal](#proposal)
+    * [User Stories](#user-stories)
+      * [Story 1](#story-1)
+      * [Story 2](#story-2)
+    * [Overall Design](#overall-design)
+    * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+      * [Naming Scheme](#naming-scheme)
+      * [API changes](#api-changes)
+        * [VirtualMachineCloneSpec CR](#virtualmachineclonespec-cr)
+      * [Implementation Details](#implementation-details)
+        * [VSphereVM clone changes](#vspherevm-clone-changes)
+    * [Resources](#resources)
+
+## Glossary
+
+**Content Library**: The content library of vSphere, which stores artifacts such as OVA, vm template, etc. which can serve as the source the VM clones from.
+
+## Summary
+
+In current CAPV implementation, we only support cloning VM from vSphere template. This works fine for normal single vCenter
+scenario as everything is under a single vCenter context. If we consider multi-vCenter scenario, the template management becomes
+a problem, because we can not reference template cross vCenter boundary. We can manually copy the template to different vCenters as a workaround, but there is no mechanism to ensure the validity/consistency of the templates, it is error-prone and not maintainable.
+
+Content library can help template management in this case, it supports subscriptions among different content libraries and provides sync capability to help distribute OVA/templates to downstream content libraries. For example, we can set up a content library on the primary vCenter that hosts the management cluster, put the OVA/templates in the content library which serves as the version of truth for the OVA/templates we intended to use. And then for other vCenters that
+hosts workload clusters, we set up a content library and subscribe to the primary content library, and sync the content library items(templates/OVAs) to it. In this way it is easy to manage templates/OVAs and have confidence in the content of the templates/OVAs.
+
+Using content library as the source of templates has another little benefit that enables using local datastore as destination for the VMs, currently we are limited to shared datastore.
+
+## Motivation
+
+Enhance CAPV to clone VM from content library, to make multi-vCenter template management easier.
+
+### Goals
+
+* Use ContentLibrary items as the VM image template.
+
+### Non-Goals
+
+* Turn content library the only supported template source.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As an admin, I'd like to manage templates/OVAs across the vCenters that can be consumed by CAPV.
+
+#### Story 2
+
+As an cluster user, I'd like to use content library items(templates/OVAs) as the source of cluster nodes.
+
+### Overall Design
+
+The design focuses on how to clone VMs from content library items. For the content library management itself users can reference vSphere documentation.
+
+### Implementation Details/Notes/Constraints
+
+#### Naming scheme
+
+We use `Template` field of `VirtualMachineCloneSpec` to specify the source we want to clone,
+since we want to augment it to also from the content library, we need to have a way to differentiate the source.
+I propose we use URI scheme for that, it has the format: `scheme://location` which
+
+* scheme, can be vsphere|library
+  * vsphere for conventional vm templates
+  * library for OVA/templates from content library
+* to make it backward compatible, the scheme part can be omitted for conventional vm templates, so if we encounter
+  template without scheme part, we assume it is vsphere template, existing CAPV does not need to do anything.
+
+#### API changes
+
+##### VirtualMachineCloneSpec CR
+
+The `VirtualMachineCloneSpec` has following changes, the spec is used to specify clone spec for `VsphereVM`, it needs the following change to deal with newly added content library capabilities.
+
+```golang
+type VirtualMachineCloneSpec struct {
+  // Template is the name or inventory path of the template used to clone
+  // the virtual machine. It can be a vSphere template, or a content library item. For example:
+  // vsphere template: [vsphere://location of the vsphere template], vsphere:// can be omitted for backward compatibility
+  // content library: [library://name of the content library item]
+  // note that this has impact on cloneMode field, since it is only possible to do full clone from content library items,
+  // and we should have validation rule for this.
+  // +kubebuilder:validation:MinLength=1
+  Template string `json:"template"`
+
+  // The secret that stores the credentials for accessing the content library.
+  // It should reside on the same namespace of the Cluster that reference it.
+  // It is optional since enabling access control for the content library is optional. It's up to content library admin to decide if
+  // it is required to enable access control.
+  // +optional
+  LibraryCredentials string `json:"library_credentials"`
+}
+```
+
+#### Implementation Details
+
+##### VSphereVM clone changes
+
+We need to enhance the [clone](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/19c5deb79e97a8cd9d546ec39db3f29231bbcb91/pkg/services/govmomi/vcenter/clone.go#L75) function to recogize the scheme we defined, and choose appropriate API to clone the VM.
+
+There are no changes to reconciling logic of various controllers.
+
+## Resources
+
+* [Using Content Libraries](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.vm_admin.doc/GUID-254B2CE8-20A8-43F0-90E8-3F6776C2C896.html)
+* [Content Library API](https://developer.vmware.com/apis/vsphere-automation/latest/content/content/)
+* [govc library clone](https://github.com/vmware/govmomi/blob/main/govc/library/clone.go)
--- a/docs/proposal/content-library-support.md
+++ b/docs/proposal/content-library-support.md
@@ -5,11 +5,14 @@
 title: Support for Content Library
 authors:
   - "@adam-jian-zhang"
+  - "@rikatz"
 reviewers:  
   - "@srm09"
   - "@yastij"
+  - "@chrischdi"
+  - "@randomvariable"
 creation-date: 2023-03-27
-last-updated: 2023-03-27
+last-updated: 2023-09-21
 status: proposed
 ---
 ```
@@ -38,7 +41,7 @@ status: proposed
 
 ## Glossary
 
-**Content Library**: The content library of vSphere, which stores artifacts such as OVA, vm template, etc. which can serve as the source the VM clones from.
+**Content Library**: The content library of vSphere, that stores artifacts such as OVA, vm template, etc. and can serve as the source for VM clones.
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,129 @@
...
+#### Naming scheme
+
+We use `Template` field of `VirtualMachineCloneSpec` to specify the source we want to clone,
+since we want to augment it to also from the content library, we need to have a way to differentiate the source.
+I propose we use URI scheme for that, it has the format: `scheme://location` which
+- scheme, can be vsphere|library
+    - vsphere for conventional vm templates
+    - library for OVA/templates from content library
+- to make it backward compatible, the scheme part can be omitted for conventional vm templates, so if we encounter the template without scheme part, we assume it is vsphere
+template, existing CPAV does not need to do anything.
```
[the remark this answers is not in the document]
adam-jian-zhang: Thanks for review. Fixed.

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,129 @@
...
+	// the virtual machine. It can be a vSphere template, or a content library item. For example:
+	// vsphere template: [vsphere://location of the vsphere template], vsphere:// can be omitted for backward compatibility
+	// content library: [library://name of the content library item]
+	// +kubebuilder:validation:MinLength=1
+	Template string `json:"template"`
+
+	// The secret that stores the credentials for accessing the content library.
+	// It is optional since enabling access control for the content library is optional. It's up to content library admin to decide if
+	// it is required to enable access control.
+	// +optional
```
srm09, line 113: Makes sense to call out that the secret needs to be in the same namespace as the Cluster.

srm09 (commented):
/lgtm
@yastij what do you think?

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,129 @@
...
+
+Using content library as the source of templates has another little benefit that enables using local datastore as destination for the VMs, currently we are limited to shared datastore.
+
+## Motivation
+
+Enhance CAPV to clone VM from content library, to make multi-vCenter template management easier.
+
+### Goals
+
+* Provision VM from content library items(OVA or template).
```
rikatz, line 60: Maybe `Use ContentLibrary items as the VM image template`?

Also, should we call out somewhere if using ContentLibrary item turns the option to use linkedClone not possible anymore? (should be a restriction, I guess? or can we do linkedClone based on content library items?)

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Non-Goals
+
+* Replacing local template as the vm cloning source.
```
rikatz, line 64: I think both non-goals can be consolidated as a single  "Turn content library the only supported template source"

srm09:
/assign @yastij 
/ok-to-test

The verify markdown job is failing, could you please fix that.

docs/proposal/content-library-support.md
adam-jian-zhang: done

adam-jian-zhang:
@srm09, @yastij , @rikatz, @vrabbi I have addressed the comments, and fixed lint, please have a review.

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,134 @@
...
+
+The `VirtualMachineCloneSpec` has following changes, the spec is used to specify clone spec for `VsphereVM`, it needs the following change to deal with newly added content library capabilities.
+
+```golang
+type VirtualMachineCloneSpec struct {
+  // Template is the name or inventory path of the template used to clone
+  // the virtual machine. It can be a vSphere template, or a content library item. For example:
+  // vsphere template: [vsphere://location of the vsphere template], vsphere:// can be omitted for backward compatibility
+  // content library: [library://name of the content library item]
+  // note that this has impact on cloneMode field, since it is only possible to do full clone from content library items,
```
zhanggbj, line 108: hi @adam-jian-zhang , I assume most users are using full clone mode, but out of curiosity, if users are using linked clone mode in an multi-vc environment, how will we handle this?

adam-jian-zhang: Linked clone is only possible within the same vc. So in multivc scenario, we can not use linked clone.

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,134 @@
...
+  // and we should have validation rule for this.
+  // +kubebuilder:validation:MinLength=1
+  Template string `json:"template"`
```
chrischdi, line 111: If we introduce the URI scheme as only valid way, we would have a breaking change.

If we add it as a breaking change, we would have to add it using a new API Version (`v1beta2`) which then also includes converting from v1beta1 to v1beta2 (adding the `vsphere://` prefix).

An alternative to introduce it as a breaking change would be: values not having a prefix will be treated as `vsphere://<value>`.

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,134 @@
...
+
+  // The secret that stores the credentials for accessing the content library.
+  // It should reside on the same namespace of the Cluster that reference it.
+  // It is optional since enabling access control for the content library is optional. It's up to content library admin to decide if
+  // it is required to enable access control.
+  // +optional
+  LibraryCredentials string `json:"library_credentials"`
```
chrischdi, line 118: Should we make this of type `corev1.TypedLocalObjectReference`?

This way we would be able to in future support other object types too without having to adjust the API.

chrischdi (commented):
This sounds like a reasonable feature and implementation looks pretty straight forward.

Should we additionally take care via validation, that if the Template refers a library, that the CloneMode is set to `fullClone`?

From testing perspective: I think we could just switch over a single e2e test to clone from a local content library in the test environment, which then uses the same template as before but just from a different source.

docs/proposal/content-library-support.md
rikatz: I think this is defined in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1839/files#diff-05d6df7ed3848d35aa732cfdeb480d8a394f42809bc3489bc87a9bcdcc0fcc6dR93 (`to make it backward compatible, the scheme part can be omitted for conventional vm templates, so if we encounter template without scheme part, we assume it is vsphere template, existing CAPV does not need to do anything.`)

chrischdi: Oh looks like I missed that 😃 thx

docs/proposal/content-library-support.md
randomvariable: Why are credentials for the library needed? A sequence diagram of interactions between VA-API and CLS that go from library item to VM creation should make this clear.

The content library will already have been subscribed to in vCenter with whatever http credentials it needs, we should only need to pass it's UUID to CreateTemplate or CreateOVF

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,134 @@
...
+}
+```
+
+#### Implementation Details
+
+##### VSphereVM clone changes
+
+We need to enhance the [clone](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/19c5deb79e97a8cd9d546ec39db3f29231bbcb91/pkg/services/govmomi/vcenter/clone.go#L75) function to recogize the scheme we defined, and choose appropriate API to clone the VM.
```
randomvariable, line 126: I'd like some more details on this before proceeding. I think you will want to use https://github.com/vmware/govmomi/blob/main/vapi/vcenter/vcenter_ovf.go#L292-L306 and then https://github.com/vmware/govmomi/blob/252e157f27b33d5cebeecdb21212e3f3fbe44039/object/virtual_machine.go#L72 

We already have this logic in VM Service mode at https://github.com/vmware-tanzu/vm-operator/blob/main/pkg/vmprovider/providers/vsphere/session/session_vm_create.go

randomvariable (changes requested):
I want to see quite a bit more detail before approving this one. Please respond to comments.

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,134 @@
...
+a problem, because we can not reference template cross vCenter boundary. We can manually copy the template to different vCenters as a workaround, but there is no mechanism to ensure the validity/consistency of the templates, it is error-prone and not maintainable.
+
+Content library can help template management in this case, it supports subscriptions among different content libraries and provides sync capability to help distribute OVA/templates to downstream content libraries. For example, we can set up a content library on the primary vCenter that hosts the management cluster, put the OVA/templates in the content library which serves as the version of truth for the OVA/templates we intended to use. And then for other vCenters that
+hosts workload clusters, we set up a content library and subscribe to the primary content library, and sync the content library items(templates/OVAs) to it. In this way it is easy to manage templates/OVAs and have confidence in the content of the templates/OVAs.
+
+Using content library as the source of templates has another little benefit that enables using local datastore as destination for the VMs, currently we are limited to shared datastore.
```
rikatz, line 52: I don't think this is a problem solved by content library, but mostly because of a problem on the way we used to clone our templates: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2134 as we always set the datastore as the same from the template (while VC does support cloning between different datastores).

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Motivation
+
+Enhance CAPV to clone VM from content library, to make multi-vCenter template management easier.
+
+### Goals
+
+* Use ContentLibrary items as the VM image template.
```
rikatz, line 60: Just to explore a bit more on goals and non goals, are we going to support at this time "OVA and OVF templates" and "VM templates"?

The "OVA and OVF templates" deploy should be the one simpler, where we just "Import the item" on the content library, while the 2nd one is when we deploy a VM, customize it on vSphere and clone it back as a template to CL.

It would be good to make it clear what we want to support right now :)

rikatz: Also: do we want to provide an official subscribable content library for OVAs? If not, probably this should be a non goal to make it explicit.

docs/proposal/content-library-support.md
rikatz: re-reading the doc, as the overall design states that `For the content library management itself users can reference vSphere documentation`, I agree the credential should not be required.

Also, we should probably add that the "Management of content library items and subscriptions" is a nongoal, so it is clear that the VI admin should first subscribe/create a content library and CAPV will only use it.

rikatz:
hey @adam-jian-zhang , just took a look into the proposal again and left some comments.

Additionally, out of the proposal but IMHO should be part of the implementation, I think we need to consider that:
* This workflow should be as easy and transparent as possible to the user. This means that from an API change, I think it is fine to have just the scheme/itemname support and nothing else. 
* While not part of the direct implementation, I think that lowering the adoption bar of this feature would be great for users, so I would expect that there's a quick start documentation on using Content Libraries, some real next/next/finish like "Create content library, import OVA item, use it on your CAPV cluster".

I'm happy and also interested to help you move this proposal forward, let me know if you need something from me to keep this moving!

docs/proposal/content-library-support.md
rikatz: Also some good direct reference in https://github.com/vmware/govmomi/blob/main/govc/library/deploy.go

chrischdi:
@rikatz maybe worth talking directly to @adam-jian-zhang to figure out when and who will continue on this one or if it makes sense to take this over :-)

docs/proposal/content-library-support.md
laozc: I believe it needs to download the image content from Content Library when deploying.
We need to be careful here as it may run multiple times and create a heavy workload to vCenter if we need multiple VM instances.
I think it would be better to import the image as a separate template and do the clone as what we have today.

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,257 @@
...
+#### Implementation Details
+
+
+##### Getting the right Content Library item
+Because content libraries can have the same name, and items on different content libraries
+can have the same name, we should have a way to disambiguate what exact template users wants to deploy.
+
+This way, given the schema of a content library item is `library://[library_name_or_id]/itemname` we should:
+* Parse the field as a URI. If just a "Host" is present, it means user is requesting an item name, so it should be queried by item name
+* If both host and item are present, then host is the content library name (or uuid) and path is the item name
```
chrischdi, line 138: Do I get it right that it would be possible to use it via (noob content library question):

* not specify the library name or id: `library://ubuntu-22.04-k8s-1.28.0`
* specify the library name or id: `library://foo/ubuntu-22.04-k8s-1.28.0`

Adding direct examples here would make it a better read :-)

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,257 @@
...
+		},
+        Target: vcenter.Target {
+            // Add target spec here. Contains selected resource Pool, hostID, VMFolder
+        },
+	})
+    if err != nil {
+        // ERROR deploying the content library item
+    }
+
+	finder := find.NewFinder(vim25Client)
```
chrischdi, line 233:
```suggestion
	finder := find.NewFinder(vim25Client)
```

I assume `DeployLibraryItem` is not instantly ready and would result in a task on vsphere? so finding the object is after the task did run right? And the logic below here should then be the same as for the normal clone?

docs/proposal/content-library-support.md
```diff
@@ -0,0 +1,257 @@
...
+// https://github.com/vmware/govmomi/blob/main/govc/library/deploy.go
+import (
+    ...
+   	"github.com/vmware/govmomi/vapi/vcenter"
+    ...
+)
+
+func deployContentLibraryAsVM(itemID string, additionalargs...) (*object.VirtualMachine, error) {
+	m := vcenter.NewManager(restclient)
+	ref, err := m.DeployLibraryItem(ctx, itemID string, vcenter.Deploy{
```
chrischdi, line 221: This function call is basically the equivalent to calling `tpl.Clone` right? (https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/19c5deb79e97a8cd9d546ec39db3f29231bbcb91/pkg/services/govmomi/vcenter/clone.go#L75)

docs/proposal/content-library-support.md
rikatz: yes you do! but if there are two items named `ubuntu-22.04-k8s-1.28.0` on different libraries it should be an error, as CAPV should not select for the user one of two IMHO

Same for 2 content libraries with same names on VC.

Let me add some examples here!

docs/proposal/content-library-support.md
rikatz: correct :) And the customization will need to happen differently, as we know what exists on a template, but the OVF deployment is different (I need to figure out actually if there's a way to get the spec and customize accordingly, I didn't wanted to go that down on the spec right now).

So, for template we call lastly `tpl.Clone` that generates a new task.

And for library we call DeployLibraryItem that does that directly (I'm checking if there's a way to generate a task as well)

docs/proposal/content-library-support.md
rikatz: So, it appears to not return a task, I'm checking here

docs/proposal/content-library-support.md
chrischdi: That sounds good, thx

adam-jian-zhang: Sounds good, we need  a way to unique identify the library item user want, the proposed scheme library://library_name_or_id/itemname should be able to address that. library_id is what we actually needed here to resolve ambiguity here, allowing user to use library_name instead or even omitting the library_id/library_name when the item name is unique is just for enhancing user experience.
Personally I think the full path is better( although it is harder to use),  as it can avoid us problem down the road. I remember we have several similar issues for templates, as we allow user to just specify the template name not the full path of the template, and use tkr-resolver to update it to the full path later, which may break some time.

docs/proposal/content-library-support.md
adam-jian-zhang: When creating a subscribed content library, there are two options for dealing with the content, download immediately/when needed. 
I think we should advise user to use eager option(download immediately).
If user choose lazy option(when needed), I think the work round here is to deploy a test vm before creating clusters, or create a small test cluster to fetch the library items to local.

docs/proposal/content-library-support.md
rikatz: I've discussed with @adam-jian-zhang and I'm also adding to the status field of vspherevm what was the "resolved" library id / item name

I was even thinking about adding the item ID, but I think in this case is too much, as we are not allowing users to add the item ID anyway, wdyt?

chrischdi: Question: what would the status field get used for? If it is just for visualization: I think having a log line would also be okay.

rikatz: good question, IIRC vspherevm is not mutable after cloned right? So it would just be a "I'm based on this image" field, which can be a log line yes (unless I'm forgetting some other condition)

chrischdi: If we don't actually use the field I slightly tend to only log it yes 👍 

I assume the image could get moved in the content library, so the status field may get outdated.

rikatz:
Question: should we add a featureflag for it?

chrischdi:
> Question: should we add a featureflag for it?

Good question: I tend to say no, because you have to explicitly use the `library://` prefix to make use of it. There's no magical "gets used".

When adding a feature gate for it I think we would also have to verify it on creation (if the feature gate is disabled) that no `library://` prefix gets set.

rikatz:
@chrischdi 👋 

I'm coming back to this (after almost 2 months of hiatus, sorry!) 

Can you take a new look just to see if I'm missing something?

I think the most important piece right now is that the "cloning" from a Content Library should happen on a different workflow, as it does not use a template/object reference but instead https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1839/files#diff-05d6df7ed3848d35aa732cfdeb480d8a394f42809bc3489bc87a9bcdcc0fcc6dR219 and also as you pointed, it is "sync" (does not return a task).

Thanks!

chrischdi:
Hi @rikatz , I think this looks good to continue.

The only question coming up for me after reading all again is this part here:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1839/files#r1335335234

About "do we have to pre download it in some way" but I'd let it up to you if there is something to think about here. It would be more obvious to me when we have code :-)

docs/proposal/content-library-support.md
rikatz: The problem with download immediately is that we may end up subscribing to CL with a lot of images that may be problematic for long term storage space.

Anyway, can we agree this is not an enhancement question but more an implementation detail?

docs/proposal/content-library-support.md
ericvmw: It does not show the task in the deploy result though it does create a vCenter task, you can find it by query vCenter task API, and may check how VM operator handle it.

sbueringer:
/close

Let's revive if there is interest in this again

PR #1843, opened 2023-03-28
Fix ginkgo mismatch

zhanggbj:
**What this PR does / why we need it**:
Run e2e locally failed due to mismatched ginkgo. So bump ginkgo in hack/tools to fix it. 
This is a just quick fix for us to run e2e. For long-term, since bot will bump dependencies automatically,  is it possible to bump hack/tools packages by bot?
```
Ginkgo detected a version mismatch between the Ginkgo CLI and the version of Ginkgo imported by your packages:
  Ginkgo CLI Version:
    2.8.3
  Mismatched package versions found:
    2.9.1 used by e2e
```

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1842

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note

```

Issue #1842, opened 2023-03-28
Run e2e locally failed due to mismatch ginkgo version


zhanggbj:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
Run e2e locally with `make e2e`

**What did you expect to happen:**
Run e2e successfully

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

It failed with error message below. It looks like the ginkgo package in go.mod is **2.9.1** but ginkgo package in hack/tools is **2.8.3** 

```
Contents of /Users/gongz/zhangg/workspace/src/sigs.k8s.io/cluster-api-provider-vsphere/hack/tools/bin:
ginkgo		govc		kind		kustomize

time /Users/gongz/zhangg/workspace/src/sigs.k8s.io/cluster-api-provider-vsphere/hack/tools/bin/ginkgo -v -focus="multivc"  -timeout=2h \
		--output-dir="/Users/gongz/zhangg/workspace/src/sigs.k8s.io/cluster-api-provider-vsphere/_artifacts" --junit-report="junit.e2e_suite.1.xml" ./test/e2e -- \
		--e2e.config=""/Users/gongz/zhangg/workspace/src/sigs.k8s.io/cluster-api-provider-vsphere/test/e2e/config/vsphere-dev.yaml"" \
		--e2e.artifacts-folder="/Users/gongz/zhangg/workspace/src/sigs.k8s.io/cluster-api-provider-vsphere/_artifacts" \
		--e2e.skip-resource-cleanup=true \
		--e2e.use-existing-cluster="false"
Ginkgo detected a version mismatch between the Ginkgo CLI and the version of Ginkgo imported by your packages:
  Ginkgo CLI Version:
    2.8.3
  Mismatched package versions found:
    2.9.1 used by e2e

  Ginkgo will continue to attempt to run but you may see errors (including flag
  parsing errors) and should either update your go.mod or your version of the
  Ginkgo CLI to match.

  To install the matching version of the CLI run
    go install github.com/onsi/ginkgo/v2/ginkgo
  from a path that contains a go.mod file.  Alternatively you can use
    go run github.com/onsi/ginkgo/v2/ginkgo
  from a path that contains a go.mod file to invoke the matching version of the
  Ginkgo CLI.

  If you are attempting to test multiple packages that each have a different
  version of the Ginkgo library with a single Ginkgo CLI that is currently
  unsupported.

[...]

PR #1844, opened 2023-03-28
🌱 Updates dependabot config to scan hack/tools directory

srm09:
Since the dependencies under the hack/tools directory are managed by go.mod, we should instruct dependabot to also look for updates to those dependencies as well. Raised as a follow up to #1843 

n/a

n/a

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,3 +27,19 @@ updates:
     prefix: ":seedling:"
   labels:
     - "ok-to-test"
+
+- package-ecosystem: "gomod"
+  directory: "/hack/tools"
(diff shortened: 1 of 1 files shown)


zhanggbj:
LGTM. This would be helpful to avoid issue like package mismatch as https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1842.

zhanggbj:
/lgtm
This would be helpful to avoid issue like package mismatch as https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1842.

PR #1857, opened 2023-04-03
:sparkles: ensure IPAddressClaims are created with a cluster annotation

adobley:
This PR ensures that IPAddressClaims get an annotation with a reference to the cluster they belong to. This will allow pausing of claim reconciliation when the cluster is paused, such as during a move from the bootstrap clusters to the management clusters.

This is related to https://github.com/telekom/cluster-api-ipam-provider-in-cluster/pull/105, which will use this annotation to check the cluster paused state when reconciling claims.

Fixes #

**Release note**:

```release-note
add cluster name label annotation to IPAddressClaims created by CAPV
```

### DIFF
--- a/controllers/vspherevm_ipaddress_reconciler.go
+++ b/controllers/vspherevm_ipaddress_reconciler.go
@@ -125,7 +125,8 @@ func (r vmReconciler) reconcileIPAddressClaims(ctx *context.VMContext) error {
 }
 
 // createOrPatchIPAddressClaim creates/patches an IPAddressClaim object for a device requesting an address
-// from an externally managed IPPool.
+// from an externally managed IPPool. Ensures that the claim has a reference to the cluster of the VM to
+// support pausing reconciliation.
 // The responsibility of the IP address resolution is handled by an external IPAM provider.
 func createOrPatchIPAddressClaim(ctx *context.VMContext, name string, poolRef corev1.TypedLocalObjectReference) (*ipamv1.IPAddressClaim, bool, error) {
 	claim := &ipamv1.IPAddressClaim{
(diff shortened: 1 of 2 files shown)


srm09:
Checkout https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1867, this logic might need to be moved into another file.

pkg/services/govmomi/service.go
```diff
@@ -326,6 +327,9 @@ func createIPAddressClaim(ctx *context.VMContext, ipAddrClaimName string, poolRe
 				},
 			},
 			Finalizers: []string{infrav1.IPAddressClaimFinalizer},
+			Annotations: map[string]string{
+				clusterv1.ClusterLabelName: ctx.VSphereVM.ObjectMeta.Annotations[clusterv1.ClusterLabelName],
+			},
```
srm09, line 332: This needs to be moved within https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/6c50469d4c890ad32e39bdd81c3c0aa932e16f97/controllers/vspherevm_ipaddress_reconciler.go#L137-L153

pkg/services/govmomi/service_test.go
```diff
@@ -140,6 +144,11 @@ func Test_reconcileIPAddressClaims_ShouldGenerateIPAddressClaims(t *testing.T) {
 		claimedCondition := conditions.Get(ctx.VSphereVM, infrav1.IPAddressClaimedCondition)
 		g.Expect(claimedCondition).NotTo(BeNil())
 		g.Expect(claimedCondition.Reason).To(Equal(infrav1.WaitingForIPAddressReason))
+
+		// Ensure that the claims have a reference to the cluster to support pauses
+		for _, claim := range ipAddrClaims.Items {
+			g.Expect(claim.ObjectMeta.Annotations).To(HaveKeyWithValue(clusterv1.ClusterLabelName, "my-cluster"))
+		}
```
srm09, line 151: This can be moved under https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/6c50469d4c890ad32e39bdd81c3c0aa932e16f97/controllers/vspherevm_ipaddress_reconciler_test.go#L90-L91

srm09:
/unhold
/ok-to-test
/lgtm
/approve

Thanks for reworking the PR to simplify the logic around api server operations for IP Address Claim objects.

flawedmatrix:
@srm09 Apologies, we realized after the fact that we should have been setting the label, not the annotation. We'll make another PR for this.

srm09:
Can we bake the logic to remove the annotation in the same PR?

PR #1862, opened 2023-04-05
🌱 CI: Use script to download golangci-lint

srm09:
The [PR](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1847) introduced the bump to v1.52.0 version of the `golangci-lint` package used to run linters. With this bump, importing golangci-lint as a depenendency in go.mod and building it is not possible due to https://github.com/ryancurrah/gomodguard/issues/33
Hence switching to using the CAPI script to download the binary and use that in the `lint` jobs.

Necessary to fix the broken `main` branch.

n/a

**Release note**:

```release-note
NONE
```


srm09:
/override pull-cluster-api-provider-vsphere-test
Failure due to a flaky test fixed in #1821

PR #1867, opened 2023-04-05
🌱 Refactor: Move IPAddressClaim reconcile logic to controller

srm09:
This patch moves the IPAddressClaim creation/update/deletion logic to the vsphereVM controller.
It also:
- Adds a watcher to IPAddressClaim objects: Introduces a watcher on the IPAddressClaim object to enqueue a reconcile request for the VSphereVM owner object.
- Reorders Ready condition calculation logic: Re-orders the conditions used to geenrate the summary condition on the VSphereVM object.
- Adds new ipam sub-package under govmomi pkg: This resturctures the logic to check for the bound IP address claims and building the corresponding IPAM state which gets passed to the metadata reconcile method.

**Release note**:

```release-note
Refactor: Move IPAddressClaim reconcile logic to controller
```


srm09:
cc: @tylerschultz @adobley

tylerschultz:
I don't think I see any logic changes, is that correct? This change is mostly the moving of the code out of services/govmomi, and the added reconcilliation registration for IPAddressClaims?

/LGTM

yastij (commented):
/approve
/lgtm

PR #1869, opened 2023-04-06
:bug: Check for bool ptrs before using it

rikatz:
This PR adds a check for boolean pointers before using it.

When a user is configuring failure domains, some fields can be specified as booleans. But those are pointers because "nil is different than false".

The controller assumes this pointer will always exists, are there are mutation webhooks that defaults this field to false when it is not set (and then the pointer starts to exists).

But in deployments were the webhook is not working properly, or no mutation happens, the object will be created with a nil pointer and controller will panic.

This fix adds a simple check on pointer existence, before allowing it to be referenced
Fixes #

**Release note**:

```release-note
Avoid null pointer on boolean pointer fields
```


srm09:
/close
Closing in favor of https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1898

Issue #1757, opened 2023-01-28
`verify-crds` test is flaky


ntoofu:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

`hack/verify-crds.sh`, which is used by `pull-cluster-api-provider-vsphere-verify-crds` prow job or `make verify-crds`, is flaky. The script occasionally returns errors, but it will run successfully when retried. For example, please see [this PR](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1739#issuecomment-1407443719).

**What did you expect to happen:**

Even if the script is executed several times, the results are the same.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

An example of a failed test: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/1739/pull-cluster-api-provider-vsphere-verify-crds/1619032549575626752

**Environment:**

- Cluster-api-provider-vsphere version: I confirmed this issue on 4aafab65d5cfcde29026d01b70ca3065124cd019
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

ntoofu:
It might be a bug of `controller-gen`. Even when I called `controller-gen` directly, the same issue still occurred.

```
$ while true; do ./hack/tools/bin/controller-gen paths=./apis/v1alpha3 paths=./apis/v1alpha4 paths=./apis/v1beta1 crd:crdVersions=v1 output:stdout | sha256sum; sleep 3; done
0ca33b5ac647051dde8e233cd0c324d5581a27cf167b690bb7b67a8854fccf9f  -
0ca33b5ac647051dde8e233cd0c324d5581a27cf167b690bb7b67a8854fccf9f  -
0ca33b5ac647051dde8e233cd0c324d5581a27cf167b690bb7b67a8854fccf9f  -
9a2448e120cb9da054283686b468f268e53914b350bcdefda56b16ee31e46d8a  -
0ca33b5ac647051dde8e233cd0c324d5581a27cf167b690bb7b67a8854fccf9f  -
0ca33b5ac647051dde8e233cd0c324d5581a27cf167b690bb7b67a8854fccf9f  -
0ca33b5ac647051dde8e233cd0c324d5581a27cf167b690bb7b67a8854fccf9f  -
[...]

PR #1870, opened 2023-04-06
⚠️ Deprecates the NodeLabeling feature flag

### LINKED ISSUE #1852: Deprecate NodeLabeling feature flag from CAPV
/kind feature

CAPV introduced the `NodeLabeling` feature flag to propagate all the labels with the prefix `cluster.x-k8s.io` from the `Machine` object onto the Kubernetes node.
Here is the PR that introduced the flag in CAPV: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1628

In CAPI 1.4, this feature is natively available. So deploying CAPV with CAPI 1.4, this feature might not need to be enabled on CAPV. With the next release introducing support for CAPI 1.4, we should officially deprecate support for the flag from CAPV deployments. And eventually plan on removing it entirely from CAPV.  

**Environment:**

- Cluster-api-provider-vsphere version: `next` minor version, probably `v1.7.0`

srm09:
Now that CAPI v1.4.0 is supporting this feature natively, we are dropping support from CAPV for this feature. Consider using CAPV with CAPI v1.4.0 to avail this feature.

Fixes #1852 

This will change the default manifests shipped with CAPV which will no longer include mention of the `NodeLabeling` feature flag.

**Release note**:
```release-note
⚠️ Deprecates the NodeLabeling feature flag
```


config/rbac/role.yaml
```diff
@@ -106,15 +106,6 @@ rules:
   - get
   - list
   - watch
-- apiGroups:
```
yastij, line 109: why are we dropping these? I don't see any corresponding kubebuilder tags being dropped?

srm09: The entire node controller is being dropped, the RBAC changes are coming from this one.

controllers/node_controller.go
```diff
@@ -1,180 +0,0 @@
...
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-
-	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/constants"
-	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
-	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/record"
-)
-
-// +kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch;update;patch
-// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines;clusters,verbs=get;watch;list
```
srm09, line 46: @yastij These are the kubebuilder directives being removed.

srm09:
The api-diff is failing due to the node controller public method being removed.
/override pull-cluster-api-provider-vsphere-apidiff-main 

The verify crds and gen jobs failure are being tracked in #1757, those are non required jobs now.
/unhold

PR #1891, opened 2023-04-25
:bug: Allow ipv6 gateway to be optional when DHCPv6 is disabled.

### LINKED ISSUE #1861: Gateway6 is not required, but says it is
/kind bug

The [description](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/0b9668541617aeb536092f9aa288381c5b9a8c03/apis/v1alpha4/types.go#L234-L237) for Gateway6 says it is required if DHCP6 is false. For Node IPAM, Gateway is also required in the case of both IPv4 and IPv6.

However, IPv6 works differently than IPv4 in that Gateway comes from the NDP and not from DHCPv6. So disabling DHCPv6 doesn't affect the auto configuration of the Gateway like it does in IPv4 and we are still able to rely on NDP to set our gateway without DHCP.

I am allowed to not set Gateway6 when DHCP6 is false in the case of Node IPAM and in the case of setting static IPAddresses on the device spec.

**Environment:**

christianang:
This PR allows ipv6 gateway to be optional. In the case of using IPAM for nodes the IPAddress can have an ipv6 address without a gateway and it will configure the machine's metadata appropriately. The comment on gateway6 says it is required when DHCPv6 is disabled. This comment was removed, but there was no logic enforcing it except in the case of Node IPAM were it would always try to parse the gateway from the `IPAddress` resource.

Fixes #1861

**Release note**:

```release-note
Allow IPv6 Gateways to be optional when DHCPv6 is disabled.
```

### DIFF
--- a/apis/v1beta1/types.go
+++ b/apis/v1beta1/types.go
@@ -281,7 +281,6 @@ type NetworkDeviceSpec struct {
 	Gateway4 string `json:"gateway4,omitempty"`
 
 	// Gateway4 is the IPv4 gateway used by this device.
-	// Required when DHCP6 is false.
 	// +optional
 	Gateway6 string `json:"gateway6,omitempty"`
 
--- a/config/default/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachines.yaml
+++ b/config/default/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachines.yaml
@@ -1075,7 +1075,6 @@ spec:
                           type: string
                         gateway6:
                           description: Gateway4 is the IPv4 gateway used by this device.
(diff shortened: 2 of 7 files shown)


christianang:
cc @schrej We made this PR to make ipv6 optional. Per the latest discussion on https://github.com/kubernetes-sigs/cluster-api/issues/8536#issuecomment-1521390064 is it still reasonable to required ipv4 gateway in CAPV or would you need this to also be optional?

srm09:
Has this code been tested in a live setup, there are no provisions for ipv6 tests upstream?
/hold

christianang:
I did manually test this on a live ipv6 env. Will fix the test and lint errors.

PR #1894, opened 2023-04-27
✨ Introduces parameter for control plane zone selection

srm09:
This patch introduces a new parameter for selecting zones for control plane nodes.
In the past, all zones having the same vCenter URL in the spec.server would be associated to a vSphereCluster object with the same vCenter URL set in the spec.server.
Now the vSphereCluster spec exposes a label based selection which can be used to choose the zones for control plane node placement. The label selector follows the same rules as described on the label selector types in https://pkg.go.dev/k8s.io/apimachinery@v0.26.1/pkg/apis/meta/v1#LabelSelector

Fixes #

n/a

**Release note**:

```release-note
Introduces parameter for control plane zone selection
```

### DIFF
--- a/apis/v1beta1/vspherecluster_types.go
+++ b/apis/v1beta1/vspherecluster_types.go
@@ -58,6 +58,12 @@ type VSphereClusterSpec struct {
 	// for each of the objects responsible for creation of VM objects belonging to the cluster.
 	// +optional
 	ClusterModules []ClusterModule `json:"clusterModules,omitempty"`
+
+	// FailureDomainSelector is the label selector to use for failure domain selection
+	// for the control plane nodes of the cluster.
+	// An empty value for the selector includes all the related failure domains.
+	// +optional
+	FailureDomainSelector *metav1.LabelSelector `json:"failureDomainSelector,omitempty"`
 }
 
 // ClusterModule holds the anti affinity construct `ClusterModule` identifier
--- a/controllers/vspherecluster_reconciler.go
+++ b/controllers/vspherecluster_reconciler.go
@@ -27,9 +27,11 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/utils/pointer"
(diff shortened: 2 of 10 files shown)


apis/v1beta1/vspherecluster_types.go
```diff
@@ -58,6 +58,12 @@ type VSphereClusterSpec struct {
 	// for each of the objects responsible for creation of VM objects belonging to the cluster.
 	// +optional
 	ClusterModules []ClusterModule `json:"clusterModules,omitempty"`
+
+	// ZoneSelector is the label selector to use for failure domain selection
+	// for the control plane nodes of the cluster.
+	// An empty zone selector includes all the related zones.
+	// +optional
+	ZoneSelector *metav1.LabelSelector `json:"zoneSelector,omitempty"`
```
rikatz, line 66: would this be a ZoneSelector or a FailureDomain selector? I think calling it a FailureDomain selector would be clearer that this is tied to a Failure Domain

controllers/vspherecluster_reconciler.go
```diff
@@ -358,15 +360,29 @@ func (r clusterReconciler) reconcileDeploymentZones(ctx *context.ClusterContext)
...
+				}
+
+				// An empty selector allows the zone to be selected
+				if !selector.Empty() && !selector.Matches(labels.Set(zone.GetLabels())) {
+					r.Logger.V(5).Info("skipping the deployment zone due to label mismatch", "name", zone.Name)
+					continue
+				}
+			}
+
 			if zone.Status.Ready == nil {
```
rikatz, line 377: Probably not related to this PR, but...
Based on this logic, if any of the "available deployment zone" is not ready, it will fail the whole deployment. Now, if I have 6 deployment zones, and a request of 3 replicas, but of this 6 zones 1 is off, wouldn't this stop the whole deployment anyway?

rikatz:
This is amazing! We are looking forward to have a way to select failure domains per labels. This would help on situations like "I want to control what subset of FDs I can use for my Control Plane, instead of deploying it all spread around the globe" and IMO a label selector is the best approach.

Thanks for this one Sagar!

controllers/vspherecluster_reconciler.go
srm09: The idea is to wait until all the zones have reported their status. Irrespective of the ready status (`true` or `false`), as long as it is reported, we let the reconciliation pass through. As for including the zones in the failure domain list, only the zone with Ready reported as `true` are included in the list.

apis/v1beta1/vspherecluster_types.go
srm09: Done, makes sense.

yastij (commented):
/approve

This looks good in general

srm09:
/retitle ✨ Introduces parameter for control plane zone selection

PR #1897, opened 2023-04-28
⚠️ Re-introduce the keep alive flag on the CAPV manager

### LINKED ISSUE #1896: Re-Introduce the keep alive feature flag
/kind feature

PR #https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1522 removed the `keep-alive` feature flag from the CAPV command line arguments since the default mode of this flag was supposed to be on by default. Although this is a good feature to be turned on by default, we have observed that in some cases long running sessions (and leaks) tend to overwhelm the vCenter resources causing important vCenter services to not function properly.
This enhancement talks about re-introducing the flag as an argument on the CAPV controller manager.

We should re-introduce the flag and keep it enabled by default, so as to not introduce any breaking behavior changes.

**Environment:**

- Cluster-api-provider-vsphere version: `v1.6.1`

srm09:
Thie patch re-introduces the keep alive flag on the CAPV manager. The flag is by enabled default so there are no breaking changes after upgrades.

Fixes #1896 

This leads to the `--enable-keep-alive` flag being exposed on the CAPV manifest.

**Release note**:
```release-note
Re-introduce the keep alive flag on the CAPV manager
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -19,6 +19,7 @@ spec:
       - args:
         - --leader-elect
         - --v=4
+        - --enable-keep-alive
         - "--feature-gates=NodeAntiAffinity=${EXP_NODE_ANTI_AFFINITY:=false}"
         image: gcr.io/cluster-api-provider-vsphere/release/manager:latest
         imagePullPolicy: IfNotPresent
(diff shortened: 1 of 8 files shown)


srm09:
/unhold
/override pull-cluster-api-provider-vsphere-verify-gen
since the failure is being tracked by a different upstream issue.

PR #1898, opened 2023-04-28
⚠️ Deprecates the auto configure functionality

### LINKED ISSUE #1899: Remove the tagging functionality of the VSphereFailureDomain API
/kind feature

Traditionally VI admins have been responsible for tagging and managing the vCenter resources. Having CAPV tag resources out of band is a departure from this behavior. From the feedback we received from our users, this is a functionality that they do not want to expose to CAPV, but have the VI admins perform resource tagging as a pre-requisite. 

Deprecate the functionality via the following steps:
- Introduce the deprecation notice on the API type
- Remove the controller logic for resource tagging

**Environment:**

srm09:
Deprecates the auto configure functionality from the VSphereDeploymentZone API.

Fixes #1899 

This would also lead to removing some permissions from the vCenter user in use by CAPV.

**Release note**:
```release-note
VSphereDeploymentZone: Deprecates the auto configure functionality
```


srm09:
/retitle ⚠️ Deprecates the auto configure functionality

apis/v1beta1/vspherefailuredomain_types.go
```diff
@@ -53,6 +53,8 @@ type FailureDomain struct {
 	// TagCategory is the category used for the tag
 	TagCategory string `json:"tagCategory"`
 
+	// Deprecated: This will be released in a future release. Setting this value
```
yastij, line 56: removed*

apis/v1beta1/vspherefailuredomain_webhook.go
```diff
@@ -80,14 +76,3 @@ func (r *VSphereFailureDomain) ValidateUpdate(old runtime.Object) error {
 func (r *VSphereFailureDomain) ValidateDelete() error {
 	return nil
 }
-
-// Default implements webhook.Defaulter so a webhook will be registered for the type.
-func (r *VSphereFailureDomain) Default() {
```
yastij, line 85: why are we removing the defaulting. This is just the deprecation announcement

srm09: The broader idea behind the PR was:
1. To add the depreciation notice to the API and
2. To remove the functionality exercised when setting the flag 

According to the PR, irrespective of what value the user sets, no resource tagging gets executed.

sbueringer:
@srm09 @yastij 

Discussed this in the offce hours yesterday.

We think it might be good to do a cycle with only deprecation first to see if we get user feedback that this feature is still used and then remove it in a subsequent release. So we would suggest to only deprecate the field in this PR.

Does this sound reasonable to you?

srm09:
@sbueringer Not objected to the deprecation of the field before we yank out the functionality entirely. Feel free to open a separate PR, and close this one.

srm09:
/close
@sbueringer to open another one with just the deprecation piece.

PR #1902, opened 2023-05-03
✨ Add VMRef moID to Object status

rikatz:
Adds a new status field to vspherevms resources, representing the fetched moid/vmref from reconciliation process.

This information is desired on situations where you create a cluster, and later may want to query on vCenter directly information about the nodes (for observability reasons, metrics collection, datastore usage, etc).

vmRef/Moid is an immutable information, so as soon as a virtual machine is created it is allocated and will persist during VM lifecycle, while biosUUID can change on specific situations (see https://kb.vmware.com/s/article/2002506).

Having biosUUID should be enough to query a virtual machine VMRef, but this needs a bunch of machinery (https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/5a027e203ceae96c9721a691e75751adec005ea0/pkg/services/govmomi/util.go#L54) while having the vmref generated during machine provisioning is desired, enough and easier to correlate vspherevms with real objects on vCenter

**Release note**:

```release-note
Issue virtual machine reference/moid on vspherevm status
```

### DIFF
--- a/apis/v1alpha3/types.go
+++ b/apis/v1alpha3/types.go
@@ -340,6 +340,9 @@ type VirtualMachine struct {
 
 	// Network is the status of the VM's network devices.
 	Network []NetworkStatus `json:"network"`
+
+	// VMRef is the the VM's Managed Object Reference on vSphere.
+	VMRef string `json:"vmRef"`
 }
 
 // SSHUser is granted remote access to a system.
--- a/apis/v1alpha3/conversion_test.go
+++ b/apis/v1alpha3/conversion_test.go
@@ -133,4 +133,5 @@ func CustomStatusNewFieldFuzzer(in *nextver.VSphereVMStatus, c fuzz.Continue) {
 
 	in.Host = ""
 	in.ModuleUUID = nil
+	in.VMRef = ""
 }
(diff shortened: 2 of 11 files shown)


rikatz:
/test pull-cluster-api-provider-vsphere-verify-gen
Works on my machine ;x

apis/v1alpha3/types.go
```diff
@@ -340,6 +340,9 @@ type VirtualMachine struct {
 
 	// Network is the status of the VM's network devices.
 	Network []NetworkStatus `json:"network"`
+
+	// VMRef is the the VM's Managed Object Reference on vSphere.
+	VMRef string `json:"vmRef"`
```
berlin-ab, line 345: VirtualMachine seems to be spelled out elsewhere. We might want to stick with that pattern for consistency.

srm09:
@yastij any thoughts on this one? 
@rikatz Do the VMRef changes need to be included on the `v1alpha3` and `v1alpha4` types?

apis/v1alpha3/types.go
rikatz: VMRef is an internal concept of vSphere :) This is why we are using this naming

rikatz:
> @rikatz Do the VMRef changes need to be included on the `v1alpha3` and `v1alpha4` types?
On the internal types, I think so (otherwise we would need to add conversions, and IIUC the types.go contains internal types only).

For previous vspherevm API, I'm not sure. We can add, as it is a noop and it will stick empty. Specifically in our case we are already using the latest CAPV so not required for us :)

yastij (commented):
/approve

@rikatz can you fix the verify gen job, will lgtm after that

rikatz:
@yastij @srm09 told me the gen error was expected due to some well known problem on generations. let me try to fix it anyway :)

PR #1903, opened 2023-05-03
:seedling: Deflake verify-crd test

rikatz:
verify-crd is flaky. This is because the yaml files can be generated out of order sometimes, getting differences. 
This approach (as recommended on [yq](https://mikefarah.gitbook.io/yq/usage/tips-and-tricks#comparing-yaml-files) will check the consistency of the CRD yaml file, not only the diffs

**Release note**:

```release-note
NONE
```


rikatz:
/close

Testing with prow image gives me the same error, locally not. Wondering if this is some different version of libs between both images

PR #1904, opened 2023-05-04
🐛 Allow thumbprint update on VSphereVM

### LINKED ISSUE #1718: Changing vCenter thumbprint causes the VSphereMachine reconciliation to be blocked
/kind bug

VSphereMachine does not have thumbprint set in its spec.
User updates the VSphereCluster `spec.thumbprint` when the certificates of the vCenter instance are updated. 
The CAPV logs show the following error:
```
\E1130 22:35:46.992024       1 vspheremachine_controller.go:469] capv-controller-manager/vspheremachine-controller/default/cluster01-control-plane-gbp6q "msg"="failed to CreateOrUpdate VSphereVM" "error"="admission webhook \"validation.vspherevm.infrastructure.x-k8s.io\" denied the request: VSphereVM.infrastructure.cluster.x-k8s.io \"cluster01-control-plane-shbbv\" is invalid: spec: Forbidden: cannot be modified"  "name"="cluster01-control-plane-shbbv" "namespace"="default"
E1130 22:35:46.992515       1 controller.go:317] controller/vspheremachine "msg"="Reconciler error" "error"="admission webhook \"validation.vspherevm.infrastructure.x-k8s.io\" denied the request: VSphereVM.infrastructure.cluster.x-k8s.io \"cluster01-control-plane-shbbv\" is invalid: spec: Forbidden: cannot be modified" "name"="cluster01-control-plane-gbp6q" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="VSphereMachine"
```

Changing the thumbprint on the VSphereCluster object should not cause the VSphereMachine reconciliation to be blocked.

When a thumbprint of the vCenter is updated, the VSphereVM object would not report an error since the controller would use a valid cached session (with the updated thumbprint, mostly because the vspherecluster controller must have created a new one with the new thumbprint). This masks the stale value on the VSphereVM object's `spec`.

We should relax the webhook validation to allow for VSphereVM's thumbprint to be updated.
- For the case where the VSphereVM uses the thumbprint from the VSphereCluster object, the VSphereVM object would be updated with the correct value of the thumbprint.  
- For the case wherein the VSphereVM object has the thumbprint set, the VSphereVM object would continue to have the incorrect value. If the thumbprint was set on the `VSphereMachineTemplate` object used to create the VSphereMachine and VSphereVM objects, it is the responsibility of the user to update the thumbprint by creating a new object with the updated thumbprint. 

When we holistically think about the set off steps to follow when updating the thumbprint of a VSphereCluster, the user updates the VSphereCluster value and all the VSphereVM objects (residing on the vCenter which has updated thumbprint) get updated irrespective of the fact whether they have `spec.thumbprint` set or not.

**Ignore this** 
For the second case above, we should also think about checking for the vCenter URLs on the VSphereVM and the VSphereCluster object, and if they are the same consider updating the thumbprint value of the VSphereVM object.

**Environment:**

zhanggbj:
Updating thumbprint of VSphereCluster will cause reconcilererror for VSphereVM, to fix it
    - Relax the VSphereVM webhook to allow the thumbprint update.
    - Keep VSphereVM thumbprint consistent with VSphereCluster

Fixes #1718

**Release note**:

```release-note
Allow thumbprint update on VSphereVM
```

### DIFF
--- a/apis/v1beta1/vspherevm_webhook.go
+++ b/apis/v1beta1/vspherevm_webhook.go
@@ -86,26 +86,22 @@ func (r *VSphereVM) ValidateUpdate(old runtime.Object) error {
 	newVSphereVMSpec := newVSphereVM["spec"].(map[string]interface{})
 	oldVSphereVMSpec := oldVSphereVM["spec"].(map[string]interface{})
 
-	// allow changes to biosUUID
-	delete(oldVSphereVMSpec, "biosUUID")
-	delete(newVSphereVMSpec, "biosUUID")
-
-	// allow changes to bootstrapRef
-	delete(oldVSphereVMSpec, "bootstrapRef")
-	delete(newVSphereVMSpec, "bootstrapRef")
+	// allow changes to biosUUID, bootstrapRef, thumbprint
+	keys := []string{"biosUUID", "bootstrapRef", "thumbprint"}
+	// allow changes to os only if the old spec has empty OS field
+	if _, ok := oldVSphereVMSpec["os"]; !ok {
+		keys = append(keys, "os")
+	}
+	r.deleteSpecKeys(oldVSphereVMSpec, keys)
+	r.deleteSpecKeys(newVSphereVMSpec, keys)
 
 	newVSphereVMNetwork := newVSphereVMSpec["network"].(map[string]interface{})
 	oldVSphereVMNetwork := oldVSphereVMSpec["network"].(map[string]interface{})
 
 	// allow changes to the network devices
-	delete(oldVSphereVMNetwork, "devices")
-	delete(newVSphereVMNetwork, "devices")
-
-	// allow changes to os only if the old spec has empty OS field
-	if _, ok := oldVSphereVMSpec["os"]; !ok {
-		delete(oldVSphereVMSpec, "os")
-		delete(newVSphereVMSpec, "os")
-	}
+	networkKeys := []string{"devices"}
+	r.deleteSpecKeys(oldVSphereVMNetwork, networkKeys)
+	r.deleteSpecKeys(newVSphereVMNetwork, networkKeys)
 
 	if !reflect.DeepEqual(oldVSphereVMSpec, newVSphereVMSpec) {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec"), "cannot be modified"))
@@ -117,4 +113,14 @@ func (r *VSphereVM) ValidateUpdate(old runtime.Object) error {
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
 func (r *VSphereVM) ValidateDelete() error {
 	return nil
+}
+
+func (r *VSphereVM) deleteSpecKeys(spec map[string]interface{}, keys []string) {
+	if len(spec) == 0 || len(keys) == 0 {
+		return
+	}
+
+	for _, key := range keys {
+		delete(spec, key)
+	}
 }
--- a/apis/v1beta1/vspherevm_webhook_test.go
+++ b/apis/v1beta1/vspherevm_webhook_test.go
@@ -34,9 +34,9 @@ const (
 func TestVSphereVM_Default(t *testing.T) {
 	g := NewWithT(t)
 
-	WindowsVM := createVSphereVM(windowsVMName, "foo.com", "", "", []string{"192.168.0.1/32", "192.168.0.3/32"}, nil, Windows)
-	LinuxVM := createVSphereVM(linuxVMName, "foo.com", "", "", []string{"192.168.0.1/32", "192.168.0.3/32"}, nil, Linux)
-	NoOSVM := createVSphereVM(linuxVMName, "foo.com", "", "", []string{"192.168.0.1/32", "192.168.0.3/32"}, nil, "")
+	WindowsVM := createVSphereVM(windowsVMName, "foo.com", "", "", "", []string{"192.168.0.1/32", "192.168.0.3/32"}, nil, Windows)
+	LinuxVM := createVSphereVM(linuxVMName, "foo.com", "", "", "", []string{"192.168.0.1/32", "192.168.0.3/32"}, nil, Linux)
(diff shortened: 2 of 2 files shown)


pkg/services/vimmachine.go
```diff
@@ -390,7 +390,10 @@ func (v *VimMachineService) createOrPatchVSphereVM(ctx *context.VIMMachineContex
 		if vm.Spec.Server == "" {
 			vm.Spec.Server = ctx.VSphereCluster.Spec.Server
 		}
-		if vm.Spec.Thumbprint == "" {
+		// Need to cover two cases:
+		// 1. vm.Spec.Thumbprint is empty after copy from VSphereMachine.Spec
+		// 2. VSphereCluster.Spec.Thumbprint is updated by user, such as when the certificates of the vCenter instance are updated
+		if vm.Spec.Thumbprint == "" || vm.Spec.Thumbprint != ctx.VSphereCluster.Spec.Thumbprint {
```
srm09, line 396: IMO this change is not necessary, but am open to listen to your thoughts here.
A bit of background, the `mutateFn` defined on line 340 aims to recreate a new copy of the `VSphereVM` object using the same values it originally used to create the `VSphereVM` object. This object is then passed to `patchOrCreate` which (in this case) updates the object stored in the API server.
So the check (on line 393, referring to the original file before the PR) is going to be true if the `VSphereMachine.Spec.thumbprint` is empty. If that is the case, then copy the value from the `VSphereCluster.Spec.thumbprint` into the VSphereVM object. 
Once the call to `patchOrCreate` is finished, you will see the VSphereVM object get updated with the newer thumbprint value.

apis/v1beta1/vspherevm_webhook.go
```diff
@@ -107,6 +107,10 @@ func (r *VSphereVM) ValidateUpdate(old runtime.Object) error {
 		delete(newVSphereVMSpec, "os")
 	}
 
+	// allow changes to thumbprint
+	delete(oldVSphereVMSpec, "thumbprint")
+	delete(newVSphereVMSpec, "thumbprint")
```
srm09, line 112: Since we have the same `delete` call being repeated over and over, consider creating a separate method which takes in a spec and a key and removes it from the spec map. List the keys which are allowed updates in an array and for each one pass it to this new function.

zhanggbj: Sure, I can refine them in next patch.

pkg/services/vimmachine.go
zhanggbj: Hi @srm09,
I'm totally agree with you about the case you mentioned, VSphereVM thumbprint is empty here so everything works fine with original logic.

I have a concern for another case, that user also specified thumbprint in VSphereMachineTemplate, so `VSphereMachine.Spec.thumbprint` has an old thumbprint, when user want to update VSphereCluster and VSphereVM with new thumbprint, according to [Line 337 ](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/pkg/services/vimmachine.go#L377) , `vm.Spec.Thumbprint` will deep copy from `VSphereMachine.Spec.thumbprint` with old thumbprint, so it is not empty and the check is fail, I'm afraid it never get the chance to update with new thumbprint. What do you think?

srm09: For the case when the user is consuming the thumbprint via the VSphereMachineTemplate, the way to move forward with the update is to create a new template with the updated thumbprint value and use that for machine creation.

srm09: This PR should handle the case when thumbprint is not set on the `VSphereMachine.Spec`, and is consumed by the VSphereCluster object.

zhanggbj: As if `VSphereMachine.Spec.thumbprint` is not empty, user shouldn't use this approach (updating `VSphereCluster`)to update VSpherevm.Spec.Thumbprint, so original logic works fine. I'll revert the change. Thanks for the clarification!

zhanggbj:
@srm09 All comments are addressed, would you please help to take another look, thanks!

/assign @srm09

zhanggbj:
I assume pull-cluster-api-provider-vsphere-verify-gen error is not related to this PR.

zhanggbj:
This PR is ready for a second review. Thanks!

srm09:
/lgtm
@yastij PTAL

yastij (approved):
/approve

samdulam:
I can't seem to be able to change thumbprint on a `vspherevm`. Editing the vspherevm accepts the new thumbprint but when inspecting again it has the old thumbprint.

sbueringer:
Does the VSphereMachine also have a thumbprint field?

samdulam:
> Does the corresponding VSphereMachine also have a thumbprint field set?

Yes.
I've changed the entry for vspherecluster but unable to change it on vspherevm.

This is causing my account to lockout in AD :(

sbueringer:
Hm, I don't have the context on why that is forbidden

@chrischdi do you know?

chrischdi:
> > Does the corresponding VSphereMachine also have a thumbprint field set?
> 
> Yes. Changing thumbprint on vspheremachine is forbidden?
> 
> I've changed the entry for vspherecluster but unable to change it on vspherevm.
> 
> This is causing my account to lockout in AD :(

This seem to be two separate issues.

A wrong thumbprint should lead to verification errors when trying to establish the client, but not lead to locking accounts in AD.

Could it be that in addition the vsphere credentials expired or are wrong?

samdulam:
erros in capv-controller-manager
`Reconciler error failed to create vcenter session: failed to create client.... thumbprint does not match`
vcenter logs show login attempts from the IP address of the host where capv-controller is running.

I can confirm that the login works when its unlocked.

samdulam:
How can I update thumbprint so I don't get verification errors please?

samdulam:
@chrischdi 
Reconcile is working even though I occasionally see thumbprint mismatch messages in capv-controller pod. I've since updated all my vspherecluster objects.

I don't see a way to unset or update the thumbprint on vspheremachine object without creating a new vspheremachinetemplate. Is that the expectation?

PR #1907, opened 2023-05-08
:bug: Add update to ipaddressclaim rbac

christianang:

This PR adds back update to the ipaddressclaim rbac. Looks like https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1893 assumed we were not using update and only using patch. Turns out update is still used during delete: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/controllers/vspherevm_ipaddress_reconciler.go#L222.

Fixes #

**Release note**:

```release-note
Add update to ipaddressclaim rbac
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -304,6 +304,7 @@ rules:
   - get
   - list
   - patch
+  - update
   - watch
 - apiGroups:
   - ipam.cluster.x-k8s.io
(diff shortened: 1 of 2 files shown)


christianang:
I think I can ignore the failure happening in verify-crds job, but let me know if there is anything I should do.

PR #1944, opened 2023-06-14
:seedling: Improve fd selection logic

rikatz:
As the cluster.spec.failureDomainSelector is a real K8s label selector, it can be used when listing the deployment zones directly.

This way, instead of having a full list to iterate and check if the label match on every interaction, we can be already sure the returned list contains only the deployment zones we care about 

Fixes #

**Release note**:
```release-note
NONE
```

### DIFF
--- a/controllers/vspherecluster_reconciler.go
+++ b/controllers/vspherecluster_reconciler.go
@@ -27,7 +27,6 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


rikatz:
ok I have no clue what's this all about :D

rikatz:
or maybe I do and I need to do it better on the options vs nil, will check again in a bit

rikatz:
/retest
will try once more and if this doesn't work will verify tomorrow what am I doing wrong :)

hack/tools/Makefile
```diff
@@ -99,7 +99,7 @@ golangci-lint: $(GOLANGCI_LINT)
 $(GOLANGCI_LINT): $(REPO_ROOT)/.github/workflows/golangci-lint.yaml
 	$(REPO_ROOT)/hack/ensure-golangci-lint.sh \
 		-b $(BIN_DIR) \
-		$(shell cat .github/workflows/golangci-lint.yaml | grep version | sed 's/.*version: //')
```
sbueringer, line 102: Let's drop this from this PR. This was changed on main to a totally different implementation
(also not sure if we can assume that yq is always available)

sbueringer:
@rikatz lgtm in general. Please drop the change to the Makefile then we should be ready for merge

rikatz:
@sbueringer sorry, missed the notification of this one.

Requested change is done :)

rikatz:
It is a valid error, I need to check it here

rikatz:
/retest

This is a flake :)

rikatz:
@randomvariable @sbueringer aaaand it's done, sorry for the delay on this one

Issue #1890, opened 2023-04-25
CAPV should respect both the Cluster's `spec.Paused` property and paused annotation consistently.


tylerschultz:
/kind bug

**What steps did you take and what happened:**

I noticed cluster paused detection is inconsistent when browsing code.

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/bac8a97e0566642c6710fc1e47a65cd63d4288ed/controllers/vspherevm_controller.go#L114-L126

The above mentioned predicate looks at the `spec.Paused` property of the watched `Cluster` during updates, and looks at the paused annotation during creates. It seems to me that the Predicate should watch both the `spec.Paused` property and the paused annotation in both cases. If either is true, for create or update, the predicate should return false.

I had a discussion with @srm09, and he points out that perhaps this predicate should be declared once in a reusable spot, allowing all controllers to skip reconciliation in a consistent way.

**What did you expect to happen:**
I expect the predicate to skip reconciling when the spec.Paused annotation is true, or the paused annotation is present for either Create or Update cases.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

schrej:
There is a util function in CAPI that covers both cases: [`annotations.IsPaused()`](https://pkg.go.dev/sigs.k8s.io/cluster-api/util/annotations#IsPaused)

lubronzhan:
Fixed

PR #1946, opened 2023-06-15
🐛 Use same check when checking cluster is pasued or not in the handler

lubronzhan:
Right now the CreateFunc and UpdateFunc use different way of checking whether a cluster is paused.

Now change it to annotations.IsPaused as a general way

Fixes #https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1890

**Release note**:

```release-note
Use same func to check whether cluster is pasued inside vspherevm_controller handler
```

### DIFF
--- a/controllers/vspherevm_controller.go
+++ b/controllers/vspherevm_controller.go
@@ -113,10 +113,9 @@ func AddVMControllerToManager(ctx *context.ControllerManagerContext, mgr manager
 		handler.EnqueueRequestsFromMapFunc(r.clusterToVSphereVMs),
 		predicate.Funcs{
 			UpdateFunc: func(e event.UpdateEvent) bool {
-				oldCluster := e.ObjectOld.(*clusterv1.Cluster)
 				newCluster := e.ObjectNew.(*clusterv1.Cluster)
 				// check whether cluster has either spec.paused or pasued annotation
-				return annotations.IsPaused(oldCluster, oldCluster) && !annotations.IsPaused(newCluster, newCluster)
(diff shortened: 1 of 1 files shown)


lubronzhan:
Why the verify-crds is failing.
I ran it locally and it's fine. And I only changed the code in the handler
```
config/default/crd/bases/infrastructure.cluster.x-k8s.io_vspherevms.yaml
10a11,12
>     categories:
>     - cluster-api

```

controllers/vspherevm_controller.go
```diff
@@ -115,13 +115,13 @@ func AddVMControllerToManager(ctx *context.ControllerManagerContext, mgr manager
 			UpdateFunc: func(e event.UpdateEvent) bool {
 				oldCluster := e.ObjectOld.(*clusterv1.Cluster)
 				newCluster := e.ObjectNew.(*clusterv1.Cluster)
-				return oldCluster.Spec.Paused && !newCluster.Spec.Paused
```
srm09, line 118: This is a discrepancy, we should check for the `paused` field (only `newCluster` for UpdateEvent) as well as the presence of the annotation.

srm09 (changes requested):
Added a comment, could you update the PR with the same?

controllers/vspherevm_controller.go
lubronzhan: Thanks!

lubronzhan:
Hi @srm09 could you review again? Thank you

PR #1951, opened 2023-06-19
🐛 Support non-multi-az cluster deployment in a multi-az enabled env

### LINKED ISSUE #1950: Creating a non-multi-az cluster in a multi-az enabled environment still forces the control plane nodes to be deployed into multiple AZs
/kind bug

1.  created a multi-az enabled management cluster with right VSphereFailureDomain and VSphereDeploymentZone settings
2. create a workload cluster with multi-az setting, everything works as expected, and nodes are placed in the correct AZs.
3. create another workload cluster without using multi-az setting, but the control plane nodes still gets placed in multi AZs.

creating a non-multi-az cluster in a multi-az enabled environment won't forces control plane nodes into multi AZs. 

Currently for control plane node zone selection, VSphereCluster has a labelSelector fort this, refer to [PR](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1894). Basically it supports two cases:
1) when it's a empty label selector, all zones will be selected.
2) If it's not empty and has labels set, so matched zones will be selected.

In the case we reported, we should support a null label selector, and no zones should be selected in this situation.

**Environment:**

zhanggbj:
Creating a non-multi-az cluster in a multi-az enabled environment still forces the control plane nodes to be deployed into multiple AZs.
In this PR, will support non-multi-az workload cluster deployment in a multi-az enabled management cluster by nil selector.

Fixes #1950

**Release note**:

```release-note
Support non-multi-az cluster deployment in a multi-az enabled env
```

### DIFF
--- a/controllers/vspherecluster_reconciler.go
+++ b/controllers/vspherecluster_reconciler.go
@@ -361,18 +361,22 @@ func (r clusterReconciler) reconcileDeploymentZones(ctx *context.ClusterContext)
 	failureDomains := clusterv1.FailureDomains{}
 	for _, zone := range deploymentZoneList.Items {
 		if zone.Spec.Server == ctx.VSphereCluster.Spec.Server {
-			// This should never fail, validating webhook should catch this first
-			if ctx.VSphereCluster.Spec.FailureDomainSelector != nil {
-				selector, err := metav1.LabelSelectorAsSelector(ctx.VSphereCluster.Spec.FailureDomainSelector)
-				if err != nil {
-					return false, errors.Wrapf(err, "zone label selector is misconfigured")
-				}
+			// If users are deploying a non-multi-az workload cluster in a multi-az enabled management cluster,
+			// FailureDomainSelector shouldn't be set, it would be a null label selector, and no zone should be
+			// selected
+			if ctx.VSphereCluster.Spec.FailureDomainSelector == nil {
+				continue
+			}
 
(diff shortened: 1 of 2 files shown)


zhanggbj:
CC @yastij @srm09 for awareness.

yastij (commented):
/approve

rikatz:
I think this makes sense from a CP perspective, where you may not want to use Failure domain. OTOH, this is a (maybe breaking?) change of current behavior.

Before the failureDomainSelector was introduced (https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1894), as soon as there was a failure domain on the environment, it would be selected and used by CP. 

Now it forces the user to set a failureDomainSelector, so previously people relying on "whatever the admin set I will use" will be broken/stuck on a single configuration. 

Again, I think this change is fine, but need to verify with people already relying on failure domain if this will be a breaking change for them, as there may be people around using failure domains for control plane that will not be using anymore as they now need to force this selector :)

yastij:
I think this is mostly a bug rather than a feature, as the selector was always a pointer in order to differentiate between `nil` and empty.

I will add a communication in the release note with an action required. I'll also add messaging on slack to let people know more broadly

/hold cancel

rikatz:
Yeah tbh it is going to make my life much easier :) 

Just wanted to callout the behavior change!

Thanks for this :)

Issue #1928, opened 2023-05-31
make verify-crds intermittently produces different output


randomvariable:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

If you run `make verify-crds` repeatedly, about 20% of the time, the generated CRDs are actually different, usually adding "cluster-api" as a category or setting an atomic map type.

**What did you expect to happen:**

Consistent CRD generation

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

chrischdi:
I can confirm that this is still the case on main.

chrischdi:
And also happens at updated controller-gen / controller-tools.

sbueringer:
We had a similar bug like this in a previous controller-tools release (but that one was fixed). Basically the problem was some non-determinism.

Unfortunately I can't find the old issue/PR anymore. But overall probably requires some deep-dive controller-gen debugging

sbueringer:
Probably was this one: https://github.com/kubernetes-sigs/controller-tools/issues/680

sbueringer:
Can you try controller-tools < v0.9.0 if it happens there?

chrischdi:
Note: I will debug into controller-gen to get down on this.

chrischdi:
I did take a further look. I can see three cases:

* case A: having `categories: cluster-api` set / not set on a top-level CRD object
* case B: having `x-kubernetes-map-type: atomic` set / not set at a property of type `object`. Often this is some kind of object reference (e.g. `bootstrapRef`).
* case C: adding/removing that the property `apiGroup` is `required`, this also seems to happen on object references, e.g. `addressFromPools`. At the same time, this may also have the case B.

Some more observations for the Case A:

* if one of the api versions does not define the `categories=cluster-api` marker, this flakyness happens. Controller-gen  seems to be flaky/non-deterministic in this case, maybe because of the random order of reading parsing the files (this is a guess)?
  * Fixable either by taking a deeper look into controller-gen to make the output deterministic
  * Or by adding the `categories=cluster-api` marker to every api version.

One more observation for all the cases:
* Output seems to be deterministic when only generating for v1beta1.

TODO: inspect more into the other cases, a workaround does not look as simple as for case A.

[...]

PR #1967, opened 2023-06-30
:sparkles: Extend Failure Domain API to support extra network configuration

### LINKED ISSUE #1964: Failure Domain network configuration should support similar fields of VMTemplate network config
/kind feature

Per [API](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/apis/v1beta1/types.go#L256) there is a set of configurations that can be applied for a Network device, not only the portgroup name.

When working on different failure domains, there may be different IP Pools (eg.: AWS you create subnetworks per a VPC region). 

Other configurations may differ as well, like DNS Server.

The original [Failure domain design proposal](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/docs/proposal/20201103-failure-domain.md#api-design) also proposed that the Network placement was added to the DeploymentZone (although today it is on failure domain, which is fine).

In fact, this API was [implemented](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/c77f5fb7dc6f8244c5d3d2adf1ea19825a14cabe/apis/v1beta1/vspheredeploymentzone_types.go#L63) but never used. There is no reference for it 

This feature request is to move forward with the implementation of per-failure domain network configuration

# Use cases
## Use case 1 - Different nameservers per Zone
Acme Ltd runs their Kubernetes clusters over a vSphere environment. This vSphere manages a set of clusters within the same datacenter. The concept of cluster, for Acme Ltd is an isolated room on the same datacenter, containing its own network resources: different network switches, different DNS servers

This way, Acme Ltd needs to be able to define different network configurations (DNS Server, routes, even NTP servers) per cluster/deployment zone

## Use case 2 - Different network requirements per zone
Acme Ltd is migrating its networking model. They understand that running Kubernetes control plane nodes with dynamic IPv4 addressing via DHCP may be risky for their operations, so they decided to establish IP pools per zones/clusters. As an example, nodes in zone 1 should receive the IPs 10.10.1.0/24, while in zone 2 should receive IPs 10.10.2.0/24. 
This migration needs to be nondisruptive, which means zone1 may already have IP Pool configured, but zone 2 may be still operating on DHCP, and at the proper moment, a new zone definition will be made with the right IP Pool, and the nodes on that zone migrated to this new zone definition.

This way, Acme Ltd needs to be able to define different Network models (DHCP vs IPAM/IPPool) per zone definition within the same datacenter

rikatz:
This PR is the initial work to extend failure domain API to support extra network configuration.

The idea is that extra network configuration can be established on failure domains, so different configurations (like DHCP, a different set of DNS servers) can be configured based on the selected region/zone

Fixes #1964

Some points to be discussed:
* Should we copy the networkspec struct or should we implement (as it is now) a different API?
* Can we drop/remove the dead code from Network struct (inside deploymentzone_types) and ignore the apidiff?
* Is there any e2e test that can be added here? I couldn't find any failure domain tests on integration or e2e

**Release note**:
```release-note
Introduce extra network configuration on Failure domain API
```


rikatz:
On the previous run: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/1967/pull-cluster-api-provider-vsphere-apidiff-main/1674828318085484544

Probably as a last commit, we should remove the dead Network struct and override this alert from apidiff.

apis/v1beta1/vspherefailuredomain_webhook.go
```diff
@@ -90,4 +90,15 @@ func (r *VSphereFailureDomain) Default() {
...
+	// Converts the old Networks field to NetworkConfigs
+	// TODO (@rkatz) - Can this be harmful for Gitops and other users that check spec and re-apply? Probably it will generate
+	// a difference between "what is expected" and "what we have" as there's going to be the additional "NetworkConfigs" field
+	//
+	if len(r.Spec.Topology.NetworkConfigs) == 0 && len(r.Spec.Topology.Networks) > 0 {
+		r.Spec.Topology.NetworkConfigs = make([]FailureDomainNetwork, len(r.Spec.Topology.Networks))
+		for i := range r.Spec.Topology.Networks {
+			r.Spec.Topology.NetworkConfigs[i].NetworkName = r.Spec.Topology.Networks[i]
+		}
+	}
```
chrischdi, line 103: I'm still ramping up on CAPV, but I think we should not convert this in the webhook to the other result.

IMHO we should:
* block if both are set (`NetworkConfigs` and `Networks`)
* Handle converting to `NetworkConfigs` internally in the controller to support both settings for now
* Handle migration to the new field, when introducing a new api version (via conversion webhook)

pkg/services/vimmachine.go
```diff
@@ -496,8 +496,17 @@ func (v *VimMachineService) generateOverrideFunc(ctx *context.VIMMachineContext)
...
+		// Some old cases may still use Networks instead of NetworkConfigs, so we should use the old fields
+		if len(vsphereFailureDomain.Spec.Topology.NetworkConfigs) == 0 && len(vsphereFailureDomain.Spec.Topology.Networks) > 0 {
+			vsphereFailureDomain.Spec.Topology.NetworkConfigs = make([]infrav1.FailureDomainNetwork, len(vsphereFailureDomain.Spec.Topology.Networks))
+			for i := range vsphereFailureDomain.Spec.Topology.Networks {
+				vsphereFailureDomain.Spec.Topology.NetworkConfigs[i].NetworkName = vsphereFailureDomain.Spec.Topology.Networks[i]
+			}
+		}
+
+		if len(vsphereFailureDomain.Spec.Topology.NetworkConfigs) > 0 {
+			vm.Spec.Network.Devices = overrideNetworkDeviceSpecs(vm.Spec.Network.Devices, vsphereFailureDomain.Spec.Topology.NetworkConfigs)
```
chrischdi, line 509: Together with the above, convert if required and use the converted value.
I would suggest to not change the original object so we don't by accident patch this change  back.

```suggestion
		networkConfigs := vsphereFailureDomain.Spec.Topology.NetworkConfigs
		// Convert the deprecated vsphereFailureDomain.Spec.Topology to networkConfigs if necessary.
		if len(networkConfigs) == 0 && len(vsphereFailureDomain.Spec.Topology.Networks) > 0 {
			networkConfigs = make([]infrav1.FailureDomainNetwork, len(vsphereFailureDomain.Spec.Topology.Networks))
			for i := range vsphereFailureDomain.Spec.Topology.Networks {
				networkConfigs[i].NetworkName = vsphereFailureDomain.Spec.Topology.Networks[i]
			}

		}

		if len(networkConfigs) > 0 {
			vm.Spec.Network.Devices = overrideNetworkDeviceSpecs(vm.Spec.Network.Devices, networkConfigs)
```

apis/v1beta1/vspherefailuredomain_webhook.go
rikatz: Sound good. I have added the conversion internally as well, but wanted to catch it early. Will change it here

pkg/services/vimmachine.go
rikatz: no problem :) will commit the change here and drop the webhook conversion, keeping only the validation of one or other field. Thanks for the review!

apis/v1beta1/vspherefailuredomain_webhook.go
rikatz: ok, dropped mutation, moved the validation to validate creation.
Internally the controller already does this conversion as well, as you pointed below :)

apis/v1beta1/vspherefailuredomain_types.go
```diff
@@ -88,6 +94,60 @@ type FailureDomainHosts struct {
...
+// this failure domain.
+// @rkatz - To be discussed with team, should we just embed NetworkDeviceSpec?
+type FailureDomainNetwork struct {
+	// NetworkName is the network name for this machine's VM.
+	NetworkName string `json:"name,omitempty"`
+
+	// @rkatz - The reason NetworkDeviceSpec is not being copied here, is because DHCP4 and DHCP6 are not pointers there
+	// This means that, in a case like "I want DHCP4 to be enabled on my template" and "I want DHCP4 to be disabled on this failure domain"
+	// we cannot verify if DHCP4 was unset (nil, we don't care about it on failure domain) vs set to false (I want to FORCE it to be false)
+	// DHCP4 is a flag that indicates whether or not to use DHCP for IPv4
```
chrischdi, line 107:
```suggestion
	// DHCP4 is a flag that indicates whether or not to use DHCP for IPv4
	// @rkatz - The reason NetworkDeviceSpec is not being copied here, is because DHCP4 and DHCP6 are not pointers there
	// This means that, in a case like "I want DHCP4 to be enabled on my template" and "I want DHCP4 to be disabled on this failure domain"
	// we cannot verify if DHCP4 was unset (nil, we don't care about it on failure domain) vs set to false (I want to FORCE it to be false)
```

Needs to get moved around to not break the godoc comment :-) (maybe you wanted to remove the note anyway?)

apis/v1beta1/vspherefailuredomain_types.go
```diff
@@ -88,6 +94,60 @@ type FailureDomainHosts struct {
 	HostGroupName string `json:"hostGroupName"`
 }
 
+// FailureDomainNetwork defines a network configuration that should be used when consuming
+// this failure domain.
+// @rkatz - To be discussed with team, should we just embed NetworkDeviceSpec?
+type FailureDomainNetwork struct {
```
chrischdi, line 100: Another point:

We may need to set unique identifiers to this to make it work with the CAPI contract for allowing it to be used with ClusterClass.

More information is here: https://main.cluster-api.sigs.k8s.io/developer/providers/migrations/v1.1-to-v1.2.html?highlight=listMapKey#required-api-changes-for-providers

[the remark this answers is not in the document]
rikatz: yeap, but let me do it on a different commit as I will have to rename it all around.

apis/v1beta1/vspherefailuredomain_types.go
rikatz: yeap, I wanted to remove :) just left it here to remember myself I wanted to discuss it

apis/v1beta1/vspherefailuredomain_types.go
rikatz: yeap, CAPV shouldn't be writing any information back to this structure. This structure should be configured just one time (in fact, the whole failure domain is immutable after created) so I think this is not needed, right?

rikatz:
Fyi, tested locally with the changes, it works fine.

Test executed:
* Created 3 Failure domains/deployment zones
* Each failure domain has a different portgroup, a different set of DNS, some with DHCPv6 disabled other enabled
* Could verify that on each node the right configuration was added to netplan

To be implemented:
* Verify the existence of the requested IPPool on the Failure Domain controller before setting the deployment zone as ready

config/default/crd/bases/infrastructure.cluster.x-k8s.io_vsphereclusters.yaml
```diff
@@ -321,7 +321,6 @@ spec:
                     description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
                     type: string
                 type: object
-                x-kubernetes-map-type: atomic
```
chrischdi, line 324: I think this should not have happened 🤔  may be related to #1928

controllers/vspheredeploymentzone_controller_domain.go
```diff
@@ -84,6 +84,16 @@ func (r vsphereDeploymentZoneReconciler) reconcileTopology(ctx *context.VSphereD
 		}
 	}
 
+	for _, networkConfig := range topology.NetworkConfigurations {
+		if networkConfig.NetworkName == "" {
```
chrischdi, line 88: Should we validate for `NetworkName` being set instead of silently ignoring the array entry?

apis/v1beta1/vspherefailuredomain_webhook.go
```diff
@@ -64,6 +64,10 @@ func (r *VSphereFailureDomain) ValidateCreate() error {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "Topology", "ComputeCluster"), fmt.Sprintf("cannot be nil if zone's Failure Domain type is %s", r.Spec.Zone.Type)))
 	}
 
+	if len(r.Spec.Topology.NetworkConfigurations) != 0 && len(r.Spec.Topology.Networks) != 0 {
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "Topology", "Networks"), "cannot be set if spec.Topology.NetworkConfigs is not empty"))
```
chrischdi, line 68:
```suggestion
		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "Topology", "Networks"), "cannot be set if spec.Topology.NetworkConfigs is already set"))
```

config/default/crd/bases/infrastructure.cluster.x-k8s.io_vsphereclusters.yaml
rikatz: Yeah, @srm09 pointed me to some issue on controller-gen IIRC, it keeps changing :)

controllers/vspheredeploymentzone_controller_domain.go
rikatz: My question actually is: should we make NetworkName required? Per https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1967/files#diff-2f616e5f5f40526609f3723e82cbcf01b16b9f240ea00fcdf3fc06b6aa6d164cR548 we just merge if it is not empty. This would allow a case where a failure domain has different nameservers, but same Portgroup name.

OTOH, there's no point IMO in validate a failure domain network config if the portgroup is not configured, so probably marking Network Name as required would be better IMHO.

WDYT?

chrischdi: If the other code which overwrites stuff from `NetworkConfig` to the `vmNetworkDeviceSpec` and is ok with not replacing the NetworkName, we should drop the if here and keep it non-mandatory.

If that is not a use case, (and I assume it isn't according your above comment), then we should mark the NetworkName as required so we ensure on api side already that it cannot be misconfigured (by not setting the NetworkName).

Happy to also hear thoughts from @yastij or others :-)

randomvariable:
/assign @lubronzhan 
for review on networking side.

@rikatz , can you rebase this onto main? this will also fix the CRD generation.

controllers/vspheredeploymentzone_controller_domain.go
lubronzhan: I think is better to make the networkName as a `key` inside the network array. Because if there is Multiple devices
 configured vsphereMachineTemplate, if the use forgot to specify the network name, it might accidentally configure Devices config on the vsphereMachineTemplate.

lubronzhan:
I think better to call out in the doc that the network devices specified in FailureDomain will conduct a `Merge` operation instead of `Overwrite` operation to the devices specified inside VSphereMachineTemplate

PR #1983, opened 2023-07-11
🐛 [release-1.6] Makefile: refactor how govc is built to workaround dependency issues building govc

chrischdi:

1. Solves issue on `release-1.6` jobs:

```
/home/prow/go/pkg/mod/github.com/vmware/govmomi@v0.23.1/sts/signer.go:38:2: github.com/google/uuid@v1.3.0: missing go.sum entry for go.mod file; to add it:
	go mod download github.com/google/uuid
```

Fixes CI for release-1.6 which is completely red.

Regarding 1.:

* On release-1.6 we import a very old version of `github.com/vmware/govmomi` in `hack/tools`.
* Together with the other dependencies defined in `hack/tools/go.mod`, `govc` (which is part of govmomi) cannot be built anymore because the dependencies refer different versions of `github.com/google/uuid`.
* Bumping `govmomi` is also not easily possible on `release-1.6` because this would also bump `golangci-lint` and result in not being able to build `govc` too.

Because of the above issues and to align with the core CAPI CI tooling, this PR introduces building `govc` in the same version as used in the top level `go.mod` without using `hack/tools/go.mod`, but the `hack/go-install.sh` script instead.

**Release note**:

```release-note

```


chrischdi:
awesome.

We now hit the same issue as on release-1.5 😄

chrischdi:
Note: the follow-up issue gets solved in #1984, which is this one:
```

Note: this PR is required to fix CI which is red due to:

Failed to run clusterctl config cluster
Unexpected error:
    <*yamlprocessor.errMissingVariables | 0xc000ccc090>: {
        Missing: [
            "VSPHERE_TLS_THUMBPRINT",
        ],
    }
    value for variables [VSPHERE_TLS_THUMBPRINT] is not set. Please set the value using os environment variables or the clusterctl config file
occurred
/home/prow/go/pkg/mod/sigs.k8s.io/cluster-api/test@v1.2.2/framework/clusterctl/client.go:216}


```

chrischdi:
Still needs a lgtm :-)

PR #2015, opened 2023-07-17
🌱 Drop redundant list webhook registrations

sbueringer:
This PR drops the registration of List webhooks. Registering webhooks for List types is a no-op as the list types do not implement validation / default interfaces (it's also not necessary as the webhooks only have to be implemented on the non-list types)

Fixes #

**Release note**:

```release-note

```


killianmuldoon (commented):
/lgtm

PR #2022, opened 2023-07-18
🌱 Bump Go to 1.20.6

chrischdi:

Bumps Go to the latest supported 1.20.6 to be on the same version as CAPI and k8s v1.27 / v1.28

Fixes #

**Release note**:

```release-note
Bump Go to 1.20.6
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 # Build the manager binary
-ARG GOLANG_VERSION=golang:1.19.3
+ARG GOLANG_VERSION=golang:1.20.6
 FROM --platform=${BUILDPLATFORM} ${GOLANG_VERSION} as builder
 WORKDIR /workspace
 
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ SHELL := /usr/bin/env bash
 
 VERSION ?= $(shell cat clusterctl-settings.json | jq .config.nextVersion -r)
 
-GO_VERSION ?=1.19.3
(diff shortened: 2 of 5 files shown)


sbueringer:
Changes look okay to me. But CI doesn't agree :).

I wonder if we also have to bump the go version in ProwJobs before we can merge this PR

.github/workflows/golangci-lint.yaml
```diff
@@ -12,11 +12,13 @@ jobs:
...
+      - name: Set up Go
+        uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # tag=v4.0.1
         with:
-          go-version: 1.19
+          go-version: ${{ steps.vars.outputs.go_version }}
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3.6.0
-        with:
-          version: v1.52.2
+        uses: golangci/golangci-lint-action@639cd343e1d3b897ff35927a75193d57cfcba299 # tag=v3.6.0
```
sbueringer, line 24: I think we should still pin the golangci-lint version

sbueringer: It's probably already part of what you're planning to do. Let's also get rid of the shell script to install golangci-lint and align to core CAPI

chrischdi: Let me take bumping the linter out to a separate PR :-)

sbueringer: Absolutely!

PR #2023, opened 2023-07-18
:bug: Return error on session creation if VCentre client not used

killianmuldoon:
Return an error if a non-VCentre client is used. The current behaviour for ESXi is not implemented.

```release-note
NONE
```


pkg/services/govmomi/create.go
```diff
@@ -17,18 +17,18 @@ limitations under the License.
...
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/vcenter"
 )
 
 // createVM creates a new VM with the data in the VMContext passed. This method does not wait
 // for the new VM to be created.
 func createVM(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Format) error {
-	if ctx.Session.IsVC() {
-		return vcenter.Clone(ctx, bootstrapData, format)
+	if !ctx.Session.IsVC() {
+		return errors.Errorf("expected VCentre client got %v", ctx.Session.ServiceContent.About.ApiType)
```
killianmuldoon, line 31: This could also be returned as an error in session `GetOrCreate`, but I don't know about stability guarantees or likely external usage of that function.

killianmuldoon: IsVC() appears to only be used here.

pkg/services/govmomi/esxi/clone.go
```diff
@@ -1,29 +0,0 @@
...
-
-import (
-	"github.com/pkg/errors"
-	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
-
-	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
-)
-
-// Clone kicks off a clone operation on ESXi to create a new virtual machine.
-func Clone(_ *context.VMContext, _ []byte, _ bootstrapv1.Format) error {
```
killianmuldoon, line 27: Again not sure of stability guarantees - but removing this one function that only returned an error seems non-disruptive. If there is some API rule for go packages documented somewhere that would be good to see.

randomvariable:
Think this is fine. I don't have the override permissions. 
/assign @srm09 @yastij 

/lgtm

pkg/services/govmomi/esxi/clone.go
sbueringer: I think it's fine

pkg/services/govmomi/create.go
sbueringer: I think the change is fine as is

sbueringer, line 31:
```suggestion
		return errors.Errorf("expected VCenter client got %v", ctx.Session.ServiceContent.About.ApiType)
```
sounds french otherwise :D

sbueringer:
@randomvariable "pull-cluster-api-provider-vsphere-apidiff-main"  is optional, no override needed

pkg/services/govmomi/create.go
killianmuldoon: Also English :facepalm: going to be making this mistake a bund.

sbueringer:
Okay I think we now have two flakes which just occur all the time.

/retest
/lgtm

PR #2036, opened 2023-07-19
🌱 Rename findVMPre7 to findVSphereVM

sbueringer:
Just want to rename the func to make it clear this works independent of the vCenter version.

The only restriction we have is that it is for the non-supervisor mode but that is clear because the func is a method on VimMachineService

Fixes #

**Release note**:

```release-note

```


sbueringer:
/retest

The IP thing is definitely frequently occuring now

PR #2041, opened 2023-07-20
🐛 hack/e2e.sh: retry getting ipaddress object name

chrischdi:

Lot's of e2e CI jobs failed fast with the following error:

```
ipclaim.ipam.metal3.io/ip-claim-0e869fd5082c7413eaadc42d54b26bb27569c203 created
Acquired Control Plane IP: 192.168.6.254
ipclaim.ipam.metal3.io/workload-ip-claim-81ba2a87a03045ce899fe07df55d5f17ae6ce671 created
error: resource name may not be empty
```

`error: resource name may not be empty` comes from trying to get a resource without a name, e.g. `kubectl get pod ""`.

In our `hack/e2e.sh` script this could happen at two places while claiming IP addresses:

When `IPADDRESS_NAME` or `WORKLOAD_IPADDRESS_NAME` are an empty string and using them in one of the following occurencies:

* https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/a9ae01e7a505712399957bce3a20b29f59038df1/hack/e2e.sh#L87

* https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/a9ae01e7a505712399957bce3a20b29f59038df1/hack/e2e.sh#L95

This could happen if the metal3 ipaddress controller did not yet update the status field of the ipclaim, where we read the ipaddress name from.

This PR fixes this by retrying:
* for up to five times
* with a backoff of 1s

to get a non-empty value for the ipaddress name.
It would also output a better error in case of still not being able to read a name from the ipclaim after the timeout.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -78,12 +78,34 @@ function append_api_group() {
   echo "${resource}.ipam.metal3.io"
 }
 
+function kubectl_get_jsonpath() {
+  local OBJECT_KIND="${1}"
+  local OBJECT_NAME="${2}"
+  local JSON_PATH="${3}"
+  local n=0
+  until [ $n -ge 30 ]; do
+    OUTPUT=$(kubectl --kubeconfig="${KUBECONFIG}" get "$(append_api_group "${OBJECT_KIND}")" "${OBJECT_NAME}" -o=jsonpath="${JSON_PATH}")
+    if [[ "${OUTPUT}" != "" ]]; then
+      break
+    fi
+    n=$((n + 1))
+    sleep 1
+  done
+
+  if [[ "${OUTPUT}" == "" ]]; then
+    echo "Received empty output getting ${JSON_PATH} from ${OBJECT_KIND}/${OBJECT_NAME}" 1>&2
+    return 1
+  else
+    echo "${OUTPUT}"
+    return 0
+  fi
+}
+
 # Retrieve an IP to be used as the kube-vip IP
 KUBECONFIG="/root/ipam-conf/capv-services.conf"
 IPCLAIM_NAME="ip-claim-$(openssl rand -hex 20)"
 sed "s/IPCLAIM_NAME/${IPCLAIM_NAME}/" "${REPO_ROOT}/hack/ipclaim-template.yaml" | kubectl --kubeconfig=${KUBECONFIG} create -f -
-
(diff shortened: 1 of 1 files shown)


chrischdi:
Ah, shellcheck did already run 🎉

sbueringer:
A bit ironic, golangci-lint
```
 Running [/home/runner/golangci-lint-1.53.3-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
```

chrischdi:
> A bit ironic, golangci-lint
> 
> ```
>  Running [/home/runner/golangci-lint-1.53.3-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
>   level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
> ```

retriggered by renaming PR 😂

hack/e2e.sh
```diff
@@ -78,20 +78,42 @@ function append_api_group() {
   echo "${resource}.ipam.metal3.io"
 }
 
+function kubectl_get_jsonpath() {
+  local OBJECT_KIND="${1}"
+  local OBJECT_NAME="${2}"
+  local JSON_PATH="${3}"
+  local n=0
+  until [ $n -ge 5 ]; do
```
sbueringer, line 86: Should we use more retries. Maybe 10 or 30? Just to avoid another itreation because 5s is sometimes not enough

sbueringer:
Thank you very much !!

/lgtm

hack/e2e.sh
chrischdi: I'll bump to 30 retries.

Normally it should be pretty fast.

randomvariable:
thx for the fix. at some point it'd be good to move to this to [ginkgo.SynchronizedBeforeSuite like there is in the CAPA repo](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/suites/conformance/conformance_suite_test.go#L45C13-L46) instead of bash.

/approve

randomvariable:
Filed https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/2044

PR #2055, opened 2023-07-20
🌱 Use pointer receiver consistently in vspheremachine controller

sbueringer:
In general it is recommended to consistently use the same receiver type. I think in this case (as usually in all reconcilers) we should use a pointer receiver.

Otherwise the machineReconciler struct is getting copied on method calls.

Fixes #

**Release note**:

```release-note

```


chrischdi (approved):
/lgtm

PR #2071, opened 2023-07-22
✨: Add guestinfo.hostname to config of a VM

### LINKED ISSUE #2013: `Guestinfo.hostname` field in vCenter not set when using ignition for VMs
/kind feature

I am using `DHCP4`  within my vSphereMachineTemplate and the vSphereVM does not have a DNSName set in the vCenter (has localhost.localdomain) and I have to override that myself.

I expect my newly created VM to have a DNSName set in the vCenter dashboard - the name of the machine itself.

I thought it might be appropriate to set the `guestinfo.hostname` extraConfig while reconciling the metadata here:
 - https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/pkg/services/govmomi/extra/config.go#L55

The necessary information is already present in the metadata under field `local-hostname`.

**Environment:**

- Cluster-api-provider-vsphere version: v1.7.0
- Kubernetes version: (use `kubectl version`): --- 
- OS (e.g. from `/etc/os-release`):  ---

odvarkadaniel:
The VMs in the vCenter are currently without a DNSName (they don't have a config data at key guestinfo.hostname) while using an ignition to set the VM up.  With setting this field up, we can also append an internal DNS address to the list of machine's addresses.

Fixes #2013

**Release note**:

```release-note
Set a DNSName for VMs to a corresponding machine's name.
```

### DIFF
--- a/controllers/vspheremachine_controller.go
+++ b/controllers/vspheremachine_controller.go
@@ -124,7 +124,7 @@ func AddMachineControllerToManager(ctx *context.ControllerManagerContext, mgr ma
 		).
 		WithOptions(controller.Options{MaxConcurrentReconciles: ctx.MaxConcurrentReconciles})
 
-	r := machineReconciler{
+	r := &machineReconciler{
 		ControllerContext: controllerContext,
 		VMService:         &services.VimMachineService{},
 		supervisorBased:   supervisorBased,
@@ -179,7 +179,7 @@ type machineReconciler struct {
 }
 
 // Reconcile ensures the back-end state reflects the Kubernetes resource state intent.
-func (r machineReconciler) Reconcile(_ goctx.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
+func (r *machineReconciler) Reconcile(_ goctx.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
 	var machineContext context.MachineContext
 	logger := r.Logger.WithName(req.Namespace).WithName(req.Name)
 	logger.V(3).Info("Starting Reconcile VSphereMachine")
@@ -259,7 +259,7 @@ func (r machineReconciler) Reconcile(_ goctx.Context, req ctrl.Request) (_ ctrl.
 	return r.reconcileNormal(machineContext)
 }
 
-func (r machineReconciler) reconcileDelete(ctx context.MachineContext) (reconcile.Result, error) {
+func (r *machineReconciler) reconcileDelete(ctx context.MachineContext) (reconcile.Result, error) {
 	ctx.GetLogger().Info("Handling deleted VSphereMachine")
 	conditions.MarkFalse(ctx.GetVSphereMachine(), infrav1.VMProvisionedCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
 
@@ -277,7 +277,7 @@ func (r machineReconciler) reconcileDelete(ctx context.MachineContext) (reconcil
 	return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 }
 
-func (r machineReconciler) reconcileNormal(ctx context.MachineContext) (reconcile.Result, error) {
+func (r *machineReconciler) reconcileNormal(ctx context.MachineContext) (reconcile.Result, error) {
 	machineFailed, err := r.VMService.SyncFailureReason(ctx)
 	if err != nil && !apierrors.IsNotFound(err) {
 		return reconcile.Result{}, err
--- a/pkg/services/vimmachine.go
+++ b/pkg/services/vimmachine.go
@@ -318,6 +318,10 @@ func (v *VimMachineService) reconcileNetwork(ctx *context.VIMMachineContext, vm 
 				Address: addr,
 			})
 		}
+		machineAddresses = append(machineAddresses, clusterv1.MachineAddress{
+			Type:    clusterv1.MachineInternalDNS,
+			Address: vm.GetName(),
+		})
 		ctx.VSphereMachine.Status.Addresses = machineAddresses
 	}
 


randomvariable:
I wouldn't mind some additional eyes from

/assign @akutz

akutz:
> I wouldn't mind some additional eyes from
> 
> 
> 
> /assign @akutz 

Is this CAPV as a stand-alone management cluster?

Also, please see https://github.com/coreos/afterburn/pull/888 as it will likely be useful in the future.

akutz:
/hold

Please see https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2071#discussion_r1274865028. Changes needed.

odvarkadaniel:
> > I wouldn't mind some additional eyes from
> > /assign @akutz
> 
> Is this CAPV as a stand-alone management cluster?
> 
> Also, please see [coreos/afterburn#888](https://github.com/coreos/afterburn/pull/888) as it will likely be useful in the future.

Do I get that right that I will need to get the necessary data from the newly introduced netplan config file which will be stored as guestinfo.metadata?

odvarkadaniel:
Hi, I just came back from PTO.

I have chatted with folks from flatcar forum and I don't really understand what needs to be done here in order for this to get merged as this file exists and I suspect [this](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/de5ee9982c677acc30388e19c88b579ea9e4b3aa/pkg/util/constants.go#L19) is going to be plugged in as `guestinfo.metadata` and the field `local-hostname` is still going to be used to get the data necessary for this change.

As of right now we just take the metadata file from `guestinfo.metadata` and parse the `local-hostname` value out of it and assign it to `guestinfo.hostname` to see it on the VM in vCenter, which, as far as I understand, stays the same.

Thank you for clarifying this situation in case I am missing something.

chrischdi:
/assign @randomvariable

for helping to clarify. Thanks!

jepio:
> > Do I get that right that I will need to get the necessary data from the newly introduced netplan config file which will be stored as guestinfo.metadata?
> 
> Yup that's correct. Given the complexity of networking on-prem environments, we decided that netplan had the required expressivity, and we'll be encouraging all bootstrap providers to adopt it. I would sync up with Flatcar counterparts in Ignition.

So if I understand this PR and #2013 correctly - @odvarkadaniel issue is not about setting the hostname of the guest, but about exposing that value through vCenter.

Given that we are aligned on using `guestinfo.metadata` to pass cloud-init/netplan yaml (with the `local-hostname` field) on the CAPV side, will vCenter be able to handle that? As in: will vCenter parse out the hostname from that location and display it in the webui or does the guest have an interface to push it's hostname to vCenter?

odvarkadaniel:
> Sorry, but I can't follow. Let's forget `guestinfo.hostname` for a second. What usecase is currently broken for you? `guestinfo.metadata.local_hostname` is where CAPV places the hostname right now. Cloud-init processes it and sets the VM hostname to that value. The Flatcar template currently has a systemd service that performs that operation, and in future this will be handled by afterburn.

Ah, totally forgot to mention that this is for Ignition, sorry. The ignition bootstrapping method currently does not set the hostname on the VM as cloud-init does. This just adds the same thing as cloud-init does, but for Ignition.

We are running into an issue, where MachineSet provisions a Machine and a new VM gets created in the vCenter, but never gets ignited due to a failure cause by empty hostname.

jepio:
Do you have an error message/log that shows what's going on? As far as I
know neither ignition, nor afterburn care about the value of
`guestinfo.hostname`.

Flatcar also uses ignition and in CAPV it has a systemd service defined in
ignition data that sets the node hostname. Without this `kubeadm join`
doesn't work, but ignition itself does. Do you use the same ignition
template from this repo? Would that approach work for your case as well?

odvarkadaniel:
> Do you have an error message/log that shows what's going on? As far as I know neither ignition, nor afterburn care about the value of `guestinfo.hostname`. Flatcar also uses ignition and in CAPV it has a systemd service defined in ignition data that sets the node hostname. Without this `kubeadm join` doesn't work, but ignition itself does. Do you use the same ignition template from this repo? Would that approach work for your case as well?

I have been testing this with our openshift downstream version of CAPV (which should be identical to upstream) with our ignition config files. We do stuff a little bit different downstream, which comes to this issue, I thought could also be useful adding upstream (assign value to `guestinfo.hostname` for ignition from CAPV), so this is not a bug in here, just a feature that would help us downstream work with MachineSets.

jepio:
Do you have a link to the downstream code that uses/depends on
`guestinfo.hostname`?

odvarkadaniel:
Sorry for confusion, it's been a long time since I worked on this.

I think that our machine-approver operator is preventing the machine from getting created as the newly created machine is currently missing an `InternalDNS` name ([link](https://github.com/openshift/cluster-machine-approver/blob/master/pkg/machinehandler/machinehandler.go#L130-L139)). This piece of code is also enforcing us to have the `InternalDNS` name set to the machine's name, which is then also used in the ignition config file. CAPV is already setting the InternalDNS for cloud-init already, but not for ignition as far as I remember.

jepio:
So it sounds like you really want this part of the PR merged:
```patch
+		machineAddresses = append(machineAddresses, clusterv1.MachineAddress{
+			Type:    clusterv1.MachineInternalDNS,
+			Address: ctx.Machine.Name,
+		})
```

odvarkadaniel:
> So it sounds like you really want this part of the PR merged:
> 
> ```diff
> +		machineAddresses = append(machineAddresses, clusterv1.MachineAddress{
> +			Type:    clusterv1.MachineInternalDNS,
> +			Address: ctx.Machine.Name,
> +		})
> ```

Yes, that seems to fix this issue for us. Didn't realise at first that upstream's ignition template already has a service that sets the hostname on the VM.

jepio:
I think that part should be OK to merge. Would you agree @akutz ?

@odvarkadaniel : in the meantime, are you able to help or find someone at Red Hat who could take care of the points blocking https://github.com/coreos/afterburn/pull/888#issuecomment-1659509299 from getting merged? Those are: reviews and packaging dependencies in Fedora. Then we can build on that PR to also handle the `local-hostname` field.

odvarkadaniel:
> I think that part should be OK to merge. Would you agree @akutz ?
> 
> @odvarkadaniel : in the meantime, are you able to help or find someone at Red Hat who could take care of the points blocking [coreos/afterburn#888 (comment)](https://github.com/coreos/afterburn/pull/888#issuecomment-1659509299) from getting merged? Those are: reviews and packaging dependencies in Fedora. Then we can build on that PR to also handle the `local-hostname` field.

We are discussing this in the original issue linked to this PR.

Of course, I will ask about the coreos/afterburner PR.

odvarkadaniel:
@jepio, Just to update you on the afterburner PR - I talked to folks from coreos team and they are currently not working on it nor on their radar as this is not a priority for them currently.

jepio:
> @jepio, Just to update you on the afterburner PR - I talked to folks from coreos team and they are currently not working on it nor on their radar as this is not a priority for them currently.

I see. That afterburn PR would help CAPV cover more scenarios, which would also benefit Openshift.

PR #2099, opened 2023-07-27
🐛 Refine finalizer handling

### LINKED ISSUE #2039: Improve finalizer handling
/kind bug

I think we have to improve our finalizer handling in CAPV.

The ideal behavior in Reconcile is the following:
* if deletionTimestamp is set
  * => do nothing
* if deletionTimestamp is unset
  * if finalizer is set
    * => do nothing
  * if finalizer is unset
    * => set finalizer and immediately requeue

I recently had a PR to fix this everywhere in core CAPI: https://github.com/kubernetes-sigs/cluster-api/pull/8949

Reasoning:
* if we don't immediately requeue we potentially create resources in vSphere that would be orphaned if the same resource is deleted before we can set the finalizer in defer
* if we try to set the finalizer when the deletionTimestamp is already set we just get an error from the apiserver because this operation is not allowed

An easy example how this should look like can be found here: https://github.com/sbueringer/cluster-api/blob/cef1cf1d9312088a785c520f49970b1c609ab9ad/internal/controllers/cluster/cluster_controller.go#L145-L158

Note: it's important that the defer patch is done in a way that the finalizer is actually written

**Environment:**

zhanggbj:

- Add finalizer if not existed
- Requeue immediately after adding finalizer to avoid the race condition between init and delete
Please find more details in #2039 description.

Fixes #2039 

**Release note**:

```release-note
Refine finalizer handling
```

### DIFF
--- a/controllers/vmware/vspherecluster_reconciler.go
+++ b/controllers/vmware/vspherecluster_reconciler.go
@@ -121,6 +121,13 @@ func (r ClusterReconciler) Reconcile(_ goctx.Context, req ctrl.Request) (_ ctrl.
 		return reconcile.Result{}, nil
 	}
 
+	// If the VSphereCluster doesn't have our finalizer, add it.
+	// Requeue immediately after adding finalizer to avoid the race condition between init and delete
+	if !controllerutil.ContainsFinalizer(vsphereCluster, vmwarev1.ClusterFinalizer) {
+		controllerutil.AddFinalizer(vsphereCluster, vmwarev1.ClusterFinalizer)
+		return ctrl.Result{}, nil
+	}
+
 	// Handle non-deleted clusters
 	return ctrl.Result{}, r.reconcileNormal(clusterContext)
 }
@@ -146,9 +153,6 @@ func (r *ClusterReconciler) reconcileDelete(ctx *vmware.ClusterContext) {
 
 func (r *ClusterReconciler) reconcileNormal(ctx *vmware.ClusterContext) error {
 	ctx.Logger.Info("Reconciling vsphereCluster")
-
-	// If the vsphereCluster doesn't have our finalizer, add it.
-	controllerutil.AddFinalizer(ctx.VSphereCluster, vmwarev1.ClusterFinalizer)
 
 	// Get any failure domains to report back to the CAPI core controller.
 	failureDomains, err := r.getFailureDomains(ctx)
--- a/controllers/vspherecluster_reconciler.go
+++ b/controllers/vspherecluster_reconciler.go
@@ -127,6 +127,13 @@ func (r clusterReconciler) Reconcile(_ goctx.Context, req ctrl.Request) (_ ctrl.
 		return r.reconcileDelete(clusterContext)
 	}
 
+	// If the VSphereCluster doesn't have our finalizer, add it.
+	// Requeue immediately after adding finalizer to avoid the race condition between init and delete
+	if !ctrlutil.ContainsFinalizer(vsphereCluster, infrav1.ClusterFinalizer) {
+		ctrlutil.AddFinalizer(vsphereCluster, infrav1.ClusterFinalizer)
+		return reconcile.Result{}, nil
+	}
+
 	// Handle non-deleted clusters
 	return r.reconcileNormal(clusterContext)
 }
@@ -216,9 +223,6 @@ func (r clusterReconciler) reconcileDelete(ctx *context.ClusterContext) (reconci
 
 func (r clusterReconciler) reconcileNormal(ctx *context.ClusterContext) (reconcile.Result, error) {
 	ctx.Logger.Info("Reconciling VSphereCluster")
-
-	// If the VSphereCluster doesn't have our finalizer, add it.
-	ctrlutil.AddFinalizer(ctx.VSphereCluster, infrav1.ClusterFinalizer)
 
 	ok, err := r.reconcileDeploymentZones(ctx)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


zhanggbj:
This is ready for a code review. 
golangci-lint is failing due to timeout, still checking, probably same issue as described here?https://kubernetes.slack.com/archives/CKFGK3SSD/p1690291588477029

Also CC @sbueringer for a review

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -121,6 +121,13 @@ func (r ClusterReconciler) Reconcile(_ goctx.Context, req ctrl.Request) (_ ctrl.
 		return reconcile.Result{}, nil
 	}
 
+	// If the VSphereCluster doesn't have our finalizer, add it.
+	// Requeue immediately after adding finalizer to avoid the race condition between init and delete
+	if !controllerutil.ContainsFinalizer(vsphereCluster, vmwarev1.ClusterFinalizer) {
```
chrischdi, line 126: Nit/optional: we could move this up a bit, above the `EmitEvent` so "just adding the finalizer" does not emit an event. (Note: we would have to add `vsphereCluster.DeletionTimestamp.IsZero()` to the if, if we are above the deletion part)

controllers/vspherecluster_reconciler.go
```diff
@@ -127,6 +127,13 @@ func (r clusterReconciler) Reconcile(_ goctx.Context, req ctrl.Request) (_ ctrl.
 		return r.reconcileDelete(clusterContext)
 	}
 
+	// If the VSphereCluster doesn't have our finalizer, add it.
+	// Requeue immediately after adding finalizer to avoid the race condition between init and delete
+	if !ctrlutil.ContainsFinalizer(vsphereCluster, infrav1.ClusterFinalizer) {
```
chrischdi, line 132: Nit/Optional: Maybe we should move this above L121 (where we call `setOwnerRefsOnVsphereMachines(...)`), because `setOwnerRefsOnVsphereMachines` may patch vspheremachines and on errors we won't have yet set the finalizer. (Note: we would have to add vsphereCluster.DeletionTimestamp.IsZero() to the if, if we are above the deletion part).

However, I think we are fine as is too.

chrischdi (commented):
Just two small optional nits / to discuss what's better, as of that lgtm :-)

/lgtm

controllers/vmware/vspherecluster_reconciler.go
sbueringer: I think let's leave it here. we need the defer patch to set the finalizer

controllers/vspheredeploymentzone_controller.go
```diff
@@ -147,12 +147,17 @@ func (r vsphereDeploymentZoneReconciler) Reconcile(ctx goctx.Context, request re
 		return ctrl.Result{}, r.reconcileDelete(vsphereDeploymentZoneContext)
 	}
 
+	// If the VSphereDeploymentZone doesn't have our finalizer, add it.
+	// Requeue immediately after adding finalizer to avoid the race condition between init and delete
+	if !ctrlutil.ContainsFinalizer(vsphereDeploymentZone, infrav1.DeploymentZoneFinalizer) {
+		ctrlutil.AddFinalizer(vsphereDeploymentZoneContext.VSphereDeploymentZone, infrav1.DeploymentZoneFinalizer)
```
sbueringer, line 153:
```suggestion
		ctrlutil.AddFinalizer(vsphereDeploymentZone, infrav1.DeploymentZoneFinalizer)
```
nit: just for consistency with the line above

controllers/vspherevm_controller.go
```diff
@@ -279,6 +279,13 @@ func (r vmReconciler) Reconcile(ctx goctx.Context, req ctrl.Request) (_ ctrl.Res
 		}
 	}
 
+	// If the VSphereVM doesn't have our finalizer, add it.
+	// Requeue immediately to avoid the race condition between init and delete
+	if !ctrlutil.ContainsFinalizer(vsphereVM, infrav1.VMFinalizer) {
+		ctrlutil.AddFinalizer(vsphereVM, infrav1.VMFinalizer)
+		return reconcile.Result{}, nil
+	}
+
```
sbueringer, line 288:
```suggestion
	if vsphereVM.ObjectMeta.DeletionTimestamp.IsZero() {
		// If the VSphereVM doesn't have our finalizer, add it.
		// Requeue immediately to avoid the race condition between init and delete
		if !ctrlutil.ContainsFinalizer(vsphereVM, infrav1.VMFinalizer) {
			ctrlutil.AddFinalizer(vsphereVM, infrav1.VMFinalizer)
			return reconcile.Result{}, nil
		}
	}
```
In this case we have to add the if to ensure we only do this if the deletionTimestamp is not set, as the check for deletionTimestamp happens within reconcile

controllers/vspherecluster_reconciler.go
sbueringer: I think it won't make much of a difference. The main goal with the finalizer is that we set it before the current reconciler creates anything for the current object

sbueringer: Maybe kepp it here as it's slightly simpler

randomvariable:
There's two suggested changes from @sbueringer . So will await the resolution of those

controllers/vspheredeploymentzone_controller.go
zhanggbj: Will fix it next patch

controllers/vspherevm_controller.go
zhanggbj: Will fix it in next patch

zhanggbj:
Thanks @chrischdi and @sbueringer for your review! All comments are addressed.

chrischdi (commented):
/lgtm

sbueringer:
Thx!

/lgtm

/assign @randomvariable 
for approval

PR #2104, opened 2023-07-28
🌱 align flags with CAPI

chrischdi:
**What this PR does / why we need it**:

Aligns flags to the ones used in CAPI:

- adds the flag `--leader-elect-lease-duration`
- adds the flag `--leader-elect-renew-deadline`
- adds the flag `--leader-elect-retry-period`
- adds the flag `--contention-profiling`
- adds the flag `--kube-api-qps`
- adds the flag `--kube-api-burst`
- adds the flag `--webhook-cert-dir`
- replaces the profiler webserver by the controller-runtime provided one in `main.go`

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2072

**Special notes for your reviewer**:

Detailed diff in help output:

```diff
0a1
>       --contention-profiling                     Enable block profiling, if profiler-address is set.
8a10,11
>       --kube-api-burst int                       Maximum number of queries that should be allowed in one burst from the controller client to the Kubernetes API server. Default 30 (default 30)
>       --kube-api-qps float32                     Maximum queries per second from the controller client to the Kubernetes API server. Defaults to 20 (default 20)
10a14,16
>       --leader-elect-lease-duration duration     Interval at which non-leader candidates will wait to force acquire leadership (duration string) (default 15s)
>       --leader-elect-renew-deadline duration     Duration that the leading controller manager will retry refreshing leadership before giving up (duration string) (default 10s)
>       --leader-elect-retry-period duration       Duration the LeaderElector clients should wait between tries of actions (duration string) (default 2s)
16d21
22c27,29
<       --sync-period duration                     The interval at which cluster-api objects are synchronized (default 10m0s)
---
>       --sync-period duration                     The minimum interval at which watched resources are reconciled (e.g. 15m) (default 10m0s)
29a37,43
>       --webhook-cert-dir string                  Webhook cert dir, only used when webhook-port is specified. (default "/tmp/k8s-webhook-server/serving-certs/")
```
_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
Align flags to core CAPI.
```

PR #2105, opened 2023-07-28
🌱 Introduce watch-filter flag as predicate filter + filter out paused objects

### LINKED ISSUE #2038: Support watch-filter flag
/kind feature

I think CAPV should support the watch-filter flag like core CAPI.

It's also part of the contract:
```md
In order to make it possible for users to deploy multiple instances of the same provider:

Providers MUST support the --namespace flag in their controllers.
Providers MUST support the --watch-filter flag in their controllers.
```

https://github.com/kubernetes-sigs/cluster-api/blob/main/main.go#L162-L163

**Environment:**

chrischdi:

Adds the flag `--watch-filter` and uses the predicate filter `predicates.ResourceNotPausedAndHasFilterLabel` from CAPI on every controller.

```
      --watch-filter string                    Label value that the controller watches to reconcile cluster-api objects. Label key is always cluster.x-k8s.io/watch-filter. If unspecified, the controller watches for all cluster-api objects.
```

Fixes #2038 

Requires #2104 to merge first

**Release note**:

```release-note
Introduce watch-filter flag as predicate filter + filter out paused objects.
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,6 +35,7 @@ import (
 	logsv1 "k8s.io/component-base/logs/api/v1"
 	_ "k8s.io/component-base/logs/json/register"
 	"k8s.io/klog/v2"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/controllers/remote"
 	"sigs.k8s.io/cluster-api/util/flags"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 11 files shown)


pkg/manager/options.go
```diff
@@ -49,10 +49,6 @@ type Options struct {
 	//
 	// Defaults to the eponymous constant in this package.
 	MaxConcurrentReconciles int
```
sbueringer, line 51:
```suggestion
	MaxConcurrentReconciles int

```
nit

sbueringer: Also looks like we should fix the godoc comment of PodNamespace. Doesn't start with "PodNamespace"

main.go
```diff
@@ -69,6 +70,7 @@ var (
 	syncPeriod                  time.Duration
 	webhookOpts                 webhook.Options
 	watchNamespace              string
+	watchFilterValue            string
```
sbueringer, line 73: We should probably use this variable somewhere :D

sbueringer:
Thx!

/lgtm
/approve

PR #2132, opened 2023-08-01
:seedling: Using ClusterCacheTracker instead of remote.NewClusterClient

### LINKED ISSUE #2053: Consider using ClusterCacheTracker instead of remote.NewClusterClient
I think we should consider using the ClusterCacheTracker from core CAPI which caches Clients and RESTConfig for a cluster across reconciles. The main advantage is that we don't have to construct a client on each reconcile.

Currently we have a few places where we are using remote.NewClusterClient instead:

image

An example on how to setup the ClusterCacheTracker can be seen here: https://github.com/kubernetes-sigs/cluster-api/blob/main/main.go#L372-L395

laozc:
Using ClusterCacheTracker instead of remote.NewClusterClient

Fixes #2053

### DIFF
--- a/controllers/controllers_suite_test.go
+++ b/controllers/controllers_suite_test.go
@@ -28,7 +28,9 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/controllers/remote"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
@@ -63,7 +65,35 @@ func setup() {
 
 	testEnv = helpers.NewTestEnvironment()
 
+	secretCachingClient, err := client.New(testEnv.Manager.GetConfig(), client.Options{
+		HTTPClient: testEnv.Manager.GetHTTPClient(),
+		Cache: &client.CacheOptions{
+			Reader: testEnv.Manager.GetCache(),
+		},
+	})
+	if err != nil {
+		panic("unable to create secret caching client")
+	}
+
+	tracker, err := remote.NewClusterCacheTracker(
+		testEnv.Manager,
+		remote.ClusterCacheTrackerOptions{
+			SecretCachingClient: secretCachingClient,
+			ControllerName:      "testenv-manager",
+		},
+	)
+	if err != nil {
+		panic(fmt.Sprintf("unable to setup ClusterCacheTracker: %v", err))
+	}
+
 	controllerOpts := controller.Options{MaxConcurrentReconciles: 10}
+
+	if err := (&remote.ClusterCacheReconciler{
+		Client:  testEnv.Manager.GetClient(),
+		Tracker: tracker,
+	}).SetupWithManager(ctx, testEnv.Manager, controllerOpts); err != nil {
+		panic(fmt.Sprintf("unable to create ClusterCacheReconciler controller: %v", err))
+	}
 
 	if err := AddClusterControllerToManager(testEnv.GetContext(), testEnv.Manager, &infrav1.VSphereCluster{}, controllerOpts); err != nil {
 		panic(fmt.Sprintf("unable to setup VsphereCluster controller: %v", err))
@@ -71,7 +101,7 @@ func setup() {
 	if err := AddMachineControllerToManager(testEnv.GetContext(), testEnv.Manager, &infrav1.VSphereMachine{}, controllerOpts); err != nil {
 		panic(fmt.Sprintf("unable to setup VsphereMachine controller: %v", err))
 	}
-	if err := AddVMControllerToManager(testEnv.GetContext(), testEnv.Manager, controllerOpts); err != nil {
+	if err := AddVMControllerToManager(testEnv.GetContext(), testEnv.Manager, tracker, controllerOpts); err != nil {
 		panic(fmt.Sprintf("unable to setup VsphereVM controller: %v", err))
 	}
 	if err := AddVsphereClusterIdentityControllerToManager(testEnv.GetContext(), testEnv.Manager, controllerOpts); err != nil {
@@ -80,10 +110,10 @@ func setup() {
 	if err := AddVSphereDeploymentZoneControllerToManager(testEnv.GetContext(), testEnv.Manager, controllerOpts); err != nil {
 		panic(fmt.Sprintf("unable to setup VSphereDeploymentZone controller: %v", err))
 	}
-	if err := AddServiceAccountProviderControllerToManager(testEnv.GetContext(), testEnv.Manager, controllerOpts); err != nil {
+	if err := AddServiceAccountProviderControllerToManager(testEnv.GetContext(), testEnv.Manager, tracker, controllerOpts); err != nil {
 		panic(fmt.Sprintf("unable to setup ServiceAccount controller: %v", err))
 	}
-	if err := AddServiceDiscoveryControllerToManager(testEnv.GetContext(), testEnv.Manager, controllerOpts); err != nil {
+	if err := AddServiceDiscoveryControllerToManager(testEnv.GetContext(), testEnv.Manager, tracker, controllerOpts); err != nil {
 		panic(fmt.Sprintf("unable to setup SvcDiscovery controller: %v", err))
 	}
 
--- a/controllers/serviceaccount_controller.go
+++ b/controllers/serviceaccount_controller.go
@@ -66,7 +66,7 @@ const (
 )
 
 // AddServiceAccountProviderControllerToManager adds this controller to the provided manager.
-func AddServiceAccountProviderControllerToManager(ctx *context.ControllerManagerContext, mgr manager.Manager, options controller.Options) error {
+func AddServiceAccountProviderControllerToManager(ctx *context.ControllerManagerContext, mgr manager.Manager, tracker *remote.ClusterCacheTracker, options controller.Options) error {
 	var (
 		controlledType     = &vmwarev1.ProviderServiceAccount{}
 		controlledTypeName = reflect.TypeOf(controlledType).Elem().Name()
@@ -82,8 +82,8 @@ func AddServiceAccountProviderControllerToManager(ctx *context.ControllerManager
 		Logger:                   ctx.Logger.WithName(controllerNameShort),
 	}
 	r := ServiceAccountReconciler{
-		ControllerContext:  controllerContext,
-		remoteClientGetter: remote.NewClusterClient,
+		ControllerContext:         controllerContext,
+		remoteClusterCacheTracker: tracker,
 	}
 
 	clusterToInfraFn := clusterToSupervisorInfrastructureMapFunc(ctx)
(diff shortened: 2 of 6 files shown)


sbueringer:
@laozc Q: No rush, just asking because of the draft status. Is this PR ready for review?

/ok-to-test

laozc:
> @laozc Q: No rush, just asking because of the draft status. Is this PR ready for review?
> 
> /ok-to-test

I was waiting for the two related PRs to merge first. 😁
Ready for review now.

main.go
```diff
@@ -409,3 +416,44 @@ func isCRDDeployed(mgr ctrlmgr.Manager, gvr schema.GroupVersionResource) (bool,
...
+func setupRemoteClusterCacheTracker(ctx *context.ControllerManagerContext, mgr ctrlmgr.Manager) (*remote.ClusterCacheTracker, error) {
+	secretCachingClient, err := client.New(mgr.GetConfig(), client.Options{
+		HTTPClient: mgr.GetHTTPClient(),
+		Cache: &client.CacheOptions{
+			Reader: mgr.GetCache(),
+		},
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to create secret caching client")
+		return nil, err
```
chrischdi, line 429: We log here and inside the above function. Let's use error wrapping instead and only log in the above function.

```suggestion
		return nil, fmt.Errorf("unable to create secret caching client: %w", err)
```

(Also for the other occurencies where we use setupLog in this function)

main.go
```diff
@@ -409,3 +416,44 @@ func isCRDDeployed(mgr ctrlmgr.Manager, gvr schema.GroupVersionResource) (bool,
...
+	if err != nil {
+		setupLog.Error(err, "unable to create cluster cache tracker")
+		return nil, err
+	}
+
+	if err := (&remote.ClusterCacheReconciler{
+		Client:           mgr.GetClient(),
+		Tracker:          tracker,
+		WatchFilterValue: managerOpts.WatchFilterValue,
+	}).SetupWithManager(ctx, mgr, concurrency(vSphereClusterConcurrency)); err != nil {
```
chrischdi, line 453: Wondering if that's the right concurrency to use.

We would pass the tracker to:
* VMController (which uses `vSphereVMConcurrency`) 
* ServiceAccountProviderController (which uses `providerServiceAccountConcurrency`)
* ServiceDiscoveryController (which uses `serviceDiscoveryConcurrency`)

@sbueringer I think we need here a separate concurrency flag, ideally it should get set the same as for the core CAPI controllers cluster concurrency? Or is this also a gap in core CAPI (I think no, not there)?, Is there any guidance?

controllers/controllers_suite_test.go
```diff
@@ -63,6 +65,29 @@ func setup() {
...
+	if err != nil {
+		panic("unable to create secret caching client")
+	}
+
+	tracker, err := remote.NewClusterCacheTracker(
+		testEnv.Manager,
+		remote.ClusterCacheTrackerOptions{
+			SecretCachingClient: secretCachingClient,
+			ControllerName:      "",
+			Log:                 nil,
```
chrischdi, line 83: We could also drop that as these are the default values.

```suggestion
```

controllers/vspherevm_controller_test.go
```diff
@@ -73,6 +74,19 @@ func TestReconcileNormal_WaitingForIPAddrAllocation(t *testing.T) {
 	}
 	defer simr.Destroy()
 
+	tracker, err := remote.NewClusterCacheTracker(
+		testEnv.Manager,
+		remote.ClusterCacheTrackerOptions{
+			SecretCachingClient: nil,
+			ControllerName:      "",
+			Log:                 nil,
```
chrischdi, line 82: Same as above
```suggestion
```

main.go
laozc: The Core CAPI controller uses the flag for cluster concurrency as it's related to cache the cluster.
Do we need a separate flag in CAPV?

sbueringer: I would use a separate flag

Would also be better if we have a separate flag in core CAPI to be honest

Let's make an example. You have a mgmt cluster with 2k+ workload clusters. The cluster controller is doing some heavy lifting so you would want to use more workers (for 2k 10 should be still fine but it's just an example :D).

So if you want to increase the concurrency of the cluster controller you don't want the side effect that the cluster controller of the cluster cache tracker also gets more workers.


(Basically every single controller should have it's own concurrency flag, in core CAPI we re-use a bunch of them, @chrischdi can you follow-up in core CAPI?)

main.go
```diff
@@ -250,14 +252,20 @@ func main() {
 
 	// Create a function that adds all the controllers and webhooks to the manager.
 	addToManager := func(ctx *context.ControllerManagerContext, mgr ctrlmgr.Manager) error {
+		tracker, err := setupRemoteClusterCacheTracker(ctx, mgr)
+		if err != nil {
+			setupLog.Error(err, "unable to create remote cluster tracker tracker")
```
sbueringer, line 257:
```suggestion
```
Let's not log here. The returned error will be eventually logged in ~ l.304 anyway

(But let's use this text for error wrapping in l.258)

controllers/controllers_suite_test.go
```diff
@@ -63,6 +65,29 @@ func setup() {
...
+			Indexes:             []remote.Index{remote.NodeProviderIDIndex},
+		},
+	)
+	if err != nil {
+		panic(fmt.Sprintf("unable to setup ClusterCacheTracker: %v", err))
+	}
```
sbueringer, line 89: Let's also add the ClusterCacheReconciler here

controllers/serviceaccount_controller.go
```diff
@@ -202,7 +202,7 @@ func (r ServiceAccountReconciler) Reconcile(_ goctx.Context, req reconcile.Reque
 	// then just return a no-op and wait for the next sync. This will occur when
 	// the Cluster's status is updated with a reference to the secret that has
 	// the Kubeconfig data used to access the target cluster.
-	guestClient, err := r.remoteClientGetter(clusterContext, ProviderServiceAccountControllerName, clusterContext.Client, client.ObjectKeyFromObject(cluster))
+	guestClient, err := r.remoteClusterCacheTracker.GetClient(clusterContext, client.ObjectKeyFromObject(cluster))
 	if err != nil {
 		clusterContext.Logger.Info("The control plane is not ready yet", "err", err)
```
sbueringer, line 207:
```suggestion
		// Requeue if the reconcile failed because the ClusterCacheTracker was locked for
		// the current cluster because of concurrent access.
		if errors.Is(err, remote.ErrClusterLocked) {
			r.Logger.V(5).Info("Requeuing because another worker has the lock on the ClusterCacheTracker")
			return ctrl.Result{Requeue: true}, nil
		}
		clusterContext.Logger.Info("The control plane is not ready yet", "err", err)
```
It's a bit annoying but we have to handle this case (like in core CAPI).

Basically when a workload cluster is down multiple reconciler will try to get the client for the workload cluster. But only one of those reconcilers will actually try to create the client. The others will get the `remote.ErrClusterLocked` error

This code helps to reduce noisy errors in the controller logs

(same in other places where we use it)

controllers/vspherevm_controller_test.go
```diff
@@ -73,6 +74,16 @@ func TestReconcileNormal_WaitingForIPAddrAllocation(t *testing.T) {
 	}
 	defer simr.Destroy()
 
+	tracker, err := remote.NewClusterCacheTracker(
```
sbueringer, line 77: Let's also do the full setup here

controllers/vspherevm_controller_test.go
sbueringer: Same as above :)

laozc: Done.

main.go
laozc: Done

controllers/controllers_suite_test.go
laozc: Done

controllers/serviceaccount_controller.go
laozc: Done.

main.go
```diff
@@ -409,3 +417,41 @@ func isCRDDeployed(mgr ctrlmgr.Manager, gvr schema.GroupVersionResource) (bool,
...
+	// Set up a ClusterCacheTracker and ClusterCacheReconciler to provide to controllers
+	// requiring a connection to a remote cluster
+	log := ctrl.Log.WithName("remote").WithName("ClusterCacheTracker")
+	tracker, err := remote.NewClusterCacheTracker(
+		mgr,
+		remote.ClusterCacheTrackerOptions{
+			SecretCachingClient: secretCachingClient,
+			ControllerName:      controllerName,
+			Log:                 &log,
+			Indexes:             []remote.Index{remote.NodeProviderIDIndex},
```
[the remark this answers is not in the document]
laozc: You mean we don't need the indexing because we're not listing the nodes at the moment, right?
Got it.

controllers/vspherevm_controller_test.go
laozc: Done

main.go
laozc: Used the same flag as CAPI PR.

controllers/vspherevm_controller.go
```diff
@@ -379,7 +390,8 @@ func (r vmReconciler) deleteNode(ctx *context.VMContext, name string) error {
 			Name: name,
 		},
 	}
-	return clusterClient.Delete(ctx, node)
+	err = clusterClient.Delete(ctx, node)
+	return ctrl.Result{}, err
```
chrischdi, line 394:
```suggestion
	return ctrl.Result{}, clusterClient.Delete(ctx, node)
```

chrischdi (commented):
Just a nit. Other than that

/lgtm

@sbueringer : I guess we can add an additional PR if the flag name changes in core CAPI?

controllers/vspherevm_controller.go
laozc: Done :)

sbueringer:
> @sbueringer : I guess we can add an additional PR if the flag name changes in core CAPI?

I posted your core CAPI PR in upstream Slack. I'm planning to merge the core CAPI PR on Monday (please ping me if I forget it)

laozc:
> > @sbueringer : I guess we can add an additional PR if the flag name changes in core CAPI?
> 
> I posted your core CAPI PR in Slack. I'm planning to merge the core CAPI PR on Monday (please ping me if I forget it)

Let me know if the flag name is changed to another one. 😊

main.go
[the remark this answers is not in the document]
sbueringer: Yes!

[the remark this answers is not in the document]
sbueringer: And just fyi. By adding the index we automatically add an informer for Nodes

sbueringer:
/lgtm

Yup will report back

sbueringer:
core CAPI is merged

/approve

Thank you!

sbueringer:
/hold
Discussion ongoing in core CAPI about the name of the flag

sbueringer:
@laozc can you drop the flag and just hard code cct concurrency to 10 for now?

This unblocks this pr for CAPV 1.8. And we can still consider adding the flag later. I'm pretty sure we don't actually need the flag as 10 should be easily enough for this reconciler

laozc:
> @laozc can you drop the flag and just hard code cct concurrency to 10 for now?
> 
> This unblocks this pr for CAPV 1.8. And we can still consider adding the flag later. I'm pretty sure we don't actually need the flag as 10 should be easily enough for this reconciler

Sure. Reverted the new flag related change.
Will add it back if the new flag needed after 1.8.

sbueringer:
Thank you very much!!

/lgtm
/approve

pending CI, but Prow will take care of it

PR #2134, opened 2023-08-01
:bug: StoragePolicy should select datastore based on connected hosts and set on VM spec

rikatz:
This PR fixes some workflows on CAPV + Storage policy:

* When using a storage policy, we need to make sure that the only selected datastores are the DS connected to the cluster/host. Today we don't validate it, which means that we may end up with an invalid datastore on a multi-cluster environment with a storage policy that selects datastores on different clusters, we can select a datastore not connected to our cluster - The new workflow will first get all the datastores connected to the cluster, and then pass this list of datastores to Storage Policy validation
*  When passing both storage policy and datastore, this datastore should be validated on the storage policy and fail if it cannot be used
* Finally, on cloning we MUST pass `spec.Location.Datastore = datastoreRef` otherwise the cloning process will not understand when this is a cross datastore/cluster cloning and will fail, as it will try to use the template datastore.

Fixes #

### DIFF
--- a/pkg/services/govmomi/vcenter/clone.go
+++ b/pkg/services/govmomi/vcenter/clone.go
@@ -25,7 +25,6 @@ import (
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/pbm"
 	pbmTypes "github.com/vmware/govmomi/pbm/types"
-	"github.com/vmware/govmomi/view"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 	"k8s.io/utils/pointer"
@@ -229,27 +228,31 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
 			return errors.Wrapf(err, "unable to get storageProfileID from name %s for %q", ctx.VSphereVM.Spec.StoragePolicyName, ctx)
 		}
 
-		kind := []string{"Datastore"}
-		m := view.NewManager(ctx.Session.Client.Client)
-
-		v, err := m.CreateContainerView(ctx, ctx.Session.Client.Client.ServiceContent.RootFolder, kind, true)
-		if err != nil {
-			return errors.Wrapf(err, "unable to create container view for Datastore for %q", ctx)
-		}
-
-		var content []types.ObjectContent
-		err = v.Retrieve(ctx, kind, []string{"name"}, &content)
-		_ = v.Destroy(ctx)
-		if err != nil {
-			return errors.Wrapf(err, "unable to retrieve container view for Datastore for %q", ctx)
-		}
-
 		var hubs []pbmTypes.PbmPlacementHub
-		for _, ds := range content {
+
+		// If there's a Datastore configured, it should be the only one for which we check if it matches the requirements of the Storage Policy
+		if datastoreRef != nil {
 			hubs = append(hubs, pbmTypes.PbmPlacementHub{
-				HubType: ds.Obj.Type,
-				HubId:   ds.Obj.Value,
+				HubType: datastoreRef.Type,
+				HubId:   datastoreRef.Value,
 			})
+		} else {
+			// Otherwise we should get just the Datastores connected to our pool
+			cluster, err := pool.Owner(ctx)
+			if err != nil {
+				return errors.Wrapf(err, "failed to get owning cluster of resourcepool %q to calculate datastore based on storage policy", pool)
+			}
+			dsGetter := object.NewComputeResource(ctx.Session.Client.Client, cluster.Reference())
+			datastores, err := dsGetter.Datastores(ctx)
+			if err != nil {
+				return errors.Wrapf(err, "unable to list datastores from owning cluster of requested resourcepool")
+			}
+			for _, ds := range datastores {
+				hubs = append(hubs, pbmTypes.PbmPlacementHub{
+					HubType: ds.Reference().Type,
+					HubId:   ds.Reference().Value,
+				})
+			}
 		}
 
 		var constraints []pbmTypes.BasePbmPlacementRequirement
@@ -263,25 +266,18 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
 			return fmt.Errorf("no compatible datastores found for storage policy: %s", ctx.VSphereVM.Spec.StoragePolicyName)
 		}
 
-		if datastoreRef != nil {
-			ctx.Logger.Info("datastore and storagepolicy defined; searching for datastore in storage policy compatible datastores")
-			found := false
-			for _, ds := range result.CompatibleDatastores() {
-				compatibleRef := types.ManagedObjectReference{Type: ds.HubType, Value: ds.HubId}
-				if compatibleRef.String() == datastoreRef.String() {
-					found = true
-				}
-			}
-			if !found {
-				return fmt.Errorf("couldn't find specified datastore: %s in compatible list of datastores for storage policy", ctx.VSphereVM.Spec.Datastore)
-			}
-		} else {
+		// If datastoreRef is nil here it means that the user didn't specify a Datastore. So we should
+		// select one of the datastores of the owning cluster of the resource pool that matched the
+		// requirements of the storage policy.
+		if datastoreRef == nil {
 			r := rand.New(rand.NewSource(time.Now().UnixNano())) //nolint:gosec
 			ds := result.CompatibleDatastores()[r.Intn(len(result.CompatibleDatastores()))]
 			datastoreRef = &types.ManagedObjectReference{Type: ds.HubType, Value: ds.HubId}
 		}
 	}
 
+	// if datastoreRef is nil here, means that user didn't specified a datastore NOR a
+	// storagepolicy, so we should select the default
 	if datastoreRef == nil {
 		// if no datastore defined through VM spec or storage policy, use default
 		datastore, err := ctx.Session.Finder.DefaultDatastore(ctx)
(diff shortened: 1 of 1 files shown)


randomvariable:
/hold

@rikatz is testing on linked clone workflow, but otherwise ready for review.

rikatz:
@randomvariable @sbueringer just confirming that the linked clone still works with this change. I will unhold, but feel free to merge it after the next minor release, as a patch if the time line needs to! :)

/hold cancel

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -229,27 +218,31 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
...
+		if datastoreRef != nil {
 			hubs = append(hubs, pbmTypes.PbmPlacementHub{
-				HubType: ds.Obj.Type,
-				HubId:   ds.Obj.Value,
+				HubType: datastoreRef.Type,
+				HubId:   datastoreRef.Value,
 			})
+		} else {
+			// Otherwise we should get just the Datastores connected to our pool
+			cluster, err := pool.Owner(ctx)
```
chrischdi, line 231: All in all the change looks good to me. Just one question for verification:

We always have a `pool` (which is a resourcePool right?) to use or are there cases where there could be no pool and the old variant would need to get applied?

(note: we use `ctx.Session.Finder.ResourcePoolOrDefault` so I guess we would always get a valid resource pool?)

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -179,7 +169,6 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
 			// the VM's UUID.
 			InstanceUuid:      string(ctx.VSphereVM.UID),
 			Flags:             newVMFlagInfo(),
-			DeviceChange:      deviceSpecs,
 			ExtraConfig:       extraConfig,
```
zhanggbj, line 172: Thanks for the PR! Code change looks good to me, only one question here, could you please help to explain why we are removing `DeviceChange`?

pkg/services/govmomi/vcenter/clone.go
rikatz: IIRC the pool will always exists. If there is no pool, it will return a default one as you stated. It should get from https://github.com/vmware/govmomi/blob/ba843e53ef62703cf15f0f263c003d720b63cc1e/find/finder.go#L924

I remember there was an issue we triaged this week on Resource Pools not existing on non DRS clusters, but I think this is already a current problem and in that case we should probably target clusters and not resource pools (like a long term problem)

pkg/services/govmomi/vcenter/clone.go
rikatz: We are not :) It has just been moved to the end https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2134/files/950f4f01fd3345532424e0ee966b697ade43c8b4#diff-f96a0a333f63e93dee3e49d8995ee7bc623e31ded19b4c4172b880b58dc3af9aR292

The movement was because we don't know yet what's going to be the target datastore for the cloned device, so we can only mutate the datastore device spec once we know where is this going to live

chrischdi (commented):
/lgtm

pkg/services/govmomi/vcenter/clone.go
zhanggbj: @rikatz thanks for the clarification, I missed that part

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -229,27 +218,31 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
...
-		err = v.Retrieve(ctx, kind, []string{"name"}, &content)
-		_ = v.Destroy(ctx)
-		if err != nil {
-			return errors.Wrapf(err, "unable to retrieve container view for Datastore for %q", ctx)
-		}
-
 		var hubs []pbmTypes.PbmPlacementHub
-		for _, ds := range content {
+
+		// If there's a datastore request, it should be the only one requested from Storage Policy
```
sbueringer, line 223:
```suggestion
		// If there's a Datastore configured, it should be the only one for which we check if it matches the requirements of the Storage Policy
```
Would this reflect what we do?

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -229,27 +218,31 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
...
-				HubType: ds.Obj.Type,
-				HubId:   ds.Obj.Value,
+				HubType: datastoreRef.Type,
+				HubId:   datastoreRef.Value,
 			})
+		} else {
+			// Otherwise we should get just the Datastores connected to our pool
+			cluster, err := pool.Owner(ctx)
+			if err != nil {
+				return errors.Wrapf(err, "failed to get owning cluster for storage policy datastore")
```
sbueringer, line 233:
```suggestion
				return errors.Wrapf(err, "failed to get owning cluster of resourcepool %q to calculate datastore based on storage policy", pool)
```
Maybe? (not sure if we can just print the pool)

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -229,27 +218,31 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
...
+			}
+			dsGetter := object.NewComputeResource(ctx.Session.Client.Client, cluster.Reference())
+			datastores, err := dsGetter.Datastores(ctx)
+			if err != nil {
+				return errors.Wrapf(err, "unable to list datastores from requested pool")
```
sbueringer, line 238:
```suggestion
				return errors.Wrapf(err, "unable to list datastores from owning cluster of requested resourcepool")
```

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -229,27 +218,31 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
...
+		// If there's a datastore request, it should be the only one requested from Storage Policy
+		if datastoreRef != nil {
 			hubs = append(hubs, pbmTypes.PbmPlacementHub{
-				HubType: ds.Obj.Type,
-				HubId:   ds.Obj.Value,
+				HubType: datastoreRef.Type,
+				HubId:   datastoreRef.Value,
 			})
+		} else {
+			// Otherwise we should get just the Datastores connected to our pool
```
sbueringer, line 230: I assume all datastores of the owning cluster of the resourcepool would work? (assuming the storage policy evaluation works out)

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -263,25 +256,17 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
...
-				if compatibleRef.String() == datastoreRef.String() {
-					found = true
-				}
-			}
-			if !found {
-				return fmt.Errorf("couldn't find specified datastore: %s in compatible list of datastores for storage policy", ctx.VSphereVM.Spec.Datastore)
-			}
-		} else {
+		// If datastoreRef is nil here means that the user didn't specified a datastore with storage policy
+		// so we should select one for the user
```
sbueringer, line 260:
```suggestion
		// If datastoreRef is nil here it means that the user didn't specify a Datastore. So we should 
		// select one of the datastores of the owning cluster of the resource pool that matched the 
		// requirements of the storage policy.
```

pkg/services/govmomi/vcenter/clone.go
rikatz: yeap, they should work as they should be connected to all hosts (but you know, edge cases will always exists)

pkg/services/govmomi/vcenter/clone.go
rikatz: will double check on monday, I think we need to print its name (not sure pool has a stringer method)

pkg/services/govmomi/vcenter/clone.go
rikatz: yeap, that's right!

pkg/services/govmomi/vcenter/clone.go
rikatz: Yeap, this should work:
```
failed to get owning cluster of resourcepool "ResourcePool:resgroup-9 @ /DC-A/host/Cluster1/Resources"
```

sbueringer: Perfect, thx!

sbueringer:
> /label tide/merge-method-squash

Would be good if we manually squash instead. The release note tool we use only surfaces "normal" merged PRs in the release notes. Squashed PRs are skipped

randomvariable:
I think this is fine.

@rikatz , can you squash the commits?

rikatz:
> @rikatz , can you squash the commits?

Done, sorry for the long delay :)

chrischdi (commented):
/llgtm

chrischdi:
Damn typo

/lgtm

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -294,6 +280,17 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
 	disks := devices.SelectByType((*types.VirtualDisk)(nil))
 	isLinkedClone := snapshotRef != nil
 	spec.Location.Disk = getDiskLocators(disks, *datastoreRef, isLinkedClone)
+	spec.Location.Datastore = datastoreRef
+	// Only non-linked clones may expand the size of the template's disk.
+	if snapshotRef == nil {
+		diskSpecs, err := getDiskSpec(ctx, devices)
```
sbueringer, line 286: @rikatz Sorry tried to figure it out but no idea. How does moving this code block change the outcome?

It seems to me that we would end up with the same deviceSpecs independent of if we run it at the top or here.

What am I missing? :)

sbueringer: I would guess something is mutating the devices in between but I have no idea what

rikatz: I think this is a left over change when I was trying to figure out the problem. Previously I was changing the datastore of the disk spec to add the assumed `datastoreRef` and that's why I've moved it.

Let me return to the previous place

rikatz: fixed :) sorry for the left over on it

sbueringer: No worries. I was just looking through it and really wondered what I'm missing. I thought everything is possible with something like passing ctx around like we do etc... Just couldn't find it :D

sbueringer:
Perfect. Thank you very much!

/lgtm
/approve

sbueringer:
@rikatz Do you have some time to open cherry-picks for other branches? I diffed the branches and thought they are all pretty similar :/

rikatz:
done for the 3 failed cherry picks

PR #2141, opened 2023-08-02
🌱 Fix link in release notes to point to CAPV instead of core CAPI

sbueringer:
Sort of not a bad idea to send users to core CAPI, but we probably shouldn't do it :joy:

Fixes #


chrischdi (commented):
/approve

PR #2146, opened 2023-08-03
⚠️ webhook: prevent changes to an already set VSphereVM.spec.biosUUID

chrischdi:

This PR improves the validation for VSphereVM objects by adjusting the webhook to:
* only allow changes to `VSphereVM.spec.biosUUID` if it is not already set
* prevent updates which try to change the biosUUID.

Identified during discussions around #2102

### DIFF
--- a/apis/v1beta1/vspherevm_webhook.go
+++ b/apis/v1beta1/vspherevm_webhook.go
@@ -100,14 +100,23 @@ func (r *VSphereVM) ValidateUpdate(old runtime.Object) (admission.Warnings, erro
 		return nil, apierrors.NewInternalError(errors.Wrap(err, "failed to convert old VSphereVM to unstructured object"))
 	}
 
+	oldTyped, ok := old.(*VSphereVM)
+	if !ok {
+		return nil, apierrors.NewInternalError(fmt.Errorf("failed to typecast old runtime object to VSphereVM"))
+	}
+
 	newVSphereVMSpec := newVSphereVM["spec"].(map[string]interface{})
 	oldVSphereVMSpec := oldVSphereVM["spec"].(map[string]interface{})
 
-	// allow changes to biosUUID, bootstrapRef, thumbprint
-	keys := []string{"biosUUID", "bootstrapRef", "thumbprint", "powerOffMode", "guestSoftPowerOffTimeout"}
-	// allow changes to os only if the old spec has empty OS field
-	if _, ok := oldVSphereVMSpec["os"]; !ok {
+	// Allow changes to bootstrapRef, thumbprint, powerOffMode, guestSoftPowerOffTimeout.
+	keys := []string{"bootstrapRef", "thumbprint", "powerOffMode", "guestSoftPowerOffTimeout"}
+	// Allow changes to os only if the old spec has empty OS field.
+	if oldTyped.Spec.OS == "" {
 		keys = append(keys, "os")
+	}
+	// Allow changes to biosUUID only if it is not already set.
+	if oldTyped.Spec.BiosUUID == "" {
+		keys = append(keys, "biosUUID")
 	}
 	r.deleteSpecKeys(oldVSphereVMSpec, keys)
 	r.deleteSpecKeys(newVSphereVMSpec, keys)
--- a/apis/v1beta1/vspherevm_webhook_test.go
+++ b/apis/v1beta1/vspherevm_webhook_test.go
@@ -191,6 +191,18 @@ func TestVSphereVM_ValidateUpdate(t *testing.T) {
 			vSphereVM:    createVSphereVM("vsphere-vm-1", "foo.com", biosUUID, "", "BB:CC:DD:EE:FF", []string{"192.168.0.1/32"}, nil, Linux, VirtualMachinePowerOpModeSoft, nil),
 			wantErr:      false,
 		},
+		{
+			name:         "biosUUID can be set to a value",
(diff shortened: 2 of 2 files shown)


zhanggbj (commented):
@chrischdi 
Code change looks good to me, only got a question here, is there any possibility that user may need to update the biosUUID, for instance, after the VM is moved to another host maybe due to disaster recovery, and the VM will get a new biosUUID as described https://docs.vmware.com/en/VMware-Workstation-Pro/17/com.vmware.ws.using.doc/GUID-533B2C4F-7BD5-41EB-8392-2B9FE687AE50.html#GUID-533B2C4F-7BD5-41EB-8392-2B9FE687AE50

apis/v1beta1/vspherevm_webhook.go
```diff
@@ -103,12 +103,16 @@ func (r *VSphereVM) ValidateUpdate(old runtime.Object) (admission.Warnings, erro
...
-	keys := []string{"biosUUID", "bootstrapRef", "thumbprint", "powerOffMode", "guestSoftPowerOffTimeout"}
-	// allow changes to os only if the old spec has empty OS field
+	// Allow changes to bootstrapRef, thumbprint, powerOffMode, guestSoftPowerOffTimeout.
+	keys := []string{"bootstrapRef", "thumbprint", "powerOffMode", "guestSoftPowerOffTimeout"}
+	// Allow changes to os only if the old spec has empty OS field.
 	if _, ok := oldVSphereVMSpec["os"]; !ok {
 		keys = append(keys, "os")
 	}
+	// Allow changes to biosUUID only if it is not already set.
+	if _, ok := oldVSphereVMSpec["biosUUID"]; !ok {
```
sbueringer, line 113: I assume when the field is not set this entry does not exist and it's not an empty string, right?

apis/v1beta1/vspherevm_webhook_test.go
```diff
@@ -191,6 +191,12 @@ func TestVSphereVM_ValidateUpdate(t *testing.T) {
 			vSphereVM:    createVSphereVM("vsphere-vm-1", "foo.com", biosUUID, "", "BB:CC:DD:EE:FF", []string{"192.168.0.1/32"}, nil, Linux, VirtualMachinePowerOpModeSoft, nil),
 			wantErr:      false,
 		},
+		{
+			name:         "biosUUID cannot be updated to a different value",
+			oldVSphereVM: createVSphereVM("vsphere-vm-1", "foo.com", "old-uuid", "", "AA:BB:CC:DD:EE", []string{"192.168.0.1/32"}, nil, Linux, VirtualMachinePowerOpModeTrySoft, &metav1.Duration{Duration: GuestSoftPowerOffDefaultTimeout}),
+			vSphereVM:    createVSphereVM("vsphere-vm-1", "foo.com", biosUUID, "", "BB:CC:DD:EE:FF", []string{"192.168.0.1/32"}, nil, Linux, VirtualMachinePowerOpModeSoft, nil),
```
sbueringer, line 197: nit: maybe let's not combine this with a change of the power off fields and only change the biosUUID

This test doesn't verify why exactly the update is blocked

apis/v1beta1/vspherevm_webhook.go
sbueringer: I know the test sort of confirms it, just want to double check

sbueringer: Q: What about checking the typed object instead for an empty string? old.Spec.BiosUUID?
(just feels slightly better)

sbueringer:
> @chrischdi Code change looks good to me, only got a question here, is there any possibility that user may need to update the biosUUID, for instance, after the VM is moved to another host maybe due to disaster recovery, and the VM will get a new biosUUID as described [docs.vmware.com/en/VMware-Workstation-Pro/17/com.vmware.ws.using.doc/GUID-533B2C4F-7BD5-41EB-8392-2B9FE687AE50.html#GUID-533B2C4F-7BD5-41EB-8392-2B9FE687AE50](https://docs.vmware.com/en/VMware-Workstation-Pro/17/com.vmware.ws.using.doc/GUID-533B2C4F-7BD5-41EB-8392-2B9FE687AE50.html#GUID-533B2C4F-7BD5-41EB-8392-2B9FE687AE50)

Good point. We were thinking that nobody should mess with this field when discussing it.

@randomvariable @srm09 @yastij ?

apis/v1beta1/vspherevm_webhook_test.go
```diff
@@ -191,6 +191,18 @@ func TestVSphereVM_ValidateUpdate(t *testing.T) {
 			vSphereVM:    createVSphereVM("vsphere-vm-1", "foo.com", biosUUID, "", "BB:CC:DD:EE:FF", []string{"192.168.0.1/32"}, nil, Linux, VirtualMachinePowerOpModeSoft, nil),
 			wantErr:      false,
 		},
+		{
+			name:         "biosUUID cannot be set to a value",
```
sbueringer, line 195:
```suggestion
			name:         "biosUUID can be set to a value",
```

sbueringer:
Thx for additional context. Sounds fine to me

sbueringer:
/lgtm
/approve

Let's keep the hold a bit for the discussion

randomvariable:
For all intent purposes, a new UUID constitutes a new machine identity, and we should treat that as a new machine. I'm therefore OK with this change.

PR #2149, opened 2023-08-03
:seedling: Add powerOffMode to templates

laozc:
Adds `powerOffMode` to the current templates to serve as documentation to control the VM power off behavior.

Fixes #

### DIFF
--- a/templates/cluster-template-external-loadbalancer.yaml
+++ b/templates/cluster-template-external-loadbalancer.yaml
@@ -55,6 +55,7 @@ spec:
           networkName: '${VSPHERE_NETWORK}'
       numCPUs: 2
       os: Linux
+      powerOffMode: hard
       resourcePool: '${VSPHERE_RESOURCE_POOL}'
       server: '${VSPHERE_SERVER}'
       storagePolicyName: '${VSPHERE_STORAGE_POLICY}'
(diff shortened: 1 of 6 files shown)


templates/cluster-template-external-loadbalancer.yaml
```diff
@@ -60,6 +60,7 @@ spec:
       storagePolicyName: '${VSPHERE_STORAGE_POLICY}'
       template: '${VSPHERE_TEMPLATE}'
       thumbprint: '${VSPHERE_TLS_THUMBPRINT}'
```
sbueringer, line 62: I would have to look up the details but I think we have to modify flavorgen and then re-generate those files

sbueringer: Related: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/2153

zhanggbj (commented):
@laozc it seems cluster-template-node-ipam.yaml also have `VSphereMachineTemplate` but not specified with powerOffMode, is this by intention?

laozc:
> @laozc it seems cluster-template-node-ipam.yaml also have `VSphereMachineTemplate` but not specified with powerOffMode, is this by intention?

Thanks for catching it.
Also set the powerOffMode explicitly.

sbueringer:
Thank you!!

/lgtm
/approve

PR #2154, opened 2023-08-03
✨ Improve session handling with a secure session key

zhanggbj:
Improve session handling with a secure session key to avoid hijack.

Fixes #

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -18,6 +18,7 @@ package session
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"net/netip"
 	"net/url"
@@ -113,14 +114,22 @@ func (p *Params) WithFeatures(feature Feature) *Params {
 // GetOrCreate gets a cached session or creates a new one if one does not
 // already exist.
 func GetOrCreate(ctx context.Context, params *Params) (*Session, error) {
-	logger := ctrl.LoggerFrom(ctx).WithName("session")
+	logger := ctrl.LoggerFrom(ctx).WithName("session").WithValues(
+		"server", params.server,
+		"datacenter", params.datacenter,
+		"username", params.userinfo.Username())
+
 	sessionMU.Lock()
 	defer sessionMU.Unlock()
 
-	sessionKey := params.server + params.userinfo.Username() + params.datacenter
+	userPassword, _ := params.userinfo.Password()
+	h := sha256.New()
+	h.Write([]byte(userPassword))
+	hashedUserPassword := h.Sum(nil)
+	sessionKey := fmt.Sprintf("%s#%s#%s#%x", params.server, params.datacenter, params.userinfo.Username(),
+		hashedUserPassword)
 	if cachedSession, ok := sessionCache.Load(sessionKey); ok {
 		s := cachedSession.(*Session)
-		logger = logger.WithValues("server", params.server, "datacenter", params.datacenter)
 
 		vimSessionActive, err := s.SessionManager.SessionIsActive(ctx)
 		if err != nil {
@@ -220,7 +229,7 @@ func newClient(ctx context.Context, logger logr.Logger, sessionKey string, url *
 			_, err := methods.GetCurrentTime(ctx, tripper)
 			if err != nil {
 				logger.Error(err, "failed to keep alive govmomi client")
-				logger.Info("clearing the session", "key", sessionKey)
+				logger.Info("clearing the session")
 				sessionCache.Delete(sessionKey)
 			}
 			return err
@@ -247,7 +256,7 @@ func newManager(ctx context.Context, logger logr.Logger, sessionKey string, clie
 				return nil
 			}
 
-			logger.Info("rest client session expired, clearing session", "key", sessionKey)
+			logger.Info("rest client session expired, clearing session")
 			sessionCache.Delete(sessionKey)
 			return errors.New("rest client session expired")
 		})


pkg/session/session.go
```diff
@@ -117,7 +118,11 @@ func GetOrCreate(ctx context.Context, params *Params) (*Session, error) {
 	sessionMU.Lock()
 	defer sessionMU.Unlock()
 
-	sessionKey := params.server + params.userinfo.Username() + params.datacenter
+	userPassword, _ := params.userinfo.Password()
+	sessionName := params.server + params.datacenter + params.userinfo.Username() + userPassword
+	h := sha256.New()
+	h.Write([]byte(sessionName))
+	sessionKey := string(h.Sum(nil))
```
chrischdi, line 125:
```suggestion
	sessionKey := fmt.Sprintf("x", h.Sum(nil))
```

To actually get something human readable.

https://go.dev/play/p/ZIY3RMdXg-L

pkg/session/session.go
```diff
@@ -220,7 +225,7 @@ func newClient(ctx context.Context, logger logr.Logger, sessionKey string, url *
 			_, err := methods.GetCurrentTime(ctx, tripper)
 			if err != nil {
 				logger.Error(err, "failed to keep alive govmomi client")
-				logger.Info("clearing the session", "key", sessionKey)
+				logger.Info("clearing the session", "user", url.User.Username())
```
chrischdi, line 228:
```suggestion
				logger.Info("clearing the session", "username", url.User.Username())
```

for consistency, also below

sbueringer: Let's log server, datacenter and username

Just makes it easier to identify a session

randomvariable: We also shouldn't log or print the sessionKey IMHO even if it's hashed.

sbueringer: As discussed in the office hours. Logging server, datacenter and username is fine. We just shouldn't log the hashed password

zhanggbj:
Thanks for your review, and all comments are addressed, including
- Only hashed password, so the sessionKey itself should be still human readable
- For logging, just log server, datacenter and username which is in plain text for debugging, but won't log sessionKey with the key structure and an hashed password
- Also adjust function parameters

pkg/session/session.go
zhanggbj: Will update in next patch.

pkg/session/session.go
zhanggbj: will update logging in next patch

pkg/session/session.go
```diff
@@ -197,11 +206,11 @@ func GetOrCreate(ctx context.Context, params *Params) (*Session, error) {
 	return &session, nil
 }
 
-func newClient(ctx context.Context, logger logr.Logger, sessionKey string, url *url.URL, thumbprint string, feature Feature) (*govmomi.Client, error) {
-	insecure := thumbprint == ""
+func newClient(ctx context.Context, logger logr.Logger, sessionKey string, url *url.URL, params *Params) (*govmomi.Client, error) {
```
chrischdi, line 209: wondering if we should do this function signature changes 🤔 

its more transparent in whats getting used from the input with the previous version. (also for `newClient`)

sbueringer: As far as I can tell we need every single field in params. So this seems simplre than passing in everything individually

pkg/session/session.go
```diff
@@ -117,10 +118,18 @@ func GetOrCreate(ctx context.Context, params *Params) (*Session, error) {
 	sessionMU.Lock()
 	defer sessionMU.Unlock()
 
-	sessionKey := params.server + params.userinfo.Username() + params.datacenter
+	userPassword, _ := params.userinfo.Password()
+	h := sha256.New()
+	h.Write([]byte(userPassword))
+	hashedUserPassword := string(h.Sum(nil))
+	sessionKey := fmt.Sprintf("%s%s%s%s", params.server, params.datacenter, params.userinfo.Username(),
```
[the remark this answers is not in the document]
sbueringer: Out of curiosity. What is the difference?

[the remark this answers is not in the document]
sbueringer: If we're touching this anyway. Let's add hyphens `"%s-%s-%s-%s"` (makes it easier to separate the various parts if we ever look at this via the debugger again)

pkg/session/session.go
```diff
@@ -117,10 +118,18 @@ func GetOrCreate(ctx context.Context, params *Params) (*Session, error) {
...
+		hashedUserPassword)
 	if cachedSession, ok := sessionCache.Load(sessionKey); ok {
 		s := cachedSession.(*Session)
-		logger = logger.WithValues("server", params.server, "datacenter", params.datacenter)
+		logger = logger.WithValues(
```
sbueringer, line 129: Can we add those values to l.117? Then we don't have to do it here and in newClient and newManager

pkg/session/session.go
sbueringer: Although I would now prefer doing this: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2154#discussion_r1284289759 and then reverting the changes to newManager & newClient

sbueringer:
Just nits

pkg/session/session.go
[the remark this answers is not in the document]
chrischdi: readable characters:

```
string:      ,&�kh�Ə��E<0A4B-pd�����^�bf��
fmt.Sprintf: 2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae
```

https://go.dev/play/p/ZIY3RMdXg-L

pkg/session/session.go
chrischdi: I totally missed out that they were all used because I did not take the logger things to consideration 😂 👍

pkg/session/session.go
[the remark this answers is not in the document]
zhanggbj: @chrischdi good catch! will update in next patch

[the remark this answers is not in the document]
zhanggbj: I used `#` as a separator, as we may have `-` in server dns name.

pkg/session/session.go
zhanggbj: @sbueringer Make sense, this looks more neat, didn't notice it before, will update in next patch

laozc (commented):
I think #2171 is going to add the SAML login which may impact the `userinfo` used here.

zhanggbj:
@laozc thanks for the highlight, checking...

sbueringer:
> I think #2171 is going to add the SAML login which may impact the `userinfo` used here. There are more fields to be added in the SAML case.

Thx, but let's tackle this one at a time. 

I think the current PR is fine as is and I also would like to backport this into all branches. We can then discuss in the other PR on how to build on top


/lgtm
/assign @chrischdi

sbueringer:
/approve

pending CI/Prow

sbueringer:
@zhanggbj if you have some time, can you please open cherry-picks for 1.5 and 1.6?

zhanggbj:
@sbueringer Sure, will open cherry-picks today

Issue #1959, opened 2023-06-22
Use a SAML token to log in to vCenter


rgcostea:
/kind feature

**Describe the solution you'd like**
We would like to be able to use a SAML token to establish a session with vCenter rather than having to specify a username and password.

**Anything else you would like to add:**
This is currently supported by govmomi per the examples here:

- https://github.com/vmware/govmomi/blob/main/session/example_test.go
- https://github.com/vmware/govmomi/blob/main/govc/session/login.go

**Environment:**

- Cluster-api-provider-vsphere version: 1.6.1
- Kubernetes version: (use `kubectl version`): 1.26.2
- OS (e.g. from `/etc/os-release`): ubuntu

lubronzhan:
Just to clarify, you want to input for example vsphere key/cert instead of the real token, and the CAPV should get token through that and kept refreshing token if necessary for authentication

rgcostea:
Yes.  That is correct.

lubronzhan:
Ok, I can look into it. CPI already has that function. Should be non trivial to move it here as well

sbueringer:
/assign @lubronzhan 
/lifecycle active
given the open WIP PR

PR #2171, opened 2023-08-07
[WIP] ✨SAML token authentication support

lubronzhan:
SAML token authentication support

User can use username password to authenticate against vsphere. By setting `username` and `password` in secret referenced by `vsphereCluster.spec.identityRef`

User now can also use certificate/key pairs for a solution user to trigger SAML token authentication. By setting `userCert` and `userKey` in secret referenced by `vsphereCluster.spec.identityRef`

When both username and userCert are set, the SAML token authentication is preferred

Fixes #https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1959

### DIFF
--- a/controllers/vspherecluster_reconciler.go
+++ b/controllers/vspherecluster_reconciler.go
@@ -336,11 +336,11 @@ func (r clusterReconciler) reconcileVCenterConnectivity(ctx *context.ClusterCont
 			return nil, err
 		}
 
-		params = params.WithUserInfo(creds.Username, creds.Password)
+		params = params.WithUserInfo(creds.Username, creds.Password).WithUserCertificate(creds.UserCert).WithUserKey(creds.UserKey)
 		return session.GetOrCreate(ctx, params)
 	}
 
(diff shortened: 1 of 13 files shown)


lubronzhan:
Any pointer how to add e2e test would be helpful!
Wonder do I just create another workload cluster using solution user's key and cert.

zhanggbj (commented):
I noticed this is still WIP, just raised a concern here as I'm also working on changes about session key in #2154.

As today we're using `server+username+datacenter` as sessionKey [here](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/pkg/session/session.go#L120), if users are using cert/key for auth, so I assume that username/password could be set as empty. If this is true, then the sessionKey would be `server+datacenter`, and sessions will be messed up for two different users. WDYT?

lubronzhan:
Thanks. I'll take the sessioin handling into account

sbueringer:
@lubronzhan Is this still something that you're interested in?

lubronzhan:
Let me take another look next week

PR #2172, opened 2023-08-07
🌱  flavorgen generates all flavors by default

### LINKED ISSUE #2153: Ensure via make verify that generate cluster-templates are always up-to-date
I think we should ensure via make verify on every PR that the generated cluster-templates in ./templates are up-to-date

Should ~ be:
* create a target to run generate-flavors with templates as FLAVOR_DIR
* add a corresponding verify target
* Add the target to ALL_VERIFY_CHECKS

laozc:

Fixes #2153

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,7 +223,7 @@ help:  # Display this help
 
 .PHONY: generate
 generate: ## Run all generate targets
-	$(MAKE) generate-modules generate-manifests generate-go-deepcopy generate-go-conversions
+	$(MAKE) generate-modules generate-manifests generate-go-deepcopy generate-go-conversions generate-flavors
 
 .PHONY: generate-manifests
 generate-manifests: $(CONTROLLER_GEN) ## Generate manifests e.g. CRD, RBAC etc.
@@ -339,7 +339,7 @@ APIDIFF_OLD_COMMIT ?= $(shell git rev-parse origin/main)
 apidiff: $(GO_APIDIFF) ## Check for API differences
 	$(GO_APIDIFF) $(APIDIFF_OLD_COMMIT) --print-compatible
 
-ALL_VERIFY_CHECKS = boilerplate modules gen conversions doctoc
+ALL_VERIFY_CHECKS = boilerplate modules gen conversions doctoc flavors
 
 .PHONY: verify
 verify: $(addprefix verify-,$(ALL_VERIFY_CHECKS)) lint-markdown lint-shell ## Run all verify-* targets
@@ -380,6 +380,13 @@ verify-boilerplate: ## Verify boilerplate text exists in each file
 .PHONY: verify-container-images
 verify-container-images: ## Verify container images
 	TRACE=$(TRACE) ./hack/verify-container-images.sh
+
+.PHONY: verify-flavors
+verify-flavors: $(FLAVOR_DIR) generate-flavors ## Verify generated flavors
+	@if !(git diff --quiet HEAD -- $(FLAVOR_DIR)); then \
+		git diff $(FLAVOR_DIR); \
+		echo "flavor files in templates directory are out of date"; exit 1; \
+	fi
 
 ## --------------------------------------
 ## Build
@@ -572,12 +579,7 @@ dev-flavors: $(OVERRIDES_DIR)
 
 .PHONY: generate-flavors
 generate-flavors: $(FLAVOR_DIR)
-	go run ./packaging/flavorgen -f vip > $(FLAVOR_DIR)/cluster-template.yaml
-	go run ./packaging/flavorgen -f external-loadbalancer > $(FLAVOR_DIR)/cluster-template-external-loadbalancer.yaml
-	go run ./packaging/flavorgen -f cluster-class > $(FLAVOR_DIR)/clusterclass-template.yaml
-	go run ./packaging/flavorgen -f cluster-topology > $(FLAVOR_DIR)/cluster-template-topology.yaml
-	go run ./packaging/flavorgen -f ignition > $(FLAVOR_DIR)/cluster-template-ignition.yaml
-	go run ./packaging/flavorgen -f node-ipam > $(FLAVOR_DIR)/cluster-template-node-ipam.yaml
+	go run ./packaging/flavorgen --output-dir $(FLAVOR_DIR)
 
 .PHONY: release-staging
 release-staging: ## Build and push container images to the staging registry
--- a/packaging/flavorgen/cmd/root.go
+++ b/packaging/flavorgen/cmd/root.go
@@ -17,10 +17,13 @@ limitations under the License.
 package cmd
 
 import (
+	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	"sigs.k8s.io/cluster-api-provider-vsphere/packaging/flavorgen/flavors"
 	"sigs.k8s.io/cluster-api-provider-vsphere/packaging/flavorgen/flavors/env"
@@ -28,6 +31,27 @@ import (
 )
 
 const flavorFlag = "flavor"
+const outputDirFlag = "output-dir"
+
+var (
+	flavorMappings = map[string]string{
+		flavors.VIP:                  "cluster-template.yaml",
+		flavors.ExternalLoadBalancer: "cluster-template-external-loadbalancer.yaml",
+		flavors.ClusterClass:         "clusterclass-template.yaml",
+		flavors.ClusterTopology:      "cluster-template-topology.yaml",
+		flavors.Ignition:             "cluster-template-ignition.yaml",
+		flavors.NodeIPAM:             "cluster-template-node-ipam.yaml",
+	}
+
+	allFlavors = []string{
+		flavors.VIP,
+		flavors.ExternalLoadBalancer,
+		flavors.ClusterClass,
+		flavors.Ignition,
+		flavors.NodeIPAM,
+		flavors.ClusterTopology,
+	}
+)
 
 func RootCmd() *cobra.Command {
 	rootCmd := &cobra.Command{
@@ -36,8 +60,11 @@ func RootCmd() *cobra.Command {
 		RunE: func(command *cobra.Command, args []string) error {
 			return RunRoot(command)
 		},
+		SilenceUsage: true,
 	}
 	rootCmd.Flags().StringP(flavorFlag, "f", "", "Name of flavor to compile")
+	rootCmd.Flags().StringP(outputDirFlag, "o", "", "Directory to store the generated flavor templates.\nBy default the current directory is used.\nUse '-' to output the result to stdout.")
+
 	return rootCmd
 }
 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -561,12 +561,11 @@ dev-flavors: $(OVERRIDES_DIR)
...
-	go run ./packaging/flavorgen -f external-loadbalancer > $(FLAVOR_DIR)/cluster-template-external-loadbalancer.yaml
-	go run ./packaging/flavorgen -f cluster-class > $(FLAVOR_DIR)/clusterclass-template.yaml
-	go run ./packaging/flavorgen -f cluster-topology > $(FLAVOR_DIR)/cluster-template-topology.yaml
-	go run ./packaging/flavorgen -f ignition > $(FLAVOR_DIR)/cluster-template-ignition.yaml
-	go run ./packaging/flavorgen -f node-ipam > $(FLAVOR_DIR)/cluster-template-node-ipam.yaml
+	go run ./packaging/flavorgen --output-dir $(FLAVOR_DIR)
+
+.PHONY: verify-flavors
+verify-flavors: $(FLAVOR_DIR)
+	go run ./packaging/flavorgen --output-dir $(FLAVOR_DIR) --verify
```
chrischdi, line 568: optional: An alternative to the verify flag could have been to use `git diff` instead, as we do for other verify targets :-) Would have the feature of also printing the actual diff.

Makefile
```diff
@@ -561,12 +561,15 @@ dev-flavors: $(OVERRIDES_DIR)
...
-	go run ./packaging/flavorgen -f external-loadbalancer > $(FLAVOR_DIR)/cluster-template-external-loadbalancer.yaml
-	go run ./packaging/flavorgen -f cluster-class > $(FLAVOR_DIR)/clusterclass-template.yaml
-	go run ./packaging/flavorgen -f cluster-topology > $(FLAVOR_DIR)/cluster-template-topology.yaml
-	go run ./packaging/flavorgen -f ignition > $(FLAVOR_DIR)/cluster-template-ignition.yaml
-	go run ./packaging/flavorgen -f node-ipam > $(FLAVOR_DIR)/cluster-template-node-ipam.yaml
+	go run ./packaging/flavorgen --output-dir $(FLAVOR_DIR)
+
+.PHONY: verify-flavors
+verify-flavors: $(FLAVOR_DIR)
+	go run ./packaging/flavorgen --output-dir $(FLAVOR_DIR)
```
chrischdi, line 568:
```suggestion
verify-flavors: $(FLAVOR_DIR) generate-flavors
```

chrischdi: Also: could we move this target to the other `verify-...` targets inside the makefile :-)

chrischdi (commented):
Last nit 👍

Makefile
laozc: Sure :-)

Makefile
```diff
@@ -370,6 +370,14 @@ verify-boilerplate: ## Verify boilerplate text exists in each file
 verify-container-images: ## Verify container images
 	TRACE=$(TRACE) ./hack/verify-container-images.sh
 
+.PHONY: verify-flavors
+verify-flavors: $(FLAVOR_DIR) generate-flavors
+	go run ./packaging/flavorgen --output-dir $(FLAVOR_DIR)
```
chrischdi, line 375:
```suggestion
```

We should be able to drop this line because we refer the target above :-)

packaging/flavorgen/cmd/root.go
```diff
@@ -52,13 +80,50 @@ func RunRoot(command *cobra.Command) error {
...
+		flavors = allFlavors
+	}
+	generateMultiFlavors := len(flavors) > 1
+	for _, f := range flavors {
+		objs, err := generateSingle(f)
+		if err != nil {
+			return err
+		}
+
+		yamlFileName := flavorMappings[f] + ".yaml"
```
chrischdi, line 100:
```suggestion
		yamlFileName, ok := flavorMappings[f] + ".yaml"
		if !ok {
			return fmt.Errorf("file mapping for flavor %q is missng inf flavorMappings", f)
		}
```

packaging/flavorgen/cmd/root.go
```diff
@@ -52,13 +80,50 @@ func RunRoot(command *cobra.Command) error {
...
+		if outputDir == "" {
```
chrischdi, line 101: I'm thinking about if we should use `-` for printing to stdout. And "" instead just writes to the current directory. WDYT? Let's reflect the behaviour at least also at the flag description.

packaging/flavorgen/cmd/root.go
```diff
@@ -69,13 +134,12 @@ func RunRoot(command *cobra.Command) error {
 			},
 		}
 		util.Replacements = append(util.Replacements, additionalReplacements...)
```
chrischdi, line 136: Wondering if this could byte us.

E.g. when generating other flavors afterwards we will also have the above replacement active. But seems like it does not make any difference in the output files currently?

chrischdi (commented):
Some more comments.

Thanks for tackling this 👍

packaging/flavorgen/cmd/root.go
laozc: Done

packaging/flavorgen/cmd/root.go
laozc: That's a good idea which makes the behavior more predictable.

packaging/flavorgen/cmd/root.go
laozc: Just made a little trick on the `util.Replacement` so the order of output files wouldn't become an issue.

Makefile
laozc: Sure.

packaging/flavorgen/cmd/root.go
chrischdi: Jep, that seems to do the trick.

I may have preferred adjusting `generateSingle` to directly return the generated manifests (by calling `util.GenerateManifestYaml`) and passing the replacements array as variable to `GenerateManifestYaml` from there. This way we would have a single place for the overwrites and not a 3rd place for adjustments.

E.g.
```
		return util.GenerateManifestYaml(flavors.ClusterClassTemplateWithKubeVIP(), util.DefaultReplacements), nil
```

Or in the special case:
```
		return util.GenerateManifestYaml(flavors.ClusterTopologyTemplateKubeVIP(), append(util.Replacements, util.Replacement{
			Kind:      "Cluster",
			Name:      "${CLUSTER_NAME}",
			Value:     env.ControlPlaneMachineCountVar,
			FieldPath: []string{"spec", "topology", "controlPlane", "replicas"},
		})), nil
```

But I'm ok with this approach too.

Makefile
```diff
@@ -370,6 +370,13 @@ verify-boilerplate: ## Verify boilerplate text exists in each file
 verify-container-images: ## Verify container images
 	TRACE=$(TRACE) ./hack/verify-container-images.sh
 
+.PHONY: verify-flavors
+verify-flavors: $(FLAVOR_DIR) generate-flavors
+	@if !(git diff --quiet HEAD -- templates); then \
+		git diff; \
```
chrischdi, line 376:
```suggestion
	@if !(git diff --quiet HEAD -- $(FLAVOR_DIR)); then \
		git diff $(FLAVOR_DIR); \
```

Let's use the parameter for both. paths :-)

chrischdi (commented):
last nits :-)

packaging/flavorgen/cmd/root.go
```diff
@@ -52,30 +79,74 @@ func RunRoot(command *cobra.Command) error {
...
+		if err != nil {
+			return errors.Wrapf(err, "failed to dump manifest content to file for flavor %s", f)
+		}
+	}
+
+	return nil
+}
+
+func generateSingle(flavor string) (string, error) {
+	util.Replacements = append([]util.Replacement{}, util.DefaultReplacements...)
```
chrischdi, line 125: Really the last improvement:

```suggestion
	replacements := util.DefaultReplacements
```

Let's make this private and make it a parameter for `util.GenerateManifestYaml(objs, replacements)` :-)

packaging/flavorgen/cmd/root.go
```diff
@@ -52,30 +79,74 @@ func RunRoot(command *cobra.Command) error {
...
-				Kind:      "Cluster",
-				Name:      "${CLUSTER_NAME}",
-				Value:     env.ControlPlaneMachineCountVar,
-				FieldPath: []string{"spec", "topology", "controlPlane", "replicas"},
-			},
-		}
-		util.Replacements = append(util.Replacements, additionalReplacements...)
-		util.PrintObjects(flavors.ClusterTopologyTemplateKubeVIP())
+		objs = flavors.ClusterTopologyTemplateKubeVIP()
+		util.Replacements = append(util.Replacements, util.Replacement{
```
chrischdi, line 137:
```suggestion
		replacements = append(replacements, util.Replacement{
```

packaging/flavorgen/flavors/util/helpers.go
```diff
@@ -63,6 +62,7 @@ var (
 			FieldPath: []string{"spec", "template", "spec"},
 		},
 	}
+	Replacements = append([]Replacement{}, DefaultReplacements...)
```
chrischdi, line 65:
```suggestion
```

With the other changes we should be able to drop this.

chrischdi (commented):
Thanks for all the work!

Last round of improvements 👍

packaging/flavorgen/cmd/root.go
laozc: This idea was considered in my last commit.
It was dropped as there are a few dependencies on the global variable `util.Replacements`.
I think it would be better to use a separate PR to move it as parameters instead of global variable.

Makefile
laozc: Done

Makefile
```diff
@@ -370,6 +370,13 @@ verify-boilerplate: ## Verify boilerplate text exists in each file
 verify-container-images: ## Verify container images
 	TRACE=$(TRACE) ./hack/verify-container-images.sh
 
+.PHONY: verify-flavors
+verify-flavors: $(FLAVOR_DIR) generate-flavors
```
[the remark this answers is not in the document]
laozc: Done

packaging/flavorgen/cmd/root.go
laozc: I was going to preserve the return value of runtime.Object.
Agreed that returning a string is better.

packaging/flavorgen/cmd/root.go
chrischdi: Hm, I thought it is only:
* the `root.go` which needs to change it for the one case 
* the `GenerateObjectYAML` function itself which reads it (and instead we could pass it to this func as parameter)
* https://github.com/chrischdi/cluster-api-provider-vsphere/blob/671aea8af0aa3a368361dae6bc1330f1b0a2d8f8/packaging/flavorgen/flavors/crs/util.go#L81 which also calls `GenerateObjectYAML`, there we could simply pass `util.DefaultReplacements` as parameter.

chrischdi: But I'm also ok with the code as is, this is more cosmetical and not a functional idea/change.

laozc: Yep. Those are the few lines I'm not 100% sure.
The generated files are the same and guessing it should be okay to adopt these changes as well?
I will push the change and let me know if you have other concerns. :)

chrischdi (commented):
/lgtm

Thanks for the patience :-)

laozc:
> LGTM label has been added.

Sure. Just feeling a bit scared when the CI is also not working :)

Makefile
```diff
@@ -335,7 +335,7 @@ APIDIFF_OLD_COMMIT ?= $(shell git rev-parse origin/main)
 apidiff: $(GO_APIDIFF) ## Check for API differences
 	$(GO_APIDIFF) $(APIDIFF_OLD_COMMIT) --print-compatible
 
-ALL_VERIFY_CHECKS = boilerplate modules gen conversions
+ALL_VERIFY_CHECKS = boilerplate modules gen conversions flavor
```
chrischdi, line 338:
```suggestion
ALL_VERIFY_CHECKS = boilerplate modules gen conversions flavors
```

chrischdi:
e2e main is currently broken due to the CI infrastructure.

packaging/flavorgen/cmd/root.go
```diff
@@ -52,30 +79,74 @@ func RunRoot(command *cobra.Command) error {
...
+				fmt.Printf("### %s\n", yamlFileName)
+			}
+			fmt.Print(manifest)
+			continue
+		}
+
+		yamlPath := filepath.Join(outputDir, yamlFileName)
+		err = os.WriteFile(yamlPath, []byte(manifest), 0600)
+		if err != nil {
+			return errors.Wrapf(err, "failed to dump manifest content to file for flavor %s", f)
```
sbueringer, line 117:
```suggestion
			return errors.Wrapf(err, "failed to write manifest content to file for flavor %s", f)
```
nit. Just sounds a bit nicer

packaging/flavorgen/cmd/root.go
```diff
@@ -17,17 +17,41 @@ limitations under the License.
...
+const outputDirFlag = "output-dir"
+
+var (
+	flavorMappings = map[string]string{
+		flavors.VIP:                  "cluster-template",
+		flavors.ExternalLoadBalancer: "cluster-template-external-loadbalancer",
+		flavors.ClusterClass:         "clusterclass-template",
+		flavors.ClusterTopology:      "cluster-template-topology",
+		flavors.Ignition:             "cluster-template-ignition",
+		flavors.NodeIPAM:             "cluster-template-node-ipam",
```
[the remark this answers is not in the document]
laozc: Done

packaging/flavorgen/cmd/root.go
```diff
@@ -52,30 +79,74 @@ func RunRoot(command *cobra.Command) error {
...
+		}
+
+		fileMapping, ok := flavorMappings[f]
+		if !ok {
+			return fmt.Errorf("file mapping for flavor %q is missng in flavorMappings", f)
+		}
+
+		yamlFileName := fileMapping + ".yaml"
+		if outputDir == "-" {
+			if generateMultiFlavors {
```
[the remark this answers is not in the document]
laozc: Would like to keep stdout as an option since the rendered result was output to stdout before the change.
I believe it has been handled in `GenerateManifestYaml` to add the `---` in YAML.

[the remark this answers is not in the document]
sbueringer: That's fine and you're correct. Didn't realize we always add a leading `---`

sbueringer:
Thx!!

/lgtm
/approve

PR #2174, opened 2023-08-07
🌱 Add verify-govulncheck and verify-vulnerabilities targets and integrate to scan action

### LINKED ISSUE #2119: Integrate govulncheck
govulncheck is a tool provided by the Go team: https://go.dev/blog/vuln

It allows scanning Go code for vulnerabilities and then reports if the vulnerable code is actually used (it could be that we have a Go Dependency with a CVE, but the affected package in the dependency is not actually used). The huge benefit here is that it can help to assess the impact of CVE's (i.e. no impact if we don't use the code).

It should be pretty easy to integrate as it can be ~ run via `go run golang.org/x/vuln/cmd/govulncheck ./...`. But let's integrate it with go install like other tools.

Limitations:
* Trivy could find CVE's that are either outside of the scope of govulncheck (e.g. base image) or that govulncheck is not aware of yet

Notes:
* The idea is to run it in our scan GitHub action after Trivy
* We can decide if we action should also fail is govulncheck fails (could be that it finds issues that are unknown to Trivy). I would say yes, we can always iterate

chrischdi:

Adds:
* `verify-govulncheck` target, to scan the code via `govulncheck`
* `verify-vulnerabilities` target which runs `./hack/verify-vulnerabilities.sh`
  * Runs `make verify-container-images`
  * Runs `make verify-govulncheck`
  * Exit code is 1 if one or both targets report vulnerabilities or errors
* Use `verify-vulnerabilities` in scan action instead of `verify-container-images`

Fixes #2119


hack/verify-vulnerabilities.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
chrischdi, line 1: another option would be to rename the `verify-container-images.sh` to maybe `verify-vulnerabilities.sh` and integrate `make govulncheck` into that.

chrischdi:
/hold

to track and merge core CAPI PR first: https://github.com/kubernetes-sigs/cluster-api/pull/9144

killianmuldoon (commented):
/lgtm
/approve

sbueringer:
@killianmuldoon We'll take a closer look tomorrow, but it looks like we have an issue with the CI env

sbueringer:
To unblock this PR. Let's try

/override pull-cluster-api-provider-vsphere-e2e-main

sbueringer:
@chrischdi Some cherry-picking for you :)

Issue #2125, opened 2023-07-31
Migrate from static shellcheck docker container to the shellcheck script used in other k repos


randomvariable:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

We currently hard pin a shellcheck docker image stored in a VMware owned GCP repository. This should be migrated to something more regularly updated and community owned.

https://github.com/kubernetes/kubernetes/blob/master/hack/verify-shellcheck.sh is a good example.

This will also affect Cloud provider vSphere and vsphere CSI repositories.

cc @lubronzhan 
cc @xing-yang 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

sbueringer:
Let's check if we can use the same as core CAPI: https://github.com/kubernetes-sigs/cluster-api/blob/main/hack/verify-shellcheck.sh

sbueringer:
@chrischdi Done? (for all branches?)

PR #2211, opened 2023-08-14
🌱 Use shellcheck binary instead of self-built docker image

chrischdi:

This PR:

* replaces the current shellcheck implementation (based on a self-built container image used via `docker run`) by the solution which is used at https://github.com/kubernetes-sigs/cluster-api.
* enables to run `make lint-shell` or `make verify-shellcheck` on arm based macs.
* this also adds the `verify-shellcheck` target for compatibility with CAPI's Makefile.

Part of #2125

/area ci


Makefile
```diff
@@ -326,8 +328,7 @@ lint-markdown: ## Lint the project's markdown
 	docker run --rm -v "$$(pwd)":/build$(DOCKER_VOL_OPTS) gcr.io/cluster-api-provider-vsphere/extra/mdlint:0.17.0 -- /md/lint -i _releasenotes .
 
 .PHONY: lint-shell
-lint-shell: ## Lint the project's shell scripts
-	docker run --rm -t -v "$$(pwd)":/build:ro gcr.io/cluster-api-provider-vsphere/extra/shellcheck
+lint-shell: verify-shellcheck ## Lint the project's shell scripts
```
sbueringer, line 331: Should we get rid of lint-shell?

chrischdi: I'd be okay with that 👍

sbueringer: Let's do it

chrischdi: Also verified we do not use this or the shell script in CI.

sbueringer:
There's not much to hold at this point :)

PR #2225, opened 2023-08-15
[release-1.8] 🌱 Grouped CI related cherry-picks

chrischdi:
This is a manual cherry-pick of:

- #2167
- #2172
  - only CI parts
- #2211


docs/release/release-tasks.md
```diff
@@ -7,13 +7,13 @@
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
-- [Prepare main branch for development of the new release](#prepare-main-branch-for-development-of-the-new-release)
```
sbueringer, line 10: This seems to diverge from main with this change now (not saying it has to be sync, but not sure about this change)

sbueringer: (although having the whole file sync probably makes cherry-picks easier)

chrischdi: I'll drop this. We don't update this that much / should have no conflicts.

PR #2229, opened 2023-08-15
🌱 Enable powerOffMode trySoft in cluster-templates and CI

sbueringer:

We introduced powerOffMode in CAPV v1.8. We ensured that the default behavior doesn't change by using "hard" as default on the CRDs. 

I think it's reasonable now to start using trySoft in our cluster-templates and in CI. It's the ideal value and just changing the value in the cluster-templates doesn't break our API behavior in any way.

Also the VM's now have a chance to "return" their DHCP leases during shutdown, which is helpful for CI stability.

P.S. I don't want to cherry-pick this change as I don't want to change the field in cluster-templates in a CAPV v1.8 patch release

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -288,7 +288,7 @@ generate-doctoc:
 	TRACE=$(TRACE) ./hack/generate-doctoc.sh
 
 .PHONY: generate-e2e-templates
-generate-e2e-templates: ## Generate e2e cluster templates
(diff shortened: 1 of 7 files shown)


chrischdi (commented):
/retest

/lgtm

sbueringer:
/hold 
for CI

PR #2235, opened 2023-08-16
:seedling: Fixes session active check with Datacenter scoped TKG admin account

### LINKED ISSUE #2066: Session validation fails with Datacenter scoped Administrator user.
/kind bug

**What steps did you take:**

1. Created a new user on VCenter.
2. Given user permissions as per https://docs.vmware.com/en/VMware-Tanzu-Kubernetes-Grid/1.6/vmware-tanzu-kubernetes-grid-16/GUID-mgmt-clusters-vsphere.html#vsphere-permissions
3. Followed the [getting started guide](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/docs/getting_started.md) to deploy capv cluster with v1.2.0.
4. Created a new user which was Administrator on DataCenter, same thing happened for that user as well.

Both users did not have ANY permissions at root level.

**What happened**
In capv logs, started seeing errors as below.
```tsv
session "msg"="unable to check if vim session is active" "error"="ServerFaultCode: Permission to perform this operation was denied." "datacenter"="" "server"="vcenter-url.com:443"
session "msg"="unable to get vim client session" "error"="ServerFaultCode: Permission to perform this operation was denied." "datacenter"="" "server"="vcenter-url.com:443"
```

Ideally there should be no errors, and capv should be able to run fine.

It is actually a bug in capv code on the way of checking session is active or not.
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/901740c18d4d984e5a31462dee03933c0771acef/pkg/session/session.go#L125
govmomi call `SessionManager.IsSessionActive()` requires permission on root level to work, so this will always fail for our case.

Correct way to check is session is active or not is to use something like below.
```go
u, err := s.SessionManager.UserSession(ctx)
if err != nil {
	if soap.IsSoapFault(err) {
		fault := soap.ToSoapFault(err).VimFault()
		// If the PropertyCollector is not found, the saved session for this URL is not valid
		if _, ok := fault.(vimtypes.ManagedObjectNotFound); ok {
			return false, nil
		}
	}
	return false, err
}
```

On a datacenter scoped vCenter account, `SessionManager.SessionIsActive()` fails with permission denied error.
Need to change it to `SessionManager.UserSession()` method, which does not require special permissions.

Even though [Documentation](https://docs.vmware.com/en/VMware-Tanzu-Kubernetes-Grid/1.6/vmware-tanzu-kubernetes-grid-16/GUID-mgmt-clusters-vsphere.html#vsphere-permissions) requires permissions at root level, it is not actually required if we fix above bug.
Asking for root permissions without actual requirement is not clean way and it is blocking multiple scenarios where users does not have root permissions.

**Environment:**

- Cluster-api-provider-vsphere version: v1.2.0
- Kubernetes version: (use `kubectl version`): v1.23.15
- OS (e.g. from `/etc/os-release`): Mariner OS

laozc:
Session active check does not work for Datacenter scoped TKG admin account.
API `SessionIsActive` requires `ValidateSession` permission which is not listed in current doc.
It can be replaced with a simple Datacenter retrieval call to do the same thing.

Fixes #2066


pkg/session/session.go
```diff
@@ -133,7 +134,18 @@ func GetOrCreate(ctx context.Context, params *Params) (*Session, error) {
 
 		vimSessionActive, err := s.SessionManager.SessionIsActive(ctx)
```
subbartt, line 135: Did you consider using "UserSession" as mentioned in the issue request, we have hit the same issue in our code and moved to the below piece of code for checking the active session and it is working well for us. Do you see any issues with the approach?

u, err := s.SessionManager.UserSession(ctx)
if err != nil {
	if soap.IsSoapFault(err) {
		fault := soap.ToSoapFault(err).VimFault()
		// If the PropertyCollector is not found, the saved session for this URL is not valid
		if _, ok := fault.(vimtypes.ManagedObjectNotFound); ok {
			return false, nil
		}
	}
	return false, err
}
On a datacenter scoped vCenter account, SessionManager.SessionIsActive() fails with permission denied error.
Need to change it to SessionManager.UserSession() method, which does not require special permissions.

laozc: Thanks.
I validated in an environment and indeed the `UserSession` would be a better choice as it does not require any permission at all.

namra98 (approved):

randomvariable:
Had to read the govmomi code for UserSession to understand what happened. It makes sense, but maybe worth leaving a comment that UserSession(blah) actually does a managed object retrieval for future devs.

sbueringer:
@laozc 

> API SessionIsActive requires ValidateSession permission which is not listed in current doc.

I think this is not correct:
> Session
> * Message
> * Validate session
> https://docs.vmware.com/en/VMware-Tanzu-Kubernetes-Grid/1.6/vmware-tanzu-kubernetes-grid-16/GUID-mgmt-clusters-vsphere.html#vsphere-permissions

If I got it correctly it is just that Validate Session is required on the root (?). And using UserSession (aka getting currentSession) requires less permissions.

I'm fine in general with the change, but is there any vCenter documentation confirming that UserSession / currentSession requires no specific pemissions? (@laozc @randomvariable)

randomvariable:
I don't know if there's something confirming that permissions *are not required*, but permissions are described here https://docs.vmware.com/en/VMware-vSphere/8.0/vsphere-security/GUID-ED56F3C4-77D0-49E3-88B6-B99B8B437B62.html 

Similar to what we do with [CloudTrail in CAPA](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a9e9e91b9bfc406e8bda74d2ddeb1484caad2356/test/e2e/shared/aws.go#L815-L840), at some point may want to enable the privilege recorder https://docs.vmware.com/en/VMware-vSphere/8.0/vsphere-security/GUID-8A5066BB-BE0C-48AD-9DDB-F3446AD3B9F6.html and capture the output, though that might be restricted in VMC.

sbueringer:
Thx for the info. That could become very useful going forward

sbueringer:
Renamed the PR as it's clearly specific that a TKG admin requires session validation. Nonetheless, makes sense to reduce the required privileges

PR #2263, opened 2023-08-21
⚠️ govmomi: verify the inventory path of a vm if detected via bios or instance uuid

### LINKED ISSUE #2102: Don't set failureMessage and failureReason if VSphereVM was not found by BiosUUID
### What would you like to be added (User Story)?

As an operator I would like VSphereVM's to auto-recover after recovering my datastore.

### Detailed Description

Scenario:

* Having a storage issue
* VSphereVM controller is not able anymore to find a `VSphereVM` by its BiosUUID in vCenter and sets `.status.failureMessage` + `.status.failureReason`, which get's synced to the `VSphereMachine` and `Machine` objects.
* Storage issues get resolved, VM's are still there
* `VSphereVM`, `VSphereMachine` and `Machine` controller won't reconcile the objects again unless the `.status.failureMessage` + `.status.failureReason` got manually removed from
  * the `VSphereVM` (which get's synced to the `VSphereMachine`)
  * the `Machine`

The code causing this is here:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/52e15471be8b4cb270b6b6960dbba1bad61c8231/pkg/services/govmomi/service.go#L85C1-L88

The fields for `failureReason` is defined as follows (and similar is `failureMessage`):
```go
	// FailureReason will be set in the event that there is a terminal problem
	// reconciling the vspherevm and will contain a succinct value suitable
	// for vm interpretation.
	//
	// This field should not be set for transitive errors that a controller
	// faces that are expected to be fixed automatically over
	// time (like service outages), but instead indicate that something is
	// fundamentally wrong with the vm.
	//
	// Any transient errors that occur during the reconciliation of vspherevms
	// can be added as events to the vspherevm object and/or logged in the
	// controller's output.
```

So the big question is: should we handle this as transient error instead and return an error here instead and only return the error here instead of setting the `failureReason` and `failureMessage` fields? 

### Anything else you would like to add?

_No response_

### Label(s) to be applied

/kind feature
One or more /area label. See https://github.com/kubernetes-sigs/cluster-api/labels?q=area for the list of labels.

chrischdi:

Adds a verification step at the `findVM` function to verify the actual inventory path of a vm with the expected one.

If a VM is found via bios or instance UUID and the path does not match the expectation an error is returned.

Fixes #2102

### DIFF
--- a/pkg/services/govmomi/util.go
+++ b/pkg/services/govmomi/util.go
@@ -17,13 +17,16 @@ limitations under the License.
 package govmomi
 
 import (
+	gocontext "context"
 	gonet "net"
 	"path"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/property"
+	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -58,6 +61,12 @@ func sanitizeIPAddrs(ctx *context.VMContext, ipAddrs []string) []string {
 //  3. If it is not found by instance UUID, fallback to an inventory path search
 //     using the vm folder path and the VSphereVM name
 func findVM(ctx *context.VMContext) (types.ManagedObjectReference, error) {
+	expectedInventoryPath, err := generateInventoryPath(ctx, ctx.Session.Finder, ctx.VSphereVM)
+	if err != nil {
+		return types.ManagedObjectReference{}, errors.Wrap(err, "get expected inventory path")
+	}
+
+	// Try to find vm by BiosUUID.
 	if biosUUID := ctx.VSphereVM.Spec.BiosUUID; biosUUID != "" {
 		objRef, err := ctx.Session.FindByBIOSUUID(ctx, biosUUID)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/services/govmomi/util.go
```diff
@@ -58,6 +61,12 @@ func sanitizeIPAddrs(ctx *context.VMContext, ipAddrs []string) []string {
 //  3. If it is not found by instance UUID, fallback to an inventory path search
 //     using the vm folder path and the VSphereVM name
 func findVM(ctx *context.VMContext) (types.ManagedObjectReference, error) {
+	expectedInventoryPath, err := generateInventoryPath(ctx, ctx.Session.Finder, ctx.VSphereVM)
```
chrischdi, line 64: Note: this always causes an API call to vCenter to check if the folder exists.

pkg/services/govmomi/util.go
```diff
@@ -67,35 +76,69 @@ func findVM(ctx *context.VMContext) (types.ManagedObjectReference, error) {
 			ctx.Logger.Info("vm not found by bios uuid", "biosuuid", biosUUID)
 			return types.ManagedObjectReference{}, errNotFound{uuid: biosUUID}
 		}
+
+		// Verify that inventory path matches the expectation
+		if err := verifyInventoryPath(ctx, ctx.Session.Client.Client, objRef.Reference(), expectedInventoryPath); err != nil {
```
chrischdi, line 81: Note: this causes an additional API call to vCenter.

pkg/services/govmomi/util.go
```diff
@@ -67,35 +76,69 @@ func findVM(ctx *context.VMContext) (types.ManagedObjectReference, error) {
...
 		return types.ManagedObjectReference{}, err
 	}
-	if objRef == nil {
-		// fallback to use inventory paths
-		folder, err := ctx.Session.Finder.FolderOrDefault(ctx, ctx.VSphereVM.Spec.Folder)
-		if err != nil {
-			return types.ManagedObjectReference{}, err
+	if objRef != nil {
+		// Verify that inventory path matches the expectation
+		if err := verifyInventoryPath(ctx, ctx.Session.Client.Client, objRef.Reference(), expectedInventoryPath); err != nil {
```
chrischdi, line 97: Note: this causes an additional API call to vCenter.

chrischdi:
This breaks existing use cases and can't be merged. As discussed in office hours.

/close

PR #2268, opened 2023-08-21
🌱 github: add edited and reopened as triggers for the GH workflow approval

chrischdi:

Adds the types `edited` and `reopened` to the triggers for the `PR approve GH Workflows` action.

Before this PR the golangci-lint action could have been triggered by the `edited` or `reopened` actions.
The approval workflow would not get triggered in that case and did not approve the linter.

Cross-repo cherry-pick of https://github.com/kubernetes-sigs/cluster-api/pull/9259

Fixes #

### DIFF
--- a/.github/workflows/pr-gh-workflow-approve.yaml
+++ b/.github/workflows/pr-gh-workflow-approve.yaml
@@ -3,7 +3,9 @@ name: PR approve GH Workflows
 on:
   pull_request_target:
     types:
+      - edited
       - labeled
+      - reopened
       - synchronize
 
 jobs:


killianmuldoon (commented):
should we wait to confirm the fix works on CAPI first? or are we confident this works as expected?

chrischdi:
> should we wait to confirm the fix works on CAPI first? or are we confident this works as expected?

I verified that it works.

The important point is to have the same triggers as the linter action to not hit the same case.

And they are now the same except:
* Not having `opened` --> because there will be no `ok-to-test` label set
* Adding `labeled` --> because we want to trigger the approval by this

killianmuldoon:
/lgtm

It makes sense to me - the existing action pretty much works except for not being triggered on updates.

PR #2271, opened 2023-08-22
:seedling: Add dependabot grouping for Kubernetes modules

### LINKED ISSUE #2202: Consider using dependabot dependency groups
Dependabot has an option to bump dependencies in groups:

> By default, Dependabot raises a single pull request for each dependency that needs to be updated to a newer version. You can use groups to create sets of dependencies (per package manager), so that Dependabot opens a single pull request to update multiple dependencies at the same time.
> https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

As we're bumping k8s.io/* patch versions in CAPV it might be worth using this feature. If I got it right the k8s.io PRs in the following screenshot could have been just one PR

image

killianmuldoon:
Add a grouping for all dependencies under `k8s.io/` to the dependabot configuration. This should reduce manual approvals for those dependencies.

Fixes #2202

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,13 +18,17 @@ updates:
   schedule:
     interval: "weekly"
     day: "monday"
+  ## group all dependencies with a k8s.io prefix into a single PR where possible
+  groups:
+    kubernetes:
+      patterns: [ "k8s.io/*" ]
   ignore:
+  # Ignore k8s and its transitives modules as they are upgraded manually
+  # together with controller-runtime.
+  - dependency-name: "k8s.io/*"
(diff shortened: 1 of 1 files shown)


killianmuldoon (commented):
Note: I did this on my fork and it seems there was a bit of a race condition when enabling it - i.e. I randomly got a grouped PR and some duplicates:

image

You can see the PRs here: https://github.com/killianmuldoon/cluster-api-provider-vsphere/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc

I think this should work better with the scheduled dependabot run, but this feature is still in beta so we might see some bugs. Still worth trying out as-is IMO

PR #1579, opened 2022-07-14
[WIP] VGPU implementation

geetikabatra:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note

```

PR #2272, opened 2023-08-22
:sparkles: vGPU implementation

### LINKED ISSUE #1972: vGPU support
/kind feature

[A clear and concise description of what you want to happen.]
Ability to add vGPUs to VMs

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/docs/proposal/20210823-gpu-support.md

**Environment:**

- Cluster-api-provider-vsphere version: 1.6.x

puneetkatyal:
- Builds on the changes in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1579

Support adding vGPUs to VMs

Fixes #1972


docs/gpu-vgpu.md
```diff
@@ -0,0 +1,107 @@
...
+See "2. Choosing the vGPU Profile for the Virtual Machine" at [Using GPUs with Virtual Machines on vSphere](https://blogs.vmware.com/apps/2018/09/using-gpus-with-virtual-machines-on-vsphere-part-3-installing-the-nvidia-grid-technology.html) to see what vGPU profiles are available for your GPU device.
+
+We are using NVIDIA Tesla V100 32GB cards for this example and will use the `grid_v100d-4c` vGPU profile for this card that allocates 4GB GPU memory to the worker node's vGPU device. 
+
+### Create the cluster template
+
+```shell
+$ make dev-flavors
+/Applications/Xcode.app/Contents/Developer/usr/bin/make generate-flavors FLAVOR_DIR=/Users/pkatyal/.cluster-api/overrides/infrastructure-vsphere/v0.0.0
+go run ./packaging/flavorgen --output-dir /Users/pkatyal/.cluster-api/overrides/infrastructure-vsphere/v0.0.0
```
chrischdi, line 60:
```suggestion
```

Let's omit the stdout output

docs/gpu-vgpu.md
```diff
@@ -0,0 +1,107 @@
...
+      numCPUs: 2
+      os: Linux
+      powerOffMode: trySoft
+      resourcePool: '${VSPHERE_RESOURCE_POOL}'
+      server: '${VSPHERE_SERVER}'
+      storagePolicyName: '${VSPHERE_STORAGE_POLICY}'
+      template: '${VSPHERE_TEMPLATE}'
+      thumbprint: '${VSPHERE_TLS_THUMBPRINT}'
+      vgpuDevices:
+        - profileName: "grid_v100d-4c"    <============ value from above
```
chrischdi, line 94: Would it be worth having this as envsubst parameter in some way?

WDYT: would it be worth having a separate flavor for vgpu?

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -453,3 +461,55 @@ func getNetworkSpecs(ctx *context.VMContext, devices object.VirtualDeviceList) (
 
 	return deviceSpecs, nil
 }
+
+func createPCIPassThroughDevice(deviceKey int32, backingInfo types.BaseVirtualDeviceBackingInfo) types.BaseVirtualDevice {
```
chrischdi, line 465: Let's move this func to the very bottom of the file, because its a helper for the ones below.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -453,3 +461,55 @@ func getNetworkSpecs(ctx *context.VMContext, devices object.VirtualDeviceList) (
...
+	device := &types.VirtualPCIPassthrough{
+		VirtualDevice: types.VirtualDevice{
+			Key:     deviceKey,
+			Backing: backingInfo,
+		},
+	}
+	return device
```
chrischdi, line 472:
```suggestion
	return &types.VirtualPCIPassthrough{
		VirtualDevice: types.VirtualDevice{
			Key:     deviceKey,
			Backing: backingInfo,
		},
	}
```

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -453,3 +461,55 @@ func getNetworkSpecs(ctx *context.VMContext, devices object.VirtualDeviceList) (
...
+	for _, vGPUDevice := range ctx.VSphereVM.Spec.VirtualMachineCloneSpec.VGPUDevices {
+		backingInfo := &types.VirtualPCIPassthroughVmiopBackingInfo{
+			Vgpu: vGPUDevice.ProfileName,
+		}
+		dynamicDirectPathDevice := createPCIPassThroughDevice(deviceKey, backingInfo)
+		deviceSpecs = append(deviceSpecs, &types.VirtualDeviceConfigSpec{
+			Device:    dynamicDirectPathDevice,
+			Operation: types.VirtualDeviceConfigSpecOperationAdd,
+		})
+		ctx.Logger.V(4).Info("created vGPU device", "vgpu-profile", vGPUDevice.ProfileName)
```
chrischdi, line 511: Should we have a smilar message for the gpu case?

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -453,3 +461,55 @@ func getNetworkSpecs(ctx *context.VMContext, devices object.VirtualDeviceList) (
...
+	device := &types.VirtualPCIPassthrough{
+		VirtualDevice: types.VirtualDevice{
+			Key:     deviceKey,
+			Backing: backingInfo,
+		},
+	}
+	return device
+}
+
+func getGpuSpecs(ctx *context.VMContext) []types.BaseVirtualDeviceConfigSpec {
```
chrischdi, line 475: Could we try to not use the `*context.VMContext` parameter here and use more specific parameters instead?

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -151,8 +151,16 @@ func Clone(ctx *context.VMContext, bootstrapData []byte, format bootstrapv1.Form
 
 	deviceSpecs = append(deviceSpecs, networkSpecs...)
 
-	if err != nil {
-		return errors.Wrapf(err, "error getting network specs for %q", ctx)
+	if len(ctx.VSphereVM.Spec.VirtualMachineCloneSpec.PciDevices) != 0 {
```
chrischdi, line 154: This adds a second codepath which adds PciDevices to a VM. The already existing code is at https://github.com/chrischdi/cluster-api-provider-vsphere/blob/234e7751f7bb09bb438eed6b4feabbc274cae842/pkg/services/govmomi/service.go#L478 .

I'd prefer only having one place to do this so either this here should get removed or the other place. And we have to take care to not break existing use cases.

pkg/services/govmomi/vcenter/clone.go
chrischdi: Also: this function seems to be a duplicate of https://github.com/chrischdi/cluster-api-provider-vsphere/blob/234e7751f7bb09bb438eed6b4feabbc274cae842/pkg/services/govmomi/pci/device.go#L59 ?

chrischdi (commented):
There are some general questions for me:

* I think we should prefer only keeping `reconcilePCIDevices` as the way for the normal PCI Passthrough. Especially because this also handles some condition handling.
* Question: instead of adding vgpu devices directly at clone: should this maybe also be done inside `reconcilePCIDevices`? (because they are just a bit more special PCI devices?)
* API conversion

Note: I'm not able to verify that all of this works because I currently don't have a environment including vgpu's available.

docs/gpu-vgpu.md
puneetkatyal: The same NVIDIA GPU supports multiple vGPU profiles, and the matrix expands when you add more GPU varieties to the mix. For example, in my testing, I use different profiles for different worker nodes for the same workload cluster. I don't think it's useful to have this as an envsubst parameter.

PR #2273, opened 2023-08-23
🌱 lint: fix the JSON encoding check error via errchkjson linter

### LINKED ISSUE #2058: Audit golangci-lint config
Let's compare the golangci-lint config with the one from core CAPI and see if we should add more linters / config etc. (also potentially in core CAPI).

Some observations:
* I saw a few places where we didn't handle returned errors in CAPV. IIRC one of the linters enabled in core CAPI is catching that (not 100% sure). (we probably also need a corresponding exclude)
* We should configure the `importas` linter for at least the most important imports to get some consistency
```[tasklist]
### Tasks
- [x] enable configuration for `importas` (@chrischdi ) #2231
- [x] enable linter `containedctx` (@zhanggbj) #2297
- [x] enable linter `dupword` (@MrDXY) #2260
- [x] enable linter `errchkjson` (@MrDXY) #2273
- [x] enable linter `ginkgolinter` (@nunnatsa) #2282
- [x] enable linter `nolintlint` (@MrDXY) #2294
- [x] enable configuration for `gocritic` (@MrDXY) #2315
- [x] remove `apis/v1alpha3` from `skip-files:` (@adityabhatia) #2378
- [x] remove all additional exclusions below `FIXME: All below excludes should get removed over time.` #2383
- [ ] do a final diff
```

MrDXY:
1. Fix the checked error by errchkjson linter. (remove the safe returns of errors from `json.Marshal`)
2. Enable `errchkjson` check in `.golangci.yml` configuration file.
3. Enable `linters-settings.errchkjson.check-error-free-encoding`. With it set to true, errchkjson does not report about the errors from json encoding functions that are safe to be ignored, because they are not possible to happen.

Part of #2058

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,7 +18,7 @@ linters:
   - dupword
   - durationcheck
   - errcheck
-  # - errchkjson
+  - errchkjson
   - exportloopref
   - gci
   # - ginkgolinter
--- a/packaging/flavorgen/flavors/flavors.go
+++ b/packaging/flavorgen/flavors/flavors.go
@@ -55,7 +55,10 @@ func ClusterClassTemplateWithKubeVIP() []runtime.Object {
 }
 
 func ClusterTopologyTemplateKubeVIP() []runtime.Object {
-	cluster := newClusterClassCluster()
+	cluster, err := newClusterTopologyCluster()
+	if err != nil {
+		return nil
+	}
 	identitySecret := newIdentitySecret()
 	clusterResourceSet := newClusterResourceSet(cluster)
 	crsResourcesCSI := crs.CreateCrsResourceObjectsCSI(&clusterResourceSet)
--- a/packaging/flavorgen/flavors/generators.go
+++ b/packaging/flavorgen/flavors/generators.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -94,7 +95,11 @@ systemd:
         EnvironmentFile=/run/metadata/*`
 )
 
-func newClusterClassCluster() clusterv1.Cluster {
+func newClusterTopologyCluster() (clusterv1.Cluster, error) {
+	variables, err := clusterTopologyVariables()
+	if err != nil {
+		return clusterv1.Cluster{}, errors.Wrap(err, "failed to create ClusterTopologyCluster template")
+	}
 	return clusterv1.Cluster{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: clusterv1.GroupVersion.String(),
@@ -121,17 +126,34 @@ func newClusterClassCluster() clusterv1.Cluster {
 						},
 					},
 				},
-				Variables: clusterTopologyVariables(),
+				Variables: variables,
 			},
 		},
-	}
+	}, nil
 }
 
-func clusterTopologyVariables() []clusterv1.ClusterVariable {
-	sshKey, _ := json.Marshal(env.VSphereSSHAuthorizedKeysVar)
-	controlPlaneIP, _ := json.Marshal(env.ControlPlaneEndpointVar)
-	secretName, _ := json.Marshal(env.ClusterNameVar)
-	kubeVipPod, _ := json.Marshal(kubeVIPPodYaml())
+func clusterTopologyVariables() ([]clusterv1.ClusterVariable, error) {
+	sshKey, err := json.Marshal(env.VSphereSSHAuthorizedKeysVar)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to json-encode variable VSphereSSHAuthorizedKeysVar: %q", env.VSphereSSHAuthorizedKeysVar)
+	}
+	controlPlaneIP, err := json.Marshal(env.ControlPlaneEndpointVar)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to json-encode variable ControlPlaneEndpointVar: %q", env.ControlPlaneEndpointVar)
+	}
+	secretName, err := json.Marshal(env.ClusterNameVar)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to json-encode variable ClusterNameVar: %q", env.ClusterNameVar)
+	}
+	kubeVipPodYaml := kubeVIPPodYaml()
+	kubeVipPod, err := json.Marshal(kubeVipPodYaml)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to json-encode variable kubeVipPod: %q", kubeVipPodYaml)
+	}
+	infraServerValue, err := getInfraServerValue()
+	if err != nil {
+		return nil, err
+	}
 	return []clusterv1.ClusterVariable{
 		{
 			Name: "sshKey",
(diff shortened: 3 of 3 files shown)


.golangci.yml
```diff
@@ -54,6 +54,8 @@ linters:
   - whitespace
 
 linters-settings:
+  errchkjson:
+    check-error-free-encoding: true
```
chrischdi, line 58:
```suggestion
```

We should not set this to be equal to CAPI and resolve the findings instead by handling the error.

chrischdi (changes requested):

.golangci.yml
MrDXY: Sure, I can make the necessary adjustments here, and it totally makes sense that we prefer to keep consistent with CAPI. 

To discuss it further:
This `check-error-free-encoding` parameter here, is just to indicate the `errchkjson` to ignore the errors that will never happen. 

Especially in some [cases:](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/9881385aeae28b7aea3d5e8aa5fedb0c48a46d33/packaging/flavorgen/flavors/generators.go#L131C2-L134C49)
```
sshKey, _ := json.Marshal(env.VSphereSSHAuthorizedKeysVar)
controlPlaneIP, _ := json.Marshal(env.ControlPlaneEndpointVar)
secretName, _ := json.Marshal(env.ClusterNameVar)
kubeVipPod, _ := json.Marshal(kubeVIPPodYaml())
```

the input of the `json.Marsha()` is just a string, which will never cause the function to fail, IIRC. To handle such a situation, is it a good idea to propagate errors to the higher-level calling function? Or do you recall any recommended approaches that I could follow?

chrischdi: I did see that the required changes would only be in the flavorgen tool. So just bubbling them up should be fine.

MrDXY: Got it. Then if I understand it correctly, all we need to do in this PR is just to enable `errchkjson` in `.golangci.yml` file. 😄

sbueringer: Yes. No worries, just ask  all good :)

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -476,10 +476,7 @@ func addVolumes(ctx *vmware.SupervisorMachineContext, vm *vmoprv1.VirtualMachine
 			topology := []map[string]string{
 				{kubeTopologyZoneLabelKey: *zone},
 			}
-			b, err := json.Marshal(topology)
-			if err != nil {
-				return errors.Errorf("failed to marshal zone topology %q: %s", *zone, err)
-			}
+			b, _ := json.Marshal(topology)
```
sbueringer, line 479: I think we have to keep the error handling (same in the suite_test.go) file. What is the linter finding?

MrDXY: I have made some changes here because the `errchkjson` linter has identified that it is safe to ignore the error that was brought by `json.Marshal()` once `check-error-free-encoding` is enabled.

Since we have chosen not to activate this parameter, it appears that no modifications are necessary here.

sbueringer: Ah got it. I'll take a closer look

sbueringer: Ah this probably might change with https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2273/files#r1302524736

.golangci.yml
sbueringer: Hm one moment. I mean when the linter can basically figure out for us that the error cannot happen, I think it makes sense to avoid the unnecessary noise from error handling.

Assuming I understood this correctly. I would prefer keeping this option and also introducing it in core CAPI. 

@chrischdi @killianmuldoon WDYT?

chrischdi: also fine for me 👍

chrischdi: Sorry @MrDXY for the back and forth :-)

sbueringer: Alright. So keeping it as is is fine :)

pkg/services/vmoperator/vmopmachine.go
sbueringer: All good

sbueringer:
/lgtm

/assign @killianmuldoon @chrischdi 

Would be nice if someone has time, to follow-up to do the same in core CAPI. I think it's really helpful if we get rid of error branches that can never happen.

.golangci.yml
killianmuldoon: I'm fine with keeping it as it's the consensus - but I don't really like relying on external linters like this for something this fundamental, at the cost of saving a few lines of `if err == nil`'{}

sbueringer: Good point on relying on an external linter. Basically if they have a bug in their logic we inherit that in CAPI as well

sbueringer: Alright, changing my opinion again. Now I'm in favor of dropping the `check-error-free-encoding` config

chrischdi:
> /lgtm
> 
> /assign @killianmuldoon @chrischdi
> 
> Would be nice, if someone has time, to follow-up to do the same in core CAPI. I think it's really helpful if we get rid of error branches that can never happen.

I can take that :-)

.golangci.yml
MrDXY: @chrischdi It's totally fine! Still, welcome any advice or suggestions you may have.
@sbueringer Got it!

killianmuldoon (commented):
/lgtm

Thanks for this!

chrischdi (commented):
/lgtm
/approve

Thanks!

chrischdi (approved):

packaging/flavorgen/flavors/flavors.go
```diff
@@ -55,7 +55,10 @@ func ClusterClassTemplateWithKubeVIP() []runtime.Object {
 }
 
 func ClusterTopologyTemplateKubeVIP() []runtime.Object {
-	cluster := newClusterClassCluster()
+	cluster, err := newClusterTopologyCluster()
+	if err != nil {
+		return nil
```
sbueringer, line 60: Should we really return nil here?

I would have assumed that if we choose to handle the error we propagate (bubble) it up the entire call hierarchy

MrDXY: I have considered this previously, and I choose to return `nil` for the following reason:

- the function `ClusterTopologyTemplateKubeVIP() []runtime.Object` does not have a return value for the error, so I'll need to modify the API structure to fit it in. While this method is `public,` I'm not sure if this will cause some trouble for other customers who might be using it to generate their templates. 

- Since we know that this method will not actually produce an error (the `json-encoding` error is safe to ignore), it may be okay not to expose it.

MrDXY: But it's also OK. If we consider it inappropriate to ignore the error. I can make some adjustments to the function.

chrischdi: Oh, totally missed that 🤦

sbueringer: Let's make the adjustment to the func to add the error type. I think for this sort of code it should be okay if we evolve it a bit over time.

It just feels consistent to handle the error if we choose to handle it and not drop it one level up the call hierarchy :)

sbueringer: Ah looks like Christian opened a PR already

chrischdi: I created https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2274 so we don't forget to fix that.

In this case its way better to bubble up even if we currently know the **current** code does not produce an error.

There should either only be the two extremes:
* bubbling up
* or ignore the error where we get it, not somewhere in-between

sbueringer: Yup, all good

PR #2294, opened 2023-08-29
🌱 lint: enable nolintlint and fix findings

MrDXY:
1. Enable the `-nolintlint` configuration in `.golangci.yaml` and correct any issues found with ill-formed or insufficiently explained `//nolint` directives.

Part of #2058


killianmuldoon (commented):
/ok-to-test

pkg/services/govmomi/vcenter/clone_test.go
```diff
@@ -16,15 +16,14 @@ limitations under the License.
 
 package vcenter
 
-//nolint:all
+//nolint:goimports
```
killianmuldoon, line 19: This can be fixed by running `make lint-fix` so we shouldn't need to ignore here.

test/helpers/vcsim/simulator.go
```diff
@@ -16,14 +16,13 @@ limitations under the License.
 
 package vcsim
 
-//nolint:all
+//nolint:goimports
```
killianmuldoon, line 19: As above - this can be fixed.

test/helpers/envtest.go
```diff
@@ -75,7 +75,7 @@ func init() {
 	utilruntime.Must(infrav1.AddToScheme(scheme))
 
 	// Get the root of the current file to use in CRD paths.
-	_, filename, _, _ := goruntime.Caller(0) //nolint
+	_, filename, _, _ := goruntime.Caller(0) //nolint:dogsled
```
killianmuldoon, line 78: Wondering if we can replace this call with something less dogsleddy - we have the same issue in core CAPI.

test/helpers/webhook.go
```diff
@@ -79,7 +79,7 @@ func appendWebhookConfiguration(configyamlFile []byte, tag string) ([]*v1.Mutati
 
 func initializeWebhookInEnvironment() {
 	// Get the root of the current file to use in CRD paths.
-	_, filename, _, _ := goruntime.Caller(0) //nolint
+	_, filename, _, _ := goruntime.Caller(0) //nolint:dogsled
```
killianmuldoon, line 82:
```suggestion
	_, filename, _, ok := goruntime.Caller(0)
	if !ok {
		klog.Fatalf("Failed to get information for current file from runtime")
	}
```

And elsewhere for the dogsled would mean we don't have to ignore this and we're handling an error earlier. IMO this is an improvement over ignoring the linter, and we can make the same change in core CAPI.

apis/v1alpha4/conversion.go
```diff
@@ -24,7 +24,7 @@ import (
 
 // Convert_v1beta1_VirtualMachineCloneSpec_To_v1alpha4_VirtualMachineCloneSpec is an autogenerated conversion function.
 //
-//nolint:golint,revive,stylecheck
+//nolint:golint
```
chrischdi, line 27:
```suggestion
```

I think we can just drop this too.

apis/v1alpha4/vspherecluster_conversion.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-//nolint:forcetypeassert,golint,revive,stylecheck
+//nolint:forcetypeassert,golint
```
chrischdi, line 17:
```suggestion
```

chrischdi: also for the other occurencies

pkg/services/govmomi/vcenter/clone_test.go
chrischdi: ack, please drop goimports everywhere and run `make lint-fix` afterwards 👍

test/helpers/envtest.go
chrischdi: maybe `os.Args[0]`?

sbueringer: I'm not too worried about it. It's test code and as long as it works. If we replace it with something else, please let's make absolutely sure that it still works in all cases

test/helpers/webhook.go
sbueringer: Sounds good!

test/helpers/envtest.go
sbueringer: So maybe let's just do the same as suggested in: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2294#discussion_r1308445183?

apis/v1alpha4/conversion.go
sbueringer: I would usually expect this stuff to be covered by exclude rules (but I'm not sure which findings we have here)

chrischdi: I removed the `//nolint:golint` here and it had no findings.

sbueringer: Ah, maybe the nolint finder doesn't find it if we ignore stuff via exludes

chrischdi: jep

pkg/services/govmomi/vcenter/clone_test.go
MrDXY: Hi, folks, I tried to drop `goimports` and use `make lint-fix` to fix this. But the diff with the previous one is it just adds a blank line between the duplicated package name. And when I re-run the `golangci-lint run`, it comes up with another error message, saying `File is not `gci`-ed with --skip-generated -s standard -s default (gci)`. Seem like those two linter attempt to break each other, do we have any suggestions on this?

```
import (
	ctx "context"
	"crypto/tls"
	"testing"

	"github.com/vmware/govmomi/object"
	"github.com/vmware/govmomi/simulator"

	// run init func to register the tagging API endpoints.
	_ "github.com/vmware/govmomi/vapi/simulator"
	"github.com/vmware/govmomi/vim25/types"

	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/session"
)
```

MrDXY: Should we 
- Add an `importas alias` for `github.com/vmware/govmomi/simulator` (Since this file is only for test purposes, and maybe it's not necessary to change the `.golangci.yml` file for it)
- Add `//nolint: gci` to ignore the new error.
- Add `//nolint: goimports`. (current solution)
or guys have any other suggestions?

test/helpers/webhook.go
MrDXY: Noted! 😄

apis/v1alpha4/conversion.go
MrDXY: Thanks for this! 👍

pkg/services/govmomi/vcenter/clone_test.go
sbueringer: What happens if you drop the "blank" line from the config?

MrDXY: > What happens if you drop the "blank" line from the config?

Then everything goes back to its original way, which is exactly the same when `local-prefixes:` is enabled.

MrDXY: Actually, if we don't change the `.golangci.yml` file, in other words, when `local-prefixes` is enabled, then running `golangci-lint run --fix` multiple times will result in the code alternating between having a blank line under the `github.com/vmware/govmomi/simulator` package and not having one.

chrischdi: I think you need to add `custom-order` too so it looks like:
```
  gci:
    sections:
      - standard # Standard section: captures all standard packages.
      - default # Default section: contains all imports that could not be matched to another section type.
      - prefix(sigs.k8s.io/cluster-api-provider-vsphere) # Custom section: groups all imports with the specified Prefix.
    custom-order: true
```

see the [docs](https://golangci-lint.run/usage/linters/#gci)

MrDXY: > I think you need to add `custom-order` too so it looks like:
> 
> ```
>   gci:
>     sections:
>       - standard # Standard section: captures all standard packages.
>       - default # Default section: contains all imports that could not be matched to another section type.
>       - prefix(sigs.k8s.io/cluster-api-provider-vsphere) # Custom section: groups all imports with the specified Prefix.
>     custom-order: true
> ```
> 
> From the docs:
> 
> > If `custom-order` is `true`, it follows the order of `sections` option.
> 
> see the [docs](https://golangci-lint.run/usage/linters/#gci)

Sure! (BTW, the default order of `gci` section is standard > default > custom > blank > dot, and enabling `custom-order` will make it more readable and robust.)

chrischdi: Note: I opened https://github.com/kubernetes-sigs/cluster-api/pull/9339 to do the same there.

MrDXY: So folks, are we good to enable `sections:blank`? 😄

sbueringer: I wouldn't add a separate section for blanks for now. Just because we have to make a case for it in core CAPI and I'm not sure if that is worth the effort

MrDXY: > I wouldn't add a separate section for blanks for now. Just because we have to make a case for it in core CAPI and I'm not sure if that is worth the effort to make the case (and if it's actually worth it introducing it)

Make sense to me, I'll find an alternate solution.

MrDXY: Just come up with an idea, Maybe there is a bug of `goimports` when the imported package has a comment above. And I tested with the below format, it works fine. Without introducing the `sections:blank`.
```
import (
	ctx "context"
	"crypto/tls"
	"testing"

	"github.com/vmware/govmomi/object"
	"github.com/vmware/govmomi/simulator"
	_ "github.com/vmware/govmomi/vapi/simulator" // run init func to register the tagging API endpoints.
	"github.com/vmware/govmomi/vim25/types"

	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/session"
)
```
Does this solution seem like a good option?

sbueringer: That works. Just out of curiosity. If you format the block like this:
```go
import (
	ctx "context"
	"crypto/tls"
	"testing"
	"github.com/vmware/govmomi/object"
	"github.com/vmware/govmomi/simulator"
	// run init func to register the tagging API endpoints.
	_ "github.com/vmware/govmomi/vapi/simulator"
	"github.com/vmware/govmomi/vim25/types"
	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/session"
)
```
And then run `make lint-fix`. It doesn't come up with a valid formatting? (without further findings)

MrDXY: if I run `make lint-fix`, the format will become like this:
```
import (
	ctx "context"
	"crypto/tls"
	"testing"

	"github.com/vmware/govmomi/object"
	"github.com/vmware/govmomi/simulator"

	// run init func to register the tagging API endpoints.
	_ "github.com/vmware/govmomi/vapi/simulator"
	"github.com/vmware/govmomi/vim25/types"

	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/session"
)
```
And when you run `golangci-lint run` command with this format, it will come up with another error, says: "pkg/services/govmomi/vcenter/clone_test.go:26: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(sigs.k8s.io/cluster-api-provider-vsphere) --custom-order (gci)"

chrischdi: I think this is because of the comment. Maybe we can inline the comment

```go
	_ "github.com/vmware/govmomi/vapi/simulator" 	// run init func to register the tagging API endpoints.
```

sbueringer: Got it, thx for trying it. Yup let's go with your suggestion (https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2294#discussion_r1309956345)

controllers/vmware/test/controllers_suite_test.go
```diff
@@ -65,7 +65,10 @@ func getTestEnv() (*envtest.Environment, *rest.Config) {
 	utilruntime.Must(vmwarev1.AddToScheme(scheme.Scheme))
 
 	// Get the root of the current file to use in CRD paths.
-	_, filename, _, _ := goruntime.Caller(0) //nolint
+	_, filename, _, ok := goruntime.Caller(0)
+	if !ok {
+		klog.Fatalf("Failed to get information for current file from runtime")
+	}
```
sbueringer, line 71:
```suggestion
	Expect(ok).To(BeTrue(), "Failed to get information for current file from runtime")
```
Small improvement. This way we trigger a regular gomega error instead of os.Exit vs klog.Fatal (same in similar cases of course)

sbueringer (commented):
Just one nit, otherwise lgtm

chrischdi (commented):
/lgtm

test/helpers/envtest.go
```diff
@@ -75,7 +76,8 @@ func init() {
 	utilruntime.Must(infrav1.AddToScheme(scheme))
 
 	// Get the root of the current file to use in CRD paths.
-	_, filename, _, _ := goruntime.Caller(0) //nolint
+	_, filename, _, ok := goruntime.Caller(0)
+	Expect(ok).To(BeTrue(), "Failed to get information for current file from runtime")
```
chrischdi, line 80: we need to use normal error handling here, no ginkgo. Or initialize ginkgo here but I don't know if this may cause other issues.

chrischdi (commented):
/lgtm cancel

test/helpers/envtest.go
sbueringer: Ah sorry. Yup only in the cases where we already use gomega.

MrDXY: My bad, I'm pretty new to go; I'll take a deeper look into Ginkgo. 
I have reviewed the code and found four occurrences of `_, filename, _, ok := goruntime.Caller(0)`. 
1. `pkg/identity/identity_suite_test.go`
2. `controllers/vmware/test/controllers_suite_test.go`
3.  `test/helpers/envtest.go`
4.  `test/helpers/webhook.go`
The first two files (1 and 2) include the Ginkgo test, while files 3 and 4 only provide APIs to support related tests.

So, for the first two cases, I utilize `Expect(ok).To(BeTrue())`, while for the last two cases, I make use of `klog.Fatalf`. Specifically for case 1, I changed the function name from `init` to `getTestEnv` so that to enable the Gomega to work. In this way, the code structure remains consistent with case 2, which uses the same function name to initialize environment value instead of the Golang original 'init' function.

PR #2299, opened 2023-08-29
:seedling: Add license scan for pull requests

killianmuldoon:
Adds a licence scan inline with what is currently being added in the core Cluster API repo with https://github.com/kubernetes-sigs/cluster-api/pull/9184

Part of https://github.com/kubernetes-sigs/cluster-api/pull/9184

### DIFF
--- a/hack/verify-container-images.sh
+++ b/hack/verify-container-images.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2023 The Kubernetes Authors.
+# Copyright 2022 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


killianmuldoon (commented):
/hold 

Ensure the PR over on the CAPI repo is merged before this one to avoid parallel changes. https://github.com/kubernetes-sigs/cluster-api/pull/9184/

chrischdi (commented):
/lgtm
/approve

hack/verify-container-images.sh
```diff
@@ -22,42 +22,20 @@ if [[ "${TRACE-0}" == "1" ]]; then
...
-
-GO_OS="$(go env GOOS)"
-if [[ "${GO_OS}" == "linux" ]]; then
-  TRIVY_OS="Linux"
-elif [[ "${GO_OS}" == "darwin"* ]]; then
-  TRIVY_OS="macOS"
-fi
-
+VERSION=${1}
 GO_ARCH="$(go env GOARCH)"
```
sbueringer, line 26: Interesting that we don't have this line in core CAPI and probably just assume that GOARCH is set?
Maybe something to fix in core CAPI? (@killianmuldoon)

sbueringer:
Thx!

Just one nit for core CAPI (also commented on the core CAPI PR)

/approve
/hold cancel

hack/verify-container-images.sh
killianmuldoon: The `GO_ARCH` here is only used to give the resulting image a name - I'm not sure if it's used for anything else at this point - the ensure-trivy script sets its own `GO_ARCH`.

Maybe it's a good idea to just remove this here?

sbueringer: I think it's not used to give it a name, it's used to match the name that is produced by make docker-build.

I think if we drop this line we either:
* have to pass it in
* have to find another way to match it correctly (I guess GOARCH might be different on macOS vs Linux (?))

I would avoid implicitly depending on a env variable being exported by someone calling that script

killianmuldoon: You're right - I'll do a PR into CAPI to make them consistent and make CAPI's use of the GO_ARCH more explicit.

PR #2310, opened 2023-08-31
[WIP] :sparkles: Create kube-state-metrics configuration from markers

### LINKED ISSUE #2061: Create kube-state-metrics configuration for CRD metrics
We recently created configuration to expose state metrics for core CAPI CRDs in core CAPI.

We should do the same in CAPV for CRDs for which it makes sense.

Then we should find a pragmatic way how we can use CAPI + CAPV state-metrics via core CAPI `tilt up` at the same time (currently kube-state-metrics only supports a single metrics config

The generated crd-config of core CAPI can be seen here: https://github.com/kubernetes-sigs/cluster-api/blob/main/hack/observability/kube-state-metrics/crd-config.yaml

chrischdi:

Makes use of kube-state-metrics PR https://github.com/kubernetes/kube-state-metrics/pull/2014 to generate the configuration for the CRD metrics.

TODO:
- [ ] wait for https://github.com/kubernetes/kube-state-metrics/pull/2014 to merge
- [ ] squash

Fixes #2061

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,6 +163,12 @@ GOVULNCHECK_VER := v1.0.0
 GOVULNCHECK := $(abspath $(TOOLS_BIN_DIR)/$(GOVULNCHECK_BIN)-$(GOVULNCHECK_VER))
 GOVULNCHECK_PKG := golang.org/x/vuln/cmd/govulncheck
 
+KUBE_STATE_METRICS_VER := e31ed9ab
+KUBE_STATE_METRICS_BIN := kube-state-metrics
+KUBE_STATE_METRICS := $(abspath $(TOOLS_BIN_DIR)/$(KUBE_STATE_METRICS_BIN)-$(KUBE_STATE_METRICS_VER))
+KUBE_STATE_METRICS_PKG := k8s.io/kube-state-metrics/v2
+KUBE_STATE_METRICS_MOD_REPLACE := $(KUBE_STATE_METRICS_PKG)=github.com/chrischdi/kube-state-metrics/v2@$(KUBE_STATE_METRICS_VER)
+
 GOVC_VER := $(shell cat go.mod | grep "github.com/vmware/govmomi" | awk '{print $$NF}')
 GOVC_BIN := govc
 GOVC := $(abspath $(TOOLS_BIN_DIR)/$(GOVC_BIN)-$(GOVC_VER))
@@ -184,6 +190,8 @@ RELEASE_NOTES_VER := $(CAPI_HACK_TOOLS_VER)
 RELEASE_NOTES_BIN := release-notes
 RELEASE_NOTES := $(abspath $(TOOLS_BIN_DIR)/$(RELEASE_NOTES_BIN)-$(RELEASE_NOTES_VER))
 RELEASE_NOTES_PKG := sigs.k8s.io/cluster-api/hack/tools/release
+# let go resolve the tools version via the CAPI version.
+RELEASE_NOTES_MOD_REQUIRE := sigs.k8s.io/cluster-api/hack/tools@$(RELEASE_NOTES_VER)
 
 # Define Docker related variables. Releases should modify and double check these vars.
 REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
@@ -237,7 +245,7 @@ help:  # Display this help
 
 .PHONY: generate
 generate: ## Run all generate targets
-	$(MAKE) generate-modules generate-manifests generate-go-deepcopy generate-go-conversions generate-flavors
+	$(MAKE) generate-modules generate-manifests generate-go-deepcopy generate-go-conversions generate-flavors generate-metrics-config
 
 .PHONY: generate-manifests
 generate-manifests: $(CONTROLLER_GEN) ## Generate manifests e.g. CRD, RBAC etc.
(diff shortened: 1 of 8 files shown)


config/metrics/crd-metrics-config.yaml
```diff
@@ -0,0 +1,676 @@
...
+      name:
+      - metadata
+      - name
+      namespace:
+      - metadata
+      - namespace
+      uid:
+      - metadata
+      - uid
+    metricNamePrefix: capv_vspherecluster
```
sbueringer, line 24:
```suggestion
    metricNamePrefix: capi_vspherecluster
```
I wonder if we should use capi as a prefix.

I'm wondering if we can make the core CAPI dashboard work across infra providers in a generic way. I think technically it should be possible, but it would help if the differences between metrics are as minimal as possible

Let's leave it as is for now. I would experiment a bit with the Grafana dashboard once we get closer to merging this PR (after merging the gen tool I guess)

apis/v1beta1/vspherecluster_types.go
```diff
@@ -108,8 +110,17 @@ type VSphereClusterStatus struct {
...
+// +Metrics:gvk:namePrefix="capv_vspherecluster"
+// +Metrics:labelFromPath:name="name",JSONPath=".metadata.name"
+// +Metrics:labelFromPath:name="namespace",JSONPath=".metadata.namespace"
+// +Metrics:labelFromPath:name="uid",JSONPath=".metadata.uid"
+// +Metrics:labelFromPath:name="cluster_name",JSONPath=.metadata.labels.cluster\.x-k8s\.io/cluster-name
+// +Metrics:info:name="info",help="Information about a vspherecluster.",labelsFromPath={status_vsphere_version:.status.vCenterVersion,spec_server:.spec.server,identity_reference_kind:".spec.identityRef.kind",identity_reference_name:".spec.identityRef.name"}
 type VSphereCluster struct {
-	metav1.TypeMeta   `json:",inline"`
+	metav1.TypeMeta `json:",inline"`
+	// +Metrics:gauge:name="created",JSONPath=".creationTimestamp",help="Unix creation timestamp."
```
sbueringer, line 121: Doesn't really matter for now and it's fine (especially to test / play around with the metrics gen tool). But I would probably prefer having these metrics together with the ones above the VSphereCluster struct instead of here inline

hack/go-tools-build.sh
```diff
@@ -37,7 +37,10 @@ if [ -z "${GOBIN}" ]; then
   exit 1
 fi
 
-CAPI_HACK_TOOLS="sigs.k8s.io/cluster-api/hack/tools"
+export GOWORK="off"
+
+GOMOD_REPLACE=${GOMOD_REPLACE:=}
```
sbueringer, line 42: nit: A bit of godoc around these variables would be good

(in case we keep them, but I guess why not)

Makefile
```diff
@@ -182,6 +188,8 @@ RELEASE_NOTES_VER := $(CAPI_HACK_TOOLS_VER)
 RELEASE_NOTES_BIN := release-notes
 RELEASE_NOTES := $(abspath $(TOOLS_BIN_DIR)/$(RELEASE_NOTES_BIN)-$(RELEASE_NOTES_VER))
 RELEASE_NOTES_PKG := sigs.k8s.io/cluster-api/hack/tools/release
+# let go resolve the tools version via the CAPI version.
+RELEASE_NOTES_MOD_REQUIRE := sigs.k8s.io/cluster-api/hack/tools@$(RELEASE_NOTES_VER)
```
sbueringer, line 192:
```suggestion
RELEASE_NOTES_GO_MOD_REQUIRE := sigs.k8s.io/cluster-api/hack/tools@$(RELEASE_NOTES_VER)
```
nice one!

sbueringer: Hm. Should this mean we can drop something as we now have this?

Makefile
```diff
@@ -807,15 +818,17 @@ $(GOLANGCI_LINT): # Build golangci-lint.
...
 $(GOVC): # Build GOVC.
 	CGO_ENABLED=0 GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) $(GOVC_PKG) $(GOVC_BIN) $(GOVC_VER)
 
 $(KIND): # Build kind.
 	GOBIN=$(TOOLS_BIN_DIR) $(GO_INSTALL) $(KIND_PKG) $(KIND_BIN) $(KIND_VER)
 
-
 $(RELEASE_NOTES): # Build release-notes.
-	GOBIN=$(TOOLS_BIN_DIR) $(GO_TOOLS_BUILD) $(RELEASE_NOTES_PKG) $(RELEASE_NOTES_BIN) $(RELEASE_NOTES_VER)
+	GOBIN=$(TOOLS_BIN_DIR) GOMOD_REQUIRE="$(RELEASE_NOTES_MOD_REQUIRE)" $(GO_TOOLS_BUILD) $(RELEASE_NOTES_PKG) $(RELEASE_NOTES_BIN) $(RELEASE_NOTES_VER)
```
sbueringer, line 831:
```suggestion
	GOBIN=$(TOOLS_BIN_DIR) GO_MOD_REQUIRE="$(RELEASE_NOTES_MOD_REQUIRE)" $(GO_TOOLS_BUILD) $(RELEASE_NOTES_PKG) $(RELEASE_NOTES_BIN) $(RELEASE_NOTES_VER)
```
I think (?)

sbueringer (commented):
Just did a high-level review. So far pretty nice. Thx for working on this!

Makefile
chrischdi: This is there so we can re-use the script for non sigs.k8s.io/hack/* binaries.

Makefile
chrischdi: Nope. Here we need `GO_MOD_REQUIRE`, otherwise the commands would have a failure and the script won't finish.

sbueringer: My suggestion is to rename the variable :) Sorry wasn't pretty clear

chrischdi: ah :D 👍

Makefile
chrischdi: We didn't need it before because it was hardcoded: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2310/files#diff-664d5bc044a3e1f516e637b4f9310cddd237799b6a304ebc37a9984d16744bfbL56

config/metrics/crd-metrics-config.yaml
sbueringer: Resolving here in favor of the other PR

chrischdi:
I'll reopen this once there is progress on the tooling PR

PR #2315, opened 2023-09-04
🌱 lint: enable `gocritic` configuration and fix findings

MrDXY:

1. Enable `gocritic` configuration in the `.golangci.yml` file's `linters-settings` section to check for potential bugs, performance improvements, and style issues.
2. Fix findings related to it.

Part of #2058

### DIFF
--- a/pkg/services/govmomi/util.go
+++ b/pkg/services/govmomi/util.go
@@ -366,7 +366,7 @@ func waitForMacAddresses(ctx *virtualMachineContext) error {
 func getMacAddresses(ctx *virtualMachineContext) ([]string, map[string]int, map[int]string, error) {
 	var (
 		vm                   mo.VirtualMachine
-		macAddresses         []string
+		macAddresses         = make([]string, 0)
 		macToDeviceSpecIndex = map[string]int{}
 		deviceSpecIndexToMac = map[int]string{}
 	)
@@ -375,13 +375,15 @@ func getMacAddresses(ctx *virtualMachineContext) ([]string, map[string]int, map[
 	}
 	i := 0
 	for _, device := range vm.Config.Hardware.Device {
-		if nic, ok := device.(types.BaseVirtualEthernetCard); ok {
-			mac := nic.GetVirtualEthernetCard().MacAddress
-			macAddresses = append(macAddresses, mac)
-			macToDeviceSpecIndex[mac] = i
-			deviceSpecIndexToMac[i] = mac
-			i++
+		nic, ok := device.(types.BaseVirtualEthernetCard)
+		if !ok {
+			continue
 		}
+		mac := nic.GetVirtualEthernetCard().MacAddress
+		macAddresses = append(macAddresses, mac)
+		macToDeviceSpecIndex[mac] = i
+		deviceSpecIndexToMac[i] = mac
+		i++
 	}
 	return macAddresses, macToDeviceSpecIndex, deviceSpecIndexToMac, nil
 }
--- a/test/helpers/vmware/intg_test_context.go
+++ b/test/helpers/vmware/intg_test_context.go
@@ -19,7 +19,6 @@ package vmware
 import (
 	"context"
 	"fmt"
-	"path/filepath"
 
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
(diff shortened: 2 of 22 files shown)


MrDXY:
I'll take a closer look at the tests that have failed.

pkg/services/govmomi/util.go
```diff
@@ -366,7 +366,7 @@ func waitForMacAddresses(ctx *virtualMachineContext) error {
 func getMacAddresses(ctx *virtualMachineContext) ([]string, map[string]int, map[int]string, error) {
 	var (
 		vm                   mo.VirtualMachine
-		macAddresses         []string
+		macAddresses         = make([]string, 0, len(vm.Config.Hardware.Device))
```
sbueringer, line 369:
```suggestion
		macAddresses         = make([]string, 0)
```
I think len is always 0?

test/helpers/vmware/intg_test_context.go
```diff
@@ -109,16 +111,23 @@ func NewIntegrationTestContextWithClusters(goctx context.Context, integrationTes
 	})
 
 	var config *rest.Config
+
+	// get crd paths
+	_, filename, _, ok := goruntime.Caller(0)
+	Expect(ok).To(BeTrue(), "Failed to get information for current file from runtime")
+	root := path.Join(path.Dir(filename), "..", "..")
+	crdPaths := []string{
+		filepath.Join(root, "config", "default", "crd"),
```
sbueringer, line 120: Looks like usually we use config/default/crd/bases

This probably means that these CRDs are not required for this test. Can you try dropping this line and then checking if the test still works?

test/helpers/vmware/intg_test_context.go
```diff
@@ -109,16 +111,23 @@ func NewIntegrationTestContextWithClusters(goctx context.Context, integrationTes
 	})
 
 	var config *rest.Config
+
+	// get crd paths
+	_, filename, _, ok := goruntime.Caller(0)
+	Expect(ok).To(BeTrue(), "Failed to get information for current file from runtime")
+	root := path.Join(path.Dir(filename), "..", "..")
```
sbueringer, line 118:
```suggestion
	root := path.Join(path.Dir(filename), "..", "..","..")
```

Otherwise root is the "test" folder

sbueringer: Slightly wondering if this entire test is either not run or it doesn't need any of the CRDs?

pkg/services/govmomi/util.go
```diff
@@ -375,13 +375,15 @@ func getMacAddresses(ctx *virtualMachineContext) ([]string, map[string]int, map[
...
 	for _, device := range vm.Config.Hardware.Device {
-		if nic, ok := device.(types.BaseVirtualEthernetCard); ok {
-			mac := nic.GetVirtualEthernetCard().MacAddress
-			macAddresses = append(macAddresses, mac)
-			macToDeviceSpecIndex[mac] = i
-			deviceSpecIndexToMac[i] = mac
-			i++
+		nic, ok := device.(types.BaseVirtualEthernetCard)
+		if !ok {
+			continue
```
zhanggbj, line 380:
```suggestion
			if nic, ok := device.(types.BaseVirtualEthernetCard); !ok {
				continue
```

test/helpers/vmware/intg_test_context.go
sbueringer: Let's fix it up here best effort. But as far as I can tell we are not running all tests in CI. I couldn't find any occurrence of "Service Discovery controller integration tests" in the test or integration-test logs. Let's open a separate issue to investigate if we run all our tests in CI.

zhanggbj (commented):
only one nit
/lgtm

zhanggbj:
/hold
(Just noticed other comments)

pkg/services/govmomi/util.go
zhanggbj: Ah, missed it. Please ignore this one

pkg/services/govmomi/util.go
MrDXY: Yes, of course, I missed that.

test/helpers/vmware/intg_test_context.go
MrDXY: Sure, I'll fix the path error first.

test/helpers/vmware/intg_test_context.go
MrDXY: When I removed the CRDs from here, and the `CRDDirectoryPaths` field in `envtest.Environment`, all tests passed successfully.

sbueringer: I also took a closer look. I confirmed that the test is actually executed (so we don't need a follow-up for that).

It just looks like we really don't need the CRDs. I think this is because this func creates an env test environment (apiserver+etcd) to simulate a workload cluster (or in supervisor mode called a guest cluster).

On these clusters we simply don't need the CAPV CRDs.

So I would suggest lets drop the CRD deployment here (~l.115-122 + l.130)

sbueringer:
Last nit, otherwise good to merge from my side

sbueringer:
Thank you!!

/approve

/hold cancel

PR #2322, opened 2023-09-05
🌱 Add finalizer for VSphereClusterIdentity

killianmuldoon:
Add a finalizer to ensure the cleanup for this type completes.

### DIFF
--- a/controllers/vsphereclusteridentity_controller.go
+++ b/controllers/vsphereclusteridentity_controller.go
@@ -119,6 +119,12 @@ func (r clusterIdentityReconciler) Reconcile(ctx _context.Context, req reconcile
 		return ctrl.Result{}, r.reconcileDelete(ctx, identity)
 	}
 
+	// Add a finalizer and requeue to ensure that the secret is deleted when the identity is deleted.
+	if !ctrlutil.ContainsFinalizer(identity, infrav1.VSphereClusterIdentityFinalizer) {
+		ctrlutil.AddFinalizer(identity, infrav1.VSphereClusterIdentityFinalizer)
+		return reconcile.Result{}, nil
+	}
+
 	// fetch secret
 	secret := &corev1.Secret{}
 	secretKey := client.ObjectKey{
@@ -171,6 +177,8 @@ func (r clusterIdentityReconciler) reconcileDelete(ctx _context.Context, identit
 	err := r.Client.Get(ctx, secretKey, secret)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
+			// The secret no longer exists. Remove the finalizer from the VSphereClusterIdentity.
+			ctrlutil.RemoveFinalizer(identity, infrav1.VSphereClusterIdentityFinalizer)
 			return nil
 		}
 		return err
@@ -185,5 +193,10 @@ func (r clusterIdentityReconciler) reconcileDelete(ctx _context.Context, identit
 	if err := r.Client.Update(ctx, secret); err != nil {
 		return err
 	}
-	return r.Client.Delete(ctx, secret)
+	if err := r.Client.Delete(ctx, secret); err != nil {
+		return err
+	}
+	// Remove the finalizer from the identity as all cleanup is complete.
+	ctrlutil.RemoveFinalizer(identity, infrav1.VSphereClusterIdentityFinalizer)
+	return nil
 }
--- a/apis/v1beta1/vsphereclusteridentity_types.go
+++ b/apis/v1beta1/vsphereclusteridentity_types.go
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	SecretIdentitySetFinalizer = "vspherecluster/infrastructure.cluster.x-k8s.io"
+	SecretIdentitySetFinalizer      = "vspherecluster/infrastructure.cluster.x-k8s.io"
+	VSphereClusterIdentityFinalizer = "vsphereclusteridentity/infrastructure.cluster.x-k8s.io"
 )
 
 type VSphereClusterIdentitySpec struct {


controllers/vsphereclusteridentity_controller.go
```diff
@@ -181,9 +189,16 @@ func (r clusterIdentityReconciler) reconcileDelete(ctx _context.Context, identit
...
 		ctrlutil.RemoveFinalizer(secret, legacyIdentityFinalizer)
 	}
+
+	// TODO (killianmuldoon): Consider whether we should have a finalizer on the secret and, if so, whether this finalizer should be renamed.
 	ctrlutil.RemoveFinalizer(secret, infrav1.SecretIdentitySetFinalizer)
 	if err := r.Client.Update(ctx, secret); err != nil {
 		return err
 	}
-	return r.Client.Delete(ctx, secret)
+	if err := r.Client.Delete(ctx, secret); err != nil {
```
killianmuldoon, line 198: It's not clear to me that there has to be a 1:1:1 relationship between the secret and the VSphereClusterIdentity and the Cluster, but it looks like the intended behaviour of this controller is to clean up the secret it references.

The secret can't be deleted until the finalizer is removed, but it's the same finalizer that is added by each VSphereClusterIdentity.

I think there's a bunch of weirdness here that IMO is a bug, but may be being relied on as a feature by someone.

Maybe the best course is to open an issue to establish rules for the lifecycle of this object?

sbueringer: I think it's not a 1:1:1 relationship based on 
> * Credentials via a VSphereClusterIdentity: `VSphereClusterIdentity` is a cluster-scoped resource and enables multiple VSphereClusters to share the same set of credentials. The namespaces that are allowed to use the VSphereClusterIdentity can also be configured via a `LabelSelector`.
> https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/0300fab21584ad0f026e15b6fb5208708b37ff88/docs/identity_management.md#L12-L11

killianmuldoon: You're right - it's many Clusters to one VSphereClusterIdentity to one Secret. The VSphereClusterIdentity is linked to the CAPV deployment rather than any individual cluster or group of clusters.

The current setup of these finalizers is still odd to me. I think this change is necessary to ensure the vsphereCluster deletes the secret as a cleanup - but do we need the secret to be protected from deletion by a finalizer as well?

sbueringer: Based on the proposal (20210311-single-controller-multitenancy.md) I think VSphereClusterIdentity was introduced partially to be re-used for multiple clusters

sbueringer: Based on this, is the question then should be allowed to use a single secret in multiple VSphereClusterIdentities?

controllers/vsphereclusteridentity_controller.go
```diff
@@ -115,6 +115,12 @@ func (r clusterIdentityReconciler) Reconcile(ctx _context.Context, req reconcile
 		}
 	}()
 
+	// Add a finalizer and requeque to ensure that the secret is deleted when the identity is deleted.
```
sbueringer, line 118:
```suggestion
	// Add a finalizer and requeue to ensure that the secret is deleted when the identity is deleted.
```

controllers/vsphereclusteridentity_controller.go
```diff
@@ -115,6 +115,12 @@ func (r clusterIdentityReconciler) Reconcile(ctx _context.Context, req reconcile
...
+	if !ctrlutil.ContainsFinalizer(identity, infrav1.VSphereClusterIdentityFinalizer) {
+		ctrlutil.AddFinalizer(identity, infrav1.VSphereClusterIdentityFinalizer)
+		return reconcile.Result{Requeue: true}, nil
+	}
+
```
sbueringer, line 123: This section should be below the deletionTimestamp check (~ l. 127).

As soon as the deletionTimestamp is set the apiserver will block adding a new finalizers (cf with other controllers)

controllers/vsphereclusteridentity_controller.go
```diff
@@ -181,9 +189,16 @@ func (r clusterIdentityReconciler) reconcileDelete(ctx _context.Context, identit
 	if ctrlutil.ContainsFinalizer(secret, legacyIdentityFinalizer) {
 		ctrlutil.RemoveFinalizer(secret, legacyIdentityFinalizer)
 	}
+
+	// TODO (killianmuldoon): Consider whether we should have a finalizer on the secret and, if so, whether this finalizer should be renamed.
 	ctrlutil.RemoveFinalizer(secret, infrav1.SecretIdentitySetFinalizer)
```
sbueringer, line 194: Don't we already have a finalizer on the Secret?

controllers/vsphereclusteridentity_controller.go
sbueringer: l.141 seems to block that a secret is used by multiple VSphereClusterIdentities

sbueringer: > IMO we should deprecate the finalizer on the Secret used by the VSphereClusterIdentity.

Why?

controllers/vsphereclusteridentity_controller.go
killianmuldoon: Q is whether it should be removed.

controllers/vsphereclusteridentity_controller.go
killianmuldoon: My question is why is it there in the first place. 

I don't think we generally have finalizers on our other secrets. And I think putting the finalizer on the bottom node in the dependency graph is an unusual pattern.

controllers/vsphereclusteridentity_controller.go
sbueringer: Missed the "should"

controllers/vsphereclusteridentity_controller.go
killianmuldoon: I would be in favor of getting rid of it, as it seems the whole finalizer flow for this controller never worked as expected. 

The flow before this PR was to (in no necessary order):
- delete the vsphereclusteridentity
- remove the finalizer from the secret manually
- delete the secret


But fine to leave it as it doesn't do any damage anymore AFAICT.

sbueringer: I think I"m missing something crucial. I would have thought the following removes the finalizer and deletes the secret:
```go
	ctrlutil.RemoveFinalizer(secret, infrav1.SecretIdentitySetFinalizer)
	if err := r.Client.Update(ctx, secret); err != nil {
		return err
	}
	return r.Client.Delete(ctx, secret)
```

Just because we didn't have a finalizer on the VSphereClusterIdentity it doesn't mean that reconcileDelete wasn't run. It basically just means that if we miss the delete event we don't hit reconcileDelete (e.g. if the controller was down). I would expect that today in 99% of the cases we go through a reconcileDelete. Probably missing something, but I think only in edge cases the secret is orphaned after the VSphereClusterIdentity is gone

killianmuldoon: > I would expect that today in 99% of the cases we go through a reconcileDelete. Probably missing something, but I think only in edge cases the secret is orphaned after the VSphereClusterIdentity is gone

Not sure how it compares, but in my testing this never happened. The secret was always orphaned and the finalizer was always left in place. I'm not sure why as I would have expected the same given the code. My guess is we hit the "not found" case for the `VSphereClusterIdentity` very quickly as there's nothing slowing down its deletion.

sbueringer: Ah interesting. That definitely gives us other options. I assumed we have a comparable case to: https://github.com/kubernetes-sigs/cluster-api/issues/6863 (+/- that in the CAPI issue it's fine to miss a delete event if the controller is done)

sbueringer: Considering this I think I have to change my opinion in the other direction. I would expect that we 100% hit the notfound case if we don't have a finalizer on the VSphereClusterIdentity (assuming my theory about how informers work is correct)

sbueringer: Okay so it should be safe to assume that reconcileDelete was basically never run. Which also means it doesn't change anything if we drop the code there.

What would you like to do?

1. add the VSphereClusterIdentity finalizer in this PR
2. open an issue/PR to get rid of the secret finalizer? (still not sure if we should fix the finalizer handling instead because I don't know for which reasons it was implemented and if we should rather make sure it works instead of dropping it)

killianmuldoon: It does look like we add the finalizer in the case we use the raw secret in the VSphereCluster, so it's not inconsistent within CAPV. I'm fine with keeping it as-is and just merging this PR (once I fix it up).

controllers/vsphereclusteridentity_controller.go
sbueringer: I think this one is still open

sbueringer: Nope it isn't :/ My bad. git pull error :D

controllers/vsphereclusteridentity_controller.go
```diff
@@ -119,6 +119,12 @@ func (r clusterIdentityReconciler) Reconcile(ctx _context.Context, req reconcile
 		return ctrl.Result{}, r.reconcileDelete(ctx, identity)
 	}
 
+	// Add a finalizer and requeue to ensure that the secret is deleted when the identity is deleted.
+	if !ctrlutil.ContainsFinalizer(identity, infrav1.VSphereClusterIdentityFinalizer) {
+		ctrlutil.AddFinalizer(identity, infrav1.VSphereClusterIdentityFinalizer)
+		return reconcile.Result{Requeue: true}, nil
```
sbueringer, line 125:
```suggestion
		return reconcile.Result{}, nil
```
The update will requeue automatically (the godoc is maybe a bit misleading in this regard)

sbueringer (commented):
Just one nit

chrischdi (commented):
/lgtm

sbueringer:
/approve


Thank you!!

sbueringer:
@killianmuldoon Bug fix => cherry-pick back was the idea, right?

PR #2345, opened 2023-09-11
🌱 config: add metrics configuration for kube-state-metrics

chrischdi:

This PR adds a custom resource config for kube-state-metrics to scrape metrics via kube-state-metrics from the CAPV resources.

This could be used together with https://github.com/kubernetes-sigs/cluster-api/pull/9390 and the following configuration in `tilt-settings.yaml`:

```yaml
deploy_kustomizations:
  kube-state-metrics-custom-resource-capv: ../cluster-api-provider-vsphere/config/metrics
```

**Note:** depends on the path to be correct here.

Fixes #

### DIFF
--- a/config/metrics/crd-metrics-config.yaml
+++ b/config/metrics/crd-metrics-config.yaml
@@ -1,0 +1,676 @@
+---
+kind: CustomResourceStateMetrics
+spec:
+  resources:
+  - errorLogV: 0
+    groupVersionKind:
+      group: infrastructure.cluster.x-k8s.io
+      kind: VSphereCluster
+      version: v1beta1
+    labelsFromPath:
+      cluster_name:
+      - metadata
+      - labels
+      - cluster.x-k8s.io/cluster-name
+      name:
+      - metadata
+      - name
+      namespace:
+      - metadata
+      - namespace
+      uid:
+      - metadata
+      - uid
+    metricNamePrefix: capi_vspherecluster
+    metrics:
+    - each:
+        info:
+          labelFromKey: ""
+          labelsFromPath:
+            paused_value: []
+          path:
+          - metadata
+          - annotations
+          - cluster.x-k8s.io/paused
+        type: Info
+      help: Whether the vspherecluster is paused and any of its resources will not
+        be processed by the controllers.
+      name: annotation_paused
+    - each:
+        gauge:
+          labelFromKey: ""
+          nilIsZero: false
+          path:
+          - metadata
+          - creationTimestamp
+          valueFrom: null
+        type: Gauge
+      help: Unix creation timestamp.
+      name: created
+    - each:
+        info:
+          labelFromKey: ""
+          labelsFromPath:
+            identity_reference_kind:
+            - spec
+            - identityRef
+            - kind
+            identity_reference_name:
+            - spec
+            - identityRef
+            - name
+            spec_server:
+            - spec
+            - server
+            status_vsphere_version:
+            - status
+            - vCenterVersion
+          path: null
+        type: Info
+      help: Information about a vspherecluster.
+      name: info
+    - each:
+        info:
+          labelFromKey: ""
+          labelsFromPath:
+            owner_is_controller:
+            - controller
+            owner_kind:
+            - kind
+            owner_name:
+            - name
+            owner_uid:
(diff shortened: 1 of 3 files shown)


config/metrics/crd-metrics-config.yaml
```diff
@@ -0,0 +1,676 @@
...
+      name:
+      - metadata
+      - name
+      namespace:
+      - metadata
+      - namespace
+      uid:
+      - metadata
+      - uid
+    metricNamePrefix: capv_vspherecluster
```
chrischdi, line 24: @sbueringer : wdyt: Use `capv_` or `capi_` as prefix?

When having `capv_` as prefix: is there a potential of overlap of names in-between providers?

sbueringer: Let's go with capi. As we always add the CRD name after we would also have a CRD name conflict if two providers have a CRD with the same name. That seems to be okay

config/metrics/crd-metrics-config.yaml
```diff
@@ -0,0 +1,676 @@
...
+    metrics:
+    - each:
+        info:
+          labelFromKey: ""
```
sbueringer, line 28: Why do we need this? Same below for nilIsZero, resourcePlural. Is there an omitempty missing for generation? Or is it intentional?

sbueringer: Same for errorLogV? (what is this doing? :))

config/metrics/crd-metrics-config.yaml
```diff
@@ -0,0 +1,676 @@
...
+            address:
+            - address
+            type:
+            - type
+          path:
+          - status
+          - addresses
+        type: Info
+      help: Address information about a vspheremachine.
+      name: addresses
```
sbueringer, line 455:
```suggestion
      name: status_addresses
```
Same in core CAPI I think

sbueringer: Same for VSphereVM

config/metrics/crd-metrics-config.yaml
```diff
@@ -0,0 +1,676 @@
...
+      - namespace
+      uid:
+      - metadata
+      - uid
+    metricNamePrefix: capv_vspherevm
+    metrics:
+    - each:
+        info:
+          labelFromKey: ""
+          labelsFromPath:
```
sbueringer, line 569: Intentionally different from VSphereMachine?
```yaml
        labelsFromPath:
          address:
          - address
          type:
          - type
```

config/metrics/crd-metrics-config.yaml
chrischdi: Jep, it is missing omitempty.

errorLogV `	// ErrorLogV defines the verbosity threshold for errors logged for this resource.` which is the log level for klog.

config/metrics/crd-metrics-config.yaml
chrischdi: Yes.

VSphereMachine has:
```
	Addresses []clusterv1.MachineAddress `json:"addresses,omitempty"`
```

VSphereVM has:
```
	Addresses []string `json:"addresses,omitempty"`
```

config/metrics/crd-metrics-config.yaml
sbueringer: Thx!

sbueringer:
Thank you!!

Fine to merge for me as is

/lgtm
/approve

/hold
(up to you)

Would be good to change the omitempty thing in the gentool so that our final configuration won't contain the unnecessary zero values. But definitely fine to merge for me as is

chrischdi:
> Thank you!!
> 
> Fine to merge for me as is
> 
> /lgtm /approve
> 
> /hold (up to you)
> 
> Would be good to change the omitempty thing in the gentool so that our final configuration won't contain the unnecessary zero values. But definitely fine to merge for me as is

This omitempty part is not on the gentool side, its the kube-state-metrics config struct itself. But yeah, I agree.

sbueringer:
> This omitempty part is not on the gentool side, its the kube-state-metrics config struct itself. But yeah, I agree.

Ah good info and yeah makes sense.

PR #2350, opened 2023-09-12
:seedling: Including all packages in coverage

laozc:

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -478,7 +478,7 @@ test-junit: $(SETUP_ENVTEST) $(GOTESTSUM) $(GOVC) ## Run unit tests
 
 .PHONY: test-cover
 test-cover: ## Run unit tests and generate a coverage report
-	$(MAKE) test TEST_ARGS="$(TEST_ARGS) -coverprofile=coverage.out"
+	$(MAKE) test TEST_ARGS="$(TEST_ARGS) -coverpkg=./... -coverprofile=coverage.out"
 	./hack/codecov-ignore.sh
 	go tool cover -func=coverage.out -o coverage.txt
 	go tool cover -html=coverage.out -o coverage.html


Makefile
```diff
@@ -470,7 +470,7 @@ test-junit: $(SETUP_ENVTEST) $(GOTESTSUM) $(GOVC) ## Run unit tests
 
 .PHONY: test-cover
 test-cover: ## Run unit tests and generate a coverage report
-	$(MAKE) test TEST_ARGS="$(TEST_ARGS) -coverprofile=coverage.out"
+	$(MAKE) test TEST_ARGS="$(TEST_ARGS) -coverpkg=./apis/...,./controllers/...,./pkg/...,./internal/... -coverprofile=coverage.out"
```
killianmuldoon, line 473: Would all packages be covered by `./...` instead of naming each package individually.

sbueringer: I think it should work

sbueringer:
Interesting that this increases our coverage :D I would have expected the reverse

Details: https://app.codecov.io/gh/kubernetes-sigs/cluster-api-provider-vsphere/pull/2350/indirect-changes

Makefile
killianmuldoon: I believe so. IMO we should include the entire codebase for our coverage baseline.

sbueringer: Yup will include that ones as well. I think it's fine to include all code. We just shouldn't use the one number to judge anything by. But we would always have the problem. There are more packages where we could discuss if they require test coverage or not. I think in the end we have to look package by package where we should add more tests and we can just skip packaging there.

laozc:
It looks like we're doing quite good with just one line change.

sbueringer:
Thank you!

Yup looks a lot better than expected already

/lgtm
/approve

Issue #2076, opened 2023-07-24
Improve CAPV logging


sbueringer:
Logging is a crucial to make it easier to understand controller behavior and troubleshoot issues.

While onboarding to CAPV we identified a few quickwins which should improve log quality significantly:
* It is a best practice in Kubernetes to add k/v pairs of the involved objects to a log line (https://cluster-api.sigs.k8s.io/developer/logging.html#keyvalue-pairs)
  * We should also add k/v pairs for all involved objects (and their owner hierarchy) that we retrieve at the beginning of the Reconcile
* We should also consistently use contextual logging (i.e. taking the logger from the ctx). This allows propagation of k/v pairs across the entire call stack
  * This also includes using the ctx that controller-runtime passes into the Reconcile func instead of some global ctx that is the same across Reconcile calls.

The overall goal of this issue is to improve the logs so troubleshooting becomes easier. Adding k/v pairs across the board will make it very easy to correlate logs e.g. for a specific Machine across controllers.

Additional notes:
* Let's take a look at how event recorders are setup in core CAPI vs CAPV
* Take a look at controller logger setup (e.g. regarding name). IIRC we shouldn't need any logger on the controllers though
* Audit all log calls for additional k/v pairs
* Ensure "Named" is set correctly on all controllers
* Let's double check the k/v pairs we add in core CAPI to ensure we can cross-reference everything
 

Prior art in core CAPI:
* https://github.com/kubernetes-sigs/cluster-api/pull/6150
* https://github.com/kubernetes-sigs/cluster-api/pull/7075
* https://github.com/kubernetes-sigs/cluster-api/pull/7152
* https://github.com/kubernetes-sigs/cluster-api/pull/7534

### Tasks

Concrete tasks for now (I have some follow-ups to audit, but let's do this afterwards):
* [x] Wait until context refactoring is done: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/2295
* Refactor controllers (concrete tasks below):
  * [x] vmware/vspherecluster_reconciler.go
  * [x] clustermodule_reconciler.go
  * [x] serviceaccount_controller.go
  * [x] servicediscovery_controller.go
  * [x] vspherecluster_controller.go
  * [x] vspherecluster_reconciler.go
  * [x] vsphereclusteridentity_controller.go @Madhur97
    * #2426
[...]

PR #2352, opened 2023-09-12
🌱 Improve logging of VSphereCluster, ProviderServiceAccount and ServiceDiscovery controllers

sbueringer:
Set up golang context when initiating controller manager and also pass it down
to VSphereCluster controller

Signed-off-by: Gong Zhang <gongz@vmware.com>

Have to do a few more things + the underlying PR has to be merged first

Part of #2076


controllers/clustermodule_reconciler.go
```diff
@@ -47,29 +49,31 @@ import (
...
+		Client:               controllerManagerCtx.Client,
+		ClusterModuleService: clustermodule.NewService(controllerManagerCtx, controllerManagerCtx.Client),
 	}
 }
 
-func (r Reconciler) Reconcile(clusterCtx *capvcontext.ClusterContext) (reconcile.Result, error) {
-	clusterCtx.Logger.Info("reconcile anti affinity setup")
+func (r Reconciler) Reconcile(ctx context.Context, clusterCtx *capvcontext.ClusterContext) (reconcile.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Reconcile cluster modules")
```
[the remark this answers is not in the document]
sbueringer: Thx, good catch! Will wait for some other reviews and then fix it

controllers/clustermodule_reconciler.go
```diff
@@ -166,28 +170,34 @@ func (r Reconciler) Reconcile(clusterCtx *capvcontext.ClusterContext) (reconcile
...
 		Namespace:       obj.GetNamespace(),
 		Labels:          obj.GetLabels(),
 		OwnerReferences: obj.GetOwnerReferences(),
 	})
 	if err != nil {
-		r.Logger.Error(err, "failed to get owner cluster")
+		log.Error(err, "failed to get owner cluster")
 		return nil
 	}
+	log = log.WithValues("Cluster", klog.KObj(cluster))
```
randomvariable, line 184: Nit: Use consistent casing for log keys. E.g. you have `Cluster` here, but `targetObjectName` above.

randomvariable:
Whilst you're doing this, i would enable `loggercheck` in golangci-lint to catch when someone hasn't completed kv pairs.

controllers/clustermodule_reconciler.go
sbueringer: Thx good point.

The conventions are a bit messy unfortunately. Upstream is using `targetObjectName`, `pod`, ... . We tried to do this in controller-runtime but there is simply no way to generate the right key for a CRD because there is no way to tell how to lower-case a kind correctly :). This is why we ended up with `Cluster`, `VSphereCluster`, etc. (we just use the kind directly).

Based on that we can either only use upper-case for kind and only be inconsistent with upstream there or we can do this consistently and be in-consistent with k/k everywhere :D (I think I would go with `Pod`, `Cluster`, `TargetObjectName`)

sbueringer:
> Whilst you're doing this, i would enable `loggercheck` in golangci-lint to catch when someone hasn't completed kv pairs.

Thx, nice one! Will do

controllers/serviceaccount_controller.go
```diff
@@ -249,74 +255,81 @@ func (r ServiceAccountReconciler) ReconcileNormal(guestClusterCtx *vmwarecontext
...
 }
 
 // Ensure service accounts from provider spec is created.
-func (r ServiceAccountReconciler) ensureProviderServiceAccounts(guestClusterCtx *vmwarecontext.GuestClusterContext, pSvcAccounts []vmwarev1.ProviderServiceAccount) error {
+func (r *ServiceAccountReconciler) ensureProviderServiceAccounts(ctx context.Context, guestClusterCtx *vmwarecontext.GuestClusterContext, pSvcAccounts []vmwarev1.ProviderServiceAccount) error {
+	log := ctrl.LoggerFrom(ctx)
+
 	for i, pSvcAccount := range pSvcAccounts {
+		// Note: We have to use := here to not overwrite log & ctx outside the for loop.
+		log := log.WithValues("ProviderServiceAccount", klog.KRef(pSvcAccount.Namespace, pSvcAccount.Name))
```
killianmuldoon, line 278: Nit: Might be less confusing if we use new variable names here instead of shadowing.

```suggestion
		log := log.WithValues("ProviderServiceAccount", klog.KRef(pSvcAccount.Namespace, pSvcAccount.Name))
```

killianmuldoon (commented):
Thanks for taking this on! Just some nits and clarifications.

Happy for the nits to be taken care of in a follow up - this PR is so large and touches so many lines I think we should prioritize merging it to avoid rebase hell. 

I would generally prefer if we tried to keep changes like this more atomic in future - this PR has so many small changes I think it's very difficult to review.

/lgtm
/approve
/hold

killianmuldoon:
/lgtm remove

Should be squashed before merge.

controllers/clustermodule_reconciler.go
sbueringer: I'll move this to a separate issue. I don't want to violate upstream conventions across the code base without a discussion

sbueringer: Issue: https://github.com/kubernetes-sigs/cluster-api/issues/9447

controllers/serviceaccount_controller.go
sbueringer: I prefer shadowing. If we don't shadow it's very easy to accidentally use the wrong log or context.

We intentionally shadow log & ctx here so that within the scope of the for loop the shadowed log/ctx has to be used

sbueringer: I think the main difficulty here is deciding when we have to use `:=` over `=`, but we would have the same problem if in the cases where we currently use `:=` we would use a different variable name. Once we made the right call we're good.

controllers/vspherecluster_reconciler.go
```diff
@@ -541,22 +556,27 @@ func setOwnerRefsOnVsphereMachines(ctx context.Context, clusterCtx *capvcontext.
 }
 
 // TODO: zhanggbj: Add golang context after refactoring ClusterModule controller and reconciler.
```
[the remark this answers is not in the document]
sbueringer: Yup, thx!

controllers/vspheremachine_controller.go
```diff
@@ -135,7 +136,7 @@ func AddMachineControllerToManager(controllerCtx *capvcontext.ControllerManagerC
 		// Watch any VirtualMachine resources owned by this VSphereMachine
 		builder.Owns(&vmoprv1.VirtualMachine{})
 		r.VMService = &vmoperator.VmopMachineService{}
-		networkProvider, err := inframanager.GetNetworkProvider(controllerCtx)
+		networkProvider, err := inframanager.GetNetworkProvider(context.TODO(), controllerCtx.Client, controllerCtx.NetworkProvider)
```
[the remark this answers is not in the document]
sbueringer: Mostly. I think we should use context.TODO() if we think that eventually a ctx should be passed through. It should not matter how long it takes us to get one there.




```
// TODO returns a non-nil, empty Context. Code should use context.TODO when
// it's unclear which Context to use or it is not yet available (because the
// surrounding function has not yet been extended to accept a Context
// parameter).
func TODO() Context {
	return todo
}
```

sbueringer:
> I would generally prefer if we tried to keep changes like this more atomic in future - this PR has so many small changes I think it's very difficult to review.

I agree. This escalated a bit and it was a mixture of keeping the PR safe and in some cases fixing up things while I'm already there. 

A majority of the "scope creep" was because the only way to safely refactor this was to remove the `*context.ControllerContext` and `Logger` entirely from `ClusterContext`. The result was that I had to refactor some other controllers as well, the same applies to packages like network & vmoperator that are used by the controllers.

I had an intermediate version where I only tried to stop using these fields and not setting them in some controller but this blew up pretty badly in some cases I missed in unit tests (panic's all over the place). Because I couldn't be sure that we have full coverage I thought it's the safer option to play it safe and drop them entirely instead of hoping that I got all relevant places and then risking panic's in rare edge cases in production later.

sbueringer:
Pushed some fixes. Will squash later after @chrischdi 's review and corresponding fixes

sbueringer:
codecov failed because of the flake in our tests

controllers/vspherecluster_controller.go
```diff
@@ -66,33 +66,30 @@ func AddClusterControllerToManager(ctx context.Context, controllerCtx *capvconte
...
 		clusterControlledTypeName = reflect.TypeOf(clusterControlledType).Elem().Name()
 		clusterControlledTypeGVK  = infrav1.GroupVersion.WithKind(clusterControlledTypeName)
-		controllerNameShort       = fmt.Sprintf("%s-controller", strings.ToLower(clusterControlledTypeName))
-		controllerNameLong        = fmt.Sprintf("%s/%s/%s", controllerCtx.Namespace, controllerCtx.Name, controllerNameShort)
+		controllerNameShort       = strings.ToLower(clusterControlledTypeName)
+		controllerNameLong        = fmt.Sprintf("%s/%s/%s", controllerManagerCtx.Namespace, controllerManagerCtx.Name, controllerNameShort)
 	)
 	if supervisorBased {
 		clusterControlledTypeGVK = vmwarev1.GroupVersion.WithKind(clusterControlledTypeName)
 		controllerNameShort = fmt.Sprintf("%s-supervisor-controller", strings.ToLower(clusterControlledTypeName))
```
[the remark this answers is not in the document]
chrischdi: or we add `-controller` again in l 69?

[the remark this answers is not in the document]
chrischdi: Or did the above result in `*-controller-controller`?

controllers/serviceaccount_controller.go
```diff
@@ -249,74 +255,81 @@ func (r ServiceAccountReconciler) ReconcileNormal(guestClusterCtx *vmwarecontext
...
-	err := util.SetControllerReferenceWithOverride(&pSvcAccount, &svcAccount, clusterCtx.Scheme)
+
+	err := util.SetControllerReferenceWithOverride(&pSvcAccount, &svcAccount, r.Client.Scheme())
 	if err != nil {
 		return err
 	}
-	logger.V(4).Info("Creating service account")
-	err = clusterCtx.Client.Create(clusterCtx, &svcAccount)
+	log.V(4).Info("Creating service account")
+	err = r.Client.Create(ctx, &svcAccount)
```
chrischdi, line 332: This now uses the client to the management cluster instead of the one received from the cluster Tracker!

chrischdi: Similar for other functions.

controllers/serviceaccount_controller.go
```diff
@@ -249,74 +255,81 @@ func (r ServiceAccountReconciler) ReconcileNormal(guestClusterCtx *vmwarecontext
...
 		if guestClusterCtx.Cluster != nil && annotations.IsPaused(guestClusterCtx.Cluster, &(pSvcAccounts[i])) {
-			r.Logger.V(4).Info("ProviderServiceAccount %s/%s linked to a cluster that is paused or has pause annotation",
-				pSvcAccount.Namespace, pSvcAccount.Name)
+			log.V(4).Info("ProviderServiceAccount linked to a cluster that is paused or has pause annotation")
 			continue
 		}
 
 		// 1. Create service accounts by the name specified in Provider Spec
-		if err := r.ensureServiceAccount(guestClusterCtx.ClusterContext, pSvcAccount); err != nil {
+		if err := r.ensureServiceAccount(ctx, pSvcAccount); err != nil {
```
chrischdi, line 287: We need a way to pass down the guestClusterCtx.Client or get a client from the cluster tracker

chrischdi (commented):
One thing to maybe take another look (I may missed something there?!)

A lot of awesome small fixes are included too! 🎉

pkg/services/network/network_test.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package network
```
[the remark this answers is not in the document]
sbueringer: Feel free to open an issue/PR for it. Probably it should be one file per provider (including dummy)

controllers/vspherecluster_controller.go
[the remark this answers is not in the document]
sbueringer: Hm no it was more that I fixed the non-supervisor to the appropriate name and didn't touch the supervisor name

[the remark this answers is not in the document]
sbueringer: I dropped the entire Named thing. Controller-runtime defaults to the lower-case Kind as controller. I think this is good enough. An example log looks like this

> I0918 10:27:20.552138       1 vspherecluster_reconciler.go:232] "Reconciling VSphereCluster" controller="vspherecluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="VSphereCluster" VSphereCluster="quick-start-f5q6kr/quick-start-mai91w" namespace="quick-start-f5q6kr" name="quick-start-mai91w" reconcileID=b0a382df-587e-464a-93ef-75c95684545d Cluster="quick-start-f5q6kr/quick-start-mai91w"

controller is the name here. Should be good enough that we can tell by the controllerGroup if it's supervisor or not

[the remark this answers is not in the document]
chrischdi: fine for me. The others will get refactored too to drop `-controller` from their names?

controllers/serviceaccount_controller.go
sbueringer: The clusterCtx.Client was also a mgmt cluster client

sbueringer: Please double check though. This would be a bad mistake (in fact I double checked exactly this because I made that mistake elsewhere during implementation)

controllers/serviceaccount_controller.go
chrischdi: ack! `guestClusterCtx.ClusterContext.Client` was the client to the MC, so using r.Client instead now which is 100% fine! Thx for claryfication

controllers/serviceaccount_controller.go
chrischdi: Jep you are right.
`ctx.GuestClient` is the only client to the guest cluster.

sbueringer:
test ?

PR #2353, opened 2023-09-14
🌱 Get rid of unstructured type in VSphereVM reconciler

### LINKED ISSUE #2056: Consider using VSphereVM instead of unstructured in VimMachineService.ReconcileNormal
In VimMachineService.ReconcileNormal we initially use VSphereVM when we call createOrPatchVSphereVM. Then suddenly we're switching to Unstructured. Even though the object will always be a VSphereVM.

I wonder if it would make sense to consistently use VSphereVM instead. I'm probably missing something but it would make a lot of the code a lot easier to just work with the typed object.

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/7c043af770a42ac6b464895a1dd513790649ae44/pkg/services/vimmachine.go#L131-L174

zhanggbj:

I'm assuming this is an intermediate state when to support supervisor mode, so it uses unstructured type in VSphereVM reconciler to support two modes. As we now have [vmoperator](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/tree/main/pkg/services/vmoperator) to reconcile VM in supervisor mode, I think we are good to get rid of unstructured type in VSphereVM reconciler, but better to have a review from folks who're more familiar with supervisor mode.

Fixes #2056

### DIFF
--- a/pkg/services/vimmachine.go
+++ b/pkg/services/vimmachine.go
@@ -25,8 +25,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/utils/integer"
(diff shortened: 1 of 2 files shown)


pkg/services/vimmachine.go
```diff
@@ -127,32 +125,17 @@ func (v *VimMachineService) ReconcileNormal(machineCtx capvcontext.MachineContex
...
-	vmObj.SetGroupVersionKind(vm.GetObjectKind().GroupVersionKind())
-	vmObj.SetAPIVersion(vm.GetObjectKind().GroupVersionKind().GroupVersion().String())
-	vmObj.SetKind(vm.GetObjectKind().GroupVersionKind().Kind)
-
 	// Waits the VM's ready state.
-	if ok, err := v.waitReadyState(vimMachineCtx, vmObj); !ok {
-		if err != nil {
-			return false, errors.Wrapf(err, "unexpected error while reconciling ready state for %s", vimMachineCtx)
-		}
+	if ok := v.waitReadyState(vimMachineCtx, vm); !ok {
```
sbueringer, line 129: Let's inline this function. Basically just check the field here directly
```
	if !vm.Status.Ready {
```
(I think the log in l.130 is okay we don't need another one)

sbueringer (commented):
Really nice cleanup. One nit.

Let's ignore the non-ideal logs for now (that we already had before this PR), we'll improve that when we improve logging of the entire controller

pkg/services/vimmachine.go
zhanggbj: Sure, will update it.

zhanggbj:
@sbueringer all comments are addressed, it's ready for a second look, thx!

sbueringer:
Thx!

/lgtm

/hold
Just want to merge https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2352 before

sbueringer:
/hold cancel
Ah nope we should be good. No overlap

chrischdi:
Very nice cleanup

/approve

/hold pending CI

sbueringer:
Entirely up to you, but e2e-full automatically blocks merging so we don't have to manage the hold manually

chrischdi:
> Entirely up to you, but e2e-full automatically blocks merging as soon as it's triggered so we don't have to manage the hold manually (Just fyi)

Ah right, this was recently changed :-)

/hold cancel

sbueringer:
Not really :D. We always had this in CAPV (since before we took over). We just changed it in core CAPI to be the same.

killianmuldoon (commented):
/retest

PR #2364, opened 2023-09-18
:seedling: Upgrade CSI driver to v3.1.0

### LINKED ISSUE #2331: CSI controller crashes in deployments using current templates
/kind bug

CSI plugin needs `insecure-flag` to be set without trusted CA even when the thumbprint is provided.
The templates in the CAPV source does not contain this flag.

CAPV does not require the insecure flag when the thumbprint is provided.
This should be aligned in the future to avoid customer confusion.

**Environment:**

laozc:
We're still using 2.1.0 in the templates while the latest CSI driver now is 3.1.0 with more features.

Fixes #2331


killianmuldoon (commented):
/test pull-cluster-api-provider-vsphere-e2e-full-main

packaging/flavorgen/flavors/crs/csi.go
```diff
@@ -36,76 +31,88 @@ import (
...
 // Also appends the resources to the CRS.
 func CreateCrsResourceObjectsCSI(crs *addonsv1.ClusterResourceSet) []runtime.Object {
-	serviceAccount := cloudprovider.CSIControllerServiceAccount()
-	serviceAccount.TypeMeta = metav1.TypeMeta{
-		Kind:       "ServiceAccount",
-		APIVersion: corev1.SchemeGroupVersion.String(),
-	}
-	serviceAccountSecret := newSecret(serviceAccount.Name, serviceAccount)
-	appendSecretToCrsResource(crs, serviceAccountSecret)
+	csiNamespace := cloudprovider.CSIDriverNamespace()
```
chrischdi, line 34: not: why not group all csi relevant objects to a single configmap?

templates/cluster-template-external-loadbalancer.yaml
```diff
@@ -448,49 +665,58 @@ data:
...
-              value: /etc/cloud/csi-vsphere.conf
+            - name: X_CSI_SPEC_DISABLE_LEN_CHECK
+              value: "true"
             - name: LOGGER_LEVEL
               value: PRODUCTION
-            - name: X_CSI_LOG_LEVEL
-              value: INFO
-            - name: NODE_NAME
+            - name: GODEBUG
+              value: x509sha1=1
```
chrischdi, line 704: Can we add a comment why this one is needed? Or is this simply copy/paste?

chrischdi (commented):
One question but no bigger opinion on this.
As of that:

/lgtm

templates/cluster-template-external-loadbalancer.yaml
laozc: It's copied from the CSI manifest.
The flag is used to enable the deprecated SHA1 certificate support in some enterprise environment for on-prem Kubernetes installations.
https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/616f90f7e38a24fceda3673819abafd20da08859/manifests/vanilla/vsphere-csi-driver.yaml#L367C1-L368C32

packaging/flavorgen/flavors/crs/csi.go
laozc: All the content merged into a new `csi-manifests` config map now.

chrischdi (commented):
Wondering of it makes sense to copy all this over and make go code out of it, or if we should instead use go embed and only add the things we need (e.g. csi-config-secret)

However for now its good as it is.

/lgtm

chrischdi (commented):
Wondering of it makes sense to copy all this over and make go code out of it, or if we should instead use go embed and only add the things we need (e.g. csi-config-secret)

However for now its good as it is.

/lgtm

sbueringer:
> Wondering of it makes sense to copy all this over and make go code out of it, or if we should instead use go embed and only add the things we need (e.g. csi-config-secret)
> 

This seems pretty had to maintain / keeping in sync going forward. @randomvariable WDYT? Should we follow the route to convert everything to go code and somehow try to keep this in sync or should we follow or more direct copying YAML over & diffing approach?

The latter seems for more maintainable to me

laozc:
There is a full manifest at
https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/master/manifests/vanilla/vsphere-csi-driver.yaml

I believe the CSI driver image in the file is generated for their CI.
For long term maintenance, it would be better to replicate the manifest content from CSI driver with our own customization.

chrischdi:
Maybe a copy + kustomize patches for things we need to change could be a good way forward.

sbueringer:
Okay let's go ahead with this. I took a look and I didn't see anything suspicious, but I also didn't diff it which seems a crazy amount of work. I'm fine with merging this PR after rebase & when all e2e tests are green.

But we have to find another way to do this. Let's create an issue to move this to using the upstream YAML + kustomizations on top. This is absolutely not sustainable (already wasn't before this PR, which is why I'm fine with merging this for now)

Issue #2396, opened 2023-09-26
IPAddress Gateway should be concidered optional


ekarlso:
/kind feature

**Describe the solution you'd like**
It was decided in previous discussions that Gateway should be optional so you can have interfaces without defualt gw here
https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/pull/118
https://github.com/kubernetes-sigs/cluster-api/issues/8536
https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/issues/70

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-vsphere version: 
1.8.x

christianang:
This is fine with me. There was a lot of back and forth on this across various repos and it seem like it was made optional everywhere, but CAPV. So this make sense to me to remove the restriction from CAPV.

cc @schrej @srm09

chrischdi:
Fore more context also see the discussion at https://kubernetes.slack.com/archives/CKFGK3SSD/p1695731228748929

As of today it is already okay for CAPV to have no gateway set for ipv6. With this use case we should do the same for ipv4 propably inside the `parseGateway` function:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/pkg/services/govmomi/ipam/parse.go#L57-L59

```go
	if ipamAddress.Spec.Gateway == "" {
		return nil, nil
	}
```

/triage accepted

PR #2397, opened 2023-09-26
:bug: Remove the conditional on ipv6

ekarlso:
Signed-off-by: ekarlso <endre.karlson@xait.com>

Fixes #

### DIFF
--- a/pkg/services/govmomi/ipam/parse.go
+++ b/pkg/services/govmomi/ipam/parse.go
@@ -51,10 +51,10 @@ func parseAddressWithPrefix(ipamAddress *ipamv1.IPAddress) (netip.Prefix, error)
 // does not conflict with the gateway addresses parsed from other
 // ipamv1.IPAddresses on the current device. Gateway addresses must be the same
 // family as the address on the ipamv1.IPAddress. Gateway addresses of one
-// family must match the other addresses of the same family. IPv4 Gateways are
-// required, but IPv6 gateways are not.
+// family must match the other addresses of the same family. A gateway address
+// is optional. If it is not set this function returns `nil, nil`.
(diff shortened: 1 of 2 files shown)


zhanggbj (commented):
Thanks for the PR!
Just one nit, as we want to make Gateway optional for both IPv4 and IPv6, could you please also update the comment in L55 to make it consistent. Thanks!

zhanggbj:
Hi @ekarlso, please also sign the CLA, you could follow the guide above :-)

chrischdi:
This is for #2396

pkg/services/govmomi/ipam/parse.go
```diff
@@ -54,7 +54,7 @@ func parseAddressWithPrefix(ipamAddress *ipamv1.IPAddress) (netip.Prefix, error)
 // family must match the other addresses of the same family. IPv4 Gateways are
 // required, but IPv6 gateways are not.
```
chrischdi, line 55: We should also update the godoc here

```suggestion
// family must match the other addresses of the same family. A gateway address
// is optional. If it is not set this function returns `nil, nil`.
```

killianmuldoon:
/retitle :bug: Remove the conditional on ipv6

chrischdi:
This PR also needs to fix the relevant tests :-)

killianmuldoon (commented):
Thanks!

/lgtm
/approve

killianmuldoon:
Should we backport this to 1.8 / 1.7?

chrischdi:
> Should we backport this to 1.8 / 1.7?

I was thinking about that too. Would make sense to me to do so. It reduces the restrictions on the field so it should be okay?!

killianmuldoon:
Yeah - absolutely fine with backporting this as a bug.

killianmuldoon:
@chrischdi  - looks like we might have an issue with the cherrypick bot :thinking: 

I'll open an issue over at test infra

Issue #2384, opened 2023-09-25
Review use of `nolint` directives across code


killianmuldoon:
CAPV uses many linters from golangci-lint, but many of these linters are skipped in some packages due to the use of `nolint` directives. The usage of these should be reviewed and as many as possible should be removed. This work should be tackled as a follow-on from https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/2058

Some initial findings:

- [x] godot: 
  - Should only be used in old APIVersions. Should be removed everywhere else.
    - #2383 
  - [x] gosec: 8 usages. @Ankitasw 
    - https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2406
- Allowed in test code.
  - [x] staticcheck: 	8  @xiujuanx 
  - https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2427
  - [x] nestif: 	6  @adityabhatia 
    - https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2410
  - [x]  gocritic: 	3  @MrDXY 
    - #2399
  - [x]  gocyclo: 	2 @Ankitasw 
    - https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2407
  - [x] errcheck: 2 @Ankitasw
    - #2409
  - [x] gocognit: 	2 @Ankitasw 
    - https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2407
  - [x] revive: 	2 @Ankitasw
    - #2409
  - [x] parallelltest: 	1  @Ankitasw 
    - https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2407
  - [x] wrapcheck: 	1 @Ankitasw 
    - https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2407
  - [x] prealloc: 	1 @Ankitasw
    - #2409
  - [x] containedctx @zhanggbj 
    - #2297 
- related to https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/2295
  - [x] goconst:	 	1 @Ankitasw
    - #2409
  - [x] ineffassign: 	1 @Ankitasw
    - #2409

The above usages should be investigated and we should remove where possible. Where not possible we should add comments beside the lint ignore directives explaining why these linters are ignored.

zhanggbj:
- [ ]  containedctx
[...]

PR #2399, opened 2023-09-27
🌱 Remove lint ignore for `gocritic`

MrDXY:
Remove `//nolint:gocritic` linter and fix findings.
1. Remove `nolint:gocritic` for `/controllers/controllers_suite_test.go`
    1. To properly execute the `teardown`, I remove the `defer` keyword and place it before the `os.Exit(code)` statement.
2. Remove `nolint:gocritic` for `/controllers/vspheremachine_controller.go`
    1. To simplify the code, I merge the conditions `r.supervisorBased` and `machineCtx.GetCluster().Status.InfrastructureReady` using an `else if` statement as recommended by `gocritic`.
3. Remove `nolint:gocritic` for `pkg/services/govmomi/service.go`
    1. Reconstruct the `switch` logic, and use an if to replace it.
    2. I noticed a TODO comment from @akutz  that suggests using a switch instead of an if statement to check metadata encoding. However, I have not seen any follow-ups on this. As everything seems to be working well, I have refactored it to an if structure. If anyone has a better solution or suggestion, or, perhaps I made a mistake, please feel free to comment on this PR.

Part of #2384

### DIFF
--- a/controllers/controllers_suite_test.go
+++ b/controllers/controllers_suite_test.go
@@ -51,9 +51,9 @@ var (
 
 func TestMain(m *testing.M) {
 	setup()
-	defer teardown()
 	code := m.Run()
-	os.Exit(code) //nolint:gocritic
+	teardown()
+	os.Exit(code)
 }
 
 func setup() {
--- a/controllers/vspheremachine_controller.go
+++ b/controllers/vspheremachine_controller.go
@@ -291,18 +291,19 @@ func (r *machineReconciler) reconcileNormal(ctx context.Context, machineCtx capv
 		return reconcile.Result{}, nil
 	}
 
-	//nolint:gocritic
-	if r.supervisorBased {
-		err := r.setVMModifiers(ctx, machineCtx)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-	} else {
+	// Cluster `.status.infrastructureReady == false is handled differently depending on if the machine is supervisor based.
+	// 1) If the Cluster is not supervisor-based mark the VMProvisionedCondition false and return nil.
+	// 2) If the Cluster is supervisor-based continue to reconcile as InfrastructureReady is not set to true until after the kube apiserver is available.
+	if !r.supervisorBased {
 		// vmwarev1.VSphereCluster doesn't set Cluster.Status.Ready until the API endpoint is available.
 		if !machineCtx.GetCluster().Status.InfrastructureReady {
 			log.Info("Cluster infrastructure is not ready yet")
 			conditions.MarkFalse(machineCtx.GetVSphereMachine(), infrav1.VMProvisionedCondition, infrav1.WaitingForClusterInfrastructureReason, clusterv1.ConditionSeverityInfo, "")
 			return reconcile.Result{}, nil
+		}
+	} else {
+		if err := r.setVMModifiers(ctx, machineCtx); err != nil {
+			return reconcile.Result{}, err
 		}
 	}
 
--- a/pkg/services/govmomi/service.go
+++ b/pkg/services/govmomi/service.go
@@ -529,17 +529,9 @@ func (vms *VMService) getMetadata(ctx *virtualMachineContext) (string, error) {
 
 	var metadataBase64 string
 	for _, ec := range obj.Config.ExtraConfig {
-		if optVal := ec.GetOptionValue(); optVal != nil {
-			// TODO(akutz) Using a switch instead of if in case we ever
-			//             want to check the metadata encoding as well.
-			//             Since the image stamped images always use
-			//             base64, it should be okay to not check.
-			//nolint:gocritic
-			switch optVal.Key {
-			case guestInfoKeyMetadata:
-				if v, ok := optVal.Value.(string); ok {
-					metadataBase64 = v
-				}
+		if optVal := ec.GetOptionValue(); optVal != nil && optVal.Key == guestInfoKeyMetadata {
+			if v, ok := optVal.Value.(string); ok {
+				metadataBase64 = v
 			}
 		}
 	}


killianmuldoon (commented):
/ok-to-test

controllers/controllers_suite_test.go
```diff
@@ -51,9 +51,9 @@ var (
 
 func TestMain(m *testing.M) {
 	setup()
-	defer teardown()
 	code := m.Run()
-	os.Exit(code) //nolint:gocritic
+	teardown()
```
killianmuldoon, line 55: Wow. That's an important change

controllers/vspheremachine_controller.go
```diff
@@ -291,19 +291,16 @@ func (r *machineReconciler) reconcileNormal(ctx context.Context, machineCtx capv
...
+	} else if !machineCtx.GetCluster().Status.InfrastructureReady {
 		// vmwarev1.VSphereCluster doesn't set Cluster.Status.Ready until the API endpoint is available.
-		if !machineCtx.GetCluster().Status.InfrastructureReady {
-			log.Info("Cluster infrastructure is not ready yet")
-			conditions.MarkFalse(machineCtx.GetVSphereMachine(), infrav1.VMProvisionedCondition, infrav1.WaitingForClusterInfrastructureReason, clusterv1.ConditionSeverityInfo, "")
-			return reconcile.Result{}, nil
-		}
+		log.Info("Cluster infrastructure is not ready yet")
+		conditions.MarkFalse(machineCtx.GetVSphereMachine(), infrav1.VMProvisionedCondition, infrav1.WaitingForClusterInfrastructureReason, clusterv1.ConditionSeverityInfo, "")
+		return reconcile.Result{}, nil
```
killianmuldoon, line 303: While we're here we should improve the comments and maybe try to simplify this bit of code. Maybe something like:

```go
	// Cluster `.status.infrastructureReady == false is handled differently depending on if the machine is supervisor based.
	// 1) If the Cluster is not supervisor-based mark the VMProvisionedCondition false and return nil.
	// 2) If the Cluster is supervisor-based continue to reconcile as InfrastructureReady is not set to true until after the kube apiserver is available.
	if !machineCtx.GetCluster().Status.InfrastructureReady {
		if !r.supervisorBased {
			log.Info("Cluster infrastructure is not ready yet")
			conditions.MarkFalse(machineCtx.GetVSphereMachine(), infrav1.VMProvisionedCondition, infrav1.WaitingForClusterInfrastructureReason, clusterv1.ConditionSeverityInfo, "")
		}
		// vmwarev1.VSphereCluster doesn't set Cluster.Status.Ready until the API endpoint is available so continue to reconcile.
		// Set modifiers on the supervisor VM.
		err := r.setVMModifiers(ctx, machineCtx)
		if err != nil {
		}
	}
```
We should definitely add commments like that - not sure if the code is really better, especially given how we have `if supervisorBased` in other parts of the code.

@chrischdi WDYT?

killianmuldoon (commented):
Looks good - one idea to add comments.

And you managed to fix what looks like a bad bug while you were at it!

controllers/vspheremachine_controller.go
chrischdi: I think for simplicity and readability we should keep the

```go
if r.supervisorBased {
  // supervisor
} else {
  // non supervisor
}
```

pattern. Otherwise it may get more complicated and reduce readability.

pkg/services/govmomi/service.go
```diff
@@ -530,17 +530,10 @@ func (vms *VMService) getMetadata(ctx *virtualMachineContext) (string, error) {
...
-			//nolint:gocritic
-			switch optVal.Key {
-			case guestInfoKeyMetadata:
-				if v, ok := optVal.Value.(string); ok {
-					metadataBase64 = v
-				}
+		if optVal := ec.GetOptionValue(); optVal != nil && optVal.Key == guestInfoKeyMetadata {
+			if v, ok := optVal.Value.(string); ok {
+				metadataBase64 = v
+				break
```
chrischdi, line 536:
```suggestion
```

We should not break here.

If `guestInfoKeyMetadata` is multiple times in `ExtraConfig`, the last entry wins. If we break here the first wins.

So this would be a breaking change.

(I did not dig into if it is possible that we have multiple entries).

killianmuldoon: You're right. I think it's best to preserve it as is, even if it seems unlikely there's a reasonable config that includes this multiple times.

controllers/controllers_suite_test.go
MrDXY: Indeed, using `os.Exit` could prevent `defer teardown` from executing. Removing `defer` and making such adjustments allows `teardown` to function properly.

pkg/services/govmomi/service.go
MrDXY: Make sense. I will remove the `break`.

controllers/vspheremachine_controller.go
MrDXY: Hi folks, I think both of your suggestions make sense to me. What about:
1. To set the outer condition, use `if !machineCtx.GetCluster().Status.InfrastructureReady`, and for the inner condition, use `if r.supervisorBased`.
    1. From my understanding, if the cluster is already ready, there may be no need to keep running 'setVMModifiers'.
2. Still using the `r.supervisorBased` structure to ensure readability.
3. Add the necessary comments as mentioned above.

And the code would be something like:
```
	// Cluster `.status.infrastructureReady == false is handled differently depending on if the machine is supervisor based.
	// 1) If the Cluster is not supervisor-based mark the VMProvisionedCondition false and return nil.
	// 2) If the Cluster is supervisor-based continue to reconcile as InfrastructureReady is not set to true until after the kube apiserver is available.
	if !machineCtx.GetCluster().Status.InfrastructureReady {
		if !r.supervisorBased {
			log.Info("Cluster infrastructure is not ready yet")
			conditions.MarkFalse(machineCtx.GetVSphereMachine(), infrav1.VMProvisionedCondition, infrav1.WaitingForClusterInfrastructureReason, clusterv1.ConditionSeverityInfo, "")
		} else {
			// vmwarev1.VSphereCluster doesn't set Cluster.Status.Ready until the API endpoint is available so continue to reconcile.
			// Set modifiers on the supervisor VM.
			err := r.setVMModifiers(ctx, machineCtx)
			if err != nil {
			}
		}
	}
```

MrDXY: Guys, I have an idea. I reversed the if condition, and the `golangci-lint` check passed. Is this okay with you?
```
	// Cluster `.status.infrastructureReady == false is handled differently depending on if the machine is supervisor based.
	// 1) If the Cluster is not supervisor-based mark the VMProvisionedCondition false and return nil.
	// 2) If the Cluster is supervisor-based continue to reconcile as InfrastructureReady is not set to true until after the kube apiserver is available.
	if !r.supervisorBased {
		// vmwarev1.VSphereCluster doesn't set Cluster.Status.Ready until the API endpoint is available.
		if !machineCtx.GetCluster().Status.InfrastructureReady {
			log.Info("Cluster infrastructure is not ready yet")
			conditions.MarkFalse(machineCtx.GetVSphereMachine(), infrav1.VMProvisionedCondition, infrav1.WaitingForClusterInfrastructureReason, clusterv1.ConditionSeverityInfo, "")
		}
	} else {
		err := r.setVMModifiers(ctx, machineCtx)
		if err != nil {
		}
	}
```

controllers/vspheremachine_controller.go
```diff
@@ -291,19 +291,21 @@ func (r *machineReconciler) reconcileNormal(ctx context.Context, machineCtx capv
...
+	if !r.supervisorBased {
 		// vmwarev1.VSphereCluster doesn't set Cluster.Status.Ready until the API endpoint is available.
 		if !machineCtx.GetCluster().Status.InfrastructureReady {
 			log.Info("Cluster infrastructure is not ready yet")
 			conditions.MarkFalse(machineCtx.GetVSphereMachine(), infrav1.VMProvisionedCondition, infrav1.WaitingForClusterInfrastructureReason, clusterv1.ConditionSeverityInfo, "")
 			return reconcile.Result{}, nil
 		}
+	} else {
+		err := r.setVMModifiers(ctx, machineCtx)
+		if err != nil {
```
chrischdi, line 306:
```suggestion
		if err := r.setVMModifiers(ctx, machineCtx); err != nil {
```

chrischdi (commented):
One last nit.

:-)

chrischdi (commented):
/lgtm

MrDXY:
Hi folks, do we still have any comments/suggestions or concerns on this PR? If everything looks good, please kindly approve at your convenience.

chrischdi:
I'm good with the change, however merging will currently fail due to issues in CI...

/approve

killianmuldoon:
>I'm good with the change, however merging will currently fail due to issues in CI...

Maybe not - do we re-run test runs when we merge? Let's see :thinking:

PR #2405, opened 2023-09-28
:bug: flatcar: Run kubeadm after containerd

johananl:

Ensure kubeadm runs after containerd to avoid a race condition between the two.

We intentionally add an `After=` directive and not a `Requires=` directive to avoid breaking things for distros which use Ignition and don't use containerd.

See https://github.com/kubernetes-sigs/image-builder/issues/939.

None

### DIFF
--- a/packaging/flavorgen/flavors/generators.go
+++ b/packaging/flavorgen/flavors/generators.go
@@ -90,6 +90,8 @@ systemd:
         # kubeadm must run after coreos-metadata populated /run/metadata directory.
         Requires=coreos-metadata.service
         After=coreos-metadata.service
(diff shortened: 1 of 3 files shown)


chrischdi:
/lgtm

Thanks for the PR. This makes sense to me.

Running the e2e-full tests to see if we hit any issue. This also includes ignition.

killianmuldoon (commented):
Ignition quickstart is also run as part of `pull-cluster-api-provider-vsphere-e2e-main` btw - but good to run the full suite here.

killianmuldoon (commented):
/lgtm
/approve

Thanks!

PR #2395, opened 2023-09-26
🌱 Improve logging for VSphereMachine

Ankitasw:
**What this PR does / why we need it**:
Improve logging for VSphereMachine controller and underlying packages.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Part of #2076

PR #2413, opened 2023-10-04
🌱 Removal of if block usage wrt logging in vsphere machine controller

Ankitasw:
Resolving open comments from #2395

Fixes #

### DIFF
--- a/pkg/services/vmoperator/vmopmachine.go
+++ b/pkg/services/vmoperator/vmopmachine.go
@@ -80,19 +80,17 @@ func (v *VmopMachineService) ReconcileDelete(ctx context.Context, machineCtx cap
 	}
 	log.V(2).Info("Destroying VM")
 
-	// If debug logging is enabled, report the number of vms in the cluster before and after the reconcile
-	if log.V(5).Enabled() {
+	vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
+	log.V(5).Info("Trace Destroy PRE: VirtualMachines", "vmcount", len(vms), "error", err)
+	defer func() {
 		vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
-		log.Info("Trace Destroy PRE: VirtualMachines", "vmcount", len(vms), "error", err)
-		defer func() {
-			vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
-			log.Info("Trace Destroy POST: VirtualMachines", "vmcount", len(vms), "error", err)
-		}()
-	}
+		log.V(5).Info("Trace Destroy POST: VirtualMachines", "vmcount", len(vms), "error", err)
+	}()
 
 	// First, check to see if it's already deleted
 	vmopVM := vmoprv1.VirtualMachine{}
 	if err := v.Client.Get(ctx, types.NamespacedName{Namespace: supervisorMachineCtx.Machine.Namespace, Name: supervisorMachineCtx.Machine.Name}, &vmopVM); err != nil {
+		// If debug logging is enabled, report the number of vms in the cluster before and after the reconcile
 		if apierrors.IsNotFound(err) {
 			supervisorMachineCtx.VSphereMachine.Status.VMStatus = vmwarev1.VirtualMachineStateNotFound
 			return err
@@ -116,7 +114,6 @@ func (v *VmopMachineService) ReconcileDelete(ctx context.Context, machineCtx cap
 		supervisorMachineCtx.VSphereMachine.Status.VMStatus = vmwarev1.VirtualMachineStateError
 		return err
 	}
-
 	supervisorMachineCtx.VSphereMachine.Status.VMStatus = vmwarev1.VirtualMachineStateDeleting
 	return nil
 }
@@ -143,14 +140,12 @@ func (v *VmopMachineService) ReconcileNormal(ctx context.Context, machineCtx cap
 	log.V(2).Info("Reconciling VM")
 
 	// If debug logging is enabled, report the number of vms in the cluster before and after the reconcile
-	if log.V(5).Enabled() {
-		vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
-		log.Info("Trace ReconcileVM PRE: VirtualMachines", "vmcount", len(vms), "error", err)
-		defer func() {
-			vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
-			log.Info("Trace ReconcileVM POST: VirtualMachines", "vmcount", len(vms), "error", err)
-		}()
-	}
+	vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
+	log.V(5).Info("Trace ReconcileVM PRE: VirtualMachines", "vmcount", len(vms), "error", err)
+	defer func() {
+		vms, err = v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
+		log.V(5).Info("Trace ReconcileVM POST: VirtualMachines", "vmcount", len(vms), "error", err)
+	}()
 
 	// Set the VM state. Will get reset throughout the reconcile
 	supervisorMachineCtx.VSphereMachine.Status.VMStatus = vmwarev1.VirtualMachineStatePending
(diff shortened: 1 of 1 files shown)


Ankitasw:
cc @killianmuldoon

chrischdi:
/retest

(flake at https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/2413/pull-cluster-api-provider-vsphere-test-main/1709492360339001344 , will take a look at this separately)

Ankitasw:
@chrischdi I think it's failing because of the changes in this PR, I will take a look locally.

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -118,6 +112,8 @@ func (v *VmopMachineService) ReconcileDelete(ctx context.Context, machineCtx cap
 	}
 
 	supervisorMachineCtx.VSphereMachine.Status.VMStatus = vmwarev1.VirtualMachineStateDeleting
+	vms, err = v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
+	log.V(5).Info("Trace Destroy POST: VirtualMachines", "vmcount", len(vms), "error", err)
```
killianmuldoon, line 116: This changes the behaviour. Previously this would log on any exit.

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -81,14 +81,8 @@ func (v *VmopMachineService) ReconcileDelete(ctx context.Context, machineCtx cap
...
 
 	// If debug logging is enabled, report the number of vms in the cluster before and after the reconcile
-	if log.V(5).Enabled() {
-		vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
-		log.Info("Trace Destroy PRE: VirtualMachines", "vmcount", len(vms), "error", err)
-		defer func() {
-			vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
-			log.Info("Trace Destroy POST: VirtualMachines", "vmcount", len(vms), "error", err)
-		}()
-	}
```
killianmuldoon, line 91:
```
	defer func() {
		vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
		log.V(5).Info("Trace Destroy POST: VirtualMachines", "vmcount", len(vms), "error", err)
	}()
```

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -143,14 +139,8 @@ func (v *VmopMachineService) ReconcileNormal(ctx context.Context, machineCtx cap
...
 
 	// If debug logging is enabled, report the number of vms in the cluster before and after the reconcile
-	if log.V(5).Enabled() {
-		vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
-		log.Info("Trace ReconcileVM PRE: VirtualMachines", "vmcount", len(vms), "error", err)
-		defer func() {
-			vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
-			log.Info("Trace ReconcileVM POST: VirtualMachines", "vmcount", len(vms), "error", err)
-		}()
-	}
```
killianmuldoon, line 153:
```
	defer func() {
		vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
		log.Info("Trace ReconcileVM POST: VirtualMachines", "vmcount", len(vms), "error", err)
	}()
```

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -226,6 +216,8 @@ func (v *VmopMachineService) ReconcileNormal(ctx context.Context, machineCtx cap
 	// Mark the VSphereMachine as Ready
 	supervisorMachineCtx.VSphereMachine.Status.Ready = true
 	conditions.MarkTrue(supervisorMachineCtx.VSphereMachine, infrav1.VMProvisionedCondition)
+	vms, err = v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
+	log.V(5).Info("Trace ReconcileVM POST: VirtualMachines", "vmcount", len(vms), "error", err)
```
killianmuldoon, line 220:
```suggestion

```

pkg/services/vmoperator/vmopmachine.go
killianmuldoon, line 116:
```suggestion

```

killianmuldoon (commented):
I think for simplicity we should probably keep the defer for the `POST` log. There's suggestions for what that should look like below.

One issue: I don't like the way we're not handling the error from `getVirtualMachinesInCluster` explicitly, but I'm not very comfortable changing this code as it's supervisor so maybe we should keep it as is? @chrischdi WDYT? 

One safe change would be to have a different log if the result is nil, but that's more complexity for little enough gain IMO.

chrischdi:
> I think for simplicity we should probably keep the defer for the `POST` log. There's suggestions for what that should look like below.
> 
> One issue: I don't like the way we're not handling the error from `getVirtualMachinesInCluster` explicitly, but I'm not very comfortable changing this code as it's supervisor so maybe we should keep it as is? @chrischdi WDYT?
> 
> One safe change would be to have a different log if the result is nil, but that's more complexity for little enough gain IMO.

I think I agree, would be maybe a good thing to get input e.g. from @srm09 ?

Removing the `if` results in two additional `List` API calls to the vmoperator cluster. But I think we can consider them cheap, **only if these are cached objects**?! However, if the called function changes, we might not take that into account.

killianmuldoon:
> only if these are cached objects?!

I hadn't considered that - I assume these objects are cached, but we should confirm

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -81,14 +81,8 @@ func (v *VmopMachineService) ReconcileDelete(ctx context.Context, machineCtx cap
...
 	// If debug logging is enabled, report the number of vms in the cluster before and after the reconcile
-	if log.V(5).Enabled() {
-		vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
-		log.Info("Trace Destroy PRE: VirtualMachines", "vmcount", len(vms), "error", err)
-		defer func() {
-			vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
-			log.Info("Trace Destroy POST: VirtualMachines", "vmcount", len(vms), "error", err)
-		}()
-	}
+	vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
```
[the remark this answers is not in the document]
killianmuldoon: You should be able to use a supervisorMachineCtx as a machineCtx in the input to the test.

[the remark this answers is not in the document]
Ankitasw: `supervisorMachineCtx` is injected in the vmopmachine_tests but the vspheremachine_controller_test has no way to input the context.

chrischdi:
Oh wow, I took a look and we actually found a real nil pointer case :-) So the tests are okay, `getVirtualMachinesInCluster` is not okay when `supervisorMachineCtx.Cluster` is nil. The test case explicitly says:

```
Create and Delete a VSphereMachine with a Machine but without a Cluster
```

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/0816e23e8fe4a5eeb854801f6c1670e09d485e0d/controllers/vmware/test/controllers_test.go#L477 

And nil for the `cluster` inside the context is an valid case for a machine which gets deleted.

I think we should add at the very top of `getVirtualMachinesInCluster`: (+ improving the error text)

```go
if supervisorMachineCtx.Cluster == nil {
  return 0, errors.New("No cluster set for machine")
}
```

chrischdi:
Pending: check that list calls inside `getVirtualMachinesInCluster` use a cached client.

Ankitasw:
> Pending: check that list calls inside getVirtualMachinesInCluster use a cached client/objects.

Noob question, how do we get to know if the object/client is cached? From what I see, the caller of `getVirtualMachinesInCluster` are `ReconcileNormal` and `ReconcileDelete` of `vspheremachine_controller`. There I dont see supervisorCtx being cached.

chrischdi:
> > Pending: check that list calls inside getVirtualMachinesInCluster use a cached client/objects.
> 
> @chrischdi Probably a noob question, how do we get to know if the object/client is cached? From what I see, the caller of `getVirtualMachinesInCluster` are `ReconcileNormal` and `ReconcileDelete` of `vspheremachine_controller`. There I dont see supervisorCtx being cached.

Let's take a look:

We use the client from the `VmopMachineService`:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/e9fdbc55341536a90a0aa84f0bafa8cfd6d58b58/pkg/services/vmoperator/vmopmachine.go#L377

This is a `client.Client`:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/e9fdbc55341536a90a0aa84f0bafa8cfd6d58b58/pkg/services/vmoperator/vmopmachine.go#L46

When creating the service, we pass in the `controllerManagerContext.Client`

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/e9fdbc55341536a90a0aa84f0bafa8cfd6d58b58/controllers/vspheremachine_controller.go#L136

This one comes from here:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/e9fdbc55341536a90a0aa84f0bafa8cfd6d58b58/pkg/manager/manager.go#L91

So we use the controller-runtime client created via the manager, and we don't use unstructured.

We also don't set any different settings at `managerOpts.Cache` (where we could restrict what is cached and what not)

So this lgtm 👍  (sorry for the delay in response)

/approve

Ankitasw:
Thankyou for detailed explanation @chrischdi 🙂

killianmuldoon (commented):
Thanks!

/lgtm

PR #2416, opened 2023-10-04
:seedling: Remove build-dev.sh

killianmuldoon:
Remove the `build-dev.sh` script. I don't know if this script is being used by anyone, but the line `# TODO(akutz) This script can probably be removed once v1alpha2 is released.` suggests it's time to remove this.


hack/build-dev.sh
```diff
@@ -1,79 +0,0 @@
...
-
-#
-# TODO(akutz) This script can probably be removed once v1alpha2 is released.
-#
-
-set -o errexit
-set -o nounset
-set -o pipefail
-
-# Change directories to the parent directory of the one in which this
```
killianmuldoon, line 25: AFAICT all of this should be taken care of  in the Makefile and tilt should be the preferred dev environment.

chrischdi:
/lgtm
/approve

It even refers very old stuff so should be save to remove.

Example: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm is not the active provider code.

PR #2417, opened 2023-10-09
🌱 Add kubetest files for upstream e2e tests

chrischdi:
Adds kubetest yaml files for #2452

### DIFF
--- a/test/e2e/data/kubetest/upstream-conformance.yaml
+++ b/test/e2e/data/kubetest/upstream-conformance.yaml
@@ -1,0 +1,5 @@
+disable-log-dump: true
+ginkgo.focus: \[Conformance\]
+# ginkgo.skip:
+
+ginkgo.trace: true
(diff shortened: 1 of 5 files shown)


chrischdi:
One observation: ubuntu package `nfs-common` is missing for some tests to succeed.

killianmuldoon (commented):
/lgtm

One Q: Do we have a reference for these configs - are they cloned from somewhere else?

chrischdi:
> /lgtm
> 
> One Q: Do we have a reference for these configs - are they cloned from somewhere else?

added a reference link.

killianmuldoon (commented):
Thanks for refs!

/lgtm
/approve

/hold
(not sure when you want to merge)

chrischdi:
Thanks for the hold. Currently running it in a test environment, and I will then compare the properties in the junit tests with the upstream tests, to ensure we use the very same configuration.

chrischdi:
/hold cancel

one more thing

chrischdi:
ah damn :D wanted to hold again.

Issue #2308, opened 2023-08-31
Increase code coverage


laozc:
/kind feature

**Describe the solution you'd like**
A few codes are not covered by unit test and we would like to increase the code coverage.

There are some improvements we could make to ensure the code quality in the workflow, including
1. Better report to understand the code coverage summary more intuitively (than the current go coverage report)
2. Generate the report regularly to observe the trends
3. Enable the coverage check against each PR (for gating in the future)
5. More test cases
...

**Anything else you would like to add:**
All the progress related to code coverage may be tracked with this issue.

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

laozc:
Package | Coverage | Details
-- | -- | --
sigs.k8s.io/cluster-api-provider-vsphere/apis/v1alpha3 | 69.3% | 190/274
sigs.k8s.io/cluster-api-provider-vsphere/controllers | 69.6% | 1062/1526
sigs.k8s.io/cluster-api-provider-vsphere/controllers/vmware | 18.4% | 25/136
sigs.k8s.io/cluster-api-provider-vsphere/internal/webhooks | 74.7% | 162/217
sigs.k8s.io/cluster-api-provider-vsphere/pkg/clustermodule | 34.4% | 43/125
sigs.k8s.io/cluster-api-provider-vsphere/pkg/identity | 86.8% | 46/53
sigs.k8s.io/cluster-api-provider-vsphere/pkg/manager | 46.2% | 37/80
sigs.k8s.io/cluster-api-provider-vsphere/pkg/record | 100.0% | 10/10
sigs.k8s.io/cluster-api-provider-vsphere/pkg/services | 33.7% | 69/205
sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi | 9.5% | 58/608
sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/cluster | 84.6% | 44/52
sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/extra | 86.7% | 13/15
sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/find | 83.3% | 15/18
sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/govmomi/ipam | 93.8% | 91/97
[...]

PR #2436, opened 2023-10-16
🌱 Add unit test cases for clustermodule

MrDXY:
1. Add unit test cases for `pkg/clustermodule/service.go`

Part of #2308

### DIFF
--- a/pkg/clustermodule/service_test.go
+++ b/pkg/clustermodule/service_test.go
@@ -28,6 +28,7 @@ import (
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context/fake"
+	"sigs.k8s.io/cluster-api-provider-vsphere/test/helpers/vcsim"
 )
 
 func TestService_Create(t *testing.T) {
@@ -81,6 +82,52 @@ func TestService_Create(t *testing.T) {
 			g.Expect(err).ToNot(gomega.HaveOccurred())
 			g.Expect(moduleUUID).To(gomega.BeEmpty())
 		})
+	})
+
+	t.Run("Create, DoesExist and Remove works", func(t *testing.T) {
+		g := gomega.NewWithT(t)
+		simr, err := vcsim.NewBuilder().Build()
+		defer simr.Destroy()
+		g.Expect(err).ToNot(gomega.HaveOccurred())
+		md := machineDeployment("md", fake.Namespace, fake.Clusterv1a2Name)
+		md.Spec.Template.Spec.InfrastructureRef = corev1.ObjectReference{
+			Kind:      "VSphereMachineTemplate",
+			Namespace: fake.Namespace,
+			Name:      "blah-template",
+		}
+
+		machineTemplate := &infrav1.VSphereMachineTemplate{
+			TypeMeta: metav1.TypeMeta{Kind: "VSphereMachineTemplate"},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "blah-template",
+				Namespace: fake.Namespace,
+			},
+			Spec: infrav1.VSphereMachineTemplateSpec{
+				Template: infrav1.VSphereMachineTemplateResource{Spec: infrav1.VSphereMachineSpec{
+					VirtualMachineCloneSpec: infrav1.VirtualMachineCloneSpec{
+						Server:       simr.ServerURL().Host,
+						Datacenter:   "*",
+						ResourcePool: "/DC0/host/DC0_C0/Resources",
+					},
+				}},
+			},
+		}
+
+		controllerCtx := fake.NewControllerContext(fake.NewControllerManagerContext(md, machineTemplate))
+		clusterCtx := fake.NewClusterContext(context.Background(), controllerCtx)
+		clusterCtx.VSphereCluster.Spec.Server = simr.ServerURL().Host
+		controllerCtx.ControllerManagerContext.Username = simr.Username()
+		controllerCtx.ControllerManagerContext.Password = simr.Password()
+
+		svc := NewService(controllerCtx.ControllerManagerContext, controllerCtx.Client)
+		moduleUUID, err := svc.Create(context.Background(), clusterCtx, mdWrapper{md})
+		g.Expect(err).ToNot(gomega.HaveOccurred())
+		g.Expect(moduleUUID).NotTo(gomega.BeEmpty())
+		exists, err := svc.DoesExist(context.Background(), clusterCtx, mdWrapper{md}, moduleUUID)
+		g.Expect(exists).To(gomega.BeTrue())
+		g.Expect(err).NotTo(gomega.HaveOccurred())
+		err = svc.Remove(context.Background(), clusterCtx, moduleUUID)
+		g.Expect(err).ToNot(gomega.HaveOccurred())
 	})
 }
 
--- a/test/helpers/vcsim/builder.go
+++ b/test/helpers/vcsim/builder.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/onsi/gomega/gbytes"
 	"github.com/vmware/govmomi/simulator"
+	_ "github.com/vmware/govmomi/vapi/cluster/simulator" // import this to register cluster module service test endpoint
 )
 
 type Builder struct {


pkg/session/fake/fake_session.go
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
zhanggbj, line 2: It should be 2023

pkg/session/fake/fake_session.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func MockDefaultSession() (*session.Session, error) {
+	return MockSession(DefaultDataCenter)
+}
+
+// MockSession GetOrCreate gets a cached session or creates a new one if one does not.
+func MockSession(dc string) (*session.Session, error) {
+	model := simulator.VPX()
+	model.Cluster = 1
```
zhanggbj, line 46: The model.Cluster returned by VPX() is already `1`.

pkg/session/fake/fake_session.go
```diff
@@ -0,0 +1,65 @@
...
+	Server   = ""
+	Username = ""
+	Password = ""
+)
+
+func MockDefaultSession() (*session.Session, error) {
+	return MockSession(DefaultDataCenter)
+}
+
+// MockSession GetOrCreate gets a cached session or creates a new one if one does not.
```
zhanggbj, line 43:
```suggestion
// MockSession gets a cached session or creates a new one if one does not already exist.
```

Makefile
```diff
@@ -454,7 +454,7 @@ setup-envtest: $(SETUP_ENVTEST) ## Set up envtest (download kubebuilder assets)
 
 .PHONY: test
 test: $(SETUP_ENVTEST) $(GOVC) ## Run unit tests
-	KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" GOVC_BIN_PATH=$(GOVC) go test -v ./apis/... ./controllers/... ./pkg/... ./internal/... $(TEST_ARGS)
```
chrischdi, line 457: Why is that needed?

pkg/clustermodule/service_test.go
```diff
@@ -21,13 +21,25 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/agiledragon/gomonkey/v2"
```
chrischdi, line 24: Question: did you consider to use `github.com/stretchr/testify/mock` instead because we already use that in other tests.
I just want to make sure to introduce a second variant of mocking stuff.

pkg/session/fake/fake_session.go
MrDXY: Yes, of course.

pkg/session/fake/fake_session.go
MrDXY: Noted, I will remove this statement.

pkg/clustermodule/service_test.go
MrDXY: Sure, I have considered using `/testify/mock`. 
1. The general idea of using mock in go has two pre-requirements/constraints:
    1. Have interfaces that allow you to switch to your mocked behavior.
    2. Inject a mocked struct in place of the real one(dependency injection).
2. But that is not always the case, especially when:
    1. Working on a code base that is unalterable.
    2. Lacks programming to interface, only structs, hindering dependency injection.
3. In our case:
    1. I found it hard to mock a `session`'s behavior. Especially when attempting to obtain a session with `govmomi` client. Certain internal calls, such as the call to `s.SessionManager.UserSession(ctx)`, could not be mocked by `testify`.
    2. Also the same for other `govmomi` client calls.
4. So I decided to use `gomonkey`, a tool that utilizes go to modify functions at runtime.
    1. With this mocking toolkit, we can easily mock the behavior of target functions. Like modifying the return objects and redefining the function execution process.
    2. And here I use it to mock some of the functions of `govmomi` client.

Makefile
MrDXY: That is because `gomonkey` doesn't support inline functions, and during testing, we need to disable Go language's inline optimization by using the command-line argument `-gcflags=-l`.

pkg/clustermodule/service_test.go
```diff
@@ -81,6 +93,172 @@ func TestService_Create(t *testing.T) {
 			g.Expect(moduleUUID).To(gomega.BeEmpty())
 		})
 	})
+
+	t.Run("creation is executed", func(t *testing.T) {
+		g := gomega.NewWithT(t)
+		_, err := fake_session.MockDefaultSession()
```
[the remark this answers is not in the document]
MrDXY: We'll use it [here](https://github.com/MrDXY/cluster-api-provider-vsphere/blob/pr_unit_test_for_clustermodule/pkg/clustermodule/service_test.go#L146C1-L147) (when you try to create a cluster module)
https://github.com/MrDXY/cluster-api-provider-vsphere/blob/4d2c987b67375a803280b248ef361469d5f870f6/pkg/clustermodule/service_test.go#L146C1-L147

My consideration here is to cache a session and use it for upcoming tests rather than create a session when using it. But it's also an option, only execute `vcsim.NewBuilder()...`, and let the case create a session.

pkg/clustermodule/service_test.go
MrDXY: Also, if you are concerned about the license issue, `agiledragon/monkey` is licensed under the MIT License. Our work on increasing ut-coverage could benefit from this, especially when mocking the interaction with `govmomi`.

chrischdi: I'm not concerned about the license. I just lean towards trying to prevent additional dependencies when there are alternatives. The only project in the k8s ecosystem which uses gomonkey is [vsphere-csi](https://cs.k8s.io/?q=github.com%2Fagiledragon%2Fgomonkey&i=nope&files=&excludeFiles=&repos=). 

All functionality should also be possible to get tested using vcsim as backend, I currently see no need to use monkey patching.

Another point is disabling code optimisations during tests. If we disable code optimisations we do test compiled code which is not the very same as what we ship.

MrDXY: > All functionality should also be possible to get tested using vcsim as backend

I attempted to utilize vcsim to establish a testing environment previously, but I encountered an error. 
- Error when I try to create a cluster module:
    - The error message indicates that vcenter does not have the requested path, namely `POST https://127.0.0.1:64753/rest/vcenter/cluster/modules`.image
- Also the same 404 error for `ListModules`.
- Have you encounter similar errors? May be I got this wrong, but that's literally the reason that I found it hard to mock a govmomi client's behavior. Maybe we can not simulate all vCenter's responses via vcsim.

As for code optimizations during tests, that's just an inner mechanism of Go, it should work identically with or without those optimizations. The result should be the same, just an efficiency issue. No need to be worried : ）

MrDXY: > I just lean towards trying to prevent additional dependencies when there are alternatives

It also makes sense to me, but one dependency should be OK, and it is only used for test purposes.

chrischdi: What I see is:

I can run vcsim and use it to create/list/remove cluster modules.

```sh
❯ vcsim &
export GOVC_URL=https://user:pass@127.0.0.1:8989/sdk GOVC_SIM_PID=39024
❯ export GOVC_URL=https://user:pass@127.0.0.1:8989/sdk GOVC_SIM_PID=39024
❯ export GOVC_INSECURE=true
❯ govc cluster.module.ls
❯ govc cluster.module.create
d5a10942-7178-4f9c-996c-cead7e8b4490
❯ govc cluster.module.ls
domain-c27  d5a10942-7178-4f9c-996c-cead7e8b4490
❯ govc cluster.module.rm d5a10942-7178-4f9c-996c-cead7e8b4490
❯ govc cluster.module.ls
```

> The result should be the same, just an efficiency issue. 

Yeah, but still a small risk that it behaves differently.

killianmuldoon: 100 percent agreed with @chrischdi on this - we should use vcsim for this testing if at all possible.

MrDXY: Got it, folks. Let me take a closer look at vcsim and try to figure out & solve the problem I met : )

MrDXY: Update: still encounter the same 404 error while using the command line with `vcsim.Simulator` (I'll go deeper into this issue.) Also, please feel free to share any of your ideas or suggestions on this matter.
image

MrDXY: I think I have identified the cause of the 404 error I encountered. The cluster module request will be dealt with [here](https://github.com/vmware/govmomi/blob/244e3ac4bfc31edacd09f191e6a5411ad0c15064/vapi/cluster/simulator/simulator.go#L71-L108). The [init](https://github.com/vmware/govmomi/blob/244e3ac4bfc31edacd09f191e6a5411ad0c15064/vapi/cluster/simulator/simulator.go#L37-L41) function is used to register the handler/endpoint to the simulator, however, the function will not execute if we only import `github.com/vmware/govmomi/simulator`. To ensure the `init` function works, it is necessary to import `github.com/vmware/govmomi/vapi/cluster/simulator` as well. 

Here is the code for the `init` function:
```
func init() {
        simulator.RegisterEndpoint(func(s *simulator.Service, r *simulator.Registry) {
                New(s.Listen).Register(s, r)
        })
}
```

After importing the required package, everything appears to be working properly. I will update the corresponding changes accordingly!

pkg/clustermodule/service_test.go
```diff
@@ -81,6 +84,148 @@ func TestService_Create(t *testing.T) {
...
+		controllerCtx := fake.NewControllerContext(fake.NewControllerManagerContext(md, machineTemplate))
+		clusterCtx := fake.NewClusterContext(controllerCtx)
+		clusterCtx.VSphereCluster.Spec.Server = simr.ServerURL().Host
+		controllerCtx.ControllerManagerContext.Username = simr.Username()
+		controllerCtx.ControllerManagerContext.Password = simr.Password()
+
+		svc := NewService(controllerCtx.ControllerManagerContext, controllerCtx.Client)
+		moduleUUID, err := svc.Create(context.Background(), clusterCtx, mdWrapper{md})
+		g.Expect(err).ToNot(gomega.HaveOccurred())
+		g.Expect(moduleUUID).NotTo(gomega.BeEmpty())
```
chrischdi, line 126: I think it's not worth having a separated `DoesExist` and `Delete` test functions.

These two also test teh `Create` part already. Let's combine them to a single test.


```suggestion
		g.Expect(moduleUUID).NotTo(gomega.BeEmpty())


		exists, err := svc.DoesExist(context.Background(), clusterCtx, mdWrapper{md}, moduleUUID)
		g.Expect(exists).To(gomega.BeTrue())
		g.Expect(err).NotTo(gomega.HaveOccurred())

		err = svc.Remove(context.Background(), clusterCtx, moduleUUID)
		g.Expect(err).ToNot(gomega.HaveOccurred())
```

chrischdi (commented):
NIce work with using vcims. This gets very clean :-)

pkg/clustermodule/service_test.go
```diff
@@ -81,6 +84,148 @@ func TestService_Create(t *testing.T) {
 			g.Expect(moduleUUID).To(gomega.BeEmpty())
 		})
 	})
+
+	t.Run("creation is executed", func(t *testing.T) {
+		g := gomega.NewWithT(t)
+		simr, err := initVCSimulator()
```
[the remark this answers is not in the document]
MrDXY: That makes sense. Let me remove this and make the necessary adjustments.

MrDXY:
Hi @chrischdi , This PR is ready for another round of review; please feel free to take a look whenever you get a chance. 😄

chrischdi:
> Hi @chrischdi , This PR is ready for another round of review; please feel free to take a look whenever you get a chance. 😄

Sorry for the late reply, forgot to send out the above comment 🤦

MrDXY:
> Sorry for the late reply, forgot to send out the above comment 🤦

No problem at all. I'll take care of this. 😄

pkg/clustermodule/service_test.go
```diff
@@ -81,6 +82,52 @@ func TestService_Create(t *testing.T) {
 			g.Expect(moduleUUID).To(gomega.BeEmpty())
 		})
 	})
+
+	t.Run("creation is executed", func(t *testing.T) {
```
chrischdi, line 86:
```suggestion
	t.Run("Create, DoesExist and Remove works", func(t *testing.T) {
```

chrischdi (commented):
Last nit

/lgtm

chrischdi (commented):
/approve

killianmuldoon (commented):
/lgtm
/approve

Thanks!

chrischdi (commented):
Only nits about comments and a optional question/thing for the test code, lgtm to it in general!

Thanks for also adding test cases!

PR #2444, opened 2023-10-16
:seedling: Add test coverage for VimMachineService

laozc:
Add test coverage for `VimMachineService` in `pkg/services`
Address #2308


killianmuldoon (commented):
Question on this @laozc  - do you have a way to run these tests individually locally? Especially through an IDE? 

I have a problem running tests formatted in this way i.e. using `Describe` instead of a `Test.*` function in Intellij/Goland, makes it hard to write tests and hard to review them.

laozc:
> Question on this @laozc - do you have a way to run these tests individually locally? Especially through an IDE?
> 
> I have a problem running tests formatted in this way i.e. using `Describe` instead of a `Test.*` function in Intellij/Goland, makes it hard to write tests and hard to review them.

I wrote the test and tested it with the help of Goland which could tell me the coverage in details.
Only one drawback that these tests could not be run independently.

What did you meet when you were running these tests?

chrischdi:
> Only one drawback that these tests could not be run independently.

I think that's the main point Killian thought of. 
Maybe we can converge to the pattern we use in CAPI to have a `TestMain` func and by that are able to run the tests independently, at least for the new tests we are introducing.
With that said: I don't know how much we would need to change for that 🤔

laozc:
I believe a plugin Ginkgo needs to be installed to support running the test separately in Goland/IntelliJ IDEA.
We have a few tests written in Ginkgo at the moment.

killianmuldoon:
The Ginkgo plugin does work which is good for existing tests. Going forward though I think we should conform to the upstream CAPI policy [described here](https://main.cluster-api.sigs.k8s.io/developer/testing#ginkgo).

>In Cluster API you MUST use ginkgo only for E2E tests, where it is required to leverage the support for running specs in parallel; in any case, developers MUST NOT use the table driven extension DSL (DescribeTable, Entry commands) which is considered unintuitive.

I'll make a comment to this effect on the umbrella issue too.

laozc:
Thanks for the info.
I took some time to move all the tests in this file away from Ginkgo.

sbueringer:
@laozc #2527 will merge in a bit. Do you have time to rebase this PR once #2527 is merged?

laozc:
Sure. Let me rebase it.

sbueringer:
Thank you very much!

/lgtm
/approve

Issue #2295, opened 2023-08-29
Refactor CAPV controller context


zhanggbj:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
Currently CAPV has a customized Controller Context which wraps context.Context along with other essential controller-related information as [here](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/pkg/context/controller_manager_context.go#L35). This context is propagated across controllers and reconcilers. However, this approach comes with drawbacks:

- Not align with the recommended practices outlined by the `containedctx` lint, which emphasizes that 'Contexts should not be stored inside a struct type, but instead passed to each function that needs it'.
- Hard to understand the function and may need further investigation for better comprehension about what it needs, for example `func (r clusterReconciler) reconcileNormal(ctx *context.ClusterContext)`
- contains mixed usage of CAPV controller context and golang context, sometimes capv will initialize new golang context to call some API. It is difficult to enable contextual logging and tracing.

Given the sensitivity of context usage, particularly its impact across controllers, propose a refactoring of the controller context, aiming to split it into golang context and the CAPV controller context. This will be achieved incrementally through the following steps:

- [x] Enable the linter `containedctx` and use '//nolint:containedctx' to exclude current findings. Refer to #2058 . (@zhanggbj #2297 )
- [x] rename all "CAPV context's" in a consistent way: e.g. ControllerContext => controllerCtx, also like imported package alias, function signature (@zhanggbj #2309 )
- [x] Introduce new ctx context.Context parameters to controller and reconciler and underlying functions, such as functions under /pkg, also fix Unit Test. We can break this down by controllers as outlined below, but let's begin and be open to adjustments as needed.  
	- [x] controllers/clustermodule_reconciler.go (@sbueringer https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2352)
	- [x] controllers/serviceaccount_controller.go (@sbueringer https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2352)
	- [x] controllers/servicediscovery_controller.go (@sbueringer https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2352)
	- [x] controllers/vspherecluster_controller.go (@zhanggbj https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2339)
[...]

PR #2431, opened 2023-10-13
🌱 Remove context from capv customized controller context

zhanggbj:
**What this PR does / why we need it**:
Remove context from capv customized controller context, also get rid of containedctx exceptions.
This is part of https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/2295.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Part of 
#2295
#2384

PR #2462, opened 2023-10-25
🌱 Setup separate context for test files to avoid flaky errors

zhanggbj:
Setup separate context for test files under /controllers instead of using the test-package context, this is to avoid flaky errors we found in CI.

Please find more details in #2295

Part of #2295

### DIFF
--- a/controllers/clustermodule_reconciler_test.go
+++ b/controllers/clustermodule_reconciler_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package controllers
 
 import (
+	"context"
 	"testing"
 	"time"
 
@@ -38,6 +39,7 @@ import (
 )
 
 func TestReconciler_Reconcile(t *testing.T) {
+	ctx := context.Background()
 	kcpUUID, mdUUID := uuid.New().String(), uuid.New().String()
 	kcp := controlPlane("kcp", metav1.NamespaceDefault, fake.Clusterv1a2Name)
 	md := machineDeployment("md", metav1.NamespaceDefault, fake.Clusterv1a2Name)
@@ -437,6 +439,7 @@ func TestReconciler_Reconcile(t *testing.T) {
 }
 
 func TestReconciler_fetchMachineOwnerObjects(t *testing.T) {
+	ctx := context.Background()
 	tests := []struct {
 		name         string
 		numOfMDs     int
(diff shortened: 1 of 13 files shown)


killianmuldoon (commented):
/retest

controllers/clustermodule_reconciler_test.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package controllers
```
sbueringer, line 17: Let's move the package-level ctx in controllers_suite_test.go into the setup func.

This way we ensure nobody is reusing it

sbueringer: Btw, I'm not sure I understand how the package-level context can lead to flaky tests. It should be only canceled when the test process is killed.

chrischdi: One state of @zhanggbj 's PR https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2431 had constantly failing tests in CI. the changes after the following comment fixed it.

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2431#issuecomment-1765858951

sbueringer: Interesting. If I find the time I try to reproduce. Would be just interesting to understand why. I'm not sure but could be that we should re-think the pattern of using a package-level ctx. I think we use it in a bunch of places in core CAPI (but not sure)

zhanggbj: It only can be reproduced in prow test, I cannot reproduce it locally, so not sure about the root cause.

sbueringer: If it's okay I would hold this PR for now. I'm hesitant to make this change if we don't know why we need it and tests are stable without this PR

zhanggbj: It's OK for me to hold it until we reproduce it, this PR is just to make test cases more isolated considering the context staff.

chrischdi:
Let's hold this work until we are sure about why it happened in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/compare/269023424b03a8c029ad95b605f0b6bf9431f9a3..b00e9811fd246af7a9e03676d9577b8ef56106d0 and if this would bring a benefit at all

/hold

chrischdi:
I think we don't need this anymore.

/close

PR #2463, opened 2023-10-25
:sparkles: Bump Cluster API to v1.6.0-beta.1

killianmuldoon:
Bump Cluster API to v1.6.0-beta.0. 

The release notes are here: https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.6.0-beta.0


killianmuldoon (commented):
/test

killianmuldoon (commented):
Running into a lot of issues - first hurdle being the supervisor unit tests. I can't seem to find a reasonable way to run them individually locally to help debug.

controllers/serviceaccount_controller_unit_test.go
```diff
@@ -54,9 +54,6 @@ func unitTestsReconcileNormal() {
 		_, err := reconciler.ReconcileNormal(ctx, controllerCtx.GuestClusterContext)
 		Expect(err).NotTo(HaveOccurred())
 
-		// Update the VSphereCluster and its status in the fake client.
-		Expect(controllerCtx.ControllerContext.Client.Update(ctx, controllerCtx.VSphereCluster)).To(Succeed())
-		Expect(controllerCtx.ControllerContext.Client.Status().Update(ctx, controllerCtx.VSphereCluster)).To(Succeed())
```
killianmuldoon, line 59: These functions weren't being used in the underlying test as the object was being checked in place. Some change in the fakeclient seems to have changed how `Update` works so that e.g. `Status` update also mutates the object in place making this function not work. Fine to get rid of it here, and there weren't that many changes needed for this bump.

I do wonder with how often `Update` functions are used across the code what kind of bugs are being missed though.

controllers/servicediscovery_controller_unit_test.go
```diff
@@ -49,10 +49,6 @@ func serviceDiscoveryUnitTestsReconcileNormal() {
 		}
 		_, err := reconciler.ReconcileNormal(ctx, controllerCtx.GuestClusterContext)
 		Expect(err).NotTo(HaveOccurred())
-
-		// Update the VSphereCluster and its status in the fake client.
-		Expect(controllerCtx.ControllerContext.Client.Update(ctx, controllerCtx.VSphereCluster)).To(Succeed())
-		Expect(controllerCtx.ControllerContext.Client.Status().Update(ctx, controllerCtx.VSphereCluster)).To(Succeed())
```
killianmuldoon, line 55: Similar to above - the object is being checked in place in this test so no need to update it in the fake client.

controllers/serviceaccount_controller_unit_test.go
sbueringer: Q: "being checked in place" this means our reconciler code is just reading it from our magic ctx and not via the client. So there is no need for the VSphereCluster object to exist in the fakeclient, right?

killianmuldoon: Exactly

main.go
```diff
@@ -402,11 +403,13 @@ func setupChecks(mgr ctrlmgr.Manager) {
 func isCRDDeployed(mgr ctrlmgr.Manager, gvr schema.GroupVersionResource) (bool, error) {
 	_, err := mgr.GetRESTMapper().KindFor(gvr)
 	if err != nil {
-		discoveryErr, ok := errors.Unwrap(err).(*discovery.ErrGroupDiscoveryFailed)
+		var discoveryErr *apiutil.ErrResourceDiscoveryFailed
```
sbueringer, line 406: Q: We have/had some sort of confirmation via integration / e2e tests that the discovery still works, right?

main.go
```diff
@@ -85,7 +86,8 @@ var (
 	vSphereClusterIdentityConcurrency int
 	vSphereDeploymentZoneConcurrency  int
 
-	tlsOptions = flags.TLSOptions{}
+	tlsOptions         = capiflags.TLSOptions{}
+	diagnosticsOptions = capiflags.DiagnosticsOptions{}
```
sbueringer, line 90: looks like we don't use diagnosticsOptions to set options on the manger

main.go
```diff
@@ -240,7 +239,9 @@ func main() {
 	managerOpts.KubeConfig.UserAgent = remote.DefaultClusterAPIUserAgent(controllerName)
 
 	if watchNamespace != "" {
-		managerOpts.Cache.Namespaces = []string{watchNamespace}
+		managerOpts.Cache.DefaultNamespaces = map[string]cache.Config{
+			watchNamespace: {},
+		}
 	}
 
 	if profilerAddress != "" && enableContentionProfiling {
```
sbueringer, line 247:
```suggestion
	if enableContentionProfiling {
```
Just to align to what we changed in core CAPI.

main.go
```diff
@@ -85,7 +86,8 @@ var (
 	vSphereClusterIdentityConcurrency int
 	vSphereDeploymentZoneConcurrency  int
```
sbueringer, line 87: A bit unrelatd to this PR. can we fully implemented the profilerAddress flag (similar to core CAPI)

Right now we only have a var, but we don't have a flag and we don't pass it to the manager options

sbueringer:
Thx! A few smaller findings. I compared a bit with core CAPI

main.go
killianmuldoon: Yeah - without this change the integration test on startup. We also have integration tests that run supervisor API types.

killianmuldoon: I don't believe we have any e2e tests for this.

main.go
killianmuldoon: Looks like we shouldn't have the var but we are setting the flag on ~L187

main.go
killianmuldoon: Should be set now if I'm understanding right - sorry I didn't pay any attention to getting the secure metrics serving feature into this PR - didn't even consider it :sweat_smile:

main.go
sbueringer: I meant more like: integration tests cover the supersvisor code path and e2e tests the other one

Sounds good!

main.go
sbueringer: Makes sense!

main.go
sbueringer: Looks good, thx!

sbueringer:
Thx!

Issue #2453, opened 2023-10-20
Check timeout on govmomi calls


chrischdi:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Recent reports did show that a request to vcenter could take forever without timing out.
Reason for that is that there seem to be no timeouts set on the reuqests.

Q: is CAPV waiting indefinitely for responses of requests to vCenter?

If it is so, we might want to take action and pass through contexts which include a timeout.

**What did you expect to happen:**

Requests getting canceled after some time.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

chrischdi:
Example: storage class query took 20 mins to answer.

rikatz:
sorry for the delay here folks, my bad, got dragged into other problems.

The issue here is, during a cluster reconciliation, I've seen a huge time between  "cloning" and "powering on the node".

Debugging, I've reached the place where CAPV was waiting for some vCenter answer, at [this line](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/c2b813cf2f7d0c2ab1fcaf9136675177bb61f72f/pkg/services/govmomi/service.go#L397). 

I don't know yet, on vCenter side why this query is taking some time (and I have reached also the right team to look at it), but this command should have probably timed out first, and this is what @chrischdi is reporting here.

sbueringer:
I think this issue is about more than just this specific case. We were wondering in general if we are setting timeouts on govmomi. So we're not getting stuck indefinitely if a request doesn't return.

PR #2467, opened 2023-10-26
:bug: Use a cheaper method for storage policy

rikatz:
During the debug of https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/2453#issuecomment-1773244913 we have figured out this call is too expensive for environments with a lot of resources, like more than 10k resources.

This may turn into a huge slow query that can take up to 20 minutes.

Instead, we should be using the "QueryAssociatedProfiles" call that, given an object (like a disk), check if it has associated storage policies

Fixes #

**Note** https://github.com/vmware/govmomi/pull/3268 implements the call directly on pbmClient, but as this is not released yet I have added it on the new utils file.

### DIFF
--- a/pkg/services/govmomi/service.go
+++ b/pkg/services/govmomi/service.go
@@ -394,10 +394,6 @@ func (vms *VMService) reconcileStoragePolicy(ctx context.Context, virtualMachine
 	if err != nil {
 		return errors.Wrap(err, "unable to retrieve storage profile ID")
 	}
-	entities, err := pbmClient.QueryAssociatedEntity(ctx, pbmTypes.PbmProfileId{UniqueId: storageProfileID}, "virtualDiskId")
-	if err != nil {
-		return err
-	}
 
 	var changes []types.BaseVirtualDeviceConfigSpec
 	devices, err := virtualMachineCtx.Obj.Device(ctx)
@@ -405,24 +401,41 @@ func (vms *VMService) reconcileStoragePolicy(ctx context.Context, virtualMachine
 		return err
 	}
 
+	disksRefs := make([]pbmTypes.PbmServerObjectRef, 0)
+	// diskMap is just an auxiliar map so we don't need to iterate over and over disks to get their configs
+	// if we realize they are not on the right storage policy
+	diskMap := make(map[string]*types.VirtualDisk)
+
 	disks := devices.SelectByType((*types.VirtualDisk)(nil))
+
+	// We iterate over disks and create an array of disks refs, so we just need to make a single call
+	// against vCenter, instead of one call per disk
+	// the diskMap is an auxiliar way of, besides the disksRefs, we have a "searchable" disk configuration
+	// in case we need to reconfigure a disk, to get its config
 	for _, d := range disks {
 		disk := d.(*types.VirtualDisk)
-		found := false
 		// entities associated with storage policy has key in the form <vm-ID>:<disk>
 		diskID := fmt.Sprintf("%s:%d", virtualMachineCtx.Obj.Reference().Value, disk.Key)
-		for _, e := range entities {
-			if e.Key == diskID {
-				found = true
-				break
-			}
-		}
+		diskMap[diskID] = disk
 
-		if !found {
-			// disk wasn't associated with storage policy, create a device change to make the association
+		disksRefs = append(disksRefs, pbmTypes.PbmServerObjectRef{
+			ObjectType: string(pbmTypes.PbmObjectTypeVirtualDiskId),
+			Key:        diskID,
+		})
+	}
+
+	diskObjects, err := pbmClient.QueryAssociatedProfiles(ctx, disksRefs)
+	if err != nil {
+		return errors.Wrap(err, "unable to query disks associated profiles")
+	}
+
+	// Ensure storage policy is set correctly for all disks of the VM
+	for k := range diskObjects {
+		if !isStoragePolicyIDPresent(storageProfileID, diskObjects[k]) {
+			virtualMachineCtx.Logger.V(5).Info("storage policy not found on disk, adding for reconciliation", "disk", diskObjects[k].Object.Key)
 			config := &types.VirtualDeviceConfigSpec{
 				Operation: types.VirtualDeviceConfigSpecOperationEdit,
-				Device:    disk,
+				Device:    diskMap[diskObjects[k].Object.Key],
 				Profile: []types.BaseVirtualMachineProfileSpec{
 					&types.VirtualMachineDefinedProfileSpec{ProfileId: storageProfileID},
 				},
@@ -431,6 +444,7 @@ func (vms *VMService) reconcileStoragePolicy(ctx context.Context, virtualMachine
 		}
 	}
 
+	// If there are pending changes for Storage Policies, do it before moving next
 	if len(changes) > 0 {
 		task, err := virtualMachineCtx.Obj.Reconfigure(ctx, types.VirtualMachineConfigSpec{
 			VmProfile: []types.BaseVirtualMachineProfileSpec{
--- a/pkg/services/govmomi/service_test.go
+++ b/pkg/services/govmomi/service_test.go
@@ -18,11 +18,14 @@ package govmomi
 
 import (
 	"context"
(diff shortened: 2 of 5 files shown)


rikatz:
/hold
I will compile it locally and run on our environment just to make sure the problem is gone

rikatz:
/hold cancel
Tested on our environment, creation is being really fast now with Storage Policies and respecting the associated storage policy specified

pkg/services/govmomi/service.go
```diff
@@ -394,35 +394,47 @@ func (vms *VMService) reconcileStoragePolicy(ctx context.Context, virtualMachine
...
-		return err
-	}
 
 	var changes []types.BaseVirtualDeviceConfigSpec
 	devices, err := virtualMachineCtx.Obj.Device(ctx)
 	if err != nil {
 		return err
 	}
 
+	arrayOfEntities := make([]pbmTypes.PbmServerObjectRef, 0)
```
sbueringer, line 404:
```suggestion
	diskRefs := make([]pbmTypes.PbmServerObjectRef, 0)
```
If I see correctly

pkg/services/govmomi/service.go
```diff
@@ -394,35 +394,47 @@ func (vms *VMService) reconcileStoragePolicy(ctx context.Context, virtualMachine
...
+	// diskMap is just an auxiliar map so we don't need to iterate over and over disks to get their configs
+	// if we realize they are not on the right storage policy
+	diskMap := make(map[string]*types.VirtualDisk)
+
 	disks := devices.SelectByType((*types.VirtualDisk)(nil))
+
+	// We iterate over disks and create an array of entities, so we just need to make a single call
```
sbueringer, line 411:
```suggestion
	// We iterate over disks and create an array of disk refs, so we just need to make a single call
```

pkg/services/govmomi/service.go
```diff
@@ -394,35 +394,47 @@ func (vms *VMService) reconcileStoragePolicy(ctx context.Context, virtualMachine
...
-		}
+		diskMap[diskID] = disk
+
+		arrayOfEntities = append(arrayOfEntities, pbmTypes.PbmServerObjectRef{
+			ObjectType: string(pbmTypes.PbmObjectTypeVirtualDiskId),
+			Key:        diskID,
+		})
+	}
+
+	entities, err := QueryAssociatedProfiles(ctx, pbmClient, arrayOfEntities)
```
sbueringer, line 427:
```suggestion
	diskObjects, err := QueryAssociatedProfiles(ctx, pbmClient, arrayOfEntities)
```
Maybe. Let's please use something less generic than entities

pkg/services/govmomi/service.go
```diff
@@ -394,35 +394,47 @@ func (vms *VMService) reconcileStoragePolicy(ctx context.Context, virtualMachine
...
+	if err != nil {
+		return err
```
sbueringer, line 429: Should we wrap the error?

pkg/services/govmomi/service_test.go
```diff
@@ -93,9 +101,135 @@ func Test_reconcilePCIDevices(t *testing.T) {
...
+				},
+			}
+			err = vms.reconcileStoragePolicy(context.Background(), vmCtx)
+			g.Expect(err.Error()).To(ContainSubstring("no pbm profile found with name"))
+			return nil
+		}, model)
+	})
+
+	t.Run("when the requested storage policy exists should pass", func(t *testing.T) {
+		t.Skip("PbmQueryAssociatedProfiles is not yet implemented on PBM simulator")
```
sbueringer, line 171: Is this also part of v0.34? If yes let's also add a hint to the other todo that we can drop the skip here

sbueringer (commented):
Thx! Just nits.

Should we backport this change/fix?

pkg/services/govmomi/service_test.go
rikatz: Not part of v0.34 yet, it needs to be implemented on the vcsim and I didn't.

Will add it as my next thing to do as well

pkg/services/govmomi/service.go
rikatz: Changed :) I didn't commited your suggestion because there are other parts to change here

rikatz:
/retest
weird

pkg/services/govmomi/service_test.go
rikatz: https://github.com/vmware/govmomi/issues/3271 here, now on the backlog :)

rikatz:
> Should we backport this change/fix?
Yes. On environments that, as an example don't reclaim PVs, there are going to be a lot of objects. Querying all of this objects is an expensive operation and may get cluster reconciliation stuck for a long time. IMHO at least for 1.7 and 1.8 this should be cherry picked.

Thanks!!

pkg/services/govmomi/service_test.go
sbueringer: Alright. I'll watch for progress! :)

sbueringer:
Thank you!

/lgtm

/assign @chrischdi

rikatz:
/hold
I will just update govmomi for 0.33.1 and drop the copied code (thanks @dougm !!)

rikatz:
/hold cancel
@sbueringer @chrischdi last changes after Stefan's review:
* bumped govmomi to v0.33.1 (it contains the required methods on pbmclient)
* squashed commits
Thanks!!

sbueringer:
/lgtm

@chrischdi PTAL :)

pkg/services/govmomi/service.go
```diff
@@ -394,35 +394,47 @@ func (vms *VMService) reconcileStoragePolicy(ctx context.Context, virtualMachine
...
+	}
+
+	diskObjects, err := pbmClient.QueryAssociatedProfiles(ctx, disksRefs)
+	if err != nil {
+		return errors.Wrap(err, "unable to query disks associated profiles")
+	}
 
-		if !found {
-			// disk wasn't associated with storage policy, create a device change to make the association
+	for k := range diskObjects {
```
chrischdi, line 432:
```suggestion
	// Ensure storage policy is set correctly for all disks of the VM
	for k := range diskObjects {
```

pkg/services/govmomi/service_test.go
```diff
@@ -93,9 +101,136 @@ func Test_reconcilePCIDevices(t *testing.T) {
...
+		g.Expect(vmCtx.VSphereVM.Status.TaskRef).To(BeEmpty())
+	})
+
+	t.Run("when the requested storage policy does not exists should fail", func(t *testing.T) {
+		g = NewWithT(t)
+		before()
+		model, err := storagePolicyModel()
+		g.Expect(err).ToNot(HaveOccurred())
+
+		simulator.Run(func(ctx context.Context, c *vim25.Client) error {
```
chrischdi, line 145: Optional: I don't know which method is better, this one or the other. Totally fine to keep as is.

Wondering if we should use the vcsim package here instead via:

```go
  simr, err := vcsim.NewBuilder().WithModel(model).Build()
  defer simr.Destroy()
  g.Expect(err).ToNot(gomega.HaveOccurred())

  ...
  // for session setup
  simr.ServerURL().Host
  simr.Username()
  simr.Password()
  ...

  vm, err := getPoweredoffVM(ctx, c)
  ...
```

chrischdi (commented):
Only nits about comments and a optional question/thing for the test code, lgtm to it in general!

Thanks for also adding test cases!

pkg/services/govmomi/service_test.go
rikatz: ah yeah, in this case I just tried to keep consistency with the other test. Happy to do any of the ways tho

pkg/services/govmomi/service.go
rikatz: Added. I tried to keep some balance between comments on blocks but forgot this one, sorry :)

pkg/services/govmomi/service_test.go
chrischdi: Lets keep as is then 👍

PR #2469, opened 2023-10-29
:seedling: Add flag to reserve memory to the virtual machine memory size

### LINKED ISSUE #2468: Add an option to pin memory for virtual machine
/kind feature

As an operator I would like to pin memory to max for workload clusters running at production environment.
Memory reservation management via the flag will allow operator to override the value of the virtual machine template for test environments in which there may be limited memory resources. As an operator, I will be able to opt out of different build scenarios for test and production environments.

Option has already available by vsphere api.

**Environment:**

- Cluster-api-provider-vsphere version: v1.6.0
- Kubernetes version: (use `kubectl version`): v1.22.17
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04.6 LTS

maximsnezhkov:
Add flag to configure memory reservation at VSphereMachineTemplate level.
Fixes #2468

### DIFF
--- a/pkg/services/govmomi/vcenter/clone.go
+++ b/pkg/services/govmomi/vcenter/clone.go
@@ -167,6 +167,12 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
 		memMiB = 2048
 	}
 
+	var memLockedToMax *bool
+	if vmCtx.VSphereVM.Spec.MemoryReservationLockedToMax != nil {
+		memLockedToMax = new(bool)
+		*memLockedToMax = *vmCtx.VSphereVM.Spec.MemoryReservationLockedToMax
+	}
+
 	// Disable the vAppConfig during VM creation to ensure Cloud-Init inside of the guest does not
 	// activate and prefer the OVF datasource over the VMware datasource.
 	vappConfigRemoved := true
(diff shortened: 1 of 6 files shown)


killianmuldoon (commented):
/ok-to-test

killianmuldoon:
Thanks for the contribution @maximsnezhkov  - it would be great to add this to an e2e test so we can ensure it works as expected. Do we know if there's limitations on which environments or versions of VSphere this is available on?

maximsnezhkov:
> Thanks for the contribution @maximsnezhkov - it would be great to add this to an e2e test so we can ensure it works as expected. Do we know if there's limitations on which environments or versions of VSphere this is available on?

You're welcome. As I understand it, this functionality has been available for use since at least 2013 (https://frankdenneman.nl/2013/02/21/reserve-all-guest-memory-all-locked/)

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -167,6 +167,12 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
 		memMiB = 2048
 	}
 
+	var memLockedToMax *bool
+	if vmCtx.VSphereVM.Spec.MemoryReservationLockedToMax != nil {
+		memLockedToMax = new(bool)
```
chrischdi, line 172:
```suggestion
```

That should not be required right?

maximsnezhkov: Yep. This is only required while using pass through devices and thats already in the code.

chrischdi: So let's remove it?

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -176,14 +182,15 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
...
-			MemoryMB:          memMiB,
-			VAppConfigRemoved: &vappConfigRemoved,
+			InstanceUuid:                 string(vmCtx.VSphereVM.UID),
+			Flags:                        newVMFlagInfo(),
+			DeviceChange:                 deviceSpecs,
+			ExtraConfig:                  extraConfig,
+			NumCPUs:                      numCPUs,
+			NumCoresPerSocket:            numCoresPerSocket,
+			MemoryMB:                     memMiB,
+			MemoryReservationLockedToMax: memLockedToMax,
```
chrischdi, line 192:
```suggestion
			MemoryReservationLockedToMax: vmCtx.VSphereVM.Spec.MemoryReservationLockedToMax,
```

We can directly reference that setting.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -167,6 +167,12 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
 		memMiB = 2048
 	}
 
+	var memLockedToMax *bool
+	if vmCtx.VSphereVM.Spec.MemoryReservationLockedToMax != nil {
+		memLockedToMax = new(bool)
+		*memLockedToMax = *vmCtx.VSphereVM.Spec.MemoryReservationLockedToMax
+	}
+
```
chrischdi, line 175:
```suggestion
```

Should not be required when directly referring below

pkg/services/govmomi/vcenter/clone.go
chrischdi: obsolete with the below.

sbueringer:
@maximsnezhkov Are you still interested in pursuing this PR (and do you have time to adress the open comments)?

sbueringer:
/close

See above

PR #2473, opened 2023-10-30
:bug: Re-add return on vimmachine error

rikatz:
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2395 removed a return from reconciliation.

When an error occurs, the caller cannot be able to verify there was an error, and moves forward, calling the pointer of object "vm", which is null

This PR re-adds https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/commit/a3f651f1606429b718a1d4583bfa827665167e0e#diff-2f616e5f5f40526609f3723e82cbcf01b16b9f240ea00fcdf3fc06b6aa6d164cL132


killianmuldoon (commented):
/lgtm
Thanks for catching this!

chrischdi:
Seems like I totally missed that removal on review. Thanks for fixing it!

/approve

chrischdi:
/hold

for @Ankitasw to maybe confirm that I did not miss something here.

PR #2505, opened 2023-11-14
🐛 Fix potential panics when retrieving VSphereCluster from Cluster.spec.infrastructureRef

sbueringer:

To be honest. Not sure how Machines can exist before Cluster.spec.infrastructureRef is set, but given someone hit this case with CAPD (https://github.com/kubernetes-sigs/cluster-api/pull/9673), I think there is no reason not to play it safe in CAPV as well.

Fixes #

### DIFF
--- a/controllers/vspherecluster_reconciler.go
+++ b/controllers/vspherecluster_reconciler.go
@@ -526,7 +526,7 @@ func (r *clusterReconciler) controlPlaneMachineToCluster(ctx context.Context, o 
 		log.Error(err, "VSphereMachine is missing cluster label or cluster does not exist")
 		return nil
 	}
-	log = log.WithValues("Cluster", klog.KObj(cluster), "VSphereCluster", klog.KRef(cluster.Namespace, cluster.Spec.InfrastructureRef.Name))
+	log = log.WithValues("Cluster", klog.KObj(cluster))
 	ctx = ctrl.LoggerInto(ctx, log)
 
 	if conditions.IsTrue(cluster, clusterv1.ControlPlaneInitializedCondition) {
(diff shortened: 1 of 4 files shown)


controllers/vspheremachine_controller.go
```diff
@@ -239,25 +239,29 @@ func (r *machineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
 
 	// Checking whether cluster is nil here as we still want to allow delete even if cluster is not found.
 	if cluster == nil {
+		log.Info("Cluster could not be fetched")
 		return reconcile.Result{}, nil
 	}
 
+	if cluster.Spec.InfrastructureRef == nil {
+		log.Info("Cluster.spec.infrastructureRef is not yet set")
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
fabriziopandini, line 248: q: why are we using requeing after here? if I'm not wrong this controller is watching for clusters, so it will reconcile again when InfrastructureRef is set.

controllers/vspherecluster_reconciler.go
```diff
@@ -537,6 +537,14 @@ func (r *clusterReconciler) controlPlaneMachineToCluster(ctx context.Context, o
 		return nil
 	}
 
+	if cluster.Spec.InfrastructureRef == nil {
+		log.Error(err, "Failed to get VSphereCluster: Cluster.spec.infrastructureRef is not yet set")
```
fabriziopandini, line 541: q: should this be log.info given that we are returning nil?

chrischdi (commented):
/lgtm

ok with the requeue for now. But let's clarify and open an issue for that if we want to add a watch instead.

sbueringer:
Thx! I'm leaning towards just changing the watch. I see no reason why we couldn't watch the cluster in the supervisor mode. But I want to give Naadir some time to respond


/hold

controllers/vspheremachine_controller.go
randomvariable: No, I don't think so. Should be fine.

sbueringer:
@fabriziopandini @chrischdi PTAL :)

chrischdi (commented):
/approve

sbueringer:
/hold cancel

@fabriziopandini @randomvariable an lgtm would be great :)

sbueringer:
@chrischdi Can you lgtm?

chrischdi (commented):
/lgtm

sbueringer:
Given the merge conflicts I would skip cherry-picking. I mostly opened this PR because we did similar changes to CAPD. There are no reports of CAPV observing this issue.

PR #2509, opened 2023-11-15
✨ Adds new field to set the minimum hardware version

### LINKED ISSUE #2212: Expose minimum hardware version on the VSphereVM spec
/kind feature

Similar to the [hardware version field](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/2bf748d66f951fb1dc518bbdd81e2ff0611240f1/apis/v1beta1/types.go#L206) on the `infrastructure.cluster.x-k8s.io` VSphereVM API types, we'd like to introduce a similar field on the API types under the `vmware.infrastructure.cluster.x-k8s.io` API group as well.

With some recent updates to the underlying VMOperator provisioner used in vCenter Workload Management, the hardware version of the VirtualMachine now gets inferred from the VMClass instead of being calculated from the hardware version set in the OVA. If the hardware version is not set on the VMClass, then it defaults to the vCenter datacenter/cluster/host compatibility version defaults.
Since hardware version dictates the hardware compatibility of VMs, CAPV users should have the ability to be able to prescribe the minimum hardware version that a VM should adhere to.

This would require pulling in the updated VMOp APIs which expose setting this field on the Spec.

srm09:
Adds a new field to set the minimum hardware version on the VSphereMachine Spec. This field is used to a similar field in the VMOperator's VirtualMachineSpec.

Fixes #2212


srm09:
/retitle ✨ Adds new field to set the minimum hardware version

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/stretchr/testify v1.8.4
 	github.com/vmware-tanzu/net-operator-api v0.0.0-20210401185409-b0dc6c297707
-	github.com/vmware-tanzu/vm-operator/api v1.8.2
+	github.com/vmware-tanzu/vm-operator/api v1.8.3-0.20231114230806-852c1641447a
```
sbueringer, line 17: Fine for now, but you might want to bump to a release tag before the next CAPV minor release

sbueringer:
Thank you!!

/lgtm
/approve

(merge pending retest after VMC is fixed)

pkg/util/vmx_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+package util
+
+import (
+	"testing"
+
+	"github.com/onsi/gomega"
+)
+
+func Test_LessThan(t *testing.T) {
```
sbueringer, line 25: Thx for the additional test coverage!!

PR #2535, opened 2023-12-08
:bug: Improve clustermodule existence check

rikatz:
NodeAntiAffinity uses an api called clustermodules. This API does not have a "GET" operation, which means that checking it existence needs to do a list and iterate over this list, being an expensive operation.

But the "membership"  API, part of clustermodules does have a Get members that, in case of the clustermodule not existing, will throw a 404 error.

Using this check instead may help improving the usage of clustermodules on large environments.

**Note from Ricardo** ~I'm testing this PR locally on a place I have found this to be a problem :) Will post more information soon~ - This works really fine and the consumption of rest endpoint has felt a lot!


pkg/services/govmomi/clustermodules/provider.go
```diff
@@ -87,19 +88,18 @@ func (cm *provider) DoesModuleExist(ctx context.Context, moduleID string, cluste
...
+	_, err := cm.manager.ListModuleMembers(ctx, moduleID)
+	if err == nil {
+		return true, nil
 	}
 
-	for _, mod := range modules {
-		if mod.Cluster == clusterRef.Value && mod.Module == moduleID {
-			return true, nil
-		}
+	// Is there a better way of unwrap vAPI errors?
```
chrischdi, line 96:
```suggestion
```

govmomi in this case returns the private type `statusError`: https://github.com/vmware/govmomi/blob/v0.33.1/vapi/rest/client.go#L150

pkg/services/govmomi/clustermodules/provider.go
```diff
@@ -87,19 +88,18 @@ func (cm *provider) DoesModuleExist(ctx context.Context, moduleID string, cluste
...
+	if strings.Contains(err.Error(), "404 Not Found") {
```
chrischdi, line 97: Not sure if that actually is a better variant.
We can't do a good type assertion because the error type is a private one

```suggestion
	if strings.Contains(err.Error(), fmt.Sprintf("%d %s", http.StatusNotFound, http.StatusText(http.StatusNotFound))) {
```

pkg/services/govmomi/clustermodules/provider.go
sbueringer: I would suggest to ask govmomi folks if they are open to expose a util func to extract more details (like the status code)

rikatz: https://github.com/vmware/govmomi/pull/3310 :)

rikatz:
ok, will wait for a new govmomi version to bump here :)

sbueringer:
/hold cancel

logging PR is merged

rikatz:
/hold
Will bump https://github.com/vmware/govmomi/releases/tag/v0.34.0 tomorrow and use new error check method

pkg/services/govmomi/clustermodules/provider.go
```diff
@@ -88,20 +91,19 @@ func (cm *provider) DoesModuleExist(ctx context.Context, moduleUUID string, clus
 		return false, nil
 	}
 
-	modules, err := cm.manager.ListModules(ctx)
-	if err != nil {
-		return false, err
+	_, err := cm.manager.ListModuleMembers(ctx, moduleUUID)
+	if err == nil {
+		log.V(4).Info("Cluster module exists", "moduleUUID", moduleUUID)
```
sbueringer, line 96:
```suggestion
		log.V(4).Info("Cluster module exists")
```

moduleUUID is already added a few levels above. If we add it here we will have a duplicate k/v pair.

I know it's not ideal that we have to care about that, but that's how klog behaves (for performance reasons, they don't want to go through k/v pair deduplication for every log line)

pkg/services/govmomi/clustermodules/provider.go
```diff
@@ -88,20 +91,19 @@ func (cm *provider) DoesModuleExist(ctx context.Context, moduleUUID string, clus
...
 	}
 
-	for _, mod := range modules {
-		if mod.Cluster == clusterRef.Value && mod.Module == moduleUUID {
-			log.V(4).Info("Cluster module does exist", "computeClusterRef", clusterRef)
-			return true, nil
-		}
+	// TODO: Once govmomi supports error checking for rest client, replace the below function
+	if strings.Contains(err.Error(), fmt.Sprintf("%d %s", http.StatusNotFound, http.StatusText(http.StatusNotFound))) {
+		log.V(4).Info("Cluster module doesn't exist", "moduleUUID", moduleUUID)
```
sbueringer, line 102:
```suggestion
		log.V(4).Info("Cluster module doesn't exist")
```

pkg/services/govmomi/clustermodules/provider.go
```diff
@@ -88,20 +91,19 @@ func (cm *provider) DoesModuleExist(ctx context.Context, moduleUUID string, clus
...
+		return true, nil
 	}
 
-	for _, mod := range modules {
-		if mod.Cluster == clusterRef.Value && mod.Module == moduleUUID {
-			log.V(4).Info("Cluster module does exist", "computeClusterRef", clusterRef)
-			return true, nil
-		}
+	// TODO: Once govmomi supports error checking for rest client, replace the below function
+	if strings.Contains(err.Error(), fmt.Sprintf("%d %s", http.StatusNotFound, http.StatusText(http.StatusNotFound))) {
```
sbueringer, line 101: When you bump govmomi and replace this. Can you please check if we can also get rid of `util.IsNotFoundError` above? (just randomly noticed it, wasn't aware of that util before)

(please check if we can get rid of both usages of that util and the util itself)

pkg/services/govmomi/clustermodules/provider.go
rikatz: ouch, had no idea :) removing it here

rikatz: btw I have also removed the clusterRef from the logging, as we don't use it anymore on any query

pkg/services/govmomi/clustermodules/provider.go
rikatz: done

pkg/services/govmomi/clustermodules/provider.go
sbueringer: Sounds good!

pkg/services/govmomi/clustermodules/provider.go
```diff
@@ -80,28 +79,26 @@ func (cm *provider) DeleteModule(ctx context.Context, moduleUUID string) error {
 }
 
 // DoesModuleExist checks whether a module with a given name exists with the passed clusterRef and moduleUUID.
-func (cm *provider) DoesModuleExist(ctx context.Context, moduleUUID string, clusterRef types.ManagedObjectReference) (bool, error) {
+func (cm *provider) DoesModuleExist(ctx context.Context, moduleUUID string, _ types.ManagedObjectReference) (bool, error) {
```
sbueringer, line 82: Let's drop the clusterRef parameter entirely.

It's slightly more works but it also means we don't have to retrieve the clusterRef in the "regular" case on the callsite of this func

rikatz: done

sbueringer:
/approve


/assign @chrischdi 
for final review


@rikatz Would be probably nice to cherry-pick into 1.8 / 1.7?

rikatz:
@sbueringer yeah, on my side we are using CAPV 1.8 so would be great, but also I'm happy to cherrypick to v1.7 (if everything fails here due to rebases, I can do it manually!)

chrischdi (commented):
/lgtm
/approve

sbueringer:
Sounds good, thank you!

I think automatic cherry-pick will fail with the logging stuff (I think you had to rebase after the logging PR was merged)

PR #2552, opened 2023-12-13
🐛 test: use GinkgoRecover in go subroutine of ownerreference test

chrischdi:

Properly uses `defer GinkgoRecover` in go routine.

Error occured in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/2550/pull-cluster-api-provider-vsphere-e2e-full-main/1734914800363769856

Fixes #

### DIFF
--- a/test/e2e/ownerreference_test.go
+++ b/test/e2e/ownerreference_test.go
@@ -264,6 +264,7 @@ func forcePeriodicReconcile(ctx context.Context, c ctrlclient.Client, namespace 
 	ticker := time.NewTicker(20 * time.Second)
 	stopTimer := time.NewTimer(5 * time.Minute)
 	go func() {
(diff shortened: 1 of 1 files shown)


sbueringer:
Thx!

Let's go with that for now.

I'm not sure if we want some sort of eventually. Right now the test will fail if c.List fails once if I understand it correctly

/lgtm
/approve

chrischdi:
> Thx!
> 
> Let's go with that for now.
> 
> I'm not sure if we additionally want some sort of eventually or retry . Right now the test will fail if c.List fails once if I understand it correctly
> 
> /lgtm /approve

Yep, same thought.

PR #2553, opened 2023-12-13
🌱 Fix live reload settings for Tilt

fabriziopandini:
Aligning with the list of folders where the CAPV manager code is read from


chrischdi (commented):
/approve

PR #2554, opened 2023-12-13
[release-1.8] :bug: Improve clustermodule existence check

rikatz:
Cherry-pick of #2535  for release 1.8


rikatz:
/assign @sbueringer 

@chrischdi @sbueringer I will try to make the cherry-pick for 1.7 tomorrow, just waiting this one to pass to make sure I didn't broke anything between releases

pkg/util/errors.go
```diff
@@ -1,26 +0,0 @@
...
-*/
-
-package util
-
-import (
-	"net/http"
-	"strings"
-)
-
-func IsNotFoundError(err error) bool {
```
chrischdi, line 24: I think we should keep this around in the old version.

sbueringer: I think we don't have to be as strict as in core CAPI. In core CAPI we have to assume that every exported func can be used by someone importing CAPI as a library.

I wouldn't be as strict in CAPV

sbueringer:
/lgtm

I'm fine either way regarding the public func

chrischdi:
If stefans okay then I am too

/approve

PR #2556, opened 2023-12-15
🌱 Implement consistent pause behavior

### LINKED ISSUE #2510: Audit pause implementation of all controllers
We should audit all CAPV controllers to ensure we have a clean implementation for pause.

I noticed that this is not the case for all of them, some examples (IIRC):
* VSphereMachine: only blocks reconcileNormal not reconcileDelete
* VSphereVM: blocks both reconciles, as long as we can get the cluster of the VSphereVM

Let's take a closer look at core CAPI, but in general when an object is paused "nothing" should be done (we should skip out of the reconcile very early).

sbueringer:
The goal of this PR is to implement consistent pause behavior, which is:
* if cluster.spec.paused is true => do nothing (if the cluster object is available)
* if reconciled object has paused annotation => do nothing

I mostly tried to keep the previous behavior regarding what should happen if we can't retrieve the cluster object. This is handled very differently across various providers. Maybe the reconciles still have to work even if the Cluster object is already gone, I have no idea.

Fixes #2510

### DIFF
--- a/controllers/serviceaccount_controller.go
+++ b/controllers/serviceaccount_controller.go
@@ -147,6 +147,20 @@ func (r *ServiceAccountReconciler) Reconcile(ctx context.Context, req reconcile.
 		return reconcile.Result{}, err
 	}
 
+	cluster, err := clusterutilv1.GetClusterFromMetadata(ctx, r.Client, vsphereCluster.ObjectMeta)
+	if err != nil {
+		return reconcile.Result{}, errors.Wrapf(err, "failed to get Cluster from VSphereCluster")
+	}
+	log = log.WithValues("Cluster", klog.KObj(cluster))
+	ctx = ctrl.LoggerInto(ctx, log)
+
+	// Pause reconciliation if entire VSphereCluster or Cluster is paused
+	// Note: Pause on the ProviderServiceAccount level is handled in ensureProviderServiceAccounts.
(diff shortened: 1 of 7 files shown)


sbueringer:
@fabriziopandini @chrischdi Would be good to get this done before the v1.9 release.

Happy to jump on a Zoom session to discuss the various cases

controllers/serviceaccount_controller.go
```diff
@@ -257,8 +259,8 @@ func (r *ServiceAccountReconciler) ensureProviderServiceAccounts(ctx context.Con
 		log := log.WithValues("EnsureProviderServiceAccount", klog.KRef(pSvcAccount.Namespace, pSvcAccount.Name))
 		ctx := ctrl.LoggerInto(ctx, log)
 
-		if guestClusterCtx.Cluster != nil && annotations.IsPaused(guestClusterCtx.Cluster, &(pSvcAccounts[i])) {
-			log.V(4).Info("Skipping ensure ProviderServiceAccount as ProviderServiceAccount is paused or belongs to a cluster that is paused ")
+		if annotations.HasPaused(&(pSvcAccounts[i])) {
```
sbueringer, line 262: .Cluster was always nil because before this PR we didn't set it in the cluster context

I think it's fine to not check it here as we already check if the cluster is paused above

PR #2581, opened 2023-12-28
🌱 Replace util/pointer with util/ptr

fabriziopandini:
Replace deprecated util/pointer with util/ptr + fix some lint issue


chrischdi (commented):
/lgtm
/approve

PR #2601, opened 2024-01-08
:bug: Remove logout from keepalive handler

rikatz:
Apparently calling Logout() from keepalive handler is prune to deadlocks. 

Instead, we should rely on calling logout from the session cache only, but ignore the other errors that may come from keepalive handler

Fixes #

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -263,9 +263,6 @@ func newClient(ctx context.Context, sessionKey string, url *url.URL, thumbprint 
 			_, err := methods.GetCurrentTime(ctx, tripper)
 			if err != nil {
 				log.Error(err, "Failed to keep alive govmomi client, Clearing the session now")
-				if errLogout := c.Logout(ctx); errLogout != nil {
-					log.Error(err, "Failed to logout keepalive failed session")
-				}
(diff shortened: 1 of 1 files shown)


rikatz:
running some scale tests again to be sure that this will not start leaking connections to VC once more :)

chrischdi (commented):
Sounds good. We would run logout anyway from outside the keepalive handler, right?

sbueringer:
/lgtm

Would be really great if we can confirm via some tests to avoid further back and forth (I know it's a super tricky issue :D)

> We would run logout anyway from outside the keepalive handler, right?

Not after we removed it from the sessionCache I guess?

rikatz:
/hold

Will post some test results here later

About testing, yeah, this is going to be a really hard one :) we can just catch this under scale and vc heavy loaded, so would need something on vcsim to inject random failures and guarantee that the login number does not grow exponentially

sbueringer:
We've disabled the keep alive handler per default: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2896

and we're going to remove it in a future release.

/close

(@rikatz feel free to reopen if you want to pursue this anyway)

PR #2605, opened 2024-01-09
🌱 Fix error log when default level cannot be set

sbueringer:

Just noticed that %v doesn't make sense when bumping CR in core CAPI (in core CAPI it will be addressed when bumping CR)

Fixes #


chrischdi (commented):
/lgtm
/approfve

PR #2612, opened 2024-01-11
🌱 e2e: Claim ip addresses dynamically so tests can run in parallel

### LINKED ISSUE #2342: Run e2e tests with higher concurrency
Currently all our e2e tests are run sequentially which is pretty slow.

We should improve them so they can run in parallel.

I tried to run with higher concurrency here: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2340

But this failed apparently because we only get one control plane IP in our e2e.sh and then cannot use it for multiple clusters of course.

I would suggest to try to move the IP allocation to a BeforeEach step.

Notes:
* I think for folks running the e2e tests outside of our VMC env we should still make it possible to run the tests sequentially as of today (i.e. that the IP is allocated up-front)
* We could also migrate to the CAPI IPAM controller when we do this

chrischdi:

Fixes #2342

### DIFF
--- a/test/e2e/capi_machine_deployment_rollout_test.go
+++ b/test/e2e/capi_machine_deployment_rollout_test.go
@@ -18,15 +18,29 @@ package e2e
 
 import (
 	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
 	capi_e2e "sigs.k8s.io/cluster-api/test/e2e"
+
+	"sigs.k8s.io/cluster-api-provider-vsphere/test/e2e/ipam"
 )
 
 var _ = Describe("ClusterAPI Machine Deployment Tests", func() {
 	Context("Running the MachineDeployment rollout spec", func() {
+		var (
+			testSpecificClusterctlConfigPath string
+			testSpecificIPAddressClaims      ipam.IPAddressClaims
+		)
+		BeforeEach(func() {
+			testSpecificClusterctlConfigPath, testSpecificIPAddressClaims = ipamHelper.ClaimIPs(ctx, clusterctlConfigPath)
+		})
+		defer AfterEach(func() {
+			Expect(ipamHelper.Cleanup(ctx, testSpecificIPAddressClaims)).To(Succeed())
+		})
+
 		capi_e2e.MachineDeploymentRolloutSpec(ctx, func() capi_e2e.MachineDeploymentRolloutSpecInput {
 			return capi_e2e.MachineDeploymentRolloutSpecInput{
 				E2EConfig:             e2eConfig,
-				ClusterctlConfigPath:  clusterctlConfigPath,
+				ClusterctlConfigPath:  testSpecificClusterctlConfigPath,
 				BootstrapClusterProxy: bootstrapClusterProxy,
 				ArtifactFolder:        artifactFolder,
 				SkipCleanup:           skipCleanup,
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -38,6 +38,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
+	. "sigs.k8s.io/cluster-api-provider-vsphere/test/e2e/helper"
+	"sigs.k8s.io/cluster-api-provider-vsphere/test/e2e/ipam"
 	vsphereframework "sigs.k8s.io/cluster-api-provider-vsphere/test/framework"
 )
 
@@ -84,6 +86,13 @@ var (
 	bootstrapClusterProxy framework.ClusterProxy
 
 	namespaces map[*corev1.Namespace]context.CancelFunc
+
+	// e2eIPAMKubeconfig is a kubeconfig to a cluster which provides IP address management via an in-cluster
+	// IPAM provider to claim IPs for the control plane IPs of created clusters.
+	e2eIPAMKubeconfig string
+
+	// ipamHelper is used to claim and cleanup IP addresses used for kubernetes control plane API Servers.
+	ipamHelper ipam.Helper
 )
 
 func init() {
@@ -92,6 +101,7 @@ func init() {
 	flag.BoolVar(&alsoLogToFile, "e2e.also-log-to-file", true, "if true, ginkgo logs are additionally written to the `ginkgo-log.txt` file in the artifacts folder (including timestamps)")
 	flag.BoolVar(&skipCleanup, "e2e.skip-resource-cleanup", false, "if true, the resource cleanup after tests will be skipped")
 	flag.BoolVar(&useExistingCluster, "e2e.use-existing-cluster", false, "if true, the test uses the current cluster instead of creating a new one (default discovery rules apply)")
+	flag.StringVar(&e2eIPAMKubeconfig, "e2e.ipam-kubeconfig", "", "path to the kubeconfig for the IPAM cluster")
 }
 
 func TestE2E(t *testing.T) {
@@ -134,8 +144,6 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	e2eConfig, err = vsphereframework.LoadE2EConfig(ctx, configPath)
 	Expect(err).NotTo(HaveOccurred())
 
-	By("Initializing the vSphere session to ensure credentials are working", initVSphereSession)
-
 	Byf("Creating a clusterctl local repository into %q", artifactFolder)
 	clusterctlConfigPath, err = vsphereframework.CreateClusterctlLocalRepository(ctx, e2eConfig, filepath.Join(artifactFolder, "repository"), true)
 	Expect(err).NotTo(HaveOccurred())
@@ -146,29 +154,51 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 
 	By("Initializing the bootstrap cluster")
 	vsphereframework.InitBootstrapCluster(ctx, bootstrapClusterProxy, e2eConfig, clusterctlConfigPath, artifactFolder)
-	namespaces = map[*corev1.Namespace]context.CancelFunc{}
+
+	ipamLabels := ipam.GetIPAddressClaimLabels()
+	var ipamLabelsRaw []string
+	for k, v := range ipamLabels {
+		ipamLabelsRaw = append(ipamLabelsRaw, fmt.Sprintf("%s=%s", k, v))
+	}
+
 	return []byte(
 		strings.Join([]string{
 			artifactFolder,
 			configPath,
 			clusterctlConfigPath,
 			bootstrapClusterProxy.GetKubeconfigPath(),
+			strings.Join(ipamLabelsRaw, ";"),
 		}, ","),
 	)
 }, func(data []byte) {
 	// Before each ParallelNode.
 	parts := strings.Split(string(data), ",")
-	Expect(parts).To(HaveLen(4))
+	Expect(parts).To(HaveLen(5))
 
 	artifactFolder = parts[0]
 	configPath = parts[1]
 	clusterctlConfigPath = parts[2]
 	kubeconfigPath := parts[3]
+	ipamLabelsRaw := parts[4]
+
+	namespaces = map[*corev1.Namespace]context.CancelFunc{}
+
+	By("Initializing the vSphere session to ensure credentials are working", initVSphereSession)
 
 	var err error
 	e2eConfig, err = vsphereframework.LoadE2EConfig(ctx, configPath)
 	Expect(err).NotTo(HaveOccurred())
 	bootstrapClusterProxy = framework.NewClusterProxy("bootstrap", kubeconfigPath, initScheme(), framework.WithMachineLogCollector(LogCollector{}))
+
+	ipamLabels := map[string]string{}
+	for _, s := range strings.Split(ipamLabelsRaw, ";") {
+		splittedLabel := strings.Split(s, "=")
+		Expect(splittedLabel).To(HaveLen(2))
+
+		ipamLabels[splittedLabel[0]] = splittedLabel[1]
+	}
+	ipamHelper, err = ipam.New(e2eIPAMKubeconfig, ipamLabels, skipCleanup)
+	Expect(err).ToNot(HaveOccurred())
 })
 
 // Using a SynchronizedAfterSuite for controlling how to delete resources shared across ParallelNodes (~ginkgo threads).
@@ -178,10 +208,16 @@ var _ = SynchronizedAfterSuite(func() {
 	// After each ParallelNode.
 }, func() {
 	// After all ParallelNodes.
+	if !skipCleanup {
+		By("Cleaning up orphaned IPAddressClaims")
+		vSphereFolderName, err := getClusterctlConfigVariable(clusterctlConfigPath, "VSPHERE_FOLDER")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(ipamHelper.Teardown(ctx, vSphereFolderName, vsphereClient)).To(Succeed())
+	}
 
 	By("Cleaning up the vSphere session", terminateVSphereSession)
-	By("Tearing down the management cluster")
 	if !skipCleanup {
+		By("Tearing down the management cluster")
 		vsphereframework.TearDown(ctx, bootstrapClusterProvider, bootstrapClusterProxy)
 	}
 })
--- a/test/e2e/govmomi_test.go
+++ b/test/e2e/govmomi_test.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"net/url"
 	"os"
+	"path/filepath"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -31,6 +32,7 @@ import (
 	"github.com/vmware/govmomi/vapi/rest"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/soap"
+	"sigs.k8s.io/yaml"
 )
 
 var (
@@ -91,4 +93,27 @@ func initVSphereSession() {
 
 func terminateVSphereSession() {
 	Expect(vsphereClient.Logout(ctx)).To(Succeed())
+}
+
+func getClusterctlConfigVariable(clusterctlConfigPath, configVariable string) (string, error) {
+	// Environment variable overwrite takes priority.
+	if val := os.Getenv(configVariable); val != "" {
+		return val, nil
+	}
+
+	data, err := os.ReadFile(filepath.Clean(clusterctlConfigPath))
+	if err != nil {
+		return "", err
+	}
+
+	type clusterctlConfig struct {
+		Variables map[string]string `json:"variables"`
+	}
+
+	config := clusterctlConfig{}
+	if err := yaml.Unmarshal(data, &config); err != nil {
+		return "", err
+	}
+
+	return config.Variables[configVariable], nil
 }
--- a/test/e2e/ipam/ipamhelper.go
+++ b/test/e2e/ipam/ipamhelper.go
@@ -1,0 +1,394 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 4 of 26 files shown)


test/e2e/capi_machine_deployment_rollout_test.go
```diff
@@ -19,14 +19,21 @@ package e2e
...
 )
 
 var _ = Describe("ClusterAPI Machine Deployment Tests", func() {
 	Context("Running the MachineDeployment rollout spec", func() {
+		var ipamCleanupFunc func()
+		var localClusterctlConfigPath string
+		defer AfterEach(func() { ipamCleanupFunc() })
+
 		capi_e2e.MachineDeploymentRolloutSpec(ctx, func() capi_e2e.MachineDeploymentRolloutSpecInput {
+			localClusterctlConfigPath, ipamCleanupFunc = ipam.ConfigureControlPlaneIPs(ctx, e2eIPAMKubeconfig, clusterctlConfigPath, skipCleanup)
```
sbueringer, line 33: Any reason this is not in a BeforeEach like in anti_affinity_test.go?

test/e2e/multivc_test.go
```diff
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
+	. "sigs.k8s.io/cluster-api-provider-vsphere/test/e2e/helper"
```
sbueringer, line 36:
```suggestion
```

sbueringer: Does this test also need a CP IP?

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+
+		return nil
+	}, time.Second*30, time.Second).Should(Succeed())
+
+	ip := &ipamv1.IPAddress{}
+
+	Expect(s.client.Get(ctx, client.ObjectKey{Namespace: claim.GetNamespace(), Name: claim.Status.AddressRef.Name}, ip)).To(Succeed())
+	Expect(ip.Spec.Address).ToNot(Equal(""))
+
```
sbueringer, line 149: With these expects we will leak any resources if the Expect fails, correct?

Would it help to return an error instead that we check one level higher?

Or maybe we get the same result if we move l.151 up?

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,225 @@
...
+// cleanup deletes the IPAddressClaims created by the service if the test succeeded.
+func (s *ipamService) cleanup() {
+	if s.skipCleanup {
+		By("Skipping cleanup of IPAddressClaims because skipCleanup is set")
+		return
+	}
+	if CurrentSpecReport().Failed() {
+		By("Skipping cleanup of IPAddressClaims because the tests failed and the IPs could still be in use")
+		return
+	}
```
sbueringer, line 165: This looks like we will leak IPAM objects on all failed tests (?)

This seems pretty bad

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,225 @@
...
+
+	for _, ipAddressClaim := range s.ipAddressClaims {
+		ipAddressClaim := ipAddressClaim
+		Byf("Deleting IPAddressClaim %s/%s", ipAddressClaim.GetNamespace(), ipAddressClaim.GetName())
+		Expect(s.client.Delete(context.Background(), &ipAddressClaim)).To(Succeed())
+	}
+}
+
+// Note: Copy-paste from CAPI below.
```
sbueringer, line 174: Makes sense for this PR and its cherry-picks. Let's follow-up and export this in core CAPI so we can get rid of the copy on CAPV main

sbueringer (commented):
Overall looks good!

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,225 @@
...
+
+func init() {
+	ipamScheme = runtime.NewScheme()
+	_ = ipamv1.AddToScheme(ipamScheme)
+}
+
+// ConfigureControlPlaneIPs claims one or more ip addresses if the environment variable CONTROL_PLANE_ENDPOINT_IP is not set.
+// It returns a path to a new clusterctl configuration, which contains the variable for CONTROL_PLANE_ENDPOINT_IP as well as
+// for additional variables passed via additionalIPVariableNames and a cleanup function which is supposed to get called
+// in an AfterEach ginkgo function.
```
sbueringer, line 55: This comment seems to be outdated (as we always get IPs when e2eIPAMKubeconfig is set)

test/e2e/e2e_suite_test.go
```diff
@@ -92,6 +97,7 @@ func init() {
 	flag.BoolVar(&alsoLogToFile, "e2e.also-log-to-file", true, "if true, ginkgo logs are additionally written to the `ginkgo-log.txt` file in the artifacts folder (including timestamps)")
 	flag.BoolVar(&skipCleanup, "e2e.skip-resource-cleanup", false, "if true, the resource cleanup after tests will be skipped")
 	flag.BoolVar(&useExistingCluster, "e2e.use-existing-cluster", false, "if true, the test uses the current cluster instead of creating a new one (default discovery rules apply)")
+	flag.StringVar(&e2eIPAMKubeconfig, "e2e.ipam-kubeconfig", "", "path to the kubeconfig for the IPAM cluster")
```
sbueringer, line 100: Let's update test/e2e/README.md

(should be clear how this flag and CONTROL_PLANE_ENDPOINT_IP interact / are mutually exclusive)

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,225 @@
+/*
```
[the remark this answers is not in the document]
chrischdi: It does not necessarily need to, but I could think of this being helpful getting imported and used by CPI and/or shared with future supervisor e2e tests.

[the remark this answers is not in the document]
sbueringer: Okay let's keep it here, makes sense

test/e2e/ipam/ipamhelper.go
chrischdi: Hmm thinking about this again: yes that's bad 🤔 

However, we need to be careful, that an IP Address is really not in-use anymore (== Control Plane VMs do not exist anymore). Otherwise re-using that IP won't work either / cause issues.

test/e2e/multivc_test.go
chrischdi: This tests can't run in upstream CI which is why I did not use the IPAM helper here.

This file uses the `Byf` function though which I moved to the seperated package.

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,225 @@
...
+	Expect(ip.Spec.Address).ToNot(Equal(""))
+
+	// Store claim inside the service so the cleanup function knows what to delete.
+	s.ipAddressClaims = append(s.ipAddressClaims, *claim)
+
+	return ip.Spec.Address
+}
+
+// cleanup deletes the IPAddressClaims created by the service if the test succeeded.
+func (s *ipamService) cleanup() {
```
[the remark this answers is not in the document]
sbueringer: If we do this let's make it very clear please, e.g. via an additional flag. Otherwise it can be a very magic side effect that other (and potentially independent) IPAddressClaims in our IPAM cluster are deleted by running e2e tests.

Wondering if it should be a separate binary altogether (that we can also run in a separate ProwJob). We have a bunch of other objects to cleanup (like VMs)

test/e2e/ipam/ipamhelper.go
chrischdi: whoops, that was by accident Oo, don't know how I did that sorry

chrischdi: I wanted to append it to my comment here 🤦  https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2612#discussion_r1452343106 wtf

test/e2e/multivc_test.go
sbueringer: Let's assume we would be able to run this test, would it then make sense to use the func? I think then I would be in favor of using the helper here as well for consistency (I assume it's still a no-op if the e2e config is not set)

test/e2e/ipam/ipamhelper.go
sbueringer: Yeah unfortunately. So we should only do this if the cluster has been deleted cleanly

test/e2e/ipam/ipamhelper.go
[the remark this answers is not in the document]
chrischdi: Yep, the general cleanup is also required for other projects we share the code + env with (In summary: CAPV, CPI, image-builder) and these require different credentials too.

We currently have to cleanup this manually but this should maybe be a different periodic or a cronjob instead.

sbueringer:
@chrischdi Just fyi. Linked a related issue I created a while back in the PR description

test/e2e/ipam/ipamhelper.go
chrischdi: Implemented a Teardown in SynchronizedAfter which handles the deletion of these orphaned ones if possible.

test/e2e/e2e_suite_test.go
```diff
@@ -84,6 +86,16 @@ var (
...
+
+	// e2eIPAMKubeconfig is a kubeconfig to a cluster which provides IP address management via an in-cluster
+	// IPAM provider to claim IPs for the control plane IPs of created clusters.
+	e2eIPAMKubeconfig string
+
+	// testSpecificClusterctlConfigPath gets written before running a test and is a copy of the clsuterctlConfigPath with
+	// the addition of variables for control plane IP addresses.
+	testSpecificClusterctlConfigPath string
+	// testSpecificIPAMCleanupFunc is used to cleanup IP addresses after successfully running a test.
+	testSpecificIPAMCleanupFunc func()
```
sbueringer, line 98: I would prefer if we keep these as variables with local scope instead of having package global variables.

I guess it works because we never run more than one test sequentially in one process, but it feels a bit dirty to continuously update package global variables when going through the tests.

test/e2e/govmomi_test.go
```diff
@@ -92,3 +94,26 @@ func initVSphereSession() {
 func terminateVSphereSession() {
 	Expect(vsphereClient.Logout(ctx)).To(Succeed())
 }
+
+func getClusterctlConfigVariable(clusterctlConfigPath, configVariable string) (string, error) {
```
sbueringer, line 98: Also seems like something that we should make a util exported by CAPI (in a follow-up PR)

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,373 @@
...
+		return err
+	}
+
+	ipAddressClaimsToDelete := []ipamv1.IPAddressClaim{}
+	// Collect errors and skip these ip address claims, but report at the end.
+	var errList []error
+
+	ip := &ipamv1.IPAddress{}
+	for _, ipAddressClaim := range ipAddressClaims.Items {
+		if err := ipamClient.Get(ctx, client.ObjectKey{Namespace: ipAddressClaim.GetNamespace(), Name: ipAddressClaim.Status.AddressRef.Name}, ip); err != nil {
```
sbueringer, line 126: Does it make sense here to check if AddressRef is nil?

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,373 @@
...
+			errList = append(errList, errors.Wrapf(err, "Getting IPAddress for IPAddressClaim %s/%s", ipAddressClaim.GetNamespace(), ipAddressClaim.GetName()))
```
sbueringer, line 127:
```suggestion
			errList = append(errList, errors.Wrapf(err, "getting IPAddress for IPAddressClaim %s/%s", ipAddressClaim.GetNamespace(), ipAddressClaim.GetName()))
```

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,373 @@
...
+		}
+
+		ipAddressClaimsToDelete = append(ipAddressClaimsToDelete, ipAddressClaim)
+	}
+
+	for _, ipAddressClaim := range ipAddressClaimsToDelete {
+		ipAddressClaim := ipAddressClaim
+		Byf("Deleting IP Addressclaim %s/%s", ipAddressClaim.GetNamespace(), ipAddressClaim.GetName())
+		if err := ipamClient.Delete(ctx, &ipAddressClaim); err != nil {
+			errList = append(errList, errors.Wrapf(err, "Deleting IPAddressClaim %s/%s", ipAddressClaim.GetNamespace(), ipAddressClaim.GetName()))
```
sbueringer, line 143:
```suggestion
			errList = append(errList, errors.Wrapf(err, "Deleting IPAddressClaim %s", klog.KObj(ipAddressClaim))
```

Not super important here in test code, just wanted to bring it up. Usually this is the idiomatic way that we're doing this (also very close to what we use when we add it as a k/v pair).

The resulting log here will be the same though

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+
+	var vm mo.VirtualMachine
+	virtualMachineIPAddresses := map[string]bool{}
+
+	// Iterate over the VirtualMachines, get the `guest.net` property and extract the IP addresses.
+	for _, mobj := range managedObjects {
+		// Get guest.net properties for mobj.
+		if err := vSphereClient.RetrieveOne(ctx, mobj.Object.Reference(), []string{"guest.net"}, &vm); err != nil {
+			// if err := vmobj.Properties(ctx, vmobj.Reference(), []string{"guest.net"}, &vm); err != nil {
```
sbueringer, line 178:
```suggestion
```
I think?

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,373 @@
...
+			return nil, errors.Wrapf(err, "Get properties of VM %s", mobj.Object.Reference())
```
sbueringer, line 179:
```suggestion
			return nil, errors.Wrapf(err, "get properties of VM %s", mobj.Object.Reference())
```

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+
+	return &ipamService{
+		ipAddressClaims: []ipamv1.IPAddressClaim{},
+		client:          ipamClient,
+		skipCleanup:     skipCleanup,
+	}, nil
+}
+
+func getClientSet(e2eIPAMKubeconfig string) (client.Client, error) {
```
sbueringer, line 214: Would it make sense to call this getClient?

ClientSet is usually the client-go thing which (IIRC) is actually a set of clients

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,373 @@
...
+		By("Skipping cleanup of IPAddressClaims because the tests failed and the IPs could still be in use")
+		return
+	}
+
+	var errList []error
+
+	for _, ipAddressClaim := range s.ipAddressClaims {
+		ipAddressClaim := ipAddressClaim
+		Byf("Deleting IPAddressClaim %s/%s", ipAddressClaim.GetNamespace(), ipAddressClaim.GetName())
+		if err := s.client.Delete(context.Background(), &ipAddressClaim, client.DryRunAll); err != nil {
```
sbueringer, line 301: client.DryRunAll, leftover?

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,373 @@
...
+			errList = append(errList, err)
+		}
+	}
+
+	if len(errList) > 0 {
+		Expect(kerrors.NewAggregate(errList)).ToNot(HaveOccurred())
+	}
+}
+
+func getIPAddressClaimLabels() map[string]string {
```
sbueringer, line 311: I wonder if we have to safeguard against this returning an empty map.

In this case we would create IPAddressClaims without labels and then try to delete all of them (even the ones unrelated to our test run) (+/- only if corresponding VMs don't exist)

sbueringer:
Mostly just nits

test/e2e/ipam/ipamhelper.go
sbueringer: If I'm not mistaken you are only adding the label key when the env var is set

chrischdi: ah damn, I should have looked at the code first. You're totally right.
My thinking was: when running it locally that you don't have to set them.

chrischdi:
Unrelated flake


/test pull-cluster-api-provider-vsphere-test-main

sbueringer:
@chrischdi Let's try to cherry-pick this with the bot. Maybe we'll get lucky

chrischdi:
/test pull-cluster-api-provider-vsphere-e2e-full-main

@sbueringer I changed some stuff. I guess this needs another review wdyt? :-)

test/e2e/e2e_suite_test.go
```diff
@@ -146,29 +154,53 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 	bootstrapClusterProxy = framework.NewClusterProxy("bootstrap", kubeconfigPath, initScheme(), framework.WithMachineLogCollector(LogCollector{}))
+
+	ipamLabels := map[string]string{}
+	for _, s := range strings.Split(ipamLabelsRaw, ";") {
+		splittedLabel := strings.Split(s, "=")
+		Expect(splittedLabel).To(HaveLen(2))
+
+		ipamLabels[splittedLabel[0]] = splittedLabel[1]
+	}
+	if e2eIPAMKubeconfig != "" {
```
sbueringer, line 200: If I see correctly this means the tests panic now in beforeeach if e2eIPAMKubeconfig is ""

sbueringer: Based on the Implementation of ipam.New it looks like it's good enough to drop l.200

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,384 @@
...
+	// Cleanup deletes the given IPAddressClaims.
+	Cleanup(ctx context.Context, claims IPAddressClaims) error
+
+	// Teardown tries to cleanup orphaned IPAddressClaims by checking if the corresponding IPs are still in use in vSphere.
+	// It identifies IPAddressClaims via labels.
+	Teardown(ctx context.Context, folderName string, vSphereClient *govmomi.Client) error
+}
+
+// New returns an ipam.Helper. If e2eIPAMKubeconfig is an empty string or skipCleanup is true
+// it will return a noop helper which does nothing so we cann fallback on setting environment variables.
```
sbueringer, line 76:
```suggestion
// it will return a noop helper which does nothing so we can fallback on setting environment variables.
```

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,384 @@
...
+func New(e2eIPAMKubeconfig string, labels map[string]string, skipCleanup bool) (Helper, error) {
+	if len(labels) == 0 {
+		return nil, fmt.Errorf("expecting labels to be set to prevent deletion of other IPAddressClaims")
+	}
+
+	if e2eIPAMKubeconfig == "" || skipCleanup {
```
sbueringer, line 82: I would have assumed skipCleanup only skips the deletion of the IPs, but this also skips the creation?

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,384 @@
...
+
+		ipAddressClaimsToDelete = append(ipAddressClaimsToDelete, &ipAddressClaim)
+	}
+
+	if err := h.Cleanup(ctx, ipAddressClaimsToDelete); err != nil {
+		// Group with possible previous errors.
+		errList = append(errList, err)
+	}
+
+	if len(errList) > 1 {
```
sbueringer, line 216:
```suggestion
	if len(errList) > 0 {
```

test/e2e/ipam/ipamhelper.go
```diff
@@ -0,0 +1,384 @@
...
+		if ip.Spec.Address == "" {
+			retryError = errors.Wrap(err, "IPAddress.Spec.Address is not set")
+			return false, nil
+		}
+
+		retryError = nil
+		return true, nil
+	})
+	if retryError != nil {
+		return "", claim, retryError
```
sbueringer, line 327:
```suggestion
		return "", nil, retryError
```

Let's return nil. Otherwise we should document what callers can do with claim in the err != nil case here

sbueringer (commented):
Getting closer :)

sbueringer:
Thx!

/lgtm

/assign @fabriziopandini

chrischdi:
Rebased + squashed, no changes.

/test pull-cluster-api-provider-vsphere-e2e-full-main

sbueringer:
Looks like you'll need another rebase

sbueringer:
Fine from my side. Not sure if Fabrizio wants to do another review

fabriziopandini (commented):
great work!
/lgtm
/approve

PR #2660, opened 2024-01-25
🐛 Fixing vmtoolsd path for/on a flatcar image

### LINKED ISSUE #2657: 🐛 Flatcar on vSphere (ova) - vmtoolsd. Path is not correct.
Hey there,
we have encountered a bug, while trying to use a flatcar self-builded image on vSphere.
Looking inside the code, the generator provides the full path to the binary, which is not correct. (At least in our cases)

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/d4e07cd3ec58d85645fd63faf2d835ba6649a31e/packaging/flavorgen/flavors/generators.go#L70

The binary for vmtoolsd is available inside the PATH, so no need to provide the full path to the binary.

I'd like to suggest, to remove the full the path and only use the binary. Something like this:

```
ExecStart=/usr/bin/bash -cv 'echo "COREOS_CUSTOM_HOSTNAME=$(vmtoolsd --cmd "info-get guestinfo.metadata" | base64 -d | grep local-hostname | awk {\'print $2\'} | tr -d \'"\')" > $${OUTPUT}'
```

Best,
Patrick

PatrickLaabs:
With this PR we do change the reference to the binary 'vmtoolsd'.
We removed the absolut path from

```
ExecStart=/usr/bin/bash -cv 'echo "COREOS_CUSTOM_HOSTNAME=$(/usr/share/oem/bin/vmtoolsd --cmd "info-get guestinfo.metadata" | base64 -d | grep local-hostname | awk {\'print $2\'} | tr -d \'"\')" > $${OUTPUT}'
```

to just use the vmtoolsd binary, which is already available inside the PATH:

```
ExecStart=/usr/bin/bash -cv 'echo "COREOS_CUSTOM_HOSTNAME=$(vmtoolsd --cmd "info-get guestinfo.metadata" | base64 -d | grep local-hostname | awk {\'print $2\'} | tr -d \'"\')" > $${OUTPUT}'
```

With this change, the kubeadm is able to do it thing and properly find the binary needed.
Without this change, the operation will fail, because the binary cannot be located at the specified path.

Fixes #2657


sbueringer:
The change look good to me.

I'm wondering why our ignition e2e test didn't catch this before (we're testing with flatcar-stable-3602.2.3-kube-v1.29.0)

sbueringer:
Looking at the CI failure I wonder if the binary is in the PATH in the image-builder flatcar image

PatrickLaabs:
Mhh.. weird.
I used the official image builder in its latest state.
I will double check my image, and might change to the absolut path where the vmtoolsd binary actually exists.

PatrickLaabs:
Ok, next try. I will build another flatcar image with the version you provided and double check.

PatrickLaabs:
Ok, now is see the problem.

On a flatcar image in version stable-3602.2.3 the binary can be found here:
Screenshot 2024-01-26 at 10 59 42

When using flatcar in version stable-3760.2.0 the binary can be found here:

chrischdi:
🤔 at best we would be compatible with both.

To summarise: we have flatcar versions where `vmtoolsd` is at :

* `/usr/share/oem/bin/vmtoolsd` (<= `3602.2.3`)
* `/usr/bin/vmtoolsd` >= `3760.2.0`

We could even pack the whole script here into a bash script and pass it via files 🤔 

```sh
      ExecStart=/usr/bin/bash -cv 'echo "COREOS_CUSTOM_HOSTNAME=$(/usr/bin/vmtoolsd --cmd "info-get guestinfo.metadata" | base64 -d | grep local-hostname | awk {\'print $2\'} | tr -d \'"\')" > $${OUTPUT}'
```

Idea: a command like `find /usr/share /usr/bin -iname "vmtoolsd" -type f | head -n 1` could maybe help to determine the binary path.

PatrickLaabs:
Ye, supporting both variants might be the better approach. I will do some testings.

chrischdi:
/retitle 🐛 Fixing vmtoolsd path for/on a flatcar image

PatrickLaabs:
This is current way, I am testing it:
```
ExecStart=/usr/bin/bash -cv 'VMTOOLSD_PATH=$(find /usr -iname 'vmtoolsd' -executable); if [[ -n "$vmtoolsd_path" ]]; then echo "COREOS_CUSTOM_HOSTNAME=$("$VMTOOLSD_PATH" --cmd "info-get guestinfo.metadata" | base64 -d | grep local-hostname | awk {\'print $2\'} | tr -d \'"\').eitco.de" > $${OUTPUT}; else echo "vmtoolsd not found in /usr/share/oem/bin/ or /usr/bin/"; fi'
```

Still doing some more testing, just to be sure.

For now: Running this modification on flatcar 3602 works. I was able to spawn 1 ControlPlane, and 3 Worker Nodes on my vSphere Instance.

packaging/flavorgen/flavors/generators.go
```diff
@@ -67,7 +67,7 @@ systemd:
       RemainAfterExit=yes
       Environment=OUTPUT=/run/metadata/coreos
       ExecStart=/usr/bin/mkdir --parent /run/metadata
-      ExecStart=/usr/bin/bash -cv 'echo "COREOS_CUSTOM_HOSTNAME=$(/usr/share/oem/bin/vmtoolsd --cmd "info-get guestinfo.metadata" | base64 -d | grep local-hostname | awk {\'print $2\'} | tr -d \'"\')" > $${OUTPUT}'
+      ExecStart=/usr/bin/bash -cv 'VMTOOLSD_PATH=$(find /usr -iname 'vmtoolsd' -executable); if [[ -n "$vmtoolsd_path" ]]; then echo "COREOS_CUSTOM_HOSTNAME=$("$VMTOOLSD_PATH" --cmd "info-get guestinfo.metadata" | base64 -d | grep local-hostname | awk {\'print $2\'} | tr -d \'"\')" > $${OUTPUT}; else echo "vmtoolsd not found in /usr/share/oem/bin/ or /usr/bin/"; fi'
```
chrischdi, line 70: Let's add a comment on why we are doing this :-) (at place X for version Y and place a for version B).

What do you think about only using if / else? The path should not change too much right?

```suggestion
      ExecStart=/usr/bin/bash -cv 'echo "COREOS_CUSTOM_HOSTNAME=$("$(find /usr/bin /usr/share/oem -name vmtoolsd -type f -executable 2>/dev/null | head -n 1)" --cmd "info-get guestinfo.metadata" | base64 -d | grep local-hostname | awk {\'print $2\'} | tr -d \'"\')" > $${OUTPUT}'
```

chrischdi: Maybe only use find to gather the command and let it fail if it fails?

we need `-type f` because directories can also be executable.

Also I think we don't have to search through the whole `/usr` directory. better to have some well-known paths to search in.

chrischdi: WDYT?

PatrickLaabs: Thanks for the suggestion!
I will try this one out. 👍

PatrickLaabs: Hey @chrischdi ,
i tried your suggestion on flatcar version 3602 and 3760. 
I was able to successfully spawn the control plane and the worker nodes.

PatrickLaabs:
The latest change on finding the vmtoolsd binary on the different pathes has been tested on my vsphere instance with CAPV v1.9.0 on the flatcar version '3602' and '3760'.

I was able successfully spawn the controlplane and worker nodes.
CNI has also successfully been deployed to make each node become ready.

Lets see, what the CI has to say about it 😄

chrischdi:
Note: failure of `pull-cluster-api-provider-vsphere-e2e-blocking-main` is unrelated. Its an CI env issue...

PatrickLaabs:
The failing test "pull-cluster-api-provider-vsphere-apidiff-main" gets me curious. Did i missed something? 😱

chrischdi:
Nope, that one is fine. It just notices us that a function signature changes. Which is fine in this case, because its only flavorgen code.

chrischdi:
/approve

/assign sbueringer 

For lgtm

PatrickLaabs:
> Nope, that one is fine. It just notices us that a function signature changes. Which is fine in this case, because its only flavorgen code.
> 
> (Edit: it does not impact merging / tide)

Thanks for the clearification 👍

sbueringer:
Looks good, thx!

PR #2661, opened 2024-01-26
🐛 kube-vip: fix prepare script to work on pre k8s v1.29

### LINKED ISSUE #2659: 🐛 kube-vip-prepare.sh - return does not work
Hey everyone,

we encountered the issue, that the script below does not work as one might expect.
Running flatcar with Kubernetes v1.28.6, the script failes to run, since a return is only valid inside a function.

We tried to modify it by replacing the "return" with "exit 0", which then starts to work.

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/d4e07cd3ec58d85645fd63faf2d835ba6649a31e/packaging/flavorgen/flavors/kubevip/kube-vip-prepare.sh#L22-L26

Here is our change:

```
KUBEADM_MINOR="$(kubeadm version -o short | cut -d '.' -f 2)"
if [[ "$KUBEADM_MINOR" -lt "29" ]]; then
  exit 0
fi
```

What you guys think?

chrischdi:

Fixes the kubve-vip-prepare.sh for kubernetes versions < v1.29

Fixes #2659

### DIFF
--- a/packaging/flavorgen/flavors/kubevip/kube-vip-prepare.sh
+++ b/packaging/flavorgen/flavors/kubevip/kube-vip-prepare.sh
@@ -22,7 +22,7 @@ set -e
 # Nothing to do for kubernetes < v1.29
 KUBEADM_MINOR="$(kubeadm version -o short | cut -d '.' -f 2)"
 if [[ "$KUBEADM_MINOR" -lt "29" ]]; then
-  return
+  exit 0
 fi
 
 IS_KUBEADM_INIT="false"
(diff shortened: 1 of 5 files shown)


PatrickLaabs:
Hi @chrischdi ,
not sure, if 'exit 1' is something that might break the actual workflow.
We did the testing with 'exit 0'.
Which will exit the script, but does not return an failed exit code to the shell.

Might this be something to consider?

chrischdi:
> Hi @chrischdi , not sure, if 'exit 1' is something that might break the actual workflow. We did the testing with 'exit 0'. Which will exit the script, but does not return an failed exit code to the shell.
> 
> Might this be something to consider?

Yes, you are totally right. The purpose is to continue, not break 👍

chrischdi:
Updated the PR, thanks for taking a look @PatrickLaabs !

chrischdi:
e2e flake, fix: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2663

sbueringer:
Okay make sense that we didn't find this issue in CI as we are using 1.29...

/lgtm
/approve

Thx!

PR #2674, opened 2024-01-29
🌱 test: fix error handling in incluster ip provider

chrischdi:

Fixes the error handling. Using errors.Wrap with a nil error results in not returning an error. In this case we want to create a new error which was why the errors were not bubbled up and the tests continued without any IP.

Fixes #

### DIFF
--- a/test/framework/ip/incluster.go
+++ b/test/framework/ip/incluster.go
@@ -309,7 +309,7 @@ func (h *inCluster) claimIPAddress(ctx context.Context) (_ string, _ *ipamv1.IPA
 		}
 
 		if claim.Status.AddressRef.Name == "" {
-			retryError = errors.Wrap(err, "IPAddressClaim.Status.AddressRef.Name is not set")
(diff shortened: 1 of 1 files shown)


fabriziopandini (commented):
/lgtm
/approve

test/framework/ip/incluster.go
```diff
@@ -318,14 +318,16 @@ func (h *inCluster) claimIPAddress(ctx context.Context) (_ string, _ *ipamv1.IPA
...
+			retryError = errors.New("IPAddress.Spec.Address is not set")
 			return false, nil
 		}
 
 		retryError = nil
 		return true, nil
 	})
 	if retryError != nil {
+		// Try best effort deletion of the unused claim before returning an error.
+		_ = h.client.Delete(ctx, claim)
```
sbueringer, line 330: Ideally we would aggregate errors if .Delete also returns one instead of hiding this one

(maybe something for a follow-up)

PR #2673, opened 2024-01-29
🌱 Introduce MODE and TARGET for E2E tests

fabriziopandini:
**What this PR does / why we need it**:
With this PR we introduce the concept of MODE (legacy, supervisor) and TARGET (VMC, BYO vCenter, vcsim) for E2E tests.
This will enable to have tests for supervisors and tests for the vcsim server in follow up prs.

However, it gives also an immediate benefit, because it allows simplifying how the test when using BYO vCenter (previously called dev) by dropping most of the duplicated E2E config file.

PR #2685, opened 2024-02-01
🌱 Git ignore config-overrides

fabriziopandini:
This is a follow-up of https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2673 ensuring that the config-overrides files, which is expected to contain potentially sensitive information about the user's vCenter is ignored by git.

Instead, we are providing a config-overrides example the users can start from.


sbueringer:
Thx!

/lgtm
/approve

PR #2690, opened 2024-02-02
🌱 Move vcsim object naming utils to vcsim helpers

fabriziopandini:
Move vcsim object naming utils to vcsim helpers so the can be re-used across the code base


chrischdi (commented):
/approve

PR #2692, opened 2024-02-02
🌱 Make PowerOff move configurable when using vcsim

fabriziopandini:
Make PowerOff move configurable when using vcsim


chrischdi (commented):
/lgtm
/approve

sbueringer:
How is this implemented and used? (future PR?)

fabriziopandini:
> How is this implemented and used? (future PR?)

Yes, but also a fix.
PowerOff mode was the only remaining hard-coded value in the list of settings that are included in the list of EnvVar and then used by templates, in this case for the upcoming tests for supervisor mode.

sbueringer:
Okay got it. Was just surprised to see it as env var as it's not even configurable in the regular Cluster templates we publish (templates/* folder)

PR #2727, opened 2024-02-08
:seedling Add option for non-mandatory IP/DHCP for network device

### LINKED ISSUE #2721: feat: make ip addresses on secondary interfaces optional
/kind feature

Currently all interfaces of a vSphereMachine need to have an IP Address for the actual VM to get created.
If you use those secondary interfaces for Multus you might not need an IP on those interfaces and having one can actually cause Issues.
Thus it would be nice to skip wating for IPAddrs on certain interfaces

n/a

**Environment:**

p-strusiewiczsurmacki-mobica:
This PR introduces `ipConfigurationNotMandatory` for for network device configuration (`deviceSpec`). If set `true` controller should not wait for IP address allocation for this network device.

Fixes #2721


apis/v1beta1/types.go
```diff
@@ -370,6 +370,10 @@ type NetworkDeviceSpec struct {
 	// For more information see the netplan reference (https://netplan.io/reference#dhcp-overrides)
 	// +optional
 	DHCP6Overrides *DHCPOverrides `json:"dhcp6Overrides,omitempty"`
+
+	// IPConfigurationNotMandatory allows for the device to not have IP address or DHCP configured.
+	// +optional
+	IPConfigurationNotMandatory bool `json:"ipConfigurationNotMandatory,omitempty"`
```
schrej, line 376: maybe we could use something a little more explicit here, e.g. `skipWaitForAddress`?

controllers/vspherevm_controller.go
```diff
@@ -473,7 +473,7 @@ func (r vmReconciler) reconcileNormal(ctx context.Context, vmCtx *capvcontext.VM
 func (r vmReconciler) isWaitingForStaticIPAllocation(vmCtx *capvcontext.VMContext) bool {
 	devices := vmCtx.VSphereVM.Spec.Network.Devices
 	for _, dev := range devices {
-		if !dev.DHCP4 && !dev.DHCP6 && len(dev.IPAddrs) == 0 && len(dev.AddressesFromPools) == 0 {
+		if !dev.DHCP4 && !dev.DHCP6 && len(dev.IPAddrs) == 0 && len(dev.AddressesFromPools) == 0 && !dev.IPConfigurationNotMandatory {
```
chrischdi, line 476: Can we have a comment here explaining this long if? Its hard to read and remember again.

We should also adjust the godoc of the function.

Maybe there's a way to write this function in a way so its more readable: (please check if the logic is correct)

```
// TODO explain
if dev.IgnoreIPAllocation {
  continue
}

// TODO explain
if len(dev.IPAddrs) > 0 || len(dev.AddressesFromPools) > 0 {
  continue
}

// TODO explain
if dev.DHCP4 || dev.DHCP6 {
  continue
}

// TODO explain
return true
```

apis/v1beta1/types.go
chrischdi: I think we should go into more detail at the godoc.

If this new variable gets set to true, this device gets ignored when checking for an existing ip address. And propably that at least one device should get an IP address (dunno if this is true but I can think of provisioning to fail at different places if there is no IP here)?

`SkipWaitForAddress` sounds better to me but not yet perfect. No strong opinion on the name or a better recommendation from my side though...

What about: `SkipIPAllocation` or `IgnoreIPAllocation`

zhanggbj: `SkipIPAllocation` or `IgnoreIPAllocation` sounds more straightforward to me. And I'm thinking perhaps adding `static` is more precise, like `SkipStaticIPAllocation` or `IgnoreStaticIPAllocation` as we're only skipping the static IP allocation but not the DHCP allocation.

controllers/vspherevm_controller.go
p-strusiewiczsurmacki-mobica: I've split this if statement into 3 separate statements as by tour suggestion.

p-strusiewiczsurmacki-mobica:
@chrischdi I've tried to add some more details into this godoc comment.
I can add a check to ensure that at least 1 device has IP configured/DHCP enabled, but I am not sure if that's necessary. If user has to explicitly set this flag to true I would assume they have at least a minimal knowledge of what and why they are doing anyway.
But maybe I am too optimistic about this, so I'll just let you decide on that. :)

chrischdi:
Looks like some generated code needs to get regenerated.

Thanks for tackling this. I'll try to come back to this tomorrow.

apis/v1beta1/types.go
```diff
@@ -316,7 +316,7 @@ type NetworkDeviceSpec struct {
 	// IPAddrs is a list of one or more IPv4 and/or IPv6 addresses to assign
 	// to this device.  IP addresses must also specify the segment length in
```
chrischdi, line 317:
```suggestion
	// to this device. IP addresses must also specify the segment length in
```

controllers/vspherevm_controller.go
```diff
@@ -473,8 +473,18 @@ func (r vmReconciler) reconcileNormal(ctx context.Context, vmCtx *capvcontext.VM
...
+			continue
+		}
+
+		// Ignore device if it is configured to use DHCP.
+		if dev.DHCP4 || dev.DHCP6 {
+			continue
+		}
+
+		// Static IP is not available yet.
+		if len(dev.IPAddrs) == 0 && len(dev.AddressesFromPools) == 0 {
```
chrischdi, line 487:
```suggestion
		if len(dev.IPAddrs) == 0 && len(dev.AddressesFromPools) == 0 {
			// One or more IPs are expected for the device but are not set yet.
```

pkg/services/govmomi/util.go
```diff
@@ -457,6 +457,11 @@ func waitForIPAddresses(
 				// Look at each IP and determine whether a reconcile has
 				// been triggered for the IP.
 				for _, discoveredIPInfo := range nic.IpConfig.IpAddress {
+					// Ignore device if SkipIPAllocation is set.
+					if deviceSpec.SkipIPAllocation {
```
chrischdi, line 461: Not sure if we want to run continue in this first case. Wouldn't this also skip discovering if IP Addresses get set on this interface?

If so, we should add it to the description that IPs for this device don't get bubbled up.

pkg/util/machines.go
```diff
@@ -109,10 +109,21 @@ func GetMachineMetadata(hostname string, vsphereVM infrav1.VSphereVM, ipamState
 			devices[i].Gateway6 = state.Gateway6
 		}
 
+		// Ignore device if SkipIPAllocation is set
```
chrischdi, line 112: Can we add why we want to skip this device here? What impact has that?

pkg/util/machines.go
```diff
@@ -109,10 +109,21 @@ func GetMachineMetadata(hostname string, vsphereVM infrav1.VSphereVM, ipamState
...
+		if vsphereVM.Spec.Network.Devices[i].SkipIPAllocation {
+			continue
+		}
+
 		if waitForIPv4 && waitForIPv6 {
 			// break early as we already wait for ipv4 and ipv6
 			continue
 		}
+
+		if vsphereVM.Spec.Network.Devices[i].SkipIPAllocation {
```
chrischdi, line 122: duplicate to above?

pkg/util/machines.go
p-strusiewiczsurmacki-mobica: After some consideration I think it's not really necessary. If the device has no static IP address or DHCP configured, then `waitForIPv4` and `waitForIPv6` should be `false` anyway.

pkg/services/govmomi/util.go
p-strusiewiczsurmacki-mobica: I think you're right. Even if there is IP address discovered but the device has no static IP or DHCP configured, the discovered address will be skipped anyway.
Deleted this if statement.

apis/v1beta1/types.go
```diff
@@ -370,6 +370,13 @@ type NetworkDeviceSpec struct {
 	// For more information see the netplan reference (https://netplan.io/reference#dhcp-overrides)
 	// +optional
 	DHCP6Overrides *DHCPOverrides `json:"dhcp6Overrides,omitempty"`
+
+	// SkipIPAllocation allows the device to not have IP address or DHCP configured.
+	// This is suitable for devices for which IP allocation is handled externally, eg. using Multus CNI.
+	// If true, both static IP address allocation and DHCP allocation will be skipped,
+	// and provider will not verify IP address allocation.
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Changed that to
```
// If true, CAPV will not verify IP address allocation.
```

chrischdi (commented):
/lgtm

@schrej could you maybe give this another look in its final state :-)

schrej:
lgtm!

Thanks @p-strusiewiczsurmacki-mobica 🎉

PR #2739, opened 2024-02-14
[WIP] 🌱 Restore v1alpha3 / v1alpha4

sbueringer:
More details to follow (I also have to think about conversion a bit)

Fixes #


sbueringer:
Tested the following locally:
* make tilt-up
* Deploy a cluster with the external-loadbalancer template
* Modify managed fields to add an entry with v1alpha3 via `kubectl apply /tmp/vspheremachinetemplate.yaml --server-side=false"
* Verified conversion webhook is called and it works when running `kubectl apply /tmp/vspheremachinetemplate.yaml --server-side=true"

So I would conclude that conversion works again and we can handle v1alpha3 entries in managedFields.

Additionally we have conversion fuzz-tests for all apiVersions now.

chrischdi:
/lgtm
/approve

/hold
so you can take it off

Do we need an issue to track again the final removal? (maybe when only k8s versions supported are the k/k fix lands)

PR #2751, opened 2024-02-19
🌱 Update Cluster API IPAM version from v1alpha1 to v1beta1

### LINKED ISSUE #2750: Update Cluster API IPAM version from v1alpha1 to v1beta1 in CAPV
While working with the CAPV, I noticed that the IPAM version used is still v1alpha1. This is despite the fact that the IPAM types have been promoted to v1beta1 in the cluster-api. Additionally, the first release of the Cluster API IPAM provider in-cluster uses v1beta1, leading to an inconsistency in the versions used.

NA

Thought it would be beneficial to align the versions used in CAPV with the promoted IPAM types in the cluster-api and the CAPI IPAM provider in-cluster.

**Environment:**

PavanNeerudu:

While working with the CAPV, I noticed that the IPAM version used is still v1alpha1. This is despite the fact that the IPAM types have been promoted to v1beta1 in the cluster-api. Additionally, the first release of the Cluster API IPAM provider in-cluster uses v1beta1, leading to an inconsistency in the versions used.

Fixes #2750


chrischdi (commented):
/ok-to-test

hack/ipclaim-template.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: ipam.cluster.x-k8s.io/v1alpha1
```
chrischdi, line 1: Can you remove this file? this was used for CI but is not used anymore at all :-)

pkg/services/govmomi/ipam/status_test.go
```diff
@@ -25,7 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apitypes "k8s.io/apimachinery/pkg/types"
-	ipamv1a1 "sigs.k8s.io/cluster-api/exp/ipam/api/v1alpha1"
+	ipamv1b1 "sigs.k8s.io/cluster-api/exp/ipam/api/v1beta1"
```
chrischdi, line 28: please rename this to `ipamv1`, maybe worth enforcing this via `.golangci.yml` like we already do for other apis?

chrischdi (commented):
Thanks for the PR, looks straight forward :-)

Some smaller nits.

test/framework/ip/incluster.go
```diff
@@ -39,7 +39,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
-	ipamv1 "sigs.k8s.io/cluster-api/exp/ipam/api/v1alpha1"
+	ipamv1 "sigs.k8s.io/cluster-api/exp/ipam/api/v1beta1"
```
chrischdi, line 42: Maybe we need to keep the old api group for now for the test code, because in CI v1beta1 is not yet deployed (it is only used for IPAM and thus may be not up to date.

PavanNeerudu: Yes, makes sense. I was just looking at why the tests were failing and how the testing framework is constructed. Thanks!

hack/ipclaim-template.yaml
PavanNeerudu: Removed in the latest patch.

pkg/services/govmomi/ipam/status_test.go
PavanNeerudu: Yeah, sure. Added a rule to .golangci.yml as well.

chrischdi:
/test pull-cluster-api-provider-vsphere-test-main

Unrelated flake, which gets hopefully fixed in: 

- #2752

zhanggbj (approved):
Thanks for the PR! 
/lgtm
Waiting for @chrischdi to have another review.

chrischdi:
Because we don't have any e2e test covering ipam: have you been able to give the implementation a try to create a cluster using IPAM?

chrischdi:
We now have a proper test (via #2757) :-)

/test pull-cluster-api-provider-vsphere-e2e-main

chrischdi:
Test is green 🎉 

Thanks for contributing this!

/tide merge-method squash

chrischdi:
/retest

flake

PR #2749, opened 2024-02-19
:bug: [WIP][DO NOT REVIEW] investigate flaky tests

chrischdi:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #2752, opened 2024-02-19
🐛 controllers/test: use APIReader to not assert on cached information

chrischdi:

This should resolve the flaky tests for the service discovery controller we have on main and release-1.9 from https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/controllers/servicediscovery_controller_intg_test.go.

The tests did use a cached client and did assert on cached client data. Using an uncached api reader / client instead makes the test stable.
Also this improves some timeouts in eventually instead of only trying it once.

Heavily tested at #2749

Should solve:

* [service discovery controller tests](https://storage.googleapis.com/k8s-triage/index.html?date=2024-02-19&job=.*-cluster-api-provider-vsphere.*&test=.*Service%20Discovery%20controller%20integration%20tests.*)

/cherry-pick release-1.9

Fixes #

### DIFF
--- a/controllers/serviceaccount_controller_suite_test.go
+++ b/controllers/serviceaccount_controller_suite_test.go
@@ -58,11 +58,11 @@ func createTargetSecretWithInvalidToken(ctx context.Context, guestClient client.
 	Expect(guestClient.Create(ctx, secret)).To(Succeed())
 }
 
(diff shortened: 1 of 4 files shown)


fabriziopandini (commented):
Nice one!
/lgtm
/approve

PR #2761, opened 2024-02-21
📖 docs: document tagging the test go module

chrischdi:

We added a go module in `test` since v1.9. We also should add tags for the test module.

Fixes #

### DIFF
--- a/docs/release/release-tasks.md
+++ b/docs/release/release-tasks.md
@@ -116,12 +116,15 @@ From this point forward changes which should land in the release have to be cher
       # Export the tag of the release to be cut, e.g.:
       export RELEASE_TAG=v1.8.0-beta.0
       # Create tags locally
-      # Warning: The test tag MUST NOT be an annotated tag.
       git tag -s -a ${RELEASE_TAG} -m ${RELEASE_TAG}
+      # Warning: The test tag MUST NOT be an annotated tag.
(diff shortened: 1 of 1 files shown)


fabriziopandini (commented):
/lgtm
/approve

sbueringer:
I don't mind but just wanted to say that we don't have a book in CAPV and we also probably don't want to keep the release tasks doc up-to-date on all active release branches

chrischdi:
I wanted to bubble up the release tasks, because the test tag was already forgotten for the last release (v1.9.2).

sbueringer:
Yup good point. But don't we have the problem for all release branches. And also in general if we rely on release-tasks.md being up-to-date on all branches?

chrischdi:
> Yup good point. But don't we have the problem for all release branches. And also in general if we rely on release-tasks.md being up-to-date on all branches?
> 
> (but definitely fine, this also might be a special case)

Ack, this special case was just the release-1.9 branch did lack this information forever, and I usually copy the tag + push commands from the branch so I don't miss the test tag / leave it out :-)

sbueringer:
All good! If you tend to do the same for other branches, let's cherry-pick this further :)

chrischdi:
No need to, the module was introduced in v1.9 :-)

PR #2767, opened 2024-02-26
🌱 Implement E2E test for vcsim in govmomi mode

fabriziopandini:
Add E2E test for vcsim in govmomi mode + one step closer to test supervisor mode

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -49,7 +49,7 @@ on_exit() {
     do
       echo "Cleaning up VSPHERE_PASSWORD from file ${file}"
       sed -i "s/${VSPHERE_PASSWORD}/REDACTED/g" "${file}"
-    done
+    done || true
     # Move all artifacts to the original artifacts location.
     mv "${ARTIFACTS}"/* "${ORIGINAL_ARTIFACTS}/"
   fi
@@ -76,7 +76,6 @@ export VSPHERE_SSH_PRIVATE_KEY="/root/ssh/.private-key/private-key"
 export E2E_CONF_FILE="${REPO_ROOT}/test/e2e/config/vsphere.yaml"
 export E2E_CONF_OVERRIDE_FILE=""
 export E2E_CAPV_MODE="${CAPV_MODE:-govmomi}"
-export E2E_TARGET_TYPE="${TARGET_TYPE:-vmc}"
 export ARTIFACTS="${ARTIFACTS:-${REPO_ROOT}/_artifacts}"
 export DOCKER_IMAGE_TAR="/tmp/images/image.tar"
 export GC_KIND="false"
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -26,6 +26,7 @@ import (
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/ginkgo/v2/types"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -42,6 +43,8 @@ import (
 	vsphereframework "sigs.k8s.io/cluster-api-provider-vsphere/test/framework"
 	vsphereip "sigs.k8s.io/cluster-api-provider-vsphere/test/framework/ip"
 	vspherelog "sigs.k8s.io/cluster-api-provider-vsphere/test/framework/log"
+	vspherevcsim "sigs.k8s.io/cluster-api-provider-vsphere/test/framework/vcsim"
+	vcsimv1 "sigs.k8s.io/cluster-api-provider-vsphere/test/infrastructure/vcsim/api/v1alpha1"
 )
 
 const (
@@ -57,10 +60,7 @@ const (
 )
 
 const (
-	// VMCTestTarget identify tests targeting VMC infrastructure used for CAPV CI.
-	VMCTestTarget string = "vmc"
-
-	// VCenterTestTarget identify tests targeting a user provided vCenter instance.
+	// VCenterTestTarget identify tests targeting a real vCenter instance, including also the VMC infrastructure used for CAPV CI.
 	VCenterTestTarget string = "vcenter"
 
 	// VCSimTestTarget identify tests targeting a vcsim instance (instead of a real vCenter).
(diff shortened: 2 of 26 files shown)


test/infrastructure/vcsim/config/default/manager_image_patch.yaml
```diff
@@ -7,5 +7,5 @@ spec:
   template:
     spec:
       containers:
-      - image: gcr.io/k8s-staging-capi-vsphere/cluster-api-vcsim-controller:main
+      - image: gcr.io/k8s-staging-capi-vsphere/cluster-api-vcsim-controller-arm64:dev
```
chrischdi, line 10: By accident?

chrischdi, line 10: commited by accident?

test/infrastructure/vcsim/controllers/virtualmachine_controller.go
```diff
@@ -138,8 +139,34 @@ func (r *VirtualMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			log.Info("Registering ResourceGroup for ControlPlaneEndpoint", "ResourceGroup", resourceGroup, "ControlPlaneEndpoint", listenerName)
+			err := r.APIServerMux.RegisterResourceGroup(listenerName, resourceGroup)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			found = true
+			break
+		}
+		if !found {
+			return ctrl.Result{}, errors.Errorf("unable to fine a ControlPlaneEndpoint for host %s, port %d", cluster.Spec.ControlPlaneEndpoint.Host, cluster.Spec.ControlPlaneEndpoint.Port)
```
chrischdi, line 167:
```suggestion
			return ctrl.Result{}, errors.Errorf("unable to find a ControlPlaneEndpoint for host %s, port %d", cluster.Spec.ControlPlaneEndpoint.Host, cluster.Spec.ControlPlaneEndpoint.Port)
```

test/e2e/quick_start_test.go
```diff
@@ -68,3 +68,36 @@ var _ = Describe("Cluster creation with [Ignition] bootstrap [PR-Blocking]", fun
 		})
 	})
 })
+
+var _ = Describe("Cluster Creation using Cluster API quick-start test on vcsim [vcsim]", func() {
```
chrischdi, line 72: Question: So we explicitly have to set `[vcsim]` here. The spec does not look too different from the normal quickstart?

Note: just want to clarify, I'm ok with this :-)

hack/e2e.sh
```diff
@@ -18,6 +18,13 @@ set -o errexit  # exits immediately on any unexpected error (does not bypass tra
 set -o nounset  # will error if variables are used without first being defined
 set -o pipefail # any non-zero exit code in a piped command causes the pipeline to fail with that code
 
+#TMP to test vcsim e2e before creating a job in test-infa
+echo "GINKGO_FOCUS=${GINKGO_FOCUS}"
+if [[ "${GINKGO_FOCUS}" = "\[PR-Blocking\]" ]]; then
+  export GINKGO_FOCUS="\[vcsim\]"
+fi
+echo "GINKGO_FOCUS=${GINKGO_FOCUS}"
+
```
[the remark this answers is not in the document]
chrischdi: (Note yes, its on the last commit)

test/infrastructure/tmp-to-be-deleted/server/listener.go
```diff
@@ -38,6 +38,8 @@ type WorkloadClusterListener struct {
 	host string
 	port int
```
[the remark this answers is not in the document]
fabriziopandini: yes

test/e2e/quick_start_test.go
fabriziopandini: We have a few constraints that make it possible to run vcsim tests together with regular test.
To enforce this, we had two options:
- Manage this manually by adding skip [vcsim] on all the existing tests
- Enforce this programmatically

I went for 2, automatically enforcing skip [vcsim] if there is no explicit focus [vcsim].
As a consequence, we have to have dedicated tests for [vcsim], but I think this is acceptable in this phase where we will have only quick start tests

test/e2e/e2e_suite_test.go
```diff
@@ -223,38 +247,60 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 	configPath = parts[1]
 	clusterctlConfigPath = parts[2]
 	kubeconfigPath := parts[3]
-	ipamLabelsRaw := parts[4]
+	ipClaimLabelsRaw := parts[4]
 
 	namespaces = map[*corev1.Namespace]context.CancelFunc{}
 
-	By("Initializing the vSphere session to ensure credentials are working", initVSphereSession)
+	if testTarget == VCenterTestTarget {
```
chrischdi, line 267: You have to pass through the `testTarget` variable via the strings.Join / split above, otherwise the ginkgo workers/nodes do not get the value when running tests in-parallel!

test/e2e/e2e_suite_test.go
```diff
@@ -223,38 +247,60 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
+		// but we are deferring this to future work (if an and when necessary).
+		By("Initializing the vSphere session to ensure credentials are working", initVSphereSession)
+	}
 
 	var err error
 	e2eConfig, err = vsphereframework.LoadE2EConfig(ctx, configPath)
 	Expect(err).NotTo(HaveOccurred())
+
+	// Add config overrides + drop vcsim relates provider/image if not necessary.
+	amendE2EConfig()
```
chrischdi, line 280: Nice catch that we need this here too :-) :+1:

test/e2e/e2e_suite_test.go
```diff
@@ -281,9 +341,51 @@ func initScheme() *runtime.Scheme {
 	sc := runtime.NewScheme()
 	framework.TryAddDefaultSchemes(sc)
 	_ = infrav1.AddToScheme(sc)
+	_ = vcsimv1.AddToScheme(sc)
 	return sc
 }
 
+func amendE2EConfig() {
```
chrischdi, line 348: optional: we could integrate this into `vsphereframework.LoadE2EConfig`, but ok to keep as is :-)

fabriziopandini: Let's keep this here for now, it depends on global variables

PR #2771, opened 2024-02-27
🌱 Make Flavor gen templates vcsim friendly

fabriziopandini:
Add CONTROL_PLANE_ENDPOINT_PORT env var / cluster class var as required by vcsim


chrischdi (commented):
/lgtm
/approve

PR #2784, opened 2024-02-29
🌱 Drop copy of the in-memory runtime and server

fabriziopandini:
In-memory runtime and server packages are now accessible in CAPI 🥳


chrischdi:
/lgtm

Only `make generate-modules` change 👍

PR #2796, opened 2024-03-04
🐛 Fixes issue with NCP versioning while using the semver package

gab-satchi:
NSX releases can have more than 3 segments. They frequently have 4. This causes an issue when there's an attempt to parse it with the semver package. 

In this PR, if a version with 4 segments is found, it will trim the last segment before parsing it as a semver. The codepath is only there to check that the version is >= 3.0.1 and < 3.1.0 so precision isn't lost by trimming down to 3 segments.

### DIFF
--- a/pkg/util/networkutil.go
+++ b/pkg/util/networkutil.go
@@ -18,6 +18,7 @@ package util
 
 import (
 	"context"
+	"strings"
 
 	"github.com/blang/semver/v4"
 	"github.com/pkg/errors"
@@ -78,6 +79,7 @@ func GetNamespaceNetSnatIP(ctx context.Context, controllerClient client.Client, 
 }
 
 // GetNCPVersion finds out the running ncp's version from its configmap.
+// If the version contains more than 3 segments, it will get trimmed down to 3.
 func GetNCPVersion(ctx context.Context, controllerClient client.Client) (string, error) {
 	configmapObj := &corev1.ConfigMap{}
 	namespacedName := apitypes.NamespacedName{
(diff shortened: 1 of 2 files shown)


zhanggbj (approved):
Thanks for fixing it!
/lgtm

pkg/util/networkutil.go
```diff
@@ -90,6 +91,13 @@ func GetNCPVersion(ctx context.Context, controllerClient client.Client) (string,
 	}
 
 	version := configmapObj.Data[NCPVersionKey]
+
+	// NSX doesn't stritcly follow SemVer and there are versions like 4.0.1.3
```
chrischdi, line 95: We should add a comment to the godoc of the function too.

pkg/util/networkutil.go
```diff
@@ -90,6 +91,13 @@ func GetNCPVersion(ctx context.Context, controllerClient client.Client) (string,
...
+	// This will cause an error if directly used in semver.Parse() and prevent the cluster from reconciling.
+	// Since GetNCPVersion is only used to check >= 3.0.1 and < 3.1.0, it's safe to trim the last segment
+	if segments := strings.Split(version, "."); len(segments) > 3 {
+		version = strings.Join(segments[:3], ".")
+	}
```
chrischdi, line 100:
```suggestion
	if segments := strings.Split(version, "."); len(segments) > 3 {
		return strings.Join(segments[:3], "."), nil
	}

```

chrischdi (commented):
/lgtm
/approve

PR #2802, opened 2024-03-08
e2e: disable log collector for vcsim testmode

chrischdi:

Disables the log collector when we're running the vcsim testtarget.

On vcsim there are no logs to collect because there is no machine which is running or we could ssh to.

Fixes #


fabriziopandini (commented):
Nice!
/lgtm
/approve

PR #2803, opened 2024-03-11
🌱 Add vm-operator tilt settings

fabriziopandini:
This PR add tilt settings for vm-operator using the same image/manifest we are using for CI.
NOTE: with this configuration we cannot debug vm-operator, but this is considered acceptable because we don't need this in most of the cases. 
If necessary to debug vm-operator it is requried to temporarily add a tilt settings in the vm-operator repo and do the same small modifications we are doing in the test/infrastructure/vm-operator folder


test/infrastructure/vm-operator/vm-operator-pull-policy.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Deployment
+metadata:
+  name: vmware-system-vmop-controller-manager
+  namespace: vmware-system-vmop
+spec:
+  template:
+    spec:
+      containers:
+        - name: manager
+          imagePullPolicy: Always
```
fabriziopandini, line 11: This aligns vm operator to all the other providers; TBD if we prefer if not present or if we want to switch to if not present in our E2E test config

fabriziopandini:
/hold
for https://github.com/kubernetes-sigs/cluster-api/pull/10244

chrischdi (commented):
/lgtm
/approve

sbueringer:
Quick note. If we realize that we have to debug the vm-operator regularly, I would explore if we can just build the vm-operator image with debug symbols. If I'm not missing anything we can then use dlv the same way.

Also IIRC there were some improvements over time in Go and dlv that some minimal debugging is even possible without including debug symbols into the binary.

But let's consider this if it ever becomes relevant

PR #2826, opened 2024-03-14
[release-1.6] 🌱 Disable govulncheck

sbueringer:

Quote from Slack: https://kubernetes.slack.com/archives/CKFGK3SSD/p1710424265062089

Hey folks,
the weekly scan found a few CVEs across all supported CAPV versions. We've fixed those in main / 1.9 / 1.8 / 1.7 by bumping the Go 1.21.8 and to a newer protobuf version.

It was not possible to fix CAPV 1.6 because it is based on Go 1.20. Go 1.20 is out of support and thus there is no patch release to fix the CVE.

We would ignore these CVEs for now as CAPV v1.10 will be released soon (a bit after the next CAPI minor versions). At this point CAPV v1.6 will be out of support and there will be no further patch releases.
If you're affected by these CVE's please consider upgrading to a newer CAPV version. Alternatively, we would recommend assessing if CAPV is actually by the CVE. Here is the scan: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/actions/runs/8232381773/job/22658011913

We think the only potentially relevant finding is
> Error:       #3: pkg/services/govmomi/vcenter/clone.go:221:34: vcenter.Clone calls pbm.NewClient, which eventually calls cookiejar.Jar.Cookies
> Error:       #4: pkg/services/govmomi/vcenter/clone.go:221:34: vcenter.Clone calls pbm.NewClient, which eventually calls cookiejar.Jar.SetCookies

To make it easier to keep an overview over our security scans that we run across all branches we'll disable govulncheck on release-1.6 (we don't have a more fine-granular way to ignore a CVE right now). Trivy will continue to run on release-1.6.

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -386,9 +386,10 @@ verify-govulncheck: $(GOVULNCHECK) ## Verify code for vulnerabilities
 
 .PHONY: verify-security
 verify-security: ## Verify code and images for vulnerabilities
+	## Note: govulncheck has been disabled on this branch, because there are CVEs in (out-of-support) Go 1.20.
 	$(MAKE) verify-container-images && R1=$$? || R1=$$?; \
-	$(MAKE) verify-govulncheck && R2=$$? || R2=$$?; \
-	if [ "$$R1" -ne "0" ] || [ "$$R2" -ne "0" ]; then \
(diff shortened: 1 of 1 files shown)


chrischdi (commented):
/lgtm
/approve

PR #2859, opened 2024-03-27
🌱 Fix noisy logs in vcsim

fabriziopandini:
Ensure a few vcsim logs line are generated only when something happens instead of every reconcile


chrischdi (commented):
/lgtm
/approve

PR #2878, opened 2024-04-04
🌱 github: still create a draft release on release notes generation failures

chrischdi:

Changes the github action which creates a draft release to not fail when generating release-notes failed.

Instead it will create the draft release with a note that the release-notes were not successfully generated.

This allows us to still manually generate the release notes and publish the release.

Fixes #

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -30,8 +30,10 @@ jobs:
         run: |
           make release
       - name: generate release notes
+        # Ignore failures for release-notes generation so they could still get
+        # generated manually before publishing.
         run: |
-          make generate-release-notes
+          make generate-release-notes || echo "Failed to generate release notes" >> _releasenotes/${{ env.RELEASE_TAG }}.md
         env:
           GH_TOKEN: ${{ github.token }}
       - name: Release


.github/workflows/release.yaml
```diff
@@ -30,8 +30,10 @@ jobs:
         run: |
           make release
       - name: generate release notes
+        # Ignore failures for release-notes generation so they could still get
+        # generated manually before publishing.
         run: |
-          make generate-release-notes
+          make generate-release-notes || echo "Failed to generate release notes" >> _releasenotes/${{ env.RELEASE_TAG }}.md
```
sbueringer, line 36: I assume this redirect only the echo into the file? (nothing from the make target)

chrischdi: oh well, let me check this, good point!

chrischdi: Yes, only redirects stdout from the echo!

chrischdi:
```sh
❯ (echo "hello" && fail) || echo "failed" >> out.txt
hello
zsh: command not found: fail

❯ cat out.txt
failed
```

sbueringer: Thx for checking!

PR #2906, opened 2024-04-15
🌱 templates: add a script to run prekubeadmcommands in order of files defined in /etc/pre-kubeadm-scripts

chrischdi:

The k8s v1.28 to v1.29 upgrade tests are currently failing on main and release-1.10, because the `prekubeadmscript.sh` gets run **after** `kube-vip-prepare.sh`. Because of that kube-vip fails on v1.28 and the kube-apiserver's public IP does not get up so provisioning including kubeadm fails.

This PR refactors the way we add additional pre-kubeadm scripts.

It changes the pattern from
* adding a script
* adding a preKubeadmCommand entry which runs the script

to
* adding scripts to `/etc/pre-kubeadm-commands`
* having a static script which runs the scripts inside that directory in alphabetical order

Fixes #


fabriziopandini (commented):
Nice!
/lgtm

sbueringer:
Nice!

/approve

PR #2917, opened 2024-04-18
🌱 Retry CRD detection during startup

fabriziopandini:
Retry CRD deletion during start-up, so the controller handle nice race conditions but it goes back to crash loopbacking when the issue is more persistent

### DIFF
--- a/main.go
+++ b/main.go
@@ -272,17 +272,17 @@ func main() {
 		govmomiGVR := infrav1.GroupVersion.WithResource(reflect.TypeOf(&infrav1.VSphereCluster{}).Elem().Name())
 		supervisorGVR := vmwarev1.GroupVersion.WithResource(reflect.TypeOf(&vmwarev1.VSphereCluster{}).Elem().Name())
 
-		var isSupervisorCRDLoaded, isGovimoniCRDLoaded bool
+		var isSupervisorCRDLoaded, isGovmomiCRDLoaded bool
 		if err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 30*time.Second, true, func(_ context.Context) (bool, error) {
-			var errGovimomi, errSupervisor error
+			var errGovmomi, errSupervisor error
 
 			// Check for non-supervisor VSphereCluster and start controller if found
-			isGovimoniCRDLoaded, errGovimomi = isCRDDeployed(mgr, govmomiGVR)
+			isGovmomiCRDLoaded, errGovmomi = isCRDDeployed(mgr, govmomiGVR)
 
 			// Check for supervisor VSphereCluster and start controller if found
 			isSupervisorCRDLoaded, errSupervisor = isCRDDeployed(mgr, supervisorGVR)
 
-			if (isGovimoniCRDLoaded && errGovimomi == nil) || (isSupervisorCRDLoaded && errSupervisor == nil) {
+			if (isGovmomiCRDLoaded && errGovmomi == nil) || (isSupervisorCRDLoaded && errSupervisor == nil) {
 				return true, nil
 			}
 			return false, nil
@@ -291,11 +291,11 @@ func main() {
 		}
 
 		// Continuing startup does not make sense without having managers added.
-		if !isSupervisorCRDLoaded && !isGovimoniCRDLoaded {
-			return errors.New("neither supervisor nor govimomi CRDs detected")
+		if !isSupervisorCRDLoaded && !isGovmomiCRDLoaded {
+			return errors.New("neither supervisor nor govmomi CRDs detected")
 		}
 
-		if isGovimoniCRDLoaded {
+		if isGovmomiCRDLoaded {
 			if err := setupVAPIControllers(ctx, controllerCtx, mgr, tracker); err != nil {
 				return fmt.Errorf("setupVAPIControllers: %w", err)
 			}
--- a/main.go
+++ b/main.go
@@ -19,7 +19,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -32,6 +31,7 @@ import (
 	"gopkg.in/fsnotify.v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	cliflag "k8s.io/component-base/cli/flag"
 	"k8s.io/component-base/logs"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -268,37 +269,46 @@ func main() {
 			return perrors.Wrapf(err, "unable to create remote cluster cache tracker")
 		}
 
-		// Check for non-supervisor VSphereCluster and start controller if found
-		gvr := infrav1.GroupVersion.WithResource(reflect.TypeOf(&infrav1.VSphereCluster{}).Elem().Name())
-		isNonSupervisorCRDLoaded, err := isCRDDeployed(mgr, gvr)
-		if err != nil {
-			return err
+		govimomiGVR := infrav1.GroupVersion.WithResource(reflect.TypeOf(&infrav1.VSphereCluster{}).Elem().Name())
```
sbueringer, line 272:
```suggestion
		govmomiGVR := infrav1.GroupVersion.WithResource(reflect.TypeOf(&infrav1.VSphereCluster{}).Elem().Name())
```

sbueringer: Same below

main.go
```diff
@@ -268,37 +269,46 @@ func main() {
...
+		if err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 30*time.Second, true, func(ctx context.Context) (bool, error) {
+			var errGovimomi, errSupervisor error
+
+			// Check for non-supervisor VSphereCluster and start controller if found
+			isGovimoniCRDLoaded, errGovimomi = isCRDDeployed(mgr, govimomiGVR)
+
+			// Check for supervisor VSphereCluster and start controller if found
+			isSupervisorCRDLoaded, errSupervisor = isCRDDeployed(mgr, supervisorGVR)
+
+			if errGovimomi != nil || errSupervisor != nil {
```
sbueringer, line 285: Not sure if we should change that logic. Was talking about this with Christian this morning and depending on the exact state of the cluster we can hit issues that we don't cover (e.g. if the apiGroup exists but not the VSphereCluster CRD).

What about changing this to return if one of the (loaded, error) pairs is (true, nil)

main.go
```diff
@@ -268,37 +269,46 @@ func main() {
...
+				return false, nil
+			}
+			return true, nil
+		}); err != nil {
+			return errors.New("failed to detect CRDs")
```
sbueringer, line 290: Should we use wrap?

main.go
sbueringer: > Not sure if we should change that logic

That might have been misleading. I didn't mean that we made a change compared to the previous version. More that we should consider changing the current version on the PR :)

main.go
```diff
@@ -268,37 +269,46 @@ func main() {
...
 
-		// Check for non-supervisor VSphereCluster and start controller if found
-		gvr := infrav1.GroupVersion.WithResource(reflect.TypeOf(&infrav1.VSphereCluster{}).Elem().Name())
-		isNonSupervisorCRDLoaded, err := isCRDDeployed(mgr, gvr)
-		if err != nil {
-			return err
+		govmomiGVR := infrav1.GroupVersion.WithResource(reflect.TypeOf(&infrav1.VSphereCluster{}).Elem().Name())
+		supervisorGVR := vmwarev1.GroupVersion.WithResource(reflect.TypeOf(&vmwarev1.VSphereCluster{}).Elem().Name())
+
+		var isSupervisorCRDLoaded, isGovimoniCRDLoaded bool
```
chrischdi, line 275:
```suggestion
		var isSupervisorCRDLoaded, isGovmomiCRDLoaded bool
```

main.go
```diff
@@ -268,37 +269,46 @@ func main() {
...
+		if err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 30*time.Second, true, func(_ context.Context) (bool, error) {
+			var errGovimomi, errSupervisor error
+
+			// Check for non-supervisor VSphereCluster and start controller if found
+			isGovimoniCRDLoaded, errGovimomi = isCRDDeployed(mgr, govmomiGVR)
```
chrischdi, line 280:
```suggestion
			isGovmomiCRDLoaded, errGovmomi = isCRDDeployed(mgr, govmomiGVR)
```

main.go
```diff
@@ -268,37 +269,46 @@ func main() {
...
+				return true, nil
+			}
+			return false, nil
+		}); err != nil {
+			return fmt.Errorf("failed to detect CRDs: %w", err)
+		}
+
+		// Continuing startup does not make sense without having managers added.
+		if !isSupervisorCRDLoaded && !isGovimoniCRDLoaded {
+			return errors.New("neither supervisor nor govimomi CRDs detected")
```
chrischdi, line 295:
```suggestion
			return errors.New("neither supervisor nor govmomi CRDs detected")
```

chrischdi (commented):
This also solves the problem when weird errors get returned for e.g. govmomi mode.

Naming nit :-), as of that LGTM

main.go
```diff
@@ -268,37 +269,46 @@ func main() {
...
 
-		// Check for non-supervisor VSphereCluster and start controller if found
-		gvr := infrav1.GroupVersion.WithResource(reflect.TypeOf(&infrav1.VSphereCluster{}).Elem().Name())
-		isNonSupervisorCRDLoaded, err := isCRDDeployed(mgr, gvr)
-		if err != nil {
-			return err
+		govmomiGVR := infrav1.GroupVersion.WithResource(reflect.TypeOf(&infrav1.VSphereCluster{}).Elem().Name())
+		supervisorGVR := vmwarev1.GroupVersion.WithResource(reflect.TypeOf(&vmwarev1.VSphereCluster{}).Elem().Name())
+
+		var isSupervisorCRDLoaded, isGovmoniCRDLoaded bool
```
chrischdi, line 275:
```suggestion
		var isSupervisorCRDLoaded, isGovmomiCRDLoaded bool
```

main.go
```diff
@@ -268,37 +269,46 @@ func main() {
...
+			}
+			return false, nil
+		}); err != nil {
+			return fmt.Errorf("failed to detect CRDs: %w", err)
+		}
+
+		// Continuing startup does not make sense without having managers added.
+		if !isSupervisorCRDLoaded && !isGovmoniCRDLoaded {
+			return errors.New("neither supervisor nor govmomi CRDs detected")
 		}
```
sbueringer: This is not possible anymore. One of them has to be true to not end up in an error in l.289

main.go
```diff
@@ -268,37 +269,46 @@ func main() {
...
+
+			// Check for supervisor VSphereCluster and start controller if found
+			isSupervisorCRDLoaded, errSupervisor = isCRDDeployed(mgr, supervisorGVR)
+
+			if (isGovmoniCRDLoaded && errGovmomi == nil) || (isSupervisorCRDLoaded && errSupervisor == nil) {
+				return true, nil
+			}
+			return false, nil
+		}); err != nil {
+			return fmt.Errorf("failed to detect CRDs: %w", err)
```
sbueringer, line 290: Should we define the errors outside of the poll and add them here with aggregate?

The problem is that we now only print the timeout error but not what actually failed

chrischdi (commented):
/lgtm
/approve

PR #2931, opened 2024-04-22
🌱 Generate vm-operator YAML instead of downloading it from GCS

sbueringer:

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -789,7 +789,7 @@ set-manifest-image:
 ##@ vm-operator:
 
 .PHONY: release-vm-operator
-release-vm-operator: docker-vm-operator-build-all vm-operator-manifest-build docker-vm-operator-push-all vm-operator-manifest-push ## Build and push the vm-operator image and manifest for usage in CI
+release-vm-operator: docker-vm-operator-build-all vm-operator-manifest-build docker-vm-operator-push-all ## Build and push the vm-operator image and manifest for usage in CI
 
 .PHONY: vm-operator-checkout
 vm-operator-checkout:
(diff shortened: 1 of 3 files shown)


sbueringer:
/hold

let's see if this actually works

Makefile
```diff
@@ -277,7 +277,7 @@ help:  # Display this help
 
 .PHONY: generate
 generate: ## Run all generate targets
-	$(MAKE) generate-modules generate-manifests generate-go-deepcopy generate-go-conversions generate-flavors
+	$(MAKE) generate-modules generate-manifests generate-vm-operator-manifest generate-go-deepcopy generate-go-conversions generate-flavors
```
sbueringer, line 280: TBD if we always want to run this target in pre-submit (takes a bit of time, but I don't thing very long)

@fabriziopandini @chrischdi WDYT?

chrischdi: I would say no, creates additional requests to clone a repo.

sbueringer: I'm fine either way. It's a trade-off between always running these comments vs potentially the files get outdated.

Given that these files should only change when we bump vm-operator and then we'll change the file anyway it should be fine to drop this target from generate / verify-gen.

I can change the output-file to contain the version. This way it becomes really hard to miss when reviewing vm-operator bumps of the vsphere.yaml

sbueringer: Done

test/e2e/config/vsphere.yaml
```diff
@@ -217,7 +217,7 @@ providers:
     versions:
       - name: v1.8.6-0-gde75746a
         # Use manifest from source files
-        value: "https://storage.googleapis.com/artifacts.k8s-staging-capi-vsphere.appspot.com/vm-operator/v1.8.6-0-gde75746a.yaml"
+        value: "file://../../../../cluster-api-provider-vsphere/test/infrastructure/vm-operator/vm-operator-v1.8.6-0-gde75746a.yaml"
```
sbueringer, line 220: Not sure why we are going so far up that we have to go back into the CAPV repo dir, but I copied this from vc-sim.

One reason could be that this config file can even be used from 4 levels deep in another repo (?)

sbueringer:
/assign @fabriziopandini @chrischdi 

I would cherry-pick this after this PR merges to all relevant branches and then cleanup the files in the bucket.

chrischdi (commented):
/lgtm
/approve

sbueringer:
/hold

so Fabrizio can take a look if he wants

test/e2e/config/vsphere.yaml
fabriziopandini: I think this is because paths are relative to the test file.
if this is true, probably we are going up too much, but this seems a generalized smell all around this file

fabriziopandini:
/lgtm
/approve

I would suggest to unhold only after the release
Also, are you planning to cleanup the published manifests as soon as they are not necessary anymore?

test/e2e/config/vsphere.yaml
sbueringer: Yup, agree :)

sbueringer:
Looks like we only need it in release-1.10. release-1.9 doesn't seem to have the vm-operator yet

Will wait with merging the cherry-pick until after the release tomorrow

PR #2955, opened 2024-04-30
🌱 test: use a daemonset and kubectl exec to provide images to remote clusters

chrischdi:
**What this PR does / why we need it**:

Removes pushing the images to gcloud and instead uses a daemonset and streams the image tarball via kubectl exec.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #2990, opened 2024-05-13
🌱 hack: remove references to GCR_KEY_FILE

chrischdi:

Follow-up to #2955

Fixes #


sbueringer:
/lgtm
/approve

I assume we do the corresponding cleanup in test-infra. I assume we don't have to care about actually deleting the secret in the Prow cluster as we can just invalidate it by maybe removing the corresponding service account?

PR #2989, opened 2024-05-13
🌱 hack/e2e don't add binary files to artifacts and also censor base64 encoded values

chrischdi:
**What this PR does / why we need it**:

Improves our censoring in e2e.sh to:

* not add binary files (e.g. gzipped files) to the output artifacts because we can't control their data
* also censor the base64 version of our secret

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #3006, opened 2024-05-15
[release-1.9] 🌱 hack/e2e don't add binary files to artifacts and also censor base64 encoded values

chrischdi:
Manual cherry-pick of #2989

Also syncs e2e.sh as good as possible to what we have on main.

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -43,18 +43,30 @@ on_exit() {
   # kill the VPN
   docker kill vpn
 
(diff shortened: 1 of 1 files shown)


chrischdi:
/cherry-pick release-1.7

cherry-picks did work out locally, let's see if this works remote here too.

sbueringer:
Interesting diff (seems like 2-3 different PRs? ;))

/lgtm
/approve

PR #3009, opened 2024-05-16
🌱 supervisor: drop ProviderServiceAccount ConfigMap

### LINKED ISSUE #2861: Rethink how the ServiceAccount controller manages the service account ConfigMap
/kind cleanup

The ProviderServiceAccount manages a list of ProviderServiceAccounts names into a ConfigMap

As of today the ServiceAccount controller actually reconciles VSphereClusters, thus it does not act for a specific ServiceAccount but it always acts for the ServiceAccount referencing the VSphereClusters being reconciled; this lead to an issue If you delete a ProviderServiceAccount while the cluster exists, it doesn’t get cleaned up on cluster deletion (deletion only cleans up ProviderServiceAccount existing at the time of delete)

Also:
- The entry in the config map does not include the cluster name in its name --> there is only one entry for each service account, no matter if used by one or more clusters.
- Consumers of CAPV are in the supervisor

Might be, that given that consumers of CAPV are outside of CAPV itself, we can drop this code from CAPV entirely (ultimately, the ConfigMap is just a sort of replacement for "List ProviderServiceAccounts")

chrischdi:

Drops the functionality of recording created ProviderServiceAccounts in a ConfigMap because this is not used anymore on dependent controllers.

Fixes #2861


sbueringer:
Apidiff shouldn't need an override

controllers/serviceaccount_controller.go
```diff
@@ -184,7 +182,8 @@ func (r *ServiceAccountReconciler) Reconcile(ctx context.Context, req reconcile.
 	}()
 
 	if !vsphereCluster.DeletionTimestamp.IsZero() {
-		return ctrl.Result{}, r.reconcileDelete(ctx, clusterContext)
+		controllerutil.RemoveFinalizer(clusterContext.VSphereCluster, vmwarev1.ProviderServiceAccountFinalizer)
```
sbueringer, line 185: I assume it's okay that someone else is responsible to cleanup existing ConfigMaps? (I think it's fine, just want to double check)

sbueringer: Looks like we never created it, so I think it should be okay that "whoever creates it" should clean it up once we don't use it anymore

sbueringer (commented):
All good, just one question

chrischdi:
> Apidiff shouldn't need an override

I know, just looks cleaner in notifications :-)

controllers/serviceaccount_controller.go
chrischdi: What we could do though is deleting it if the env vars are set and it still exists.

sbueringer: I would probably prefer if we just tell people to delete it. The way some people upgrade (jumping over various versions) we have to keep this forever otherwise

fabriziopandini, line 185: q: should we 
1. deprecate ProviderServiceAccountFinalizer (and remove it in future versions)
2. stop adding the finalizer and instead remove it as a cleanup from older versions?

sbueringer: Not sure if it's worth it. We have to keep it around for a few years in any case, and maybe we'll have some cleanup logic again in the future :)

fabriziopandini: No strong opinions, but It would be nice to avoid having a finalizer added to an object if we don't need it.

In case we keep it around, let's just add a comment on the Finalizer/when it is used about the fact that we are keeping this around for compatibility with older reasons

sbueringer: Also no strong opinions. I'm fine with deprecating. We just really have to be aware that we have to keep it for a while.

If we stop removing the finalizer too early we have a problem.

fabriziopandini: ack, lets just deprecate and add the add a comment on the Finalizer/when it is used about the fact that we are keeping this around for compatibility with older versions

chrischdi: Deprecated and changed the code to always remove the finalizer. I think when deprecating, we should already start removing it. We won't update it anyway.

chrischdi:
rebased + squashed

PR #3010, opened 2024-05-16
🌱 test: fix collector for machines not having an IP in status

chrischdi:

In case of test failures where machines do not yet have an IP set in status, we are currently not able to collect information like cloud-init log.

This implements a fallback to get the IP address from vSphere instead to still be able to gather information via ssh.

Fixes #


test/framework/log/collector.go
```diff
@@ -62,7 +66,18 @@ func (collector MachineLogCollector) CollectMachineLog(_ context.Context, _ clie
 				return err
 			}
 			defer f.Close()
-			return executeRemoteCommand(f, hostIPAddr, command, args...)
+			var lastErr error
+			// Try with all available IPs unless it succeeded.
+			for _, machineIPAddress := range machineIPAddresses {
+				lastErr = executeRemoteCommand(f, machineIPAddress, command, args...)
+				if lastErr == nil {
+					break
```
sbueringer, line 74:
```suggestion
					return nil
```

test/framework/log/collector.go
```diff
@@ -62,7 +66,18 @@ func (collector MachineLogCollector) CollectMachineLog(_ context.Context, _ clie
 				return err
 			}
 			defer f.Close()
-			return executeRemoteCommand(f, hostIPAddr, command, args...)
+			var lastErr error
+			// Try with all available IPs unless it succeeded.
+			for _, machineIPAddress := range machineIPAddresses {
+				lastErr = executeRemoteCommand(f, machineIPAddress, command, args...)
```
sbueringer, line 72: Let's collect/append the errors and then return the aggregation if len > 0 in l.78

sbueringer: (would be good if we include the actual error messages in the returned error)

test/framework/log/collector.go
```diff
@@ -78,10 +93,53 @@ func (collector MachineLogCollector) CollectMachineLog(_ context.Context, _ clie
...
+
+	vmObj, err := c.Finder.VirtualMachine(ctx, m.GetName())
+	if err != nil {
+		return nil, err
+	}
+
+	var vm mo.VirtualMachine
+
+	if err := c.Client.RetrieveOne(ctx, vmObj.Reference(), []string{"guest.net"}, &vm); err != nil {
+		// We cannot get the properties e.g. when the machine already got deleted or is getting deleted.
```
sbueringer, line 116:
```suggestion
		// We cannot get the properties e.g. when the vm already got deleted or is getting deleted.
```
I think?

test/framework/log/collector.go
```diff
@@ -78,10 +93,53 @@ func (collector MachineLogCollector) CollectMachineLog(_ context.Context, _ clie
...
+		return nil, errors.Errorf("Error retrieving properties for machine %s", klog.KObj(m))
```
sbueringer, line 117:
```suggestion
		return nil, errors.Errorf("error retrieving properties for machine %s", klog.KObj(m))
```
nit: Same below

sbueringer (commented):
Nice! just a few nits

test/framework/log/collector.go
chrischdi: adjusted, PTAL

sbueringer: Thx!

test/e2e/data/infrastructure-vsphere-govmomi/main/ignition/kustomization.yaml
```diff
@@ -14,3 +14,10 @@ patches:
   - target:
       kind: VSphereMachineTemplate
     path: ../commons/remove-storage-policy.yaml
+  # Replace ssh user to match expected user by the e2e machine collector
```
fabriziopandini, line 17: q: is it possible to change it cluster-template-ignition.yaml instead?

test/framework/log/collector.go
```diff
@@ -62,7 +67,20 @@ func (collector MachineLogCollector) CollectMachineLog(_ context.Context, _ clie
 				return err
 			}
 			defer f.Close()
-			return executeRemoteCommand(f, hostIPAddr, command, args...)
+			var errs []error
+			// Try with all available IPs unless it succeeded.
+			for _, machineIPAddress := range machineIPAddresses {
+				if err := executeRemoteCommand(f, machineIPAddress, command, args...); err != nil {
```
fabriziopandini, line 73: Q: does this imply we do expect ssh connectivity from the prow container to the machine? is this a new constraint introduced by this PR or it already exists for something else (thinking about constraints for the new test environment)

test/e2e/data/infrastructure-vsphere-govmomi/main/ignition/kustomization.yaml
sbueringer: I think we should keep the core user there. It's sort of standard with Ignition I think?

test/framework/log/collector.go
sbueringer: We already did this for a while to get logs

test/e2e/data/infrastructure-vsphere-govmomi/main/ignition/kustomization.yaml
chrischdi: Agree, `core` is the standard in ignition.

test/framework/log/collector.go
chrischdi: This PR just fixes the usage where we try to pull data via SSH to make it work in more failure cases. Before we tried but often did not get any data because there was no IP address yet set on the Machine's status object.

sbueringer:
Merging, so we can rebase the bump PR. We can follow-up if necessary

/lgtm
/approve

Issue #2995, opened 2024-05-13
Improve e2e testing


sbueringer:
### Improve e2e tests

We compared CAPV govmomi & supervisor e2e tests with core CAPI e2e tests. 
As a result we want to make the following improvements.

Adding e2e tests:
* [x] P0: Add v1.30 => v1.31 upgrade e2e test + check ci-latest (supervisor+govmomi) (@sbueringer)
  * PR: https://github.com/kubernetes/test-infra/pull/32610
  * PR: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2999
  * Plan is to add more and more upgrades jobs for new Kubernetes versions until we have the same test coverage as core CAPI.
* [x] P0: Add test "When testing ClusterClass changes [ClusterClass]" (supervisor) (@zhanggbj)
  * PR: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3011
* [x] P1: Add test "Ensure OwnerReferences and Finalizers are resilient" (supervisor) (@fabriziopandini)
  * [x] CAPI PR https://github.com/kubernetes-sigs/cluster-api/pull/10693
  * [x] CAPV PR https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3059
* [x] P1: Also use `ValidateResourceVersionStable` func (https://github.com/kubernetes-sigs/cluster-api/pull/10530) to ownerRef & finalizer tests (supervisor+govmomi) (@fabriziopandini) https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3033
* [x] P1: Let's check if we have unit test coverage for the following tests for supervisor @chrischdi 
  * capv-e2e.[It] Hardware version upgrade creates a cluster with VM hardware versions upgraded
    * already covered via https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/b935f0aca9e1fa6c8e3d8749f0bfd9a307f9cad7/pkg/services/vmoperator/vmopmachine_test.go#L152
  * capv-e2e.[It] Label nodes with ESXi host info creates a workload cluster whose nodes have the ESXi host info
    * PR: #3085
* [x] P1: Add clusterctl upgrade tests n-2, n-1 (supervisor) (@zhanggbj)
    * CAPV 1.9=>current, CAPI 1.6=>1.7, CAPV 1.10=>current, CAPI 1.7=>1.7 (same we already have on main for govmomi)
[...]

PR #3011, opened 2024-05-20
🌱 Fill up supervisor e2e test - ClusterClass changes

zhanggbj:

- Enable test "When testing ClusterClass changes"

Part of #2995

### DIFF
--- a/test/e2e/clusterclass_changes_test.go
+++ b/test/e2e/clusterclass_changes_test.go
@@ -21,11 +21,11 @@ import (
 	capie2e "sigs.k8s.io/cluster-api/test/e2e"
 )
 
-var _ = Describe("When testing ClusterClass changes [ClusterClass]", func() {
+var _ = Describe("When testing ClusterClass changes [supervisor] [ClusterClass]", func() {
 	const specName = "clusterclass-changes" // copied from CAPI
 	Setup(specName, func(testSpecificSettingsGetter func() testSettings) {
 		capie2e.ClusterClassChangesSpec(ctx, func() capie2e.ClusterClassChangesSpecInput {
-			return capie2e.ClusterClassChangesSpecInput{
+			clusterClassChangesSpecInput := capie2e.ClusterClassChangesSpecInput{
 				E2EConfig:             e2eConfig,
 				ClusterctlConfigPath:  testSpecificSettingsGetter().ClusterctlConfigPath,
 				BootstrapClusterProxy: bootstrapClusterProxy,
@@ -34,15 +34,25 @@ var _ = Describe("When testing ClusterClass changes [ClusterClass]", func() {
 				Flavor:                testSpecificSettingsGetter().FlavorForMode("topology"),
 				PostNamespaceCreated:  testSpecificSettingsGetter().PostNamespaceCreatedFunc,
 				ModifyControlPlaneFields: map[string]interface{}{
-					"spec.machineTemplate.nodeDrainTimeout": "10s",
+					"spec.kubeadmConfigSpec.verbosity": int64(4),
 				},
 				ModifyMachineDeploymentBootstrapConfigTemplateFields: map[string]interface{}{
 					"spec.template.spec.verbosity": int64(4),
 				},
-				ModifyMachineDeploymentInfrastructureMachineTemplateFields: map[string]interface{}{
-					"spec.template.spec.numCPUs": int64(4),
-				},
 			}
+
+			if testMode == GovmomiTestMode {
+				clusterClassChangesSpecInput.ModifyMachineDeploymentInfrastructureMachineTemplateFields = map[string]interface{}{
+					"spec.template.spec.numCPUs": int64(4),
+				}
+			}
+
+			if testMode == SupervisorTestMode {
+				clusterClassChangesSpecInput.ModifyMachineDeploymentInfrastructureMachineTemplateFields = map[string]interface{}{
+					"spec.template.spec.powerOffMode": "trySoft",
+				}
+			}
+			return clusterClassChangesSpecInput
 		})
 	})
 })


test/e2e/clusterclass_changes_test.go
```diff
@@ -46,3 +46,29 @@ var _ = Describe("When testing ClusterClass changes [ClusterClass]", func() {
 		})
 	})
 })
+
+var _ = Describe("When testing ClusterClass changes [vcsim] [supervisor] [PR-Blocking] [ClusterClass]", func() {
```
fabriziopandini, line 50: q: is it intentional to run this test on vcsim? (the corresponding govmomi test runs on real infrastructure)
q: is it intentional to run it as a PR blocking? (the corresponding govmomi test is not blocking)

test/e2e/clusterclass_changes_test.go
```diff
@@ -46,3 +46,29 @@ var _ = Describe("When testing ClusterClass changes [ClusterClass]", func() {
...
+			return capie2e.ClusterClassChangesSpecInput{
+				E2EConfig:             e2eConfig,
+				ClusterctlConfigPath:  testSpecificSettingsGetter().ClusterctlConfigPath,
+				BootstrapClusterProxy: bootstrapClusterProxy,
+				ArtifactFolder:        artifactFolder,
+				SkipCleanup:           skipCleanup,
+				Flavor:                testSpecificSettingsGetter().FlavorForMode("topology"),
+				PostNamespaceCreated:  testSpecificSettingsGetter().PostNamespaceCreatedFunc,
+				ModifyControlPlaneFields: map[string]interface{}{
+					"spec.machineTemplate.nodeDrainTimeout": "10s",
```
fabriziopandini, line 63: q: This field is now propagated in place, should we pick another field for the sake of this test? or should we pick two fields, one propagated in place, the other not... (cc @sbueringer @chrischdi)

test/e2e/clusterclass_changes_test.go
```diff
@@ -46,3 +46,29 @@ var _ = Describe("When testing ClusterClass changes [ClusterClass]", func() {
...
+				},
+				ModifyMachineDeploymentBootstrapConfigTemplateFields: map[string]interface{}{
+					"spec.template.spec.verbosity": int64(4),
```
fabriziopandini, line 66: q: This field now triggers a rotation which seems un-necessary, should we add this to the list of fields propagated in place?  (cc @sbueringer @chrischdi)
Might be there are other fields too...

fabriziopandini (commented):
Nice!
Just a few questions, feel free to ignore them if some of the are due to the PR being still in WIP state

test/e2e/clusterclass_changes_test.go
zhanggbj: And what's our criteria to label a test case as PR blocking? In this case, I'll fix it and align with the govmomi test as not blocking

test/e2e/clusterclass_changes_test.go
zhanggbj: Thanks for catching this! So I think we also need to update this field for the govmomi test, how about we change it as following?
- Pick another field for govmomi test
- Pick two fields for supervisor test, add another new field which will lead to rollout

test/e2e/clusterclass_changes_test.go
fabriziopandini: > How to decide whether a case should be labeled to run with a) or b)

It is our decision if to run a test using vcsim or real infrastructure, with the caveat that the option to run tests on vcsim was recently introduced.
Generally speaking I would prefer to run tests on real infrastructure, and use vcsim only if we have constraints about time or resources (which is not the current case I think).

So I will use a) for this case.

> And what's our criteria to label a test case as PR blocking? 

We usually consider PR blocking the minimal set of tests that we run on every PR.
Considering that this test is more focused on validating if a regression happens in CAPI than validating CAPV changes, I'm +1 to not have this as a PR blocking test.

Also, please note that as a maintainer we can dynamically expand the set or E2E to pass on each PR with the /test prow command, so it is ok to have very few E2E tests as PR blocking .

test/e2e/clusterclass_changes_test.go
fabriziopandini: Let's hear  @sbueringer @chrischdi feedback first, we can also eventually defer this change to a follow up PR if we want to take some time to think about it

test/e2e/clusterclass_changes_test.go
chrischdi: So this is kubeadms `-v` flag at the end? Makes sense to have this in-place 👍

chrischdi: Also then: so we should propably pick a different field for this test?!

test/e2e/clusterclass_changes_test.go
chrischdi: Does it even make sense to set this field in the test when its propagated in place? I think we should pick a.diferent one only so I'm personally 👍 for

> * Pick another field for govmomi test

test/e2e/clusterclass_changes_test.go
sbueringer: Agree, we don't have to run this test with vcsim and we shouldn't make it PR blocking.

This then leaves: `When testing ClusterClass changes [supervisor] [ClusterClass]`

But if we add an additional test like this, this test will run twice for govmomi.

This is because the govmomi test runs basically without any focus/skip: https://prow.k8s.io/prowjob?prowjob=0ed84718-7065-4355-9c4a-2242f7616e04
And supervisor runs with focus on [supervisor]: https://prow.k8s.io/prowjob?prowjob=207d878f-5aa4-47a2-b29b-1e2a81808393

So what we have to do instead is to add [supervisor] to the existing test

But now you'll need a bit of branching for ModifyMachineDeploymentInfrastructureMachineTemplateFields, this can be easily done with `testMode == GovmomiTestMode`

test/e2e/clusterclass_changes_test.go
sbueringer: It doesn't make much difference. This test doesn't wait until these properties are propagated down.

But I would pick a different field:
* changing something that leads to a Machine rollout increases test coverage basically by providing more chances for potential failure (and thus, us finding bugs)
* this field specifically is already covered by the ClusterClass rollout test anyway

I don't know why this field was chosen for CAPV, let's use the same as in core CAPI: 
```go
			ModifyControlPlaneFields: map[string]interface{}{
				"spec.kubeadmConfigSpec.verbosity": int64(4),
			},
```

test/e2e/clusterclass_changes_test.go
sbueringer: Based on comments above, I would keep this field to intentianlly trigger a Machine rollout

sbueringer: Should we really start patching KubeadmConfigs in-place for existing Machines in the Machine controller?

I get why the verbosity doesn't matter, but technically all fields of KubeadmConfig don't really matter, because kubeadm is only run once.

fabriziopandini: taking my words back.
to change kubeadm verbosity we actually need a rollout

test/e2e/clusterclass_changes_test.go
zhanggbj: Thanks for the clarification! Now it's more clear to me.
And this is even better to leverage the existing test case instead of adding duplicate code :-)

test/e2e/clusterclass_changes_test.go
zhanggbj: Fixed it.

test/e2e/clusterclass_changes_test.go
```diff
@@ -43,6 +43,15 @@ var _ = Describe("When testing ClusterClass changes [ClusterClass]", func() {
 					"spec.template.spec.numCPUs": int64(4),
```
sbueringer, line 43: Let's drop this here. It's enough if we set it below

test/e2e/clusterclass_changes_test.go
```diff
@@ -43,6 +43,15 @@ var _ = Describe("When testing ClusterClass changes [ClusterClass]", func() {
...
 				},
 			}
+			if testMode == GovmomiTestMode {
+				clusterClassChangesSpecInput.ModifyMachineDeploymentInfrastructureMachineTemplateFields = map[string]interface{}{
+					"spec.template.spec.numCPUs": int64(4),
+				}
+			}
+			clusterClassChangesSpecInput.ModifyMachineDeploymentInfrastructureMachineTemplateFields = map[string]interface{}{
+				"spec.template.spec.powerOffMode": "hard",
```
sbueringer, line 52:
```suggestion
				"spec.template.spec.powerOffMode": "trySoft",
```

hard is default, so setting it to hard is a no-op

sbueringer: Let's please add an if around this statement (like above just with SupervisorTestMode)

Otherwise ModifyMachineDeploymentInfrastructureMachineTemplateFields will always set powerOffMode

test/e2e/clusterclass_changes_test.go
zhanggbj: Forgot to drop it😂 Thanks!

test/e2e/clusterclass_changes_test.go
zhanggbj: Fixed it!

sbueringer: Thx!

fabriziopandini (commented):
Nice work!
/lgtm
/approve

PR #3023, opened 2024-05-24
🌱 Fill up supervisor e2e test - ClusterClass rollouts

zhanggbj:

- Enable test "When testing ClusterClass rollouts"

Part of #2995


zhanggbj:
Tests failed as CAPV will add an extra label `node.cluster.x-k8s.io/esxi-host` on Machine by [here](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/controllers/vspheremachine_controller.go#L407), so labels won't be exactly equal to the Machine labels listed in the CAPI E2E tests [here](https://github.com/kubernetes-sigs/cluster-api/blob/main/test/e2e/clusterclass_rollout.go#L436).

I'm thinking we may need to update the CAPI E2E tests, to make it check if the machine contains the 4 required labels instead of using `BeEquivalentTo`, does it make sense to you @fabriziopandini @sbueringer @chrischdi ? Thanks!


- Test Failure
```
      <map[string]string | len:5>: {
          "topology.cluster.x-k8s.io/owned": "",
          "cluster.x-k8s.io/cluster-name": "clusterclass-rollout-gnfvcg",
          "cluster.x-k8s.io/control-plane": "",
          "cluster.x-k8s.io/control-plane-name": "clusterclass-rollout-gnfvcg-cp-vngr8",
          "node.cluster.x-k8s.io/esxi-host": "10.2.32.5",
      }
  to be equivalent to
      <e2e.unionMap | len:4>: {
          "topology.cluster.x-k8s.io/owned": "",
          "cluster.x-k8s.io/cluster-name": "clusterclass-rollout-gnfvcg",
          "cluster.x-k8s.io/control-plane": "",
          "cluster.x-k8s.io/control-plane-name": "clusterclass-rollout-gnfvcg-cp-vngr8",
      }
  In [It
```

sbueringer:
I think we should look into how we can extend the core CAPI test spec to allow for validation of additional labels. (basically a CAPV func would validate the esxi-host label and then core CAPI all the other ones)

zhanggbj:
As we need to extend the CAPI E2E , I just opened a CAPI issue for further discussion, and I think it's the same situation for annotations check. https://github.com/kubernetes-sigs/cluster-api/pull/10715

zhanggbj:
For the above failures, I think it should be fixed in capie2e side when checking object annotations, please find more details here --> https://github.com/kubernetes-sigs/cluster-api/issues/10839

sbueringer:
Sounds good. Feel free to bump to CAPI main :D

zhanggbj:
Bumped CAPI version.

test/go.mod
```diff
@@ -2,9 +2,9 @@ module sigs.k8s.io/cluster-api-provider-vsphere/test
 
 go 1.22.0
```
sbueringer, line 3: We should also update the top-level go.mod

sbueringer:
Otherwise lgtm

test/go.mod
zhanggbj: Missed, let me fix it.

sbueringer: Ah nice, but test is green 🎉

sbueringer:
Thank you very much!!

/lgtm
/approve

PR #3025, opened 2024-05-24
✨ vGPU implementation

birksl:
Builds on the changes in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/2272

Support adding vGPUs to VMs

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1972


Makefile
```diff
@@ -381,6 +381,8 @@ generate-e2e-templates-main: $(KUSTOMIZE) ## Generate test templates for the mai
 	"$(KUSTOMIZE)" --load-restrictor LoadRestrictionsNone build "$(E2E_GOVMOMI_TEMPLATE_DIR)/main/install-on-bootstrap" > "$(E2E_GOVMOMI_TEMPLATE_DIR)/main/cluster-template-install-on-bootstrap.yaml"
 	# for PCI passthrough template
 	"$(KUSTOMIZE)" --load-restrictor LoadRestrictionsNone build "$(E2E_GOVMOMI_TEMPLATE_DIR)/main/pci" > "$(E2E_GOVMOMI_TEMPLATE_DIR)/main/cluster-template-pci.yaml"
+	# for vGPU template
```
chrischdi, line 384: no need to add this if we don't have any e2e tests for it.

docs/gpu-vgpu.md
```diff
@@ -0,0 +1,106 @@
...
+
+See "2. Choosing the vGPU Profile for the Virtual Machine" at [Using GPUs with Virtual Machines on vSphere](https://blogs.vmware.com/apps/2018/09/using-gpus-with-virtual-machines-on-vsphere-part-3-installing-the-nvidia-grid-technology.html) to see what vGPU profiles are available for your GPU device.
+
+We are using NVIDIA Tesla V100 32GB cards for this example and will use the `grid_v100d-4c` vGPU profile for this card that allocates 4GB GPU memory to the worker node's vGPU device. 
+
+### Create the cluster template
+
+```shell
+$ make dev-flavors
+go run ./packaging/flavorgen --output-dir /Users/pkatyal/.cluster-api/overrides/infrastructure-vsphere/v0.0.0
```
chrischdi, line 59:
```suggestion
go run ./packaging/flavorgen --output-dir $HOME/.cluster-api/overrides/infrastructure-vsphere/v0.0.0
```

apis/v1beta1/types.go
```diff
@@ -265,6 +268,15 @@ type PCIDeviceSpec struct {
 	CustomLabel string `json:"customLabel,omitempty"`
 }
 
+// VGPUSpec defines virtual machine's VGPU configuration
+type VGPUSpec struct {
+	// ProfileName is the ProfileName of a virtual machine's vGPU, in string.
+	// Defaults to the eponymous property value in the template from which the
+	// virtual machine is cloned.
+	// +kubebuilder:validation:Required
+	ProfileName string `json:"profileName,omitempty"`
```
chrischdi, line 277: we handle this in `reconcilePCIDevices`. Should this instead of being a separate field next to PciDevices be integrated into PciDevices instead?

test/e2e/config/vsphere-dev.yaml
```diff
@@ -0,0 +1,164 @@
+---
```
chrischdi, line 1: This should not get added.

test/e2e/data/infrastructure-vsphere-govmomi/main/vgpu/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
chrischdi, line 1: no need for this as long as there are no e2e tests, and we don't have gpu's in our e2e test-environment.

chrischdi (commented):
One major question: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3025/files#r1617231776

apis/v1beta1/types.go
birksl: That makes sense. I'll try to integrate the vgpu profile name into PCIDeviceSpec.

birksl: Worked well in manual test with only vgpu profilename and omitting deviceID and vendorID.

test/e2e/config/vsphere-dev.yaml
```diff
@@ -0,0 +1,163 @@
+# E2E test scenario using local dev images and manifests built from the source tree for following providers:
```
chrischdi, line 1: please drop this file

test/e2e/config/vsphere.yaml
```diff
@@ -279,6 +279,7 @@ variables:
   # These IDs correspond to Tesla T4s, they are the decimal representation of the hex values.
   DEVICE_ID: 7864
   VENDOR_ID: 4318
+  PROFILE_NAME: grid_v100d-4c
```
chrischdi, line 282: please drop this change as there is no e2e test using it.

internal/webhooks/vspheremachinetemplate.go
```diff
@@ -84,6 +84,13 @@ func (webhook *VSphereMachineTemplateWebhook) ValidateCreate(_ context.Context,
 			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "guestSoftPowerOffTimeout"), spec.GuestSoftPowerOffTimeout, "should be greater than 0"))
 		}
 	}
+	for _, device := range spec.PciDevices {
+		hasVGPU := device.VGPUProfile != ""
+		hasPCI := device.DeviceID != nil && device.VendorID != nil
+		if (hasPCI && hasVGPU) || (!hasPCI && !hasVGPU) {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "pciDevices"), spec.PciDevices, "should have either deviceID + vendorID or vgpuProfile"))
```
chrischdi, line 91:
```suggestion
	for i, device := range spec.PciDevices {
		hasVGPU := device.VGPUProfile != ""
		hasPCI := device.DeviceID != nil && device.VendorID != nil
		if (hasPCI && hasVGPU) || (!hasPCI && !hasVGPU) {
			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "pciDevices", fmt.Sprintf("%d, i)), device, "should have either deviceID + vendorID or vgpuProfile"))
```

pkg/services/govmomi/pci/device.go
```diff
@@ -76,18 +76,28 @@ func ConstructDeviceSpecs(pciDeviceSpecs []infrav1.PCIDeviceSpec) []types.BaseVi
...
 }
 
-func createBackingInfo(spec infrav1.PCIDeviceSpec) *types.VirtualPCIPassthroughDynamicBackingInfo {
-	return &types.VirtualPCIPassthroughDynamicBackingInfo{
-		AllowedDevice: []types.VirtualPCIPassthroughAllowedDevice{
-			{
-				VendorId: *spec.VendorID,
-				DeviceId: *spec.DeviceID,
+func createBackingInfo(spec infrav1.PCIDeviceSpec) types.BaseVirtualDeviceBackingInfo {
+	if spec.DeviceID != nil && spec.VendorID != nil {
```
chrischdi, line 80: Let's maybe turn it around to have less checks (just nitpicking / optics) .
```go
	if spec.VGPUProfile != "" {
		return &types.VirtualPCIPassthroughVmiopBackingInfo{
			Vgpu: spec.VGPUProfile,
		}
	}
	
	... same as before
```

pkg/services/govmomi/pci/device.go
```diff
@@ -76,18 +76,28 @@ func ConstructDeviceSpecs(pciDeviceSpecs []infrav1.PCIDeviceSpec) []types.BaseVi
...
+	}
+
+	return &types.VirtualPCIPassthroughVmiopBackingInfo{
+		Vgpu: spec.VGPUProfile,
 	}
 }
 
 func constructKey(pciDeviceSpec infrav1.PCIDeviceSpec) string {
-	return fmt.Sprintf("%d-%d", *pciDeviceSpec.DeviceID, *pciDeviceSpec.VendorID)
+	if pciDeviceSpec.DeviceID != nil && pciDeviceSpec.VendorID != nil {
```
chrischdi, line 98: Same here :-)

internal/webhooks/vspheremachinetemplate.go
birksl: Rewrite it to something like this? Which has better optics in your opinion?
```go
if device.VGPUProfile == "" {
  if device.DeviceID == nil || device.VendorID == nil {
    allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "pciDevices", fmt.Sprintf("%d", i)), device, "should have both deviceId and vendorId set"))
  
  }
} else {
  if device.DeviceID != nil || device.VendorID != nil {
    allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "pciDevices", fmt.Sprintf("%d", i)), device, "should have either deviceId + vendorId or vgpuProfile"))
  
  }
}
```

```go
if device.VGPUProfile == "" && (device.DeviceID == nil || device.VendorID == nil) {
  allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "pciDevices", fmt.Sprintf("%d", i)), device, "should have both deviceId and vendorId set"))
} else if device.VGPUProfile != "" && (device.DeviceID != nil || device.VendorID != nil) {
  allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "pciDevices", fmt.Sprintf("%d", i)), device, "should have either deviceId + vendorId or vgpuProfile"))
}
```

pkg/services/govmomi/pci/device.go
birksl: I think checking the pointers is technically more safe because an empty VGPUProfile string makes it so the VM cannot find a host, avoiding nil-pointer panic, but is essentially setting a default which is not ideal.
The validation hook makes this point obsolete though, so is that enough to not bother checking them?

internal/webhooks/vspheremachine_test.go
```diff
@@ -48,52 +48,86 @@ func TestVSphereMachine_ValidateCreate(t *testing.T) {
...
+		},
+
+		{
+			name:           "empty pciDevice",
+			vsphereMachine: createVSphereMachine("foo.com", nil, "", []string{"192.168.0.1/32", "192.168.0.3/32"}, infrav1.VirtualMachinePowerOpModeTrySoft, nil, []infrav1.PCIDeviceSpec{{VGPUProfile: ""}}),
+			wantErr:        true,
+		},
+		{
+			name:           "incorrect pciDevice",
+			vsphereMachine: createVSphereMachine("foo.com", nil, "", []string{"192.168.0.1/32", "192.168.0.3/32"}, infrav1.VirtualMachinePowerOpModeTrySoft, nil, []infrav1.PCIDeviceSpec{{VGPUProfile: "vgpu", DeviceID: new(int32)}}),
```
chrischdi, line 87: Let's set an actual value instead of an empty int32 pointer.
```suggestion
			vsphereMachine: createVSphereMachine("foo.com", nil, "", []string{"192.168.0.1/32", "192.168.0.3/32"}, infrav1.VirtualMachinePowerOpModeTrySoft, nil, []infrav1.PCIDeviceSpec{{VGPUProfile: "vgpu", DeviceID: ptr.To[int32](1)}}),
```

Also for the other occurencies.

internal/webhooks/vspheremachine.go
```diff
@@ -92,6 +92,17 @@ func (webhook *VSphereMachineWebhook) ValidateCreate(_ context.Context, raw runt
...
+	for i, device := range spec.PciDevices {
+		if device.VGPUProfile == "" {
+			if device.DeviceID == nil || device.VendorID == nil {
+				allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "pciDevices", fmt.Sprintf("%d", i)), device, "should have both deviceId and vendorId set"))
+			}
+		} else {
+			if device.DeviceID != nil || device.VendorID != nil {
+				allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "pciDevices", fmt.Sprintf("%d", i)), device, "should have either deviceId + vendorId or vgpuProfile"))
+			}
+		}
```
chrischdi, line 104: The first case could be that all is set to nothing, so both should have the same message.

Let's change it maybe the following way:

```suggestion
		if device.VGPUProfile != "" && device.DeviceID == nil && device.VendorID == nil {
			// Valid case for vGPU.
			continue
		}
		if device.VGPUProfile == "" && device.DeviceID != nil && device.VendorID != nil {
			// Valid case for PCI Passthrough.
			continue
		}
		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec", "pciDevices", fmt.Sprintf("%d", i)), device, "should have either deviceId + vendorId or vgpuProfile set"))
```

chrischdi: Also: let's make a shared `validatePCIDevices` function to use in both places (vspheremachine + vspheremachinetemplate).

docs/gpu-vgpu.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Create the cluster
+
+Set the size of the GPU nodes appropriately, since the Nvidia gpu-operator requires additional CPU and memory to install the device drivers on the VMs.
+
+Note: For GPU nodes (PCI Passthrough or vGPU), all memory of the nodes must be reserved. CAPV will automatically do this for nodes that have a PCI Passthrough GPU or a vGPU device in the spec. See "Memory Reservation" at [Using GPUs with Virtual Machines on vSphere](https://blogs.vmware.com/apps/2018/09/using-gpus-with-virtual-machines-on-vsphere-part-2-vmdirectpath-i-o.html)
+
+Apply the manifest from the previous step to your management cluster to have CAPV create a workload cluster with worker nodes that have vGPUs.
+
+From this point on, the setup is exactly the same as [GPU enabled clusters via PCI Passthrough](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/docs/gpu-pci.md#create-the-cluster).
```
chrischdi, line 106:
```suggestion
From this point on, the setup is exactly the same as [GPU enabled clusters via PCI Passthrough](./gpu-pci.md#create-the-cluster). 
```

chrischdi (commented):
Sorry for the long round-trip between reviews.

Another round of nits then I think this is good!

cc @neolit123 maybe this is interesting for you.

birksl:
Thanks for the guidance. I learnt a lot :)

internal/webhooks/vspheremachine_test.go
```diff
@@ -190,6 +231,7 @@ func createVSphereMachine(server string, providerID *string, preferredAPIServerC
 					PreferredAPIServerCIDR: preferredAPIServerCIDR,
 					Devices:                []infrav1.NetworkDeviceSpec{},
 				},
+				PciDevices: pciDevices,
```
neolit123, line 234: note, the correct PascalCase for this is PCIDevices, but that seems to already in the codebase.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -69,7 +69,7 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
 		}
 	}
 	if vmCtx.VSphereVM.Spec.CustomVMXKeys != nil {
-		log.Info("Applied custom vmx keys o VM clone spec")
+		log.Info("Applied custom vmx keys to VM clone spec")
```
neolit123, line 72:
```suggestion
		log.Info("Applied custom VMX keys to VM clone spec")
```

pkg/services/govmomi/pci/device.go
```diff
@@ -76,7 +76,12 @@ func ConstructDeviceSpecs(pciDeviceSpecs []infrav1.PCIDeviceSpec) []types.BaseVi
 	return pciDevices
 }
 
-func createBackingInfo(spec infrav1.PCIDeviceSpec) *types.VirtualPCIPassthroughDynamicBackingInfo {
+func createBackingInfo(spec infrav1.PCIDeviceSpec) types.BaseVirtualDeviceBackingInfo {
+	if spec.VGPUProfile != "" {
+		return &types.VirtualPCIPassthroughVmiopBackingInfo{
+			Vgpu: spec.VGPUProfile,
```
neolit123, line 82:
```suggestion
			VGPU: spec.VGPUProfile,
```

birksl: `Vgpu` is the name of the key in the `VirtualPCIPassthroughVmiopBackingInfo` struct directly from govmomi in this case.

apis/v1beta1/types.go
```diff
@@ -251,13 +251,21 @@ type PCIDeviceSpec struct {
...
 	// VendorId is the vendor ID of a virtual machine's PCI, in integer.
 	// Defaults to the eponymous property value in the template from which the
 	// virtual machine is cloned.
+	// Mutually exclusive with VGPUProfile.
 	// +kubebuilder:validation:Required
 	VendorID *int32 `json:"vendorId,omitempty"`
+	// VGPUProfile is the profile name of a virtual machine's vGPU, in string.
+	// Defaults to the eponymous property value in the template from which the
+	// virtual machine is cloned.
+	// Mutually exclusive with DeviceID and VendorID.
```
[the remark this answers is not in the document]
birksl: Something like this?
```
// Mutually exclusive with VGPUProfile as VGPUProfile and DeviceID + VendorID
// are two independent ways to define PCI devices.
```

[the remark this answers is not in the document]
neolit123: SGTM

neolit123:
don't forget to squash all commits before the final merge.

lubronzhan (approved):
lgtm

sbueringer:
> don't forget to squash all commits before the final merge.

Not needed. We can just use tide to squash (we improved our release notes tool a while ago, so it can handle squash commits now)

neolit123:
> > don't forget to squash all commits before the final merge.
> 
> Not needed. We can just use tide to squash (we improved our release notes tool a while ago, so it can handle squash commits now)

probably the blow scenario will never happen, but note that squash nukes gpg signatures, so some contributor might decide to create a problem.

chrischdi:
/lgtm

Thanks @neolit123 for also taking a look!

sbueringer:
@birksl Thank you very much!

Sorry for the delay. I'll create a PR to fix my findings. I don't want to delay merging this PR further.

/lgtm
/approve

PR #3052, opened 2024-06-11
🌱 Makefile: drop OVERRIDES_DIR and dependent make targets

chrischdi:

Some cleanup to our Makefile to drop:

* Variable `VERSION` (only used for `OVERRIDES_DIR`)
* Variable `OVERRIDES_DIR`
* Make targets which did use `OVERRIDES_DIR`:
  * `dev-manifests`: generated manifests and copied them to `$HOME/.cluster-api/overrides/infrastructure-vsphere/...`
  * `dev-flavors`: generated flavors and copied them to `$HOME/.cluster-api/overrides/infrastructure-vsphere/...`

Folks should use tilt instead which also allows to add the templates directory to tilt.

Fixes #


sbueringer:
/lgtm

/assign @zhanggbj 
(not sure if you're using that)

Issue #3040, opened 2024-06-07
Use vSphere projects from Boskos


sbueringer:
This is an umbrella issue for the migration to use vSphere projects only from Boskos.

For more context, see: https://github.com/kubernetes/k8s.io/issues/6877

### Prerequisites

* [x] Setup Boskos in VMC
  * Done, configuration used can be seen here: https://github.com/sbueringer/k8s.io/pull/1

### Migration

* [x] CAPV
  * [x] PR to update test-infra to set BOSKOS_HOST: https://github.com/kubernetes/test-infra/pull/32727
  * [x] PR for main: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3034
  * [x] cherry-pick to all branches with CI
    * [x] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3053
    * [x] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3054
    * [x] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3055
    * [x] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3056
* [x] CPI
  * [x] PR to update test-infra to set BOSKOS_HOST: https://github.com/kubernetes/test-infra/pull/32727
  * [x] PR for main: https://github.com/kubernetes/cloud-provider-vsphere/pull/1067
  * [x] cherry-pick to all branches with CI (delegated to CPI folks)
* [x] image-builder
  * [x] PR to update test-infra to set BOSKOS_HOST: https://github.com/kubernetes/test-infra/pull/32727
  * [x] PR for main: https://github.com/kubernetes-sigs/image-builder/pull/1487
  * [x] no release branches => no cherry-picks necessary

### Misc

* [x] Increase # of CAPV projects to 40 (verify state "initializing" works)

### Cleanup

* [x] Cleanup janitor: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3057
* [x] Remove IPAM controller and corresponding objects from CAPV services cluster in VMC
* [x] Remove old resource pools and folders in VMC (+ corresponding permissions) (also audit & cleanup the Google doc)
* [x] Cleanup secrets:
  * [x] Cleanup presets: https://github.com/kubernetes/test-infra/pull/32771
  * [x] Clean up ExternalSecrets: https://github.com/kubernetes/test-infra/pull/32688
[...]

PR #3057, opened 2024-06-11
🌱 Cleanup janitor

sbueringer:

Please wait with reviews, need a bit more time

Part of #3040


sbueringer:
/hold 
until everyone is moved over to Boskos (mostly CAPV cherry-picks + the CPI main PR)

hack/tools/pkg/janitor/janitor_test.go
```diff
@@ -348,36 +328,30 @@ func Test_janitor_CleanupVSphere(t *testing.T) {
...
 
-	deleteAll := time.Now().Add(time.Hour * 1)
-
 	tests := []struct {
 		name               string
 		dryRun             bool
-		maxCreationDate    time.Time
 		objects            []*vcsimObject
 		wantAfterFirstRun  map[string]bool
 		wantAfterSecondRun map[string]bool
```
chrischdi, line 336: Having these two does not make sense anymore and the tests should fail currently :think:

chrischdi: (only one of them should be needed)

sbueringer: Not sure. There is a case where it needs to runs to also cleanup folders/resourcePools. Do we have to change our implementation?

sbueringer: Turns out the tests are not run in CI. Maybe because govc is not setup there? (or because of hack/tools)

chrischdi (commented):
/approve

PR #3059, opened 2024-06-12
🌱 Test OwnerReferences and Finalizers for Supervisor

fabriziopandini:
Add tests OwnerReferences and Finalizers for Supervisor, and makes it possible to run those tests with vcsim as well

Part of https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/2995


fabriziopandini:
@sbueringer @chrischdi this PR is ready for review now.

Note: for this iteration, I'm keeping the owner ref test separated from the quick-start test, but in a follow up iteration we can probably merge the two tests like we have in CAPI

test/e2e/data/infrastructure-vsphere-supervisor/main/ownerrefs-finalizers/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../base
+patchesStrategicMerge:
+  - ../commons/cluster-resource-set-label.yaml
+  - ../commons/cluster-network-CIDR.yaml
+  - ../commons/cluster-resource-set-csi-insecure.yaml
```
chrischdi, line 8: base already adds this patches if I did see it right

we could maybe even just generate the tempate out of base without needing this kustomization?

```suggestion
```

test/e2e/ownerrefs_finalizers_test.go
```diff
@@ -354,3 +441,26 @@ func forcePeriodicReconcile(ctx context.Context, c ctrlclient.Client, namespace
...
+	return func(u unstructured.Unstructured) bool {
+		// When using vcsim CRS cannot be applied (not supported by the fake API server), so ignoring.
+		if testTarget == VCSimTestTarget {
+			if u.GetKind() == "ClusterResourceSet" {
+				return false
+			}
+		}
+
+		// Following objects are for vm-operator (not managed by CAPV), so checking finalizers/resourceVersion is not relevant.
+		// Note: we are excluding also VirtualMachines, which instead are considered for the owenerReference tests.
```
chrischdi, line 457:
```suggestion
		// Note: we are excluding also VirtualMachines, which instead are considered for the ownerReference tests.
```

chrischdi (commented):
Just nits

test/e2e/data/infrastructure-vsphere-supervisor/main/ownerrefs-finalizers/kustomization.yaml
sbueringer: Yeah we don't need these. Which also means we don't need a different template. We can just use the base / default template for the test instead

test/e2e/e2e_suite_test.go
```diff
@@ -210,8 +211,9 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 
 	if testTarget == VCSimTestTarget {
 		Byf("Creating a vcsim server")
-		err := vspherevcsim.Create(ctx, bootstrapClusterProxy.GetClient())
-		Expect(err).ToNot(HaveOccurred(), "Failed to create VCenterSimulator")
+		Eventually(func() error {
+			return vspherevcsim.Create(ctx, bootstrapClusterProxy.GetClient())
```
sbueringer, line 215: Is this really retryable? I would expect the Create call to fail on the second try (because we don't ignore AlreadyExists).

So this only helps if the first try fails at the Create call with network issues

test/e2e/ownerrefs_finalizers_test.go
```diff
@@ -104,39 +111,56 @@ var _ = Describe("Ensure OwnerReferences and Finalizers are resilient with Failu
...
+	csiConfigSecretName     = "vsphere-config-secret" //nolint: gosec
+	cpiCredentialSecretName = "cloud-provider-vsphere-credentials"
+)
+
 var (
 	VSphereKubernetesReferenceAssertions = map[string]func(types.NamespacedName, []metav1.OwnerReference) error{
 		// Need custom Kubernetes assertions for secrets. Secrets in the CAPV tests can also be owned by the vSphereCluster.
-		"Secret": func(_ types.NamespacedName, owners []metav1.OwnerReference) error {
+		"Secret": func(s types.NamespacedName, owners []metav1.OwnerReference) error {
+			// When using vcsim CRS cannot be applied (not supported by the fake API server), so ignoring all the Secrets that should be deployed by CRS.
```
sbueringer, line 157: Can we update the note to link to the PR that fixes this? (or potentially just directly drop it if the CAPI PR merges soon and this can be solved by a CAPI bump)

test/e2e/ownerrefs_finalizers_test.go
```diff
@@ -354,3 +441,26 @@ func forcePeriodicReconcile(ctx context.Context, c ctrlclient.Client, namespace
 		}
 	}()
 }
+
+func FilterObjectsWithKindAndName(clusterName string) func(u unstructured.Unstructured) bool {
+	f := clusterctlcluster.FilterClusterObjectsWithNameFilter(clusterName)
+
+	return func(u unstructured.Unstructured) bool {
+		// When using vcsim CRS cannot be applied (not supported by the fake API server), so ignoring.
```
sbueringer, line 449: Same here, consider dropping this if the CAPI PR merges soon

test/e2e/ownerrefs_finalizers_test.go
```diff
@@ -218,13 +281,35 @@ var (
...
+var vSphereFinalizers = func() map[string]func(types.NamespacedName) []string {
+	if testMode == SupervisorTestMode {
+		return map[string]func(types.NamespacedName) []string{
+			"VirtualMachine": func(_ types.NamespacedName) []string {
+				// When using vcsim additional finalizers are added.
+				if testTarget == VCSimTestTarget {
+					return []string{"virtualmachine.vmoperator.vmware.com", vcsimv1.VMFinalizer}
+				}
+				return []string{"virtualmachine.vmoperator.vmware.com"}
+			},
```
sbueringer, line 293: Do we need this?

FilterObjectsWithKindAndName looks like we are filtering VIrtualMachine objects out in SupervisorTestMode

test/e2e/data/infrastructure-vsphere-supervisor/main/ownerrefs-finalizers/kustomization.yaml
fabriziopandini: we need separated template because the corresponding govimovi template is different from govimomi base.

test/e2e/e2e_suite_test.go
fabriziopandini: That's correct, from what I observed, this call failed because controllers were not ready yet (and thus the webhook was not answering, which is a sort of network error).

test/e2e/ownerrefs_finalizers_test.go
fabriziopandini: https://github.com/kubernetes-sigs/cluster-api/pull/10756 should fix it, I'm ok to wait for the CAPI PR to merge first

test/e2e/ownerrefs_finalizers_test.go
fabriziopandini: Yes we need this, it is for vcsim vs vsphere (not supervisor vs govmomi)

chrischdi:
/retest
vSphere TLS thumbprint was updated

test/e2e/e2e_suite_test.go
sbueringer: Got it!

test/e2e/ownerrefs_finalizers_test.go
sbueringer: _approved_

test/e2e/ownerrefs_finalizers_test.go
sbueringer: Hm. But If I understand the Filter func correctly we are never validating finalizers for VirtualMachine

* with govmomi: VirtualMachine doesn't exist
* with supervisor: we filter VirtualMachines out

I might be misinterpreting something though :)

sbueringer:
Just ping me once the PR is rebased on top of latest CAPI main with the merged PR and then I'll take another look

test/e2e/ownerrefs_finalizers_test.go
fabriziopandini: in CAPV E2E tests we are using CAPV published images, so we actually need to wait for 1.7.3 😅 
what about merging and then follow up, so we unblock the work on supervisor tests with runtime sdk?

test/e2e/ownerrefs_finalizers_test.go
fabriziopandini: fixed + added some comments

test/e2e/ownerrefs_finalizers_test.go
fabriziopandini: fixed usign nightly images

test/e2e/ownerrefs_finalizers_test.go
```diff
@@ -74,17 +79,17 @@ var _ = Describe("Ensure OwnerReferences and Finalizers are resilient with Failu
...
 
-					// Set up a periodic patch to ensure the DeploymentZone anc ClusterResourceSets are reconciled.
+					// Set up a periodic patch to ensure the DeploymentZone are reconciled.
 					// Note: this is required because DeploymentZone are not watching for clusters, and thus the DeploymentZone controller
 					// won't be triggered when we un-pause clusters after modifying objects ownerReferences & Finalizers to test resilience.
-					// WRT to ClusterResourceSets, we are forcing reconcile to avoid the issue described in https://github.com/kubernetes-sigs/cluster-api/pull/10656;
-					// we should reconsider if possible to drop force reconcile after this PR is merged.
 					forceCtx, forceCancelFunc := context.WithCancel(ctx)
-					forcePeriodicReconcile(forceCtx, proxy.GetClient(), namespace)
+					forcePeriodicReconcile(forceCtx, proxy.GetClient())
```
sbueringer, line 92: Can we move this into the if branch from l.82? VSphereDeploymentZones don't exist in supervisor

(I know we have to keep l.91 outside of the if so we can cancel below, but I think it's still worth it not having that running in supervisor tests)

test/e2e/ownerrefs_finalizers_test.go
```diff
@@ -104,31 +109,39 @@ var _ = Describe("Ensure OwnerReferences and Finalizers are resilient with Failu
 						framework.KubeadmControlPlaneOwnerReferenceAssertions,
 						framework.ExpOwnerReferenceAssertions,
 						VSphereKubernetesReferenceAssertions,
-						VSphereReferenceAssertions,
+						VSphereReferenceAssertions(),
 					)
 					// This check ensures that finalizers are resilient - i.e. correctly re-reconciled, when removed.
+					// Note: we are not checking finalizers on VirtualMachine (finalizers are added by VM-Operator / vcsim controller)
+					// as well as other VM Operaror related kinds.
```
sbueringer, line 116:
```suggestion
					// as well as other VM Operator related kinds.
```
nit (same in l.133)

test/e2e/ownerrefs_finalizers_test.go
```diff
@@ -156,36 +169,68 @@ var (
...
+
+				// Following objects are for vm-operator (not managed by CAPV), so checking ownerReferences is not relevant.
+				"VirtualMachineImage":             func(_ types.NamespacedName, _ []metav1.OwnerReference) error { return nil },
+				"NetworkInterface":                func(_ types.NamespacedName, _ []metav1.OwnerReference) error { return nil },
+				"ContentSourceBinding":            func(_ types.NamespacedName, _ []metav1.OwnerReference) error { return nil },
+				"VirtualMachineSetResourcePolicy": func(_ types.NamespacedName, _ []metav1.OwnerReference) error { return nil },
+				"VirtualMachineClassBinding":      func(_ types.NamespacedName, _ []metav1.OwnerReference) error { return nil },
+				"VirtualMachineClass":             func(_ types.NamespacedName, _ []metav1.OwnerReference) error { return nil },
+				"VMOperatorDependencies":          func(_ types.NamespacedName, _ []metav1.OwnerReference) error { return nil },
+			}
```
sbueringer, line 200: Q: Do they all still exist after Christians PR to bump to v1alpha2? (just remember that we got rid of a bunch of CRs)

test/e2e/ownerrefs_finalizers_test.go
```diff
@@ -218,13 +265,35 @@ var (
...
-	"VSphereVM":              func(_ types.NamespacedName) []string { return []string{infrav1.VMFinalizer} },
-	"VSphereClusterIdentity": func(_ types.NamespacedName) []string { return []string{infrav1.VSphereClusterIdentityFinalizer} },
-	"VSphereDeploymentZone":  func(_ types.NamespacedName) []string { return []string{infrav1.DeploymentZoneFinalizer} },
-	"VSphereMachine":         func(_ types.NamespacedName) []string { return []string{infrav1.MachineFinalizer} },
-	"IPAddressClaim":         func(_ types.NamespacedName) []string { return []string{infrav1.IPAddressClaimFinalizer} },
-	"VSphereCluster":         func(_ types.NamespacedName) []string { return []string{infrav1.ClusterFinalizer} },
+var vSphereFinalizers = func() map[string]func(types.NamespacedName) []string {
+	if testMode == SupervisorTestMode {
+		return map[string]func(types.NamespacedName) []string{
+			"VirtualMachine": func(_ types.NamespacedName) []string {
```
sbueringer, line 271: Sorry I think I still don't understand why we are defining finalizers for VirtualMachine. We are not using this information after FilterObjectsWithKindAndName is filtering out VirtualMachines

Happy to take a look via Zoom together

test/e2e/ownerrefs_finalizers_test.go
```diff
@@ -323,26 +392,21 @@ func checkClusterIdentitySecretOwnerRefAndFinalizer(ctx context.Context, c ctrlc
...
-				for _, zone := range deploymentZoneList.Items {
-					annotationPatch := ctrlclient.RawPatch(types.MergePatchType, []byte(fmt.Sprintf("{\"metadata\":{\"annotations\":{\"cluster.x-k8s.io/modifiedAt\":\"%v\"}}}", time.Now().Format(time.RFC3339))))
-					Expect(c.Patch(ctx, zone.DeepCopy(), annotationPatch)).To(Succeed())
-				}
-				// TODO: check if this can be dropped after https://github.com/kubernetes-sigs/cluster-api/pull/10656 is merged
-				Expect(c.List(ctx, crsList, ctrlclient.InNamespace(namespace))).To(Succeed())
-				for _, crs := range crsList.Items {
-					annotationPatch := ctrlclient.RawPatch(types.MergePatchType, []byte(fmt.Sprintf("{\"metadata\":{\"annotations\":{\"cluster.x-k8s.io/modifiedAt\":\"%v\"}}}", time.Now().Format(time.RFC3339))))
-					Expect(c.Patch(ctx, crs.DeepCopy(), annotationPatch)).To(Succeed())
+				if testMode == GovmomiTestMode {
```
sbueringer, line 404: I would prefer not running this func in supervisor mode (see comment above). If we go with that change we could rename the func e.g. to forcePeriodicDeploymentZoneReconcile and then drop this if inside the func

test/e2e/ownerrefs_finalizers_test.go
fabriziopandini: I have a follow-up to this PR, that is to check why in the first chance those objects are included in the object graph.
I will review the list as soon as I get to the bottom of this.

test/e2e/ownerrefs_finalizers_test.go
fabriziopandini: I think I got your point now, PTAL

test/e2e/ownerrefs_finalizers_test.go
sbueringer: Thx!

chrischdi:
/lgtm

/hold cancel

🎉

PR #3076, opened 2024-06-27
🌱 use sha256 thumbprint in tests

chrischdi:

sha1 support for certificates gets removed in Go 1.24 ([xref](https://github.com/golang/go/issues/62048)).

Due to its insecurity we should also use sha256 for thumbprints to ensure this works.

Note: govmomi directly verifies the thumbprint and does not use the library for it, so its not going to break with v1.24.

Fixes #


sbueringer:
Thx!

/lgtm
/approve

PR #3099, opened 2024-07-09
✨ Implement VirtualMachine namingStrategy

### LINKED ISSUE #3100: Allow customization of VirtualMachine names (supervisor-mode)
/kind feature

We have the requirement to customize names of VMs in vCenter to support Windows. As Windows with AD has certain restrictions (name length etc.). See https://learn.microsoft.com/en-us/windows/win32/adschema/a-samaccountname
Especially:
* hostnames must be 20 characters or less and *unique*
* cannot contain any of these characters: "/ \ [ ] : ; | = , + * ? < >

Today we have this implemented for govmomi (https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/77abd94018c71d46a012b48b6353f840e4d43852/pkg/services/vimmachine.go#L396), but not for supervisor mode.

In contrast to govmomi, supervisor is not aware of the OS used in the vm though. So we would like to implement a more flexible solution.

The idea is the following:
* Add a field in VSphereMachineTemplate: VSphereMachineTemplate.spec.template.spec.namingStrategy.template
  * default: `{{ .machine.name }}`
  * for Windows someone could configure: `{{ if le (len .machine.name) 20 }}{{ .machine.name }}{{else}}{{ trimSuffix "-" (trunc 14 .machine.name) }}-{{ trunc -5 .machine.name }}{{end}}`

Example Machine name generation for Windows:
* original:   quick-start-d34gt4 - md-0-wqc85 - 8nxwc     - gfd5v    
* generated:  quick-start-d3 - gfd5v

sbueringer:

Fixes #3100


apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -87,6 +87,23 @@ type VSphereMachineSpec struct {
...
+	// +optional
+	NamingStrategy *VirtualMachineNamingStrategy `json:"namingStrategy,omitempty"`
+}
+
+// VirtualMachineNamingStrategy defines the naming strategy for the VirtualMachines.
+type VirtualMachineNamingStrategy struct {
+	// Template defines the template to use for generating the name of the VirtualMachine object.
+	// If not defined, it will fall back to `{{ .machine.name }}`.
+	// The templating has the following data available:
+	// * `.machine.name`: The name of the Machine object.
```
fabriziopandini, line 101: q: should we add something else? thinking about cluster.name, cluster.namespace, virtualMachine.name, vsphereMachine.name

apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -87,6 +87,23 @@ type VSphereMachineSpec struct {
...
+	// The templating also has the following funcs available:
```
fabriziopandini, line 102: q: should we add sprig funcs?

internal/webhooks/vmware/vspheremachinetemplate.go
```diff
@@ -0,0 +1,107 @@
...
+		if err != nil {
+			allErrs = append(allErrs,
+				field.Invalid(
+					templateFldPath,
+					*namingStrategy.Template,
+					fmt.Sprintf("invalid VirtualMachine name template: %v", err),
+				),
+			)
+		} else {
+			for _, err := range validation.IsDNS1123Subdomain(name) {
```
fabriziopandini, line 86: q: for my understanding. is this a constraint we inherit from VM Operator or else?

internal/webhooks/vmware/vspheremachinetemplate.go
```diff
@@ -0,0 +1,107 @@
...
+				allErrs = append(allErrs,
+					field.Invalid(
+						templateFldPath,
+						*namingStrategy.Template,
+						fmt.Sprintf("invalid VirtualMachine name template, generated name is not valid: %v", err),
```
fabriziopandini, line 91:
```suggestion
						fmt.Sprintf("invalid VirtualMachine name template, generated name is not valid DNS1123Subdomain: %v", err),
```

randomvariable:
not done a deep review, mostly from an API perspective, but it's a 

/lgtm

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -270,6 +283,76 @@ func (v *VmopMachineService) ReconcileNormal(ctx context.Context, machineCtx cap
...
+	}
+
+	return &client.ObjectKey{
+		Namespace: machineNamespace,
+		Name:      name,
+	}, nil
+}
+
+// GenerateVirtualMachineName generates the name of a VirtualMachine based on the naming strategy.
+func GenerateVirtualMachineName(machineName string, namingStrategy *vmwarev1.VirtualMachineNamingStrategy) (string, error) {
```
fabriziopandini, line 321: should this func return a book in case the name is truncated, so we can eventually give a warning in the web hook

apis/vmware/v1beta1/vspheremachine_types.go
sbueringer: Not sure if folks should just freely generate entirely new and different VirtualMachine names. This feature currently assumes that we want to calculate based on the Machine name but we have to make a few changes to it to comply with restrictions.

apis/vmware/v1beta1/vspheremachine_types.go
sbueringer: I was intentionally not adding sprig functions to not expose a huge amount of functions that we don't need of a library that is not supported anymore

internal/webhooks/vmware/vspheremachinetemplate.go
sbueringer: This constraint comes from Kubernetes object names

sbueringer: (we have the same validation in core CAPI for the namingStrategies to generate control plane names etc.)

sbueringer: Added a commet about it

internal/webhooks/vmware/vspheremachinetemplate.go
sbueringer: Added something similar but more useful for users

pkg/services/vmoperator/vmopmachine.go
sbueringer: I would only implement that once/if we decide to return a warning.

Pleaes note that if we return a warning in the VSphereMachineTemplate webhook the logs of our Cluster topology controller will be full with these warnings. I'm not sure if we really want that (especially as that's an entirely expected scenario and nothing concerning for folks reading the logs)

sbueringer:
@fabriziopandini fixed / answered, PTAL :)

pkg/services/vmoperator/vmopmachine.go
sbueringer: Also not sure how to do it. We only have the VSphereMachineTemplate webhook, at this time we don't know if VirtualMachine names will be later truncated in the controller (we just run through generation here with a test Machine name)

apis/vmware/v1beta1/vspheremachine_types.go
fabriziopandini: ACK, might be just add a comment that even if template offers some flexibility, we would like the name to somehow link the Machine name for ensuring better user experience when troubleshooting

apis/vmware/v1beta1/vspheremachine_types.go
fabriziopandini: ACK, let's just make sure we use the same func name/signature sprig is using, so we are consistent with what we already have in CAPI

pkg/services/vmoperator/vmopmachine.go
fabriziopandini: ACK, if it is not already there, then let's add a note on the template field that clarifies that generated names must be valid K8s node names and that CAPV automatically truncates names exceeding max Lenght

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -270,6 +283,76 @@ func (v *VmopMachineService) ReconcileNormal(ctx context.Context, machineCtx cap
...
+	if err := tpl.Execute(&buf, data); err != nil {
+		return "", errors.Wrap(err, "failed to generate name for VirtualMachine")
+	}
+
+	name := buf.String()
+
+	// If the name exceeds the maxNameLength, trim to maxNameLength.
+	// Note: we're not adding a random suffix as the name has to be deterministic.
+	if len(name) > maxNameLength {
+		name = name[:maxNameLength]
```
fabriziopandini, line 350: Thinking about this.
Truncating without introducing a sort of randomness might lead to conflicts...
wdyt?

sbueringer: If I introduce randomness here we have to do a lot more things. I can't start using a different random name for create / get. 

We can consider this but it requires some logic that we are able to get VirtualMachines that we created before so that we are using the Generate func 100% only on create (which is currently not the case)

apis/vmware/v1beta1/vspheremachine_types.go
sbueringer: I copied the function from sprig, see the comment:
```go

// Note: Inlining these functions from sprig to avoid introducing a dependency.
var nameTemplateFuncs = map[string]any{
	"trimSuffix": func(a, b string) string { return strings.TrimSuffix(b, a) },
	"trunc": func(c int, s string) string {
		if c < 0 && len(s)+c > 0 {
			return s[len(s)+c:]
		}
		if c >= 0 && len(s) > c {
			return s[:c]
		}
		return s
	},
}
```

sbueringer:
Added the comments, not sure about introducing randomness

pkg/services/vmoperator/vmopmachine.go
sbueringer: Added a comment to the API field

sbueringer:
Forgot to regenerate

/test pull-cluster-api-provider-vsphere-e2e-govmomi-conformance-ci-latest-main
/test pull-cluster-api-provider-vsphere-e2e-govmomi-conformance-main
/test pull-cluster-api-provider-vsphere-e2e-govmomi-upgrade-1-30-1-31-main
/test pull-cluster-api-provider-vsphere-e2e-supervisor-conformance-ci-latest-main
/test pull-cluster-api-provider-vsphere-e2e-supervisor-conformance-main
/test pull-cluster-api-provider-vsphere-e2e-supervisor-upgrade-1-30-1-31-main
/test pull-cluster-api-provider-vsphere-e2e-vcsim-govmomi-main
/test pull-cluster-api-provider-vsphere-e2e-vcsim-supervisor-main

sbueringer:
@chrischdi I had this flake a few times: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/3099/pull-cluster-api-provider-vsphere-e2e-vcsim-supervisor-main/1810995558580490240 Wondering if that's related to recent changes. Didn't get it before

internal/webhooks/vmware/vspheremachinetemplate.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (webhook *VSphereMachineTemplateWebhook) validate(_ context.Context, _, newVSphereMachineTemplate *vmwarev1.VSphereMachineTemplate) (admission.Warnings, error) {
+	var allErrs field.ErrorList
+
+	// Validate namingStrategy
+	namingStrategy := newVSphereMachineTemplate.Spec.Template.Spec.NamingStrategy
+	if namingStrategy != nil &&
+		namingStrategy.Template != nil {
+		name, err := vmoperator.GenerateVirtualMachineName("machine", namingStrategy)
```
chrischdi, line 75: Optional: I was wondering if we have a way to bubble up a warning for the truncating case, when passing a long name, but that's not straight forward.

chrischdi (commented):
Just an optional question :-)

/lgtm

internal/webhooks/vmware/vspheremachinetemplate.go
sbueringer: Good question, see: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3099#discussion_r1671947310

chrischdi: Good enough 👍

chrischdi:
Feel free to merge :-)

PR #3101, opened 2024-07-10
✨ vGPU implementation - follow-ups

sbueringer:
Follow-up to #3025

Fixes #

### DIFF
--- a/apis/v1beta1/types.go
+++ b/apis/v1beta1/types.go
@@ -253,21 +253,21 @@ type PCIDeviceSpec struct {
 	// virtual machine is cloned.
 	// Mutually exclusive with VGPUProfile as VGPUProfile and DeviceID + VendorID
 	// are two independent ways to define PCI devices.
-	// +kubebuilder:validation:Required
+	// +optional
 	DeviceID *int32 `json:"deviceId,omitempty"`
 	// VendorId is the vendor ID of a virtual machine's PCI, in integer.
 	// Defaults to the eponymous property value in the template from which the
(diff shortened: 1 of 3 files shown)


sbueringer:
Will rebase once #3025 merges

neolit123 (commented):
/lgtm

apis/v1beta1/types.go
```diff
@@ -253,21 +253,21 @@ type PCIDeviceSpec struct {
...
-	// +kubebuilder:validation:Required
+	// +optional
 	VendorID *int32 `json:"vendorId,omitempty"`
 	// VGPUProfile is the profile name of a virtual machine's vGPU, in string.
 	// Defaults to the eponymous property value in the template from which the
 	// virtual machine is cloned.
 	// Mutually exclusive with DeviceID and VendorID as VGPUProfile and DeviceID + VendorID
 	// are two independent ways to define PCI devices.
-	// +kubebuilder:validation:Required
+	// +optional
```
neolit123, line 270: makes sense

sbueringer: I was confused at first why the field was not required in the CRD, but then I realized that omitempty overwrites the explicit marker..

neolit123: what made sense to me is that deviceID + vendorID are mut.ex. with VGPUProfile, if i remember the original PR logic.

sbueringer: Yup, having them all as optional and then having validation in the webhook definitely makes sense

PR #3109, opened 2024-07-18
🌱 test: enable vcsim to run clusterctl upgrade e2e tests

chrischdi:

Part of:
- #2995

Fixes #

### DIFF
--- a/test/e2e/clusterctl_upgrade_test.go
+++ b/test/e2e/clusterctl_upgrade_test.go
@@ -25,6 +25,8 @@ import (
 	capi_e2e "sigs.k8s.io/cluster-api/test/e2e"
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
+
+	"sigs.k8s.io/cluster-api-provider-vsphere/test/framework/vcsim"
 )
 
 var (
@@ -67,11 +69,11 @@ var _ = Describe("When testing clusterctl upgrades using ClusterClass (CAPV 1.10
 				InitWithKubernetesVersion:                "v1.30.0",
 				WorkloadKubernetesVersion:                "v1.30.0",
 				WorkloadFlavor:                           testSpecificSettingsGetter().FlavorForMode("workload"),
-				UseKindForManagementCluster:              true,
+				UseKindForManagementCluster:              testSpecificSettingsGetter().UseKindForManagementCluster,
 				KindManagementClusterNewClusterProxyFunc: kindManagementClusterNewClusterProxyFunc,
 			}
 		})
-	}, WithIP("WORKLOAD_CONTROL_PLANE_ENDPOINT_IP"), WithUseKindForManagementCluster())
+	}, WithIP("WORKLOAD_CONTROL_PLANE_ENDPOINT_IP"), WithAdditionalVCSimServer(true))
 })
 
 var _ = Describe("When testing clusterctl upgrades using ClusterClass (CAPV 1.9=>current, CAPI 1.6=>1.8) [vcsim] [supervisor] [ClusterClass]", func() {
@@ -105,11 +107,11 @@ var _ = Describe("When testing clusterctl upgrades using ClusterClass (CAPV 1.9=
 				InitWithKubernetesVersion:                "v1.29.0",
 				WorkloadKubernetesVersion:                "v1.29.0",
 				WorkloadFlavor:                           testSpecificSettingsGetter().FlavorForMode("workload"),
-				UseKindForManagementCluster:              true,
+				UseKindForManagementCluster:              testSpecificSettingsGetter().UseKindForManagementCluster,
 				KindManagementClusterNewClusterProxyFunc: kindManagementClusterNewClusterProxyFunc,
 			}
 		})
-	}, WithIP("WORKLOAD_CONTROL_PLANE_ENDPOINT_IP"), WithUseKindForManagementCluster())
+	}, WithIP("WORKLOAD_CONTROL_PLANE_ENDPOINT_IP"), WithAdditionalVCSimServer(true))
 })
 
 // getStableReleaseOfMinor returns the latest stable version of minorRelease.
@@ -119,5 +121,8 @@ func getStableReleaseOfMinor(ctx context.Context, releaseMarkerPrefix, minorRele
 }
 
 func kindManagementClusterNewClusterProxyFunc(name string, kubeconfigPath string) framework.ClusterProxy {
+	if testTarget == VCSimTestTarget {
+		return vcsim.NewClusterProxy(name, kubeconfigPath, initScheme())
+	}
 	return framework.NewClusterProxy(name, kubeconfigPath, initScheme())
 }
--- a/test/e2e/clusterctl_upgrade_test.go
+++ b/test/e2e/clusterctl_upgrade_test.go
@@ -36,7 +36,7 @@ var (
 	capvReleaseMarkerPrefix = "go://sigs.k8s.io/cluster-api-provider-vsphere@v%s"
 )
 
-var _ = Describe("When testing clusterctl upgrades using ClusterClass (CAPV 1.10=>current, CAPI 1.7=>1.8) [supervisor] [ClusterClass]", func() {
+var _ = Describe("When testing clusterctl upgrades using ClusterClass (CAPV 1.10=>current, CAPI 1.7=>1.8) [vcsim] [supervisor] [ClusterClass]", func() {
 	const specName = "clusterctl-upgrade-1.10-current" // prefix (clusterctl-upgrade) copied from CAPI
 	Setup(specName, func(testSpecificSettingsGetter func() testSettings) {
 		capi_e2e.ClusterctlUpgradeSpec(ctx, func() capi_e2e.ClusterctlUpgradeSpecInput {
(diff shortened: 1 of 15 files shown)


test/e2e/e2e_setup_test.go
```diff
@@ -185,23 +135,126 @@ func Setup(specName string, f func(testSpecificSettings func() testSettings), op
...
-			runtimeExtensionProviders = append(runtimeExtensionProviders, "vm-operator", "net-operator")
-		}
+		postNamespaceCreatedFunc = func(managementClusterProxy framework.ClusterProxy, workloadClusterNamespace string) {
+			var ipVariables map[string]string
 
-		if testTarget == VCSimTestTarget {
-			runtimeExtensionProviders = append(runtimeExtensionProviders, "vcsim")
+			if testTarget == VCSimTestTarget && options.useKindForManagementCluster {
+				Byf("Creating a vcsim server")
+				Eventually(func() error {
```
fabriziopandini, line 143: q: why are we creating another VC sim server? is it the one running on the kind cluster?

might be we need to check that it is reconciled as something pops up in status

test/e2e/e2e_setup_test.go
```diff
@@ -73,6 +74,13 @@ func WithPrefix(variableName string) SetupOption {
 	}
 }
 
+// WithUseKindForManagementCluster instructs Setup to run extra steps for the separate kind management cluster.
+func WithUseKindForManagementCluster() SetupOption {
```
fabriziopandini, line 78: might be: withAdditionalVCSimServer(bool)

and the bool is a var that we assign as a input to this function and as a value for ClusterctlUpgradeSpecInput.UseKindForManagementCluster, so it surfaces the correlation between the two settings

test/e2e/e2e_setup_test.go
chrischdi: This is on the second kind cluster.

test/e2e/data/infrastructure-vsphere-govmomi/v1.9/topology/kustomization.yaml
```diff
@@ -6,4 +6,8 @@ resources:
 patchesStrategicMerge:
   - ../commons/cluster-resource-set-label.yaml
   - ../commons/cluster-network-CIDR.yaml
-  - ../commons/cluster-resource-set-csi-insecure.yaml
\ No newline at end of file
+  - ../commons/cluster-resource-set-csi-insecure.yaml
+patches:
```
sbueringer, line 10: Is there a reason why we have to do this here with a patch while in 1.10 (& supervisor) it's just directly in the YAML?

test/e2e/e2e_setup_test.go
```diff
@@ -185,23 +144,126 @@ func Setup(specName string, f func(testSpecificSettings func() testSettings), op
 			Variables:            testSpecificVariables,
 		})
 
-		if testMode == SupervisorTestMode {
-			runtimeExtensionProviders = append(runtimeExtensionProviders, "vm-operator", "net-operator")
-		}
+		postNamespaceCreatedFunc = func(managementClusterProxy framework.ClusterProxy, workloadClusterNamespace string) {
```
sbueringer, line 147: Please add a comment to explain why all of this has to happen after namespace creation

test/e2e/e2e_setup_test.go
```diff
@@ -224,15 +286,16 @@ func Setup(specName string, f func(testSpecificSettings func() testSettings), op
 				}
 				return flavor
 			},
-			RuntimeExtensionProviders: runtimeExtensionProviders,
+			RuntimeExtensionProviders:   runtimeExtensionProviders,
+			UseKindForManagementCluster: options.additionalVCSimServer,
```
sbueringer, line 290: Why is UseKindForManagementCluster set to additionalVCSimServer? Shouldn't this be an entirely separate field?

sbueringer: Also does this mean we now have cases again where we can't use kind for the additional management cluster?

test/e2e/e2e_setup_test.go
```diff
@@ -185,23 +144,126 @@ func Setup(specName string, f func(testSpecificSettings func() testSettings), op
...
-		switch testTarget {
-		case VCenterTestTarget:
-			// cleanup IPs/controlPlaneEndpoint created by the in cluster ipam provider.
-			Expect(inClusterAddressManager.Cleanup(ctx, testSpecificIPAddressClaims)).To(Succeed())
-		case VCSimTestTarget:
-			// cleanup IPs/controlPlaneEndpoint created by the vcsim controller manager.
-			Expect(vcsimAddressManager.Cleanup(ctx, testSpecificIPAddressClaims)).To(Succeed())
+		if !skipCleanup {
+			Byf("Cleaning up test env for %s", specName)
+			// We can't cleanup when a kind management cluster is used, because it won't exist anymore.
```
sbueringer, line 262: > because it won't exist anymore.

Why?

I assume because the test itself cleans it up? Please document

(This seems very specific to the clusterctl upgrade test, right? not necessarily in general when we would set additionalVCSimServer for other tests in the future)

test/e2e/ownerrefs_finalizers_test.go
```diff
@@ -53,31 +53,32 @@ var _ = Describe("Ensure OwnerReferences and Finalizers are resilient [vcsim] [s
...
 			return capi_e2e.QuickStartSpecInput{
 				E2EConfig:             e2eConfig,
 				ClusterctlConfigPath:  testSpecificSettingsGetter().ClusterctlConfigPath,
 				BootstrapClusterProxy: bootstrapClusterProxy,
 				ArtifactFolder:        artifactFolder,
 				SkipCleanup:           skipCleanup,
 				Flavor:                ptr.To(testSpecificSettingsGetter().FlavorForMode("ownerrefs-finalizers")),
-				PostNamespaceCreated:  testSpecificSettingsGetter().PostNamespaceCreatedFunc,
+				PostNamespaceCreated: func(managementClusterProxy framework.ClusterProxy, workloadClusterNamespace string) {
+					testSpecificSettingsGetter().PostNamespaceCreatedFunc(managementClusterProxy, workloadClusterNamespace)
```
sbueringer, line 64: Why is this now in the PostNamespaceCreated func? Please document

sbueringer (commented):
~ looks okay to me

test/e2e/e2e_setup_test.go
sbueringer: I find this very confusing to be honest

test/e2e/clusterctl_upgrade_test.go
```diff
@@ -67,14 +69,14 @@ var _ = Describe("When testing clusterctl upgrades using ClusterClass (CAPV 1.10
 				InitWithKubernetesVersion:                "v1.30.0",
 				WorkloadKubernetesVersion:                "v1.30.0",
 				WorkloadFlavor:                           testSpecificSettingsGetter().FlavorForMode("workload"),
-				UseKindForManagementCluster:              true,
+				UseKindForManagementCluster:              testSpecificSettingsGetter().UseKindForManagementCluster,
```
sbueringer, line 72: Under which circumstances is this not true anymore?

test/e2e/e2e_setup_test.go
chrischdi: We cannot create two VCSimServers in-parallel at the same management cluster. So `additionalVCSimServer` only makes sense to be used when we have a separate management cluster.

We don't block other cases because we are not forced to use the testsettings getter.

I added it here to follow https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3109#discussion_r1683333712

I'm also fine to not expose the `UseKindForManagementCluster` bool and just explicitly set it.

test/e2e/e2e_setup_test.go
chrischdi: Feels a bit off to document it but okay for me. It is not documented elsewhere for other things "why something needs to happen here".

test/e2e/ownerrefs_finalizers_test.go
chrischdi: added a comment.

test/e2e/clusterctl_upgrade_test.go
chrischdi: Currently: if `additionalVCSimServer` is set to true, this is also true.

test/e2e/e2e_setup_test.go
```diff
@@ -185,23 +143,58 @@ func Setup(specName string, f func(testSpecificSettings func() testSettings), op
...
+			// For VCSim we use the ControlPlaneEndpoint CRD inside the managementClusterProxy.
+			testSpecificIPAddressManager, testSpecificIPAddressClaims, testSpecificVariables = allocateIPAddresses(managementClusterProxy, options)
+
+			// Get variables required when running on VCSim like VSphere Server address, user, etc.
+			if testTarget == VCSimTestTarget {
+				addVCSimTestVariables(managementClusterProxy, specName, testSpecificIPAddressClaims, testSpecificVariables)
+			}
+
+			// Re-write the clusterctl config file and add the new variables created above (ip addresses, VCSim variables).
+			testSpecificClusterctlConfigPath = fmt.Sprintf("%s-%s.yaml", strings.TrimSuffix(clusterctlConfigPath, ".yaml"), specName)
```
sbueringer, line 170: Nit: This seems to be the same value as above. I would probably re-use the variable from the outer scope above, especially because it's important that it's the same value as we also use this to read the previous config (ClusterctlConfigPath parameter)

sbueringer (commented):
Thank you very much!
Just a few nits

test/e2e/data/infrastructure-vsphere-govmomi/v1.9/topology/kustomization.yaml
sbueringer: (up to you)

chrischdi: Ah now I got your question.

> If I understand correctly, what you're saying is that the cluster-template-topology.yaml on release-1.10 already had the controlPlanePort while the one in release-1.9 didn't have it and that's why we use a separate patch instead of modifying cluster-template-topology.yaml? (basically to preserve it in it's original state)

That's was the point, but I did only follow it for the cluster template.

> Overall fine to me, just a nit. I was just wondering why it's different :) (especiallly because we also just modify the ClusterClass instead of having a patch there as well)

Fair point :D

I'll change the base file directly instead for consistency then.

sbueringer:
Thx!

/lgtm
/approve

/retest

/assign @fabriziopandini 

/hold
(up to you, feel free to merge from my side)

fabriziopandini:
/lgtm
/approve

Great work!

PR #3128, opened 2024-07-31
✨ Enable NSX LB healthcheck on kube-apiserver

DanielXiao:
In the VPC network, the vm-operator Pod and Kubernetes Clusters are in different VPCs. TCP readiness probe does not work across VPC so use LB 's healthcheck instead. This annotation is agnostic to LB providers. NSX LB supports it to enable healthcheck while AVI LB has healthcheck enabled by default. vm-operator will propagate this annotation to Service annotation.

Fixes #

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ replace github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels => github.c
 
 require (
 	github.com/vmware-tanzu/net-operator-api v0.0.0-20240326163340-1f32d6bf7f9d
-	github.com/vmware-tanzu/nsx-operator/pkg/apis v0.1.0
+	github.com/vmware-tanzu/nsx-operator/pkg/apis v0.0.0-20240816055841-8081971d43ea
 	// The version of vm-operator should be kept in sync with the manifests at: config/deployments/integration-tests
 	github.com/vmware-tanzu/vm-operator/api v1.8.6
 	github.com/vmware-tanzu/vm-operator/external/ncp v0.0.0-20240404200847-de75746a9505
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/pkg/errors"
 	netopv1 "github.com/vmware-tanzu/net-operator-api/api/v1alpha1"
-	vpcapisv1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
+	nsxvpcv1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
 	vmoprv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
 	ncpv1 "github.com/vmware-tanzu/vm-operator/external/ncp/api/v1alpha1"
 	"gopkg.in/fsnotify.v1"
@@ -65,7 +65,7 @@ func New(ctx context.Context, opts Options) (Manager, error) {
 	_ = vmoprv1.AddToScheme(opts.Scheme)
 	_ = ncpv1.AddToScheme(opts.Scheme)
 	_ = netopv1.AddToScheme(opts.Scheme)
-	_ = vpcapisv1.AddToScheme(opts.Scheme)
+	_ = nsxvpcv1.AddToScheme(opts.Scheme)
 	_ = topologyv1.AddToScheme(opts.Scheme)
 	_ = ipamv1.AddToScheme(opts.Scheme)
 
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/pkg/errors"
 	netopv1 "github.com/vmware-tanzu/net-operator-api/api/v1alpha1"
-	nsxopv1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+	vpcapisv1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
 	vmoprv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
 	ncpv1 "github.com/vmware-tanzu/vm-operator/external/ncp/api/v1alpha1"
 	"gopkg.in/fsnotify.v1"
@@ -65,7 +65,7 @@ func New(ctx context.Context, opts Options) (Manager, error) {
 	_ = vmoprv1.AddToScheme(opts.Scheme)
 	_ = ncpv1.AddToScheme(opts.Scheme)
 	_ = netopv1.AddToScheme(opts.Scheme)
-	_ = nsxopv1.AddToScheme(opts.Scheme)
+	_ = vpcapisv1.AddToScheme(opts.Scheme)
 	_ = topologyv1.AddToScheme(opts.Scheme)
 	_ = ipamv1.AddToScheme(opts.Scheme)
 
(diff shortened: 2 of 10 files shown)


sbueringer:
Looks good so far

go.mod
```diff
@@ -8,7 +8,7 @@ replace github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels => github.c
 
 require (
 	github.com/vmware-tanzu/net-operator-api v0.0.0-20240326163340-1f32d6bf7f9d
-	github.com/vmware-tanzu/nsx-operator/pkg/apis v0.1.0
+	github.com/vmware-tanzu/nsx-operator/pkg/apis v0.0.0-20240816055841-8081971d43ea
```
chrischdi, line 11: While this is only an API package:
Question: should we pick a random commit for this? Will they have a released version soon?

pkg/manager/manager.go
```diff
@@ -22,7 +22,7 @@ import (
 
 	"github.com/pkg/errors"
 	netopv1 "github.com/vmware-tanzu/net-operator-api/api/v1alpha1"
-	nsxopv1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+	vpcapisv1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
```
chrischdi, line 25:
```suggestion
	nsxvpcv1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
```

go.mod
chrischdi: The other way would be to do it similar to how we do for vm-operators topologyv1 and have an internal copy.

cc @sbueringer @fabriziopandini .

pkg/services/network/network_test.go
```diff
@@ -630,10 +630,10 @@ var _ = Describe("Network provider", func() {
 				Expect(createdSubnetSet.Spec).To(Equal(initialSubnetSet.Spec), "SubnetSetSpec should not have been modified")
 			})
 
-			It("should successfully retrieve VM service annotations, confirm they are empty", func() {
+			It("should successfully retrieve VM service annotations, including the annotation to enable LB healthcheck", func() {
 				annotations, err := np.GetVMServiceAnnotations(ctx, clusterCtx)
 				Expect(err).ToNot(HaveOccurred())
-				Expect(annotations).To(Equal(make(map[string]string)))
+				Expect(annotations).To(HaveKey("lb.iaas.vmware.com/enable-endpoint-health-check"))
```
chrischdi, line 636: let's use the const here

pkg/services/network/nsxt_vpc_provider.go
```diff
@@ -168,7 +173,7 @@ func (vp *nsxtVPCNetworkProvider) GetClusterNetworkName(ctx context.Context, clu
 
 // The GetVMServiceAnnotations method always returns an empty map representing annotations.
 func (vp *nsxtVPCNetworkProvider) GetVMServiceAnnotations(_ context.Context, _ *vmware.ClusterContext) (map[string]string, error) {
-	return map[string]string{}, nil
+	return map[string]string{AnnotationEnableEndpointHealthCheckKey: ""}, nil
```
chrischdi, line 176: Should we have a value here or is it defined that it must be empty?

```suggestion
	return map[string]string{AnnotationEnableEndpointHealthCheckKey: "true"}, nil
```

pkg/services/network/nsxt_vpc_provider.go
```diff
@@ -40,6 +40,11 @@ import (
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services"
 )
 
+const (
+	// AnnotationEnableEndpointHealthCheckKey is the key of the annotation that is used to enable health check on the VMService endpoint port.
+	AnnotationEnableEndpointHealthCheckKey = "lb.iaas.vmware.com/enable-endpoint-health-check"
```
chrischdi, line 45:
```suggestion
	// NSXTVPCEnableEndpointHealthCheckAnnotationKey is the key of the annotation that is used to enable health check on the VMService endpoint port.
	NSXTVPCEnableEndpointHealthCheckAnnotationKey = "lb.iaas.vmware.com/enable-endpoint-health-check"
```

Do we have a link to some code which more describes what this annotation will do? And maybe add some more to the godoc what it does?

go.mod
DanielXiao: There will not be a tag in a short time and they only promise 3 CRDs we are using can have the final version this week. Internal copy does not work because they will change spec, status field names.

pkg/manager/manager.go
DanielXiao: I will refactor the name.

pkg/services/network/nsxt_vpc_provider.go
DanielXiao: This is expected to be empty

go.mod
sbueringer: If we are only using the API types from there I would prefer using a commit over copying the content of the commit.

The reason why we did this in vm-operator was because we are using vm-operator in a **lot** of other places as well

DanielXiao: Could you point me the example? I only see vm-operator tag in the go mod https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/75a5e72691d58f48027203b2c0afbc989390b8aa/go.mod#L13

pkg/services/network/nsxt_vpc_provider.go
chrischdi: Let's add a comment please that it needs to be empty :-)

pkg/services/network/constants.go
```diff
@@ -19,7 +19,7 @@ package network
 
 import (
 	netopv1 "github.com/vmware-tanzu/net-operator-api/api/v1alpha1"
-	nsxopv1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/v1alpha1"
+	nsxvpcv1 "github.com/vmware-tanzu/nsx-operator/pkg/apis/vpc/v1alpha1"
```
chrischdi, line 22: let's add a linter rule for this import in `.golangci.yml`

pkg/services/network/nsxt_vpc_provider.go
```diff
@@ -40,6 +40,13 @@ import (
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services"
 )
 
+const (
+	// AnnotationEnableEndpointHealthCheckKey is the key of the annotation that is used to enable health check on the
+	// Service endpoint port. vm-operator propagates annotations in VMService to Service and LB providers like NSX-T
+	// will enable health check on the endpoint target port when this annotation is present on the Service.
+	AnnotationEnableEndpointHealthCheckKey = "lb.iaas.vmware.com/enable-endpoint-health-check"
```
chrischdi, line 47: Let's move this definition to `constants.go` and rename the key + some adjustments for the godoc:

```suggestion
	// NSXTVPCEnableEndpointHealthCheckAnnotationKey is used by VM Operator to enable health checks for the Service's endpoint port.
	NSXTVPCEnableEndpointHealthCheckAnnotationKey = "lb.iaas.vmware.com/enable-endpoint-health-check"
```

pkg/services/network/constants.go
DanielXiao: Done. Please take a look if I set it correctly.

pkg/services/network/nsxt_vpc_provider.go
DanielXiao: Done.

pkg/services/network/constants.go
sbueringer: Looks perfect, thx!

sbueringer:
/lgtm
/approve

/hold
@chrischdi Let's merge?

PR #3149, opened 2024-08-09
🌱 packaging: write kubeadm script log for ignition

chrischdi:

Debatable if we want to have this. It was handy for debugging. Cloud-init does it simiarily.

/hold

Fixes #

### DIFF
--- a/test/framework/log/collector.go
+++ b/test/framework/log/collector.go
@@ -96,6 +96,8 @@ func (c *MachineLogCollector) CollectMachineLog(ctx context.Context, ctrlClient 
 			"sudo", "cat", "/var/log/cloud-init.log"),
 		captureLogs("cloud-init-output.log",
 			"sudo", "cat", "/var/log/cloud-init-output.log"),
+		captureLogs("kubeadm-service.log",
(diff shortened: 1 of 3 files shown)


test/framework/log/collector.go
```diff
@@ -96,6 +96,8 @@ func (c *MachineLogCollector) CollectMachineLog(ctx context.Context, ctrlClient
 			"sudo", "cat", "/var/log/cloud-init.log"),
 		captureLogs("cloud-init-output.log",
 			"sudo", "cat", "/var/log/cloud-init-output.log"),
+		captureLogs("kubeadm-service.log",
```
sbueringer, line 99: I assume this doesn't fail the test because we only bubble up the error to a certain point, but we don't fail the test because of that
(also the reason why the cloud-init files above don't lead to a failure of the ignition test)

sbueringer:
Feel free to merge once tests are green

/lgtm
/approve

test/framework/log/collector.go
chrischdi: Exact, it does not fail on errors, just logs it.

PR #3167, opened 2024-08-22
:sparkles: add function to remove VSphere VM object from VM Group

abhinandanbaheti:

This PR adds a function to remove VSphere VM object from VM Group

Fixes #

### DIFF
--- a/pkg/services/govmomi/cluster/vmgroup.go
+++ b/pkg/services/govmomi/cluster/vmgroup.go
@@ -84,3 +84,25 @@ func (vg VMGroup) HasVM(vmObj types.ManagedObjectReference) (bool, error) {
 func (vg VMGroup) listVMs() []types.ManagedObjectReference {
 	return vg.ClusterVmGroup.Vm
 }
+
+// Remove a VSphere VM object from the VM Group.
+func (vg VMGroup) Remove(ctx context.Context, vmObj types.ManagedObjectReference) (*object.Task, error) {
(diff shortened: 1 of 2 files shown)


sbueringer:
Why are we adding a function that we are seemingly not using?

abhinandanbaheti:
> Why are we adding a function that we are seemingly not using?

@sbueringer 

I have a use case where I need to add / remove VM's to/from VM Group through Golang API, andI didn't find a direct way to for these in the govmomi, and then found this suggestion in the (https://github.com/vmware/govmomi/issues/3033#issuecomment-1402194541) to use CAPV.  It has wrapper method to add VM in VM Group, check if VM Group exists, but doesn't have method to remove the VM from VM Group. Therefore, I have added this PR.

sbueringer:
@abhinandanbaheti 

I think Doug is not suggesting to use CAPV as a library
He's just saying that in the CAPV code there "is a good example of adding VMs to a group"

I would highly recommend not depending on CAPV as a library to communicate with vSphere, because it is not a library
We only implement what we use in CAPV and there are absolutely no guarantees that anything stays stable for consumers of CAPV Go code (or even that it stays exported/public)

You also get a huge amount of Go dependencies and coupling to use exactly the same Go dependencies as CAPV if you use us as a library (e.g. controller-runtime, Cluster API, ...)

sbueringer:
Closing this for now ^^ (also talked about this with @abhinandanbaheti)

/close

Issue #2596, opened 2024-01-08
Tracking issue for kube-vip workaround introduced for v1.29


chrischdi:
This issue is for tracking the following two kube-vip issues:

* https://github.com/kube-vip/kube-vip/issues/692
* https://github.com/kube-vip/kube-vip/issues/684

We did hit this issue while implementing support for v1.29 for #2564 in:

* #2561

As a follow-up we removed a part of the above PR because it caused flakes in e2e tests in:

* #2583

chrischdi:
We should try to get rid of `packaging/flavorgen/flavors/kubevip/kube-vip-prepare.sh` for k8s >= v1.31, because CAPI enables the `ControlPlaneKubeletLocalMode` feature-gate in kubeadm and that should make us not depend on the super-admin.conf anymore.

More information: https://github.com/kube-vip/kube-vip/issues/684#issuecomment-2303984047

neolit123:
> because CAPI enables the `ControlPlaneKubeletLocalMode` feature-gate in kubeadm and that should make us not depend on the super-admin.conf anymore.
> 
> More information: [kube-vip/kube-vip#684 (comment)](https://github.com/kube-vip/kube-vip/issues/684#issuecomment-2303984047)

cc @christianang @lubronzhan

lubronzhan:
Good. I can work on this

sbueringer:
Great! We were thinking about having an if/else based on Kubernetes version to stop adding these workaround patches for >= 1.31

sbueringer:
@chrischdi could this (https://github.com/kubernetes/kubernetes/pull/126460) somehow fix the first issue?

chrischdi:
Not really sure, but may be worth a try.

sbueringer:
Context: https://github.com/kubernetes/kubernetes/pull/126460 was done because of this bug report related to kube-vip https://github.com/kubernetes/kubernetes/issues/119852

PR #3168, opened 2024-08-23
:seedling: Skip kube-vip-prepare for 1.31+ k8s since CAPI won't depend on super-admin.conf

lubronzhan:
Skip kube-vip-prepare for 1.31+ k8s since CAPI won't depend on super-admin.conf

Fixes #https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/2596


lubronzhan:
I manually modified the file. Looks like I shouldn't
```
         localhost.localdomain localhost4 localhost4.localdomain4" >>/etc/hosts
       - mkdir -p /etc/pre-kubeadm-commands
       - for script in $(find /etc/pre-kubeadm-commands/ -name '*.sh' -type f | sort);
-        do echo "Running script $script"; "$script"; done
+        do echo "Running script $script"; "$script"; done
\ No newline at end of file
```

lubronzhan:
cc @sbueringer @chrischdi

chrischdi:
Looks like this does not work, there seems to still be some dependency on the loadbalancer IP in this case:

```
E0826 09:31:30.638835       1 leaderelection.go:332] error retrieving resource lock kube-system/plndr-cp-lock: leases.coordination.k8s.io "plndr-cp-lock" is forbidden: User "kubernetes-admin" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "kube-system"
```

chrischdi:
More details:

https://github.com/kube-vip/kube-vip/issues/684#issuecomment-2309781000

chrischdi:
cc @sbueringer

chrischdi:
v1.29 introduced the `super-admin.conf`. In v1.28 and before, the `admin.conf` immediately had the required permissions.

lubronzhan:
IC. so kubelet is able to talk to api through local ip, but not kubeadm. Kubeadm still talks to api through control plane ip. Is there use case that kubeadm init will be run not on the control plane node? Should it also support this mode to talk to api through localhost?
```
With ControlPlaneKubeletLocalMode and when referncing admin.conf for kube-vip:

The kubelet got started
The kubelet bootstraps itself using the local control-plane IP (not depending on kube-vip being up)
The admin.conf gets created
The kubelet should be able to start kube-vip now
```

lubronzhan:
> IC. so kubelet is able to talk to api through local ip, but not kubeadm. Kubeadm still talks to api through control plane ip. Is there use case that kubeadm init will be run not on the control plane node? Should it also support this mode to talk to api through localhost?
> 
> ```
> With ControlPlaneKubeletLocalMode and when referncing admin.conf for kube-vip:
> 
> The kubelet got started
> The kubelet bootstraps itself using the local control-plane IP (not depending on kube-vip being up)
> The admin.conf gets created
> The kubelet should be able to start kube-vip now
> ```

@neolit123 what do you think

sbueringer:
@lubronzhan @neolit123 @chrischdi *bump* :)

neolit123:
missed the prior ping.

> IC. so kubelet is able to talk to api through local ip, but not kubeadm. Kubeadm still talks to api through control plane ip. Is there use case that kubeadm init will be run not on the control plane node?

not in CAPI (automated) workflows. the user can do it out of band though.
e.g. ssh on a CP machine and call `kubeadm init phase ...something` as an utility for something.

> Should it also support this mode to talk to api through localhost?
> 

the admin.conf and super-admin.conf always talk to the CPE (control plane endpoint), that is because we want the user to be able to reach any API server.

> ```
> With ControlPlaneKubeletLocalMode and when referncing admin.conf for kube-vip:
> 
> The kubelet got started
> The kubelet bootstraps itself using the local control-plane IP (not depending on kube-vip being up)
> The admin.conf gets created
> The kubelet should be able to start kube-vip now
> ```

IIUC, at that point admin.conf still doesn't have the permissions.
thus initial kube-vip bootstrap must use super-admin.conf and then move to admin.conf, that is what @chrischdi workaround proposal is doing.

these are points a brought before but, ideally kube-vip should stop using the *admin.conf (because it's not an admin) and stop requiring any RBAC on bootstrap. it could have delayed RBAC requirements IIUC.

chrischdi:
/close 

Because this won't work

PR #3178, opened 2024-08-29
🐛 Small test fix. Remove OwnerReference & Check for Finalizers being empty

erwin-kok:
Two small fixes:

- The secret was created with an OwnerReference. Later on (line ~141) it is checked if the controller sets the OwnerReference by counting the number of OwnerReferences. However, this is always > 0 since the OwnerReference was explicitly set before, so this check always passes. Removed the initial OwnerReference and counting the ones set by the controller.

- Line ~237 it is stated: "should remove vspherecluster finalizer if the secret does not exist". However, it was not checked whether the Finalizers are actually empty. Added this check.

### DIFF
--- a/controllers/vspherecluster_reconciler_test.go
+++ b/controllers/vspherecluster_reconciler_test.go
@@ -60,14 +60,6 @@ var _ = Describe("VIM based VSphere ClusterReconciler", func() {
 				ObjectMeta: metav1.ObjectMeta{
 					GenerateName: "secret-",
 					Namespace:    "default",
-					OwnerReferences: []metav1.OwnerReference{
-						{
-							APIVersion: "bitnami.com/v1alpha1",
-							Kind:       "SealedSecret",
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
/approve

neolit123:
/ok-to-test
/kind cleanup

thanks

erwin-kok:
Thanks @neolit123 !
This PR still needs a lgtm, who can have a look and provide one? @chrischdi perhaps? Thanks!

erwin-kok:
You are right, better to check the expected OwnerReference. I changed the code accordingly, and also for the Finalizer (check the expected Finalizer i.s.o. relying on >0). Can you have another look @chrischdi ? Thanks!

chrischdi (commented):
/lgtm

chrischdi:
Thanks @erwin-kok for contributing this improvement! 🎉

PR #3214, opened 2024-10-03
:sparkles: Add ability to add new data disks to VM during clone process

### LINKED ISSUE #3213: Add data disks not present in VM template to VM during cloning
/kind feature

As a cluster admin, I would like to create machines with additional data disks so that I can have dedicated disks for such things as etcd, container runtime, image storage, swap, etc.

The idea here is to add new disks to a VM during the clone process that are not part of the vSphere VM Template.  This way we can share one template across multiple machine sets but tweak the additional data disks across each machine set based on the needs of that group.

Initial goal is to support the govmomi mode.

Currently most vSphere VM Templates have a single defined disk for the OS.  This feature should still check to see if any other disks are present in template and make sure new data disks are added.

There currently is a property `additionalDisksGiB` that will change the size of existing disks found in the vSphere VM Template, but is not related to adding disks not present in the template.  This may cause confusion.  Current proposal is to have a new field that will start simple and grow over time w/ additional capabilities.  One example would be to have new disks only use same controller (usually scsi controller) as the primary disk.

**Environment:**

- Cluster-api-provider-vsphere version: 1.11
- Kubernetes version: (use `kubectl version`): 1.30.3
- OS (e.g. from `/etc/os-release`):  Red Hat Enterprise Linux CoreOS 418

vr4manta:
Add API changes that adds the ability to add new data disks to VMs during the cloning process.  These new disks are not present in the vSphere VM Template.

Fixes #3213


neolit123 (commented):
please ping once this is ready for review.

vr4manta:
@neolit123 
This is ready to start reviewing.

chrischdi:
cc @lubronzhan @neolit123 @rikatz , this one should be ready for review (xref CAPV office hours) :-) Thanks @vr4manta for driving this!

apis/v1beta1/types.go
```diff
@@ -203,6 +203,16 @@ type VirtualMachineCloneSpec struct {
...
 	// +optional
 	HardwareVersion string `json:"hardwareVersion,omitempty"`
+	// DataDisks holds information for additional disks to add to the VM that are not part of the VM's OVA template.
+	// +optional
+	DataDisks []VSphereDisk `json:"dataDisks,omitempty"`
+}
+
+// VSphereDisk describes additional disks for vSphere to be added to VM that are not part of the VM OVA template.
+type VSphereDisk struct {
+	// SizeGiB is the size of the disk (in GiB).
```
neolit123, line 213:
```suggestion
	// SizeGiB is the size of the disk in GiB.
```

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
...
+        * [Implementation Details](#implementation-details)
+        * [Notes/Constraints](#notesconstraints)
+    * [Upgrade Strategy](#upgrade-strategy)
+
+## Glossary
+
+
+## Summary
+
+As the use of kubernetes clusters grows, admins are needing more and more improvements to the VMs themselves to make sure they run as smoothly as possible.  The number of cores and memory continue to increase for each machine and this is causing the amount of workloads to increase on each virtual machine.  This growth is now causing the base VM image to not provide enough storage for OS needs.  In some cases, users just increase the size of the primary disk using the existing configuration options for machines; however, this does not allows for all desired configuration choices.  Admins are now wanting the ability to add additional disks to these VMs for things such as etcd storage, image storage, container runtime and even swap.
```
neolit123, line 44:
```suggestion
As the use of Kubernetes clusters grows, admins are needing more and more improvements to the VMs themselves to make sure they run as smoothly as possible.  The number of cores and memory continue to increase for each machine and this is causing the amount of workloads to increase on each virtual machine.  This growth is now causing the base VM image to not provide enough storage for OS needs.  In some cases, users just increase the size of the primary disk using the existing configuration options for machines; however, this does not allows for all desired configuration choices.  Admins are now wanting the ability to add additional disks to these VMs for things such as etcd storage, image storage, container runtime and even swap.  
```

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
+# CAPV Additional Data Disks When Creating New Machines
```
neolit123, line 1: a bit surprising to have the proposal document as part of the implementation PR.
this sort of implies "here is the implementation, and there is a proposal for it too".

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
...
+### Goals
+
+* Add new configuration for machines that are used to define new data disks to add to a VM.
+* Align new property naming to be similar or even match other providers if possible.
+* Do not boil the ocean with the initial implementation of this feature.
+
+### Non-Goals/Future Work
+
+* Add abiltiy to advance configure additions disks (such as define controller type (IDE, scsi, etc) or unit number in the controller)
+* Any disk management features such as encryption
```
neolit123, line 63:
```suggestion
* Add abiltiy to advance configure additions disks, such as, define controller type (IDE, scsi, etc) or unit number in the controller.
* Any disk management features such as encryption.
```

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As an admin, I’d like my control plane machines to have an extra disk added so I can dedicate that disk for etcd databse through my own means of mount management
```
neolit123, line 71:
```suggestion
As an admin, I’d like my control plane machines to have an extra disk added so I can dedicate that disk for etcd database through my own means of mount management.
```

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### Story 2
+
+As an admin, I’d like my compute machine set to have an extra disk added to each machine so that I can use it as a dedicated disk for container image storage.
```
neolit123, line 75: machine / VM ?

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Requirements
+
+#### Functional Requirements
+
+### Overall Design
+
+The CAPV vsphere machine clone spec (VirtualMachineCloneSpec) contains a new array field to specify all additional data disks (dataDisks).  These disks will be created and added to the VM during the clone process.
```
neolit123, line 83:
```suggestion
The CAPV vSphere machine clone spec (VirtualMachineCloneSpec) contains a new array field to specify all additional data disks (dataDisks).  These disks will be created and added to the VM during the clone process.
```

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
...
+
+Once the machine has been created, any change to the DataDisks field will follow existing change behaviors for machine sets and single machine definitions.
+
+CAPV will not be responsible for any custom mounting of the data disks.  CAPV will create the disk and attach to the VM.  Each OS may assign them differently so admin will need to understand the OS image they are using for their template and then configure each node.
+
+#### For User Story 1
+
+CAPV will create the new disks for Control plane nodes during cluster creation.
```
neolit123, line 91:
```suggestion
CAPV will create the new disks for control plane nodes during cluster creation.  
```

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
...
+
+CAPV will not be responsible for any custom mounting of the data disks.  CAPV will create the disk and attach to the VM.  Each OS may assign them differently so admin will need to understand the OS image they are using for their template and then configure each node.
+
+#### For User Story 1
+
+CAPV will create the new disks for Control plane nodes during cluster creation.  
+
+The new disks will be placed in the same location as the primary disk (datastore) and will use the controller as the primary disk.  
+
+Creating new controllers is out of scope for the first pass of this enhancement.  The new disk will use the same controller as the primary disks.  The unit number for the disks on that controller will be in the order in which the disks are defined in the machine spec configuration.
```
neolit123, line 95:
```suggestion
Creating new controllers is out of scope for this enhancement.  The new disk will use the same controller as the primary disks.  The unit number for the disks on that controller will be in the order in which the disks are defined in the machine spec configuration.
```

a potential second pass will update where needed.

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
...
+    ...
+    // DataDisks holds information for additional disks to add to the VM that are not part of the VM's OVA template.
+    // +optional
+    DataDisks []VSphereDisk `json:"dataDisks,omitempty"`
+}
+```
+
+### Implementation Details
+
+The cloning process will now be able to add data disks to a VM.  Using the config provided in the VSphereCloneTemplate, the defined disks will be added to the virtual machine.  Each disk is added to the controller used by the primary disk.  Currently, there are no plans to allow the user to define a new controller (scsi/ide/etc) for these disks.
```
neolit123, line 128:
```suggestion
The cloning process will now be able to add data disks to a VM.  Using the config provided in the VSphereCloneTemplate, the defined disks will be added to the virtual machine.  Each disk is added to the controller used by the primary disk.  Currently, there are no plans to allow the user to define a new controller (SCSI/IDE/etc) for these disks. 
```

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
...
+      memoryMiB: 16384
+      network:
+        devices:
+        - dhcp4: true
+          networkName: ci-vlan-1240
+```
+
+In the above example, two data disks will be created during the clone process and placed after the OS disk.  The example shows a 50 GiB and a 90 GiB disk being configured and added.
+
+For each `dataDisks` definition, the clone procedure will attempt to generate a device VirtualDeviceConfigSpec that will be used to create the new disk device.  Each disk will be attached in the order in which they are defined in the template.  All disks defined in the vSphere OVA template will come first with the new disks being attached after.
```
neolit123, line 161: in the spec the user can only pass a size for a disk.
how can the user know which disk is which if the sizes are the same?

does the govmomi API types support names / indices that can be set from the source CAPV type optionally, so that the user can better tract what's going on in logs and API calls?

docs/proposal/20241003-data-disk-support.md
```diff
@@ -0,0 +1,170 @@
...
+
+The clone procedure will assign each new disk to the same controller being used by the primary (OS) disk of the OVA template.  The current behavior of the clone procedure will not be able to create any new controllers; however, in the future, the template may be enhanced to define new controllers and then assign them during the disk creation portion of the clone procedure. 
+
+### Notes/Constraints
+
+## Upgrade Strategy
+
+No Upgrade needed.
+Data disks are optional. Existing clusters do not have additional data disks configured, which will continue to work without changes.  Once upgraded, the new feature will be available for existing machines to be reconfigured / recreate.
```
neolit123, line 170: if the CAPV template doesn't have dataDisks set, but it's added later would the webhook block the edit?
i.e. one needs to create a new template.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -369,6 +371,18 @@ func getDiskSpec(vmCtx *capvcontext.VMContext, devices object.VirtualDeviceList)
 			diskSpecs = append(diskSpecs, additionalDiskConfigSpec)
 		}
 	}
+
+	// Now if we have increased the disk size of any additional disks that were in the template, we can now add new disks
+	// that are present in the additionalDisks list.
```
neolit123, line 376:
```suggestion
	// If there are DataDisk in the VM template create disk specs for them too.
```
seem clearer

but i think this sentence is confusing. please revise it.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -369,6 +371,18 @@ func getDiskSpec(vmCtx *capvcontext.VMContext, devices object.VirtualDeviceList)
...
+	if len(vmCtx.VSphereVM.Spec.DataDisks) > 0 {
+		additionalDisks, diskErr := createDataDisks(ctx, vmCtx.VSphereVM.Spec.DataDisks, primaryDisk, devices)
```
neolit123, line 378:
```suggestion
		additionalDisks, err := createDataDisks(ctx, vmCtx.VSphereVM.Spec.DataDisks, primaryDisk, devices)
```

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -369,6 +371,18 @@ func getDiskSpec(vmCtx *capvcontext.VMContext, devices object.VirtualDeviceList)
...
+		if diskErr != nil {
+			klog.Errorf("Unable to add additional disks: %v", diskErr)
```
neolit123, line 380: don't import klog, use `log := ctrl.LoggerFrom(ctx)` instead in this function.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +404,105 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
 	}, nil
 }
 
+// createDataDisks parses through the list of VSphereDisk objects and generates the VirtualDeviceConfigSpec for each disk needing to be created during clone process.
```
neolit123, line 407:
```suggestion
// createDataDisks parses through the list of VSphereDisk objects and generates the VirtualDeviceConfigSpec for each one.
```

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +404,105 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+func createDataDisks(ctx context.Context, disks []infrav1.VSphereDisk, primaryDisk *types.VirtualDisk, devices object.VirtualDeviceList) ([]types.BaseVirtualDeviceConfigSpec, error) {
+	log := ctrl.LoggerFrom(ctx)
+	additionalDisks := []types.BaseVirtualDeviceConfigSpec{}
+	unit := int32(1)
+
+	for i, dataDisk := range disks {
+		log.Info("Adding disk", "spec", dataDisk)
```
neolit123, line 414: this is an example of what i meant from earlier. seems having some name / id will definitely help especially during debugging.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +404,105 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+// assignUnitNumber assigns a controller unit number to a device.
+func assignUnitNumber(ctx context.Context, device types.BaseVirtualDevice, existingDevices object.VirtualDeviceList, newDevices []types.BaseVirtualDeviceConfigSpec, controller types.BaseVirtualController, offset int32) {
+	log := ctrl.LoggerFrom(ctx)
+	vd := device.GetVirtualDevice()
+	vd.ControllerKey = controller.GetVirtualController().Key
+	vd.UnitNumber = &offset
+
+	units := make([]bool, 30)
+
+	for i := 0; i < int(offset); i++ {
```
neolit123, line 464: what is the significance of 30 here?
offset is of type int32, isn't doing len(offset) returning an error here?
> invalid argument: offset (variable of type int32) for built-in len

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +404,105 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+		log.V(4).Info(fmt.Sprintf("Marking SCSI Controller's unit number: %d", sc.GetVirtualSCSIController().ScsiCtlrUnitNumber))
+		units[sc.GetVirtualSCSIController().ScsiCtlrUnitNumber] = true
+	}
+
+	key := controller.GetVirtualController().Key
+
+	// Check all existing devices
+	for _, device := range existingDevices {
+		d := device.GetVirtualDevice()
+
```
neolit123, line 480:
```suggestion
```

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +404,105 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+		if d.ControllerKey == key && d.UnitNumber != nil {
+			units[int(*d.UnitNumber)] = true
+		}
+	}
+
+	// Check new devices
+	for _, device := range newDevices {
+		d := device.GetVirtualDeviceConfigSpec().Device.GetVirtualDevice()
+
```
neolit123, line 489:
```suggestion
```

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +404,105 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+		log.Info("Generated device", "dev", dev)
+
+		additionalDisks = append(additionalDisks, &diskConfigSpec)
+	}
+
+	return additionalDisks, nil
+}
+
+// assignUnitNumber assigns a controller unit number to a device.
+func assignUnitNumber(ctx context.Context, device types.BaseVirtualDevice, existingDevices object.VirtualDeviceList, newDevices []types.BaseVirtualDeviceConfigSpec, controller types.BaseVirtualController, offset int32) {
```
neolit123, line 456: can this function be unit tested explicitly?

docs/proposal/20241003-data-disk-support.md
vr4manta: Sorry, I wasn't sure the exact process for this.  I can remove this and do separately or just remove entirely.

neolit123: np. yeah, normally the feature author sends the PR with the proposal and maintainers rally on it with comments.
then the implementation PR follows after the design is polished.

but you can move it as the first commit, i guess.

pkg/services/govmomi/vcenter/clone.go
vr4manta: Sure.  I'll work on making some tests for this method.

docs/proposal/20241003-data-disk-support.md
vr4manta: Sounds good.  Thanks!

docs/proposal/20241003-data-disk-support.md
vr4manta: I did not see any way to tag or name a disk from govmomi.  I can look a little more at govc to see what it also does.  I originally had a deviceName field so mentally you can put a name on the disk; however, here internally a lot of people here were hesitant on that idea.  If we can put a name on it, then that would be ideal for sure.

Also for the case where there are multiple disks with the same size, i think one the admin goes to create the machine config or whatever technique they are gonna use to assign the disk to a location, they can refer the disk to its DiskBySequence if preferred.  Depending on OS, these disks can also get mounted in Linux in a way that could potentially change between reboots.  RHEL did that for me, so mounting in my tests I did by `/dev/disk/by-diskseq`.  We are still looking into potentially making a boostrap-controller to handle the mounting / assigning of the disk locations.

docs/proposal/20241003-data-disk-support.md
vr4manta: I was just starting to look at that and forgot to revisit.  In OCP, I have our MAPI operator allowing the change to a machineset which then requires the admin to roll out the changes manually (either delete the machine they wish to remove first, or scale down then up).  I'll look back at this and add a section on this here.

pkg/services/govmomi/vcenter/clone.go
vr4manta: Sure.  The intent was for me to remember that we got to handle the OVA template disks first before doing the additional disks.

docs/proposal/20241003-data-disk-support.md
neolit123: yes, just a Name or an Index field that can better help with tracking.
i'm sure users can track disks with the same size, but having the Name / Index would just make it easier.
so hopefully it's possible somehow.

pkg/services/govmomi/vcenter/clone.go
vr4manta: Sorry, I thought I removed these already.  I had others too, but just missed this one.  Thanks!

pkg/services/govmomi/vcenter/clone.go
vr4manta: Would it make sense to add `deviceName` or `deviceId` as a field for each disk entry?

neolit123: if we can define an optional string name on each disk in CAPV and that can propagate in the govmomi API types that would make the UX a little bit nicer i would argue.
if its not possible i think it's worth printing the index of the disk as defined in the slice of disks ini the CAPV types in all logged entries in controllers.

e.g. `doing X with disk at index Y`

docs/proposal/20241003-data-disk-support.md
vr4manta: I switch commit order

pkg/services/govmomi/vcenter/clone.go
neolit123: they are doing this though:
```
	units := make([]bool, 30)

	for i := 0; i < offset; i++ {
		units[i] = true
	}
```

i think `len(offset)` would trigger a compiler error.

pkg/services/govmomi/vcenter/clone.go
vr4manta: I updated the comment to say 
`Process all DataDisks definitions to dynamically create and add disks to the VM`

pkg/services/govmomi/vcenter/clone.go
vr4manta: Sounds good.  I'll go ahead and add the index output in the log message and will update this later if we find a way to leverage a name.

pkg/services/govmomi/vcenter/clone.go
vr4manta: I think what we are wanting to do here is two things.

1.) create an array that is the max number of units any controller can hold (or maybe the one we are targeting, but not sure how to figure that).  I think I read that vSphere for some controllers max is 16, but I think special cases are much much higher.  We can adjust what the default array size is here to the max of any, but will leave future issues popping up where we now exceeded that controller.
2.) The loop using offset is just gonna prefill the array up to the offset point as true as to prevent us from assigning an index below that for the current disk we are attempting to add / create.  This loop seems to match both govmomi and this function.  The govmomi is doing len(offset) because their function is using int32 as type for offset, but mine is just int.

The main function that calls this one, we default the starting point to 0 (which should always be the primary disk).  After that, its any additional disks in the vSphere OVA template or previously added data disks.  If we have more disks being added then supported, I do want to make sure that something triggers a failure.  We could add a check here to error out if offset is > 30 and state it exceeds the max unit number.

I am happy for any suggestions.  This is probably a weak spot, but for the average user of adding 1-5 extra disks (i'm just guessing 5 is max we want to add for each node for OS usage that is not included w/ any of the PVs) I think 30 is an ok number to default with initially.  What do you think?

neolit123: > create an array that is the max number of units any controller can hold 

aren't there exported constants for that? would the govmomi maintainers know?

i think adding a few comment in the code to clarify the choices of 30 and the rest of the logic will clarify the picture for readers. but we should avoid panics if arrays are accessed out of bounds.

my main point was that the code as is cannot work. please try this:
https://go.dev/play/p/fWud2MEYR0d

neolit123: i think docs.vmware.com might be going down in a few days due to broadcom docs migration, but you can still link that URL and a quote from it. if someone needs the original page they can use web archived version.

vr4manta: ah good point.  I'll see if I can find the broadcom equivelent.  I keep forgetting that things are in flux with these sites.

docs/proposal/20241003-data-disk-support.md
vr4manta: VM.  Let me update that.  Thanks!

docs/proposal/20241003-data-disk-support.md
vr4manta: Ok, i've added a name field to the vsphere disk type.  let me know your thoughts.

pkg/services/govmomi/vcenter/clone.go
vr4manta: I've added some tests around this method.  I think its covering most scenarios.

vr4manta:
@neolit123 @lubronzhan lots of updates are in.  Ready for next review.  Thanks!

Also, e2e is in, but for some reason, it is not working against the e2e cluster in other PR.  When I run these tests locally against my vcenter, it all seems to work.  I cannot access the logs from the e2e run due to what looks like IP / VPN.  Is someone able to check that and help me diagnose it?  Thanks!

apis/v1beta1/types.go
```diff
@@ -203,6 +203,21 @@ type VirtualMachineCloneSpec struct {
 	// Check the compatibility with the ESXi version before setting the value.
 	// +optional
 	HardwareVersion string `json:"hardwareVersion,omitempty"`
+	// DataDisks are additional disks to add to the VM that are not part of the VM's OVA template.
+	// +optional
```
chrischdi, line 207:
```suggestion
	// +optional
	// +listType=map
	// +listMapKey=name
```

chrischdi: I was guessing 29 being a good fit as that's the maximum we currently handle in the code?

Note: we can start lower and increase later, but cannot decrease without be api breaking.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +408,117 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+		}
+		unit = *dev.UnitNumber
+
+		diskConfigSpec := types.VirtualDeviceConfigSpec{
+			Device:        dev,
+			Operation:     types.VirtualDeviceConfigSpecOperationAdd,
+			FileOperation: types.VirtualDeviceConfigSpecFileOperationCreate,
+		}
+
+		log.Info("Generated device", "dev", dev)
```
chrischdi, line 454: Let's set a proper log level, also for the other log entries

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +408,117 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+}
+
+// assignUnitNumber assigns a controller unit number to a device.
+func assignUnitNumber(ctx context.Context, device types.BaseVirtualDevice, existingDevices object.VirtualDeviceList, newDevices []types.BaseVirtualDeviceConfigSpec, controller types.BaseVirtualController, offset int32) error {
+	if offset > maxUnitNumber {
+		return errors.Errorf("unable to assign unit number due to offset %v exceeding max allowed of %v", offset, maxUnitNumber)
+	}
+
+	if controller == nil {
+		return errors.Errorf("unable to assign unit number due to controller parameter being nil")
```
chrischdi, line 469:
```suggestion
		return errors.New("controller parameter cannot be nil")
```

Let's use errors.New also we will use error wrapping above to not repeat too much

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +408,117 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+				},
+				ControllerKey: controller.GetVirtualController().Key,
+			},
+			CapacityInKB: int64(dataDisk.SizeGiB) * 1024 * 1024,
+		}
+
+		// Assign unit number to the next slot on the controller.
+		if err := assignUnitNumber(ctx, dev, devices, additionalDisks, controller, unit); err != nil {
+			log.Error(err, "unable to assign unit number")
+			return nil, err
```
chrischdi, line 444:
```suggestion
			return nil, errors.Wrap(err, "failed to assign device unit number to disk")
```

Errors are logged by controller runtime, having the log here would log it twice.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +408,117 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+		additionalDisks = append(additionalDisks, &diskConfigSpec)
+	}
+
+	return additionalDisks, nil
+}
+
+// assignUnitNumber assigns a controller unit number to a device.
+func assignUnitNumber(ctx context.Context, device types.BaseVirtualDevice, existingDevices object.VirtualDeviceList, newDevices []types.BaseVirtualDeviceConfigSpec, controller types.BaseVirtualController, offset int32) error {
+	if offset > maxUnitNumber {
+		return errors.Errorf("unable to assign unit number due to offset %v exceeding max allowed of %v", offset, maxUnitNumber)
```
chrischdi, line 465:
```suggestion
		return errors.Errorf("%d exceeds maximum number of units %d", offset, maxUnitNumber)
```

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +408,117 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+		if !used {
+			unit32 := int32(unit)
+			vd.UnitNumber = &unit32
+			log.V(4).Info(fmt.Sprintf("Determined next available unit number: %d", unit32))
+			return nil
+		}
+	}
+
+	// If we are here, we did not find a unit number.  Return error.
+	return errors.Errorf("unable to find available unit number")
```
chrischdi, line 519:
```suggestion
	return errors.New("all unit numbers are already in-use")
```

Not sure if that's a good error message though.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -369,6 +376,17 @@ func getDiskSpec(vmCtx *capvcontext.VMContext, devices object.VirtualDeviceList)
 			diskSpecs = append(diskSpecs, additionalDiskConfigSpec)
 		}
 	}
+
+	// Process all DataDisks definitions to dynamically create and add disks to the VM
+	if len(vmCtx.VSphereVM.Spec.DataDisks) > 0 {
+		additionalDisks, err := createDataDisks(ctx, vmCtx.VSphereVM.Spec.DataDisks, primaryDisk, devices)
+		if err != nil {
+			log.Error(err, "Unable to add additional disks.")
```
chrischdi, line 384: same here, returning the error already logs it.
```suggestion
```

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -138,10 +142,11 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
 
 	// Only non-linked clones may expand the size of the template's disk.
 	if snapshotRef == nil {
-		diskSpecs, err := getDiskSpec(vmCtx, devices)
+		diskSpecs, err := getDiskSpec(ctx, vmCtx, devices)
 		if err != nil {
 			return errors.Wrapf(err, "error getting disk spec for %q", ctx)
 		}
+		log.Info("Got the following disks", "disk", diskSpecs)
```
chrischdi, line 149: Also here, use a log level, this otherwise gets verbose

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +408,117 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
 	}, nil
 }
 
+// createDataDisks parses through the list of VSphereDisk objects and generates the VirtualDeviceConfigSpec for each one.
+func createDataDisks(ctx context.Context, disks []infrav1.VSphereDisk, primaryDisk *types.VirtualDisk, devices object.VirtualDeviceList) ([]types.BaseVirtualDeviceConfigSpec, error) {
+	log := ctrl.LoggerFrom(ctx)
+	additionalDisks := []types.BaseVirtualDeviceConfigSpec{}
+	unit := int32(1)
```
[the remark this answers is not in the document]
vr4manta: I'll rename that variable.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +408,117 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+			},
+			CapacityInKB: int64(dataDisk.SizeGiB) * 1024 * 1024,
+		}
+
+		// Assign unit number to the next slot on the controller.
+		if err := assignUnitNumber(ctx, dev, devices, additionalDisks, controller, unit); err != nil {
+			log.Error(err, "unable to assign unit number")
+			return nil, err
+		}
+		unit = *dev.UnitNumber
```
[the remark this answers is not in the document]
vr4manta: Added comment

test/e2e/multi-disk_test.go
```diff
@@ -0,0 +1,121 @@
...
+type DiskSpecInput struct {
+	InfraClients
+	Global    GlobalInput
+	SpecName  string
+	Namespace *corev1.Namespace
+	Template  string
+	ToVersion string
+}
+
+var _ = Describe("Multi-Disk support", func() {
```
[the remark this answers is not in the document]
vr4manta: I would assume it can also run with vsim.  I'll verify that is working.  I'll also update the name.

test/e2e/multi-disk_test.go
```diff
@@ -0,0 +1,121 @@
...
+					Finder:     vsphereFinder,
+				},
+				Global: GlobalInput{
+					BootstrapClusterProxy: bootstrapClusterProxy,
+					ClusterctlConfigPath:  testSpecificSettingsGetter().ClusterctlConfigPath,
+					E2EConfig:             e2eConfig,
+					ArtifactFolder:        artifactFolder,
+				},
+			})
+		})
```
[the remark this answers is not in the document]
vr4manta: I'll see if I can figure it out.  Thanks.

test/e2e/multi-disk_test.go
```diff
@@ -0,0 +1,121 @@
...
+		vmObj, err := input.Finder.VirtualMachine(ctx, vm.Name)
+		Expect(err).NotTo(HaveOccurred())
+
+		devices, err := vmObj.Device(ctx)
+		Expect(err).NotTo(HaveOccurred())
+
+		// We expect control plane VMs to have 3 disks, and the compute VMs will have 2.
+		disks := devices.SelectByType((*types.VirtualDisk)(nil))
+		Expect(disks).To(HaveLen(diskCount), fmt.Sprintf("Disk count of VM should be %d", diskCount))
+	}
```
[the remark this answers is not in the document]
vr4manta: ok

pkg/services/govmomi/vcenter/clone.go
vr4manta: Ok

pkg/services/govmomi/vcenter/clone.go
vr4manta: Ok

pkg/services/govmomi/vcenter/clone.go
vr4manta: Ok, I will update this. Thanks!

pkg/services/govmomi/vcenter/clone.go
vr4manta: I'll update this

pkg/services/govmomi/vcenter/clone.go
vr4manta: Yeah, I would almost think the template image being already full before attempting to add an additional should never happen, but this is a good safety net in case something strange happens.  I was almost leaning towards `no unit numbers are available for use", but if your suggestion makes sense to you, I am ok with it.  I'll change this to the new text unless someone else has a better suggestion.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -138,10 +142,11 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
 
 	// Only non-linked clones may expand the size of the template's disk.
 	if snapshotRef == nil {
-		diskSpecs, err := getDiskSpec(vmCtx, devices)
+		diskSpecs, err := getDiskSpec(ctx, vmCtx, devices)
```
chrischdi, line 145: This only runs on non-linked clones.

But adding disks may be fine for linked clones too?!

Othwerwise: if this only works for non-linked clones, we should catch that in a validation webhook to ensure linkedClone is not set. (.spec.cloneMode)

vr4manta: Refactored code to allow for all clone types.  Seems to be working now for PR e2e as well.  Thanks!

apis/v1beta1/types.go
chrischdi: good point.

> [...] Order is preserved upon merge. [...]

https://github.com/kubernetes/kube-openapi/blob/master/pkg/idl/doc.go#L48-L51

test/e2e/multi-disk_test.go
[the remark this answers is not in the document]
vr4manta: I got this working.  So far so good.

test/e2e/multi-disk_test.go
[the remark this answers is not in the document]
vr4manta: It seems that using finder against vcsim doesn't like finding the VMs for some reason.  The test works fine against real vCenter, but getting a bad error when trying to work against vcsim.  I'm still looking into it, but I may just avoid vcsim for now.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +405,126 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+			unit32 := int32(unit)
+			vd.UnitNumber = &unit32
+			log.V(4).Info(fmt.Sprintf("Determined next available unit number: %d", unit32))
+			return nil
+		}
+	}
+
+	// If we are here, we did not find a unit number.  Return error.
+	return errors.Errorf("all unit numbers are already in-use")
+}
```
chrischdi, line 526: Sorry for the big refactor, but I think this simplifies it a bit.

```suggestion
func createDataDisks(ctx context.Context, dataDiskDefs []infrav1.VSphereDisk, devices object.VirtualDeviceList) ([]types.BaseVirtualDeviceConfigSpec, error) {
	log := ctrl.LoggerFrom(ctx)
	additionalDisks := []types.BaseVirtualDeviceConfigSpec{}

	disks := devices.SelectByType((*types.VirtualDisk)(nil))
	if len(disks) == 0 {
		return nil, errors.Errorf("Invalid disk count: %d", len(disks))
	}

	// There is at least one disk
	primaryDisk := disks[0].(*types.VirtualDisk)

	// Get the controller of the primary disk.
	controller, ok := devices.FindByKey(primaryDisk.ControllerKey).(types.BaseVirtualController)
	if !ok {
		return nil, errors.Errorf("unable to find controller with key=%v", primaryDisk.ControllerKey)
	}
	controllerKey := controller.GetVirtualController().Key

	unitNumberAssigner, err := newUnitNumberAssigner(controller, devices, additionalDisks)
	if err != nil {
		return nil, err
	}

	for i, dataDisk := range dataDiskDefs {
		dev := &types.VirtualDisk{
			VirtualDevice: types.VirtualDevice{
				Key: devices.NewKey() - int32(i),
				Backing: &types.VirtualDiskFlatVer2BackingInfo{
					DiskMode:        string(types.VirtualDiskModePersistent),
					ThinProvisioned: types.NewBool(true),
					VirtualDeviceFileBackingInfo: types.VirtualDeviceFileBackingInfo{
						FileName: "",
					},
				},
				ControllerKey: controller.GetVirtualController().Key,
			},
			CapacityInKB: int64(dataDisk.SizeGiB) * 1024 * 1024,
		}

		vd := dev.GetVirtualDevice()
		vd.ControllerKey = controllerKey
		unitNumber, err := unitNumberAssigner.assign()
		if err != nil {
			return nil, err
		}
		vd.UnitNumber = &unitNumber

		log.V(4).Info("Created device for data disk device", "name", dataDisk.Name, "spec", dataDisk, "device", dev)

		additionalDisks = append(additionalDisks, &types.VirtualDeviceConfigSpec{
			Device:        dev,
			Operation:     types.VirtualDeviceConfigSpecOperationAdd,
			FileOperation: types.VirtualDeviceConfigSpecFileOperationCreate,
		})
	}

	return additionalDisks, nil
}

type unitNumberAssigner struct {
	used   []bool
	offset int32
}

func newUnitNumberAssigner(controller types.BaseVirtualController, existingDevices object.VirtualDeviceList, newDevices []types.BaseVirtualDeviceConfigSpec) (*unitNumberAssigner, error) {
	if controller == nil {
		return nil, errors.New("controller parameter cannot be nil")
	}

	used := make([]bool, maxUnitNumber)

	// SCSIControllers also use a unit.
	if scsiController, ok := controller.(types.BaseVirtualSCSIController); ok {
		used[scsiController.GetVirtualSCSIController().ScsiCtlrUnitNumber] = true
	}

	controllerKey := controller.GetVirtualController().Key

	// Mark all unit numbers of existing devices as used
	for _, device := range existingDevices {
		d := device.GetVirtualDevice()
		if d.ControllerKey == controllerKey && d.UnitNumber != nil {
			used[*d.UnitNumber] = true
		}
	}

	// Mark all unit numbers of new devices as used
	for _, device := range newDevices {
		d := device.GetVirtualDeviceConfigSpec().Device.GetVirtualDevice()
		if d.ControllerKey == controllerKey && d.UnitNumber != nil {
			used[*d.UnitNumber] = true
		}
	}

	// Set offset to 0, it will auto-increment on the first assignment.
	return &unitNumberAssigner{used: used, offset: 0}, nil
}

func (a *unitNumberAssigner) assign() (int32, error) {
	if int(a.offset) > len(a.used) {
		return -1, fmt.Errorf("all unit numbers are already in-use")
	}

	for i, isInUse := range a.used[a.offset:] {
		unit := int32(i) + a.offset
		if !isInUse {
			a.used[unit] = true
			a.offset++
			return unit, nil
		}
	}
	return -1, fmt.Errorf("all unit numbers are already in-use")
}
```

test/e2e/data/infrastructure-vsphere-govmomi/main/multi-disk/data-disks-patch.yaml
```diff
@@ -0,0 +1,25 @@
...
+  name: '${CLUSTER_NAME}'
+  namespace: '${NAMESPACE}'
+spec:
+  template:
+    spec:
+      dataDisks:
+      - name: "etcd"
+        sizeGiB: 10
+      - name: "container-images"
+        sizeGiB: 20
```
chrischdi, line 13: Question: in test these disks won't actually get mounted right? They would just get added to the VM?

If so, let's rename them to something which does not make it seem like this is a full example :-) (just give it some more random names)

chrischdi (commented):
Looks great, I think that's the last round for me :-)

Sorry for the big suggestion, that escalated a bit quickly on my side 😂

test/e2e/multi-disk_test.go
```diff
@@ -0,0 +1,100 @@
+/*
```
chrischdi, line 1: also please rename this file to

`multi_disk_test.go`

test/e2e/data/infrastructure-vsphere-govmomi/main/multi-disk/data-disks-patch.yaml
vr4manta: Sure, I can change the names to generics.

test/e2e/multi-disk_test.go
vr4manta: I totally meant to rename him.  Sorry about that.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -142,9 +146,19 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
 		if err != nil {
 			return errors.Wrapf(err, "error getting disk spec for %q", ctx)
 		}
+		log.V(4).Info("Got the following disks", "disks", diskSpecs)
```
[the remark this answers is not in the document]
vr4manta: I'll move this log statement into the new if statement below this one.  I had added it originally only to log old and new disks.  Since I moved the data disk logic out of this if block, this is really needed below in the other if block.

test/e2e/multi-disk_test.go
```diff
@@ -0,0 +1,100 @@
...
+Copyright 2022 The Kubernetes Authors.
```
neolit123, line 2: 2024 (it's a new file)

test/e2e/multi-disk_test.go
```diff
@@ -0,0 +1,100 @@
...
+	. "sigs.k8s.io/cluster-api/test/framework/ginkgoextensions"
+)
+
+type DiskSpecInput struct {
+	InfraClients
+	Global      GlobalInput
+	SpecName    string
+	Namespace   string
+	ClusterName string
+}
```
neolit123, line 38: can it be local with non exported fields?
```suggestion
type diskSpecInput struct {
	InfraClients
	global      GlobalInput
	specName    string
	namespace   string
	clusterName string
}
```

test/e2e/multi-disk_test.go
vr4manta: I'll update that.

test/e2e/multi-disk_test.go
vr4manta: Sure.  I can make those changes.

vr4manta:
> Big Kudos for the e2e test! That's a very efficient one and also easy to understand!

Thanks!

vr4manta:
> Looks great, I think that's the last round for me :-)
> 
> Sorry for the big suggestion, that escalated a bit quickly on my side 😂

No problem.  I do like making it an object that maintains the state.  I've made the changes and am testing more thoroughly.

pkg/services/govmomi/vcenter/clone_test.go
```diff
@@ -114,32 +121,268 @@ func TestGetDiskSpec(t *testing.T) {
...
+				// Check number of disks present
+				if len(deviceResults) != tc.expectedDiskCount {
+					t.Fatalf("Expected device count to be %v, but found %v", tc.expectedDiskCount, len(deviceResults))
+				}
+			}
+		})
+	}
+}
+
+func TestCreateDiskSpec(t *testing.T) {
```
chrischdi, line 156:
```suggestion
func TestCreateDataDisks_CreateDiskSpec(t *testing.T) {
```

pkg/services/govmomi/vcenter/clone_test.go
```diff
@@ -114,32 +121,268 @@ func TestGetDiskSpec(t *testing.T) {
...
+				// Check number of disks present
+				if len(deviceResults) != len(tc.dataDisks) {
+					t.Fatalf("Expected device count to be %v, but found %v", tc.expectedDiskCount, len(deviceResults))
+				}
+			}
+		})
+	}
+}
+
+func TestAddingDataDisks(t *testing.T) {
```
chrischdi, line 240:
```suggestion
func TestCreateDataDisks_AddingDataDisks(t *testing.T) {
```

pkg/services/govmomi/vcenter/clone_test.go
chrischdi: Note: should we combine this and the next test as both are testing CreateDataDisks?

chrischdi:
I'm good with the change! Let's get this over the finish line :-)

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -390,6 +406,115 @@ func getDiskConfigSpec(disk *types.VirtualDisk, diskCloneCapacityKB int64) (type
...
+	if err != nil {
+		return nil, err
+	}
+
+	for i, dataDisk := range dataDiskDefs {
+		log.V(2).Info("Adding disk", "name", dataDisk.Name, "spec", dataDisk)
+
+		dev := &types.VirtualDisk{
+			VirtualDevice: types.VirtualDevice{
+				Key: devices.NewKey() - int32(i),
```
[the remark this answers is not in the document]
vr4manta: Sure.  You cannot have the keys the same for devices.  If the new disks have same key, the clone will fail.  That utility method always returns the same value for some reason.  GOVC uses this logic when adding a disk, so by subtracting one, we keep it unique until vcenter adds the disk and gives it a real unique key.

pkg/services/govmomi/vcenter/clone_test.go
vr4manta: I merged these tests.  Should be cleaner now.

pkg/services/govmomi/vcenter/clone_test.go
```diff
@@ -121,25 +128,187 @@ func TestGetDiskSpec(t *testing.T) {
...
+				// Check number of disks present
+				if len(deviceResults) != tc.expectedDiskCount {
+					t.Fatalf("Expected device count to be %v, but found %v", tc.expectedDiskCount, len(deviceResults))
+				}
 			}
 		})
 	}
 }
 
+func TestAddingDataDisks(t *testing.T) {
```
chrischdi, line 155:
```suggestion
func TestCreateDataDisks(t *testing.T) {
```

To match the function we add test coverage for

vr4manta: Makes sense.  I renamed the function.

PR #3219, opened 2024-10-08
✨  enables storage policy in failure domain 

RenilRaj-BR:
**What this PR does:**
Currently failure domain doesn't support storage policy and it supports specifying a datastore only. We want to add a feature  to enable specifying a storage policy for the failure domain. This provides flexibility and we don't have to specify the exact datastore name. It can also target multiple datastores when we specify a storage policy instead of a specific datastore


apis/v1alpha3/vspherefailuredomain_types.go
```diff
@@ -78,6 +78,11 @@ type Topology struct {
 	// virtual machine is created/located.
 	// +optional
 	Datastore string `json:"datastore,omitempty"`
+
+	// StoragePolicy is the name of the policy that is used to target a datastore
+	// in which the virtual machine is created/located
+	// +optional
+	StoragePolicy string `json:"storagePolicy,omitempty"`
```
chrischdi, line 85: We should only add this to v1beta1.

controllers/vspherecluster_reconciler.go
```diff
@@ -359,7 +359,12 @@ func (r *clusterReconciler) reconcileDeploymentZones(ctx context.Context, cluste
 }
 
 func (r *clusterReconciler) reconcileClusterModules(ctx context.Context, clusterCtx *capvcontext.ClusterContext) (reconcile.Result, error) {
-	if feature.Gates.Enabled(feature.NodeAntiAffinity) {
+	// We want to enable cluster module only when NodeAntiAffinity flag is true and failure domains are not used.
+	// When failure domain is used and an empty value is passed for data center or the resource pool in a vm template,
+	// cluster module creation will fail as there are no values in dc or resource pool.
+	// These values in the vm template gets replaced at a later stage by the values in the failure domain.
+	// When failure domain is used, we do not need the cluster module
+	if feature.Gates.Enabled(feature.NodeAntiAffinity) && clusterCtx.VSphereCluster.Spec.FailureDomainSelector == nil {
```
chrischdi, line 367: You note:

> [...] During the creation of cluster module, it validates the values in the vm template like data-center, resource-pool etc. When a failure domain is used, these values in the vm template can be empty and it get replaced by the values in the failure domain/deployment zone at a later stage. So, when the value of data-center, resource-pool etc are empty in vm template, the cluster module creation fails and cluster module doesn't make sense when failure domains are being used. To mitigate this issue, we have introduced a check in the vSphere cluster reconciler to enable cluster module creation only when the node anti-affinity is enabled and the failure domains are not being used

Is it **always** the case that cluster modules don't make sense or are there cases where they make sense?

RenilRaj-BR: @chrischdi,
when failure domain is used it is not making sense as the cluster module creation happens much before the vm template override function. The override function will replace the values in vm template with values in the failure domain. So control plane vms can end up in different zones(vSphere cluster - not vSphereCluster CR) with different values for resource pool, vm folder etc whereas cluster module gets created from the template with wrong values for resource pool, folder etc.

When failure domains are not used, cluster module makes sense

chrischdi: But it still makes sense if the different zones are in the same vcenter right?

chrischdi: @RenilRaj-BR it feels like this change is orthogonal to adding support for Storage Policies per Failure Domain.

Please consider to move this to a separate PR/issue so we can discuss on this topic properly.

chrischdi:
CC @lubronzhan , @neolit123 , as this falls into govmomi territory :-)

controllers/vspherecluster_reconciler.go
RenilRaj-BR: sure. I will move it out

apis/v1alpha3/vspherefailuredomain_types.go
RenilRaj-BR: @chrischdi ,
I tried adding the storage policy only in v1beta1, but the conversion-gen fails to generate the `zz_generated.conversion.go` without compilation error due to missing field.
Could you please suggest

internal/webhooks/vspherefailuredomain.go
```diff
@@ -61,6 +61,11 @@ func (webhook *VSphereFailureDomainWebhook) ValidateCreate(_ context.Context, ra
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "Topology", "ComputeCluster"), "cannot be empty if Hosts is not empty"))
 	}
 
+	// We should either pass a datastore or a storage policy, not both at the same time
+	if obj.Spec.Topology.Datastore != "" && obj.Spec.Topology.StoragePolicy != "" {
```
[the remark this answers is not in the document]
lubronzhan: Yeah looks like the priority of these two are handled at the vm creation level https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/a2bac7b1984fac89adc4f8ae54a5aeea5b828c11/pkg/services/govmomi/vcenter/clone.go#L206-L274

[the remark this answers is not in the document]
RenilRaj-BR: removed the validation as it is handled in the vm creation phase

rikatz:
/lgtm

From a provisioning perspective, looks good :) 

Will leave for CAPV approvers to do the final review.

Thanks!

lubronzhan:
LGTM

apis/v1beta1/vspherefailuredomain_types.go
```diff
@@ -85,6 +85,11 @@ type Topology struct {
 	// virtual machine is created/located.
 	// +optional
 	Datastore string `json:"datastore,omitempty"`
+
+	// StoragePolicy is the name of the policy that is used to target a datastore
+	// in which the virtual machine is created/located
```
neolit123, line 90: missing dot at the end.

config/default/crd/bases/infrastructure.cluster.x-k8s.io_vspherefailuredomains.yaml
```diff
@@ -368,6 +368,11 @@ spec:
                     items:
                       type: string
                     type: array
+                  storagePolicy:
+                    description: |-
+                      StoragePolicy is the name of the policy that is used to target a datastore
+                      in which the virtual machine is created/located
```
neolit123, line 374: same

RenilRaj-BR: fixed

apis/v1beta1/vspherefailuredomain_types.go
RenilRaj-BR: fixed

neolit123 (commented):
this is a small change, so please squash the commits to 1.

neolit123 (commented):
/lgtm

rikatz:
/lgtm
Thanks

apis/v1alpha4/zz_generated.conversion.go
```diff
@@ -896,14 +896,10 @@ func autoConvert_v1beta1_Topology_To_v1alpha4_Topology(in *v1beta1.Topology, out
 	out.Hosts = (*FailureDomainHosts)(unsafe.Pointer(in.Hosts))
 	out.Networks = *(*[]string)(unsafe.Pointer(&in.Networks))
 	out.Datastore = in.Datastore
+	// WARNING: in.StoragePolicy requires manual conversion: does not exist in peer-type
```
chrischdi, line 899: Don't we need to do the manual conversion for this? (in both api versions)

In both vspherefailuredomain_conversion.go functions we have to do things similar to:

For the ConvertTo function:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/apis/v1alpha4/vspheremachine_conversion.go#L33-L48

For the ConvertFrom function:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/apis/v1alpha4/vspheremachine_conversion.go#L61-L64

fabriziopandini:
/hold

> Currently failure domain doesn't support storage policy and it supports specifying a datastore only.

It seems that it is not required to have different storagePolicy on each failure domain, and the storagePolicy setting in the vsphere machine template already cover this use case

rikatz:
I have discussed internally with @RenilRaj-BR and we agreed this can be closed for now.

/close

PR #3229, opened 2024-10-21
:sparkles: Allow using moid

rikatz:
**What this PR does / why we need it**: Allow users passing MOID as part of the spec

In some cases, passing the name of an object may not be enough. There may be disambiguation problems on name usage, or special characters that are accepted by vSphere but not properly encoded to be used during VM provisioning.

This change allows users that can rely on passing MOID  (eg.: automation, terraform, vSphere UI) to pass the MOID of an object instead of passing the canonical name

Note: This change doesn't cover yet Network Name and the Failure Domain fields, that can be either covered by this same change, or on a follow up PR.

### DIFF
--- a/pkg/services/govmomi/create_test.go
+++ b/pkg/services/govmomi/create_test.go
@@ -72,7 +72,7 @@ func TestCreate(t *testing.T) {
 
 		vimClient, err := vim25.NewClient(ctx, vmContext.Session.RoundTripper)
 		if err != nil {
-			t.Fatal("could not make vim25 client.")
+			t.Fatalf("could not create vim25 client: %v", err)
 		}
 
 		if replaceFunc != nil {
@@ -89,9 +89,8 @@ func TestCreate(t *testing.T) {
 		}
 
 		task := object.NewTask(vimClient, taskRef)
-		err = task.Wait(ctx)
-		if err != nil {
-			t.Fatal("error waiting for task:", err)
+		if err := task.Wait(ctx); err != nil {
+			t.Fatalf("error waiting for task: %v", err)
 		}
 	}
 
@@ -101,7 +100,7 @@ func TestCreate(t *testing.T) {
 		t.Error("failed to clone vm")
 	}
 
-	t.Log("executing with moid")
+	t.Log("executing with MOID")
 	replaceFuncMOID := func(vmContext *vmcontext.VMContext, vimClient *vim25.Client) {
 		finderClient := find.NewFinder(vimClient)
 		dc, err := finderClient.DatacenterOrDefault(ctx, "")
--- a/pkg/services/govmomi/create_test.go
+++ b/pkg/services/govmomi/create_test.go
@@ -20,17 +20,21 @@ import (
 	"context"
 	"testing"
 
+	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/simulator"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/types"
 
 	"sigs.k8s.io/cluster-api-provider-vsphere/internal/test/helpers/vcsim"
+	vmcontext "sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context/fake"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/session"
 )
 
 func TestCreate(t *testing.T) {
+	ctx := context.Background()
+
 	model := simulator.VPX()
 	model.Host = 0 // ClusterHost only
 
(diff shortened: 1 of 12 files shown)


pkg/session/session.go
```diff
@@ -206,7 +206,14 @@ func GetOrCreate(ctx context.Context, params *Params) (*Session, error) {
 
 	// Assign the datacenter if one was specified.
 	if params.datacenter != "" {
-		dc, err := session.Finder.Datacenter(ctx, params.datacenter)
+		var dc *object.Datacenter
+		dcRef := object.ReferenceFromString(params.datacenter)
+		if dcRef != nil {
+			dc = object.NewDatacenter(client.Client, *dcRef)
+			_, err = dc.ObjectName(ctx)
```
neolit123, line 213: should it be verified if the returned obj name matches the expected one?

pkg/session/session.go
```diff
@@ -218,6 +225,7 @@ func GetOrCreate(ctx context.Context, params *Params) (*Session, error) {
 			}
 			return nil, errors.Wrapf(err, "failed to create vCenter session: failed to find datacenter %q", params.datacenter)
 		}
+
```
neolit123, line 228:
```suggestion
```

neolit123 (commented):
> In some cases, passing the name of an object may not be enough. There may be disambiguation problems on name usage, or special characters that are accepted by vSphere but not properly encoded to be used during VM provisioning.

> This change allows users that can rely on passing MOID (eg.: automation, terraform, vSphere UI) to pass the MOID of an object instead of passing the canonical name

could you provide an example of users passing a name that doesn't work and how they would pass MOID instead?

should API field docs be updated to reflect that?

> Note: This change doesn't cover yet Network Name and the Failure Domain fields, that can be either covered by this same change, or on a follow up PR.

best to include it in the same PR, IMO.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -118,14 +118,27 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
...
-	folder, err := vmCtx.Session.Finder.FolderOrDefault(ctx, vmCtx.VSphereVM.Spec.Folder)
-	if err != nil {
-		return errors.Wrapf(err, "unable to get folder for %q", ctx)
+	var folder *object.Folder
+
+	folderRef := object.ReferenceFromString(vmCtx.VSphereVM.Spec.Folder)
+	if folderRef != nil {
+		folder = object.NewFolder(vmCtx.Session.Client.Client, *folderRef)
+	} else {
+		folder, err = vmCtx.Session.Finder.FolderOrDefault(ctx, vmCtx.VSphereVM.Spec.Folder)
```
viveksyngh, line 127: Should we also add status field and update the names of folders, resource pool  etc which MOIDs  refers to ?

This should be very helpful someone looking at the object to figure out where in VC these resources are created.

rikatz: I honestly would avoid changing the API right now, but I am happy to do it if we think it really adds more value.

If so, I think the best place would be on vSphereVM which IIRC is considered an "internal API" (like intermediate between VsphereMachine and vSphere calls)

pkg/session/session.go
rikatz: Actually the idea here was less to check the DC name, but instead force a vsphere API call to happen to guarantee that the passed MOID ref for the datacenter is valid.

We cannot match the expected because in this case we don't know the expected DC name, (eg RicardoDC), but the MOID. What I can do anyway is maybe add an info log here saying the Ref was resolved to DC name returned by that ObjectName call, wdyt?

neolit123: +1 for adding the log

pkg/services/govmomi/vcenter/clone.go
neolit123: like i said on this review summary comment, we should probably update API field go doc comments?
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3229#pullrequestreview-2384048320

rikatz:
I will add the api docs as well, didn't finished taking care of all the comments yet :)

pkg/session/session.go
lubronzhan: + 1. Debug level log would be helpful, whether it's using the string as MOID or real name.

fabriziopandini:
I will try to get to this PR soon, but currently trying to finish some work in CAPI first, but in the meantime it will be great if we can add an E2E test to this PR.

e.g. we can take [this E2E test](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/ba821a54fd8b8dfb7aa2436ae916aa330950cce6/test/e2e/quick_start_test.go#L34-L49), duplicate it, write code that resolve names in env vars into Moid, then sets env vars with the Moid so the test will use them

pkg/services/govmomi/util.go
```diff
@@ -83,6 +83,22 @@ func findVM(ctx context.Context, vmCtx *capvcontext.VMContext) (types.ManagedObj
 		return types.ManagedObjectReference{}, err
 	}
 	if objRef == nil {
+		folderRef := object.ReferenceFromString(vmCtx.VSphereVM.Spec.Folder)
+		if folderRef != nil {
+			// We should find the VM inside this vmfolder
+			folder := object.NewFolder(vmCtx.Session.Client.Client, *folderRef)
+			objs, err := folder.Children(ctx)
+			if err != nil {
+				return types.ManagedObjectReference{}, err
```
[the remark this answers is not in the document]
rikatz: changed to "log.Error(err, "could not enumerate children for folder", "folder", vmCtx.VSphereVM.Spec.Folder)" and applied here

pkg/services/govmomi/util.go
```diff
@@ -83,6 +83,22 @@ func findVM(ctx context.Context, vmCtx *capvcontext.VMContext) (types.ManagedObj
...
+			for _, obj := range objs {
+				if vm, ok := obj.(*object.VirtualMachine); ok && vm.Name() == vmCtx.VSphereVM.Name {
+					log.Info("VM found by moid folder", "vmRef", vm.Reference())
+					return vm.Reference(), nil
+				}
+			}
+			return types.ManagedObjectReference{}, errNotFound{byInventoryPath: folder.Reference().Value}
+		}
 		// fallback to use inventory paths
 		folder, err := vmCtx.Session.Finder.FolderOrDefault(ctx, vmCtx.VSphereVM.Spec.Folder)
```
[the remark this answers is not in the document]
rikatz: no no, actually what happens before is:

```golang
folderRef := object.ReferenceFromString(vmCtx.VSphereVM.Spec.Folder)
		if folderRef != nil {
```

Which means "if user passed a valid moRef (so it is not null) then follow the moref path, otherwise it fallbacks to the named folder.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -118,14 +118,27 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
...
+		if err != nil {
+			return errors.Wrapf(err, "unable to get folder for %q", ctx)
```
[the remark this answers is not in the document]
rikatz: RA! It came from the old code, and I guess no one (including me) saw it! nice catch.

pkg/services/govmomi/vcenter/clone.go
rikatz: updated. My concern is not about the API field doc (sorry) but adding new status fields.

rikatz:
@fabriziopandini just an update. Setting the e2e here is showing to be harder than I though, because in "theory" my e2e suite cannot access vcsim, which is running as a Pod on my cluster.

I am not sure we are expecting this to be possible or not, but in this case I will need to, instead of making my e2e test go to vcsim and get the MOIDs, make the vcsim envvar controller to add on its status the moid instead of the name.

wdyt? Which one should I follow?

rikatz:
hum, weird this error, it passes locally. Will have to check the issue

rikatz:
/hold

I want to wait for https://github.com/vmware/govmomi/pull/3609 to be merged and a new release made, so instead we can bump govmomi here and rely on the new behavior of being able to search by name or moid directly from the library

rikatz:
This is validated on my tests, works fine :) 

@sbueringer @fabriziopandini @chrischdi for final review :) 

Thanks!

pkg/services/govmomi/create_test.go
```diff
@@ -39,51 +43,155 @@ func TestCreate(t *testing.T) {
...
+		taskRef := types.ManagedObjectReference{
+			Type:  morefTypeTask,
+			Value: vmContext.VSphereVM.Status.TaskRef,
+		}
+
+		task := object.NewTask(vimClient, taskRef)
+		err = task.Wait(ctx)
+		if err != nil {
+			t.Fatal("error waiting for task:", err)
+		}
```
neolit123, line 95:
```suggestion
		if err := task.Wait(ctx); err != nil {
			t.Fatalf("error waiting for task: %v", err)
		}
```

pkg/services/govmomi/create_test.go
```diff
@@ -39,51 +43,155 @@ func TestCreate(t *testing.T) {
...
+	t.Log("executing with canonical name")
+	executeTest("canonical", nil)
+	if model.Machine+1 != model.Count().Machine {
+		t.Error("failed to clone vm")
 	}
-	vimClient, err := vim25.NewClient(ctx, vmContext.Session.RoundTripper)
-	if err != nil {
-		t.Fatal("could not make vim25 client.")
+
+	t.Log("executing with moid")
```
neolit123, line 104:
```suggestion
	t.Log("executing with MOID")
```

pkg/services/govmomi/create_test.go
```diff
@@ -39,51 +43,155 @@ func TestCreate(t *testing.T) {
...
+		vmContext.Session = authSession
+
+		vmContext.VSphereVM.Spec.Template = vm.Name
+
+		disk := object.VirtualDeviceList(vm.Config.Hardware.Device).SelectByType((*types.VirtualDisk)(nil))[0].(*types.VirtualDisk)
+		disk.CapacityInKB = int64(vmContext.VSphereVM.Spec.DiskGiB) * 1024 * 1024
+
+		vimClient, err := vim25.NewClient(ctx, vmContext.Session.RoundTripper)
+		if err != nil {
+			t.Fatal("could not make vim25 client.")
```
neolit123, line 75:
```suggestion
			t.Fatalf("could not create vim25 client: %v", err)
```

test/e2e/e2e_setup_test.go
```diff
@@ -82,6 +83,14 @@ func WithAdditionalVCSimServer(t bool) SetupOption {
 	}
 }
 
+// WithMOID instructs Setup to use the MOID of resources instead of
+// the real name.
```
neolit123, line 87:
```suggestion
// WithMOID instructs Setup to use the MOID of resources instead of the real name.
```

test/infrastructure/vcsim/api/v1alpha1/envvar_types.go
```diff
@@ -38,6 +38,9 @@ type EnvVarSpec struct {
 	// - the system automatically picks the first VirtualMachine class defined in the VMOperatorDependencies
 	// - the system automatically picks the first Image from the content library defined in the VMOperatorDependencies
 	VMOperatorDependencies *NamespacedRef `json:"vmOperatorDependencies,omitempty"`
+
+	// UseMOID defines if envvar status should be filled with the name or MOID of objects
```
neolit123, line 42:
```suggestion
	// UseMOID defines if envvar status should be filled with the name or MOID of objects.
```

test/infrastructure/vcsim/controllers/envvar_controller.go
```diff
@@ -287,14 +293,74 @@ func clusterEnvVarSpecGovmomiVariables(c *vcsimv1.ClusterEnvVarSpec) map[string]
 	}
 
 	// NOTE: omitting cluster Name intentionally because E2E tests provide this value in other ways
-	vars["VSPHERE_DATACENTER"] = vcsimhelpers.DatacenterName(datacenter)
+	// Vars below can be named only so they should be used as is
```
neolit123, line 296: named -> names?
as is -> as they are
potentially reword this sentence for better clarity

test/infrastructure/vcsim/controllers/envvar_controller.go
```diff
@@ -287,14 +293,74 @@ func clusterEnvVarSpecGovmomiVariables(c *vcsimv1.ClusterEnvVarSpec) map[string]
...
-	return vars
+
+	if !moid {
+		vars["VSPHERE_DATACENTER"] = vcsimhelpers.DatacenterName(datacenter)
+		vars["VSPHERE_DATASTORE"] = vcsimhelpers.DatastoreName(datastore)
+		vars["VSPHERE_FOLDER"] = vcsimhelpers.VMFolderName(datacenter)
+		vars["VSPHERE_RESOURCE_POOL"] = vcsimhelpers.ResourcePoolPath(datacenter, cluster)
+		vars["VSPHERE_NETWORK"] = vcsimhelpers.NetworkPath(datacenter, vcsimhelpers.DefaultNetworkName)
+		vars["VSPHERE_TEMPLATE"] = vcsimhelpers.VMPath(datacenter, template)
+
```
neolit123, line 306:
```suggestion
```

test/infrastructure/vcsim/controllers/envvar_controller.go
```diff
@@ -287,14 +293,74 @@ func clusterEnvVarSpecGovmomiVariables(c *vcsimv1.ClusterEnvVarSpec) map[string]
...
+	}
+
+	var dc *object.Datacenter
+	dcRef := object.ReferenceFromString(vcsimhelpers.DatacenterName(datacenter))
+	if dcRef != nil {
+		dc = object.NewDatacenter(s.Client.Client, *dcRef)
+	} else {
+		dc, err = s.Finder.Datacenter(ctx, vcsimhelpers.DatacenterName(datacenter))
+		if err != nil {
+			return vars, fmt.Errorf("failed to locate datacenter reference: %w", err)
```
neolit123, line 327:
```suggestion
			return vars, errors.Errorf("failed to locate datacenter reference: %v", err)
```

test/infrastructure/vcsim/controllers/envvar_controller.go
```diff
@@ -287,14 +293,74 @@ func clusterEnvVarSpecGovmomiVariables(c *vcsimv1.ClusterEnvVarSpec) map[string]
...
+		}
+	}
+	vars["VSPHERE_DATACENTER"] = dc.Reference().String()
+	s.Finder.SetDatacenter(dc)
+
+	dsRef, err := s.Finder.Datastore(ctx, vcsimhelpers.DatastoreName(datastore))
+	if err != nil {
+		return vars, fmt.Errorf("failed to locate datastore reference: %w", err)
```
neolit123, line 335: for consistency, please use pkg/errors with %v instead of introducing fmt.Error with %w
applies to more places below...

test/infrastructure/vcsim/controllers/envvar_controller.go
rikatz: reworded to `	// The var VSPHERE_COMPUTE_CLUSTER cannot be used as moid during the tests.`

test/infrastructure/vcsim/controllers/envvar_controller.go
```diff
@@ -287,14 +293,73 @@ func clusterEnvVarSpecGovmomiVariables(c *vcsimv1.ClusterEnvVarSpec) map[string]
...
+		WithThumbprint(existing["VSPHERE_TLS_THUMBPRINT"]).
+		WithUserInfo(existing["VSPHERE_USERNAME"], existing["VSPHERE_PASSWORD"])
+
+	s, err := session.GetOrCreate(ctx, params)
+	if err != nil {
+		return vars, err
+	}
+
+	var dc *object.Datacenter
+	dcRef := object.ReferenceFromString(vcsimhelpers.DatacenterName(datacenter))
```
rikatz, line 320: Rolled back this as govmomi now supports searching by moid

lubronzhan (approved):

test/e2e/quick_start_test.go
```diff
@@ -48,6 +48,23 @@ var _ = Describe("Cluster Creation using Cluster API quick-start test [vcsim] [s
 	})
 })
 
+var _ = Describe("Cluster Creation using Cluster API quick-start test and MOID [vcsim]", func() {
```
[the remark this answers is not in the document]
sbueringer: I think we can ensure that via
```go
	if testTarget != VCSimTestTarget || testMode != GovmomiTestMode {
		Skip("This test is only implemented for vcsim+govmomi")
		return
	}
```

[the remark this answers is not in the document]
rikatz: I think it is fine to run for both, as during the test itself we do assertions to convert the requested resources to MOID and replace them. But I can change this way if you prefer as well :)

test/infrastructure/vcsim/controllers/envvar_controller.go
```diff
@@ -287,14 +292,68 @@ func clusterEnvVarSpecGovmomiVariables(c *vcsimv1.ClusterEnvVarSpec) map[string]
...
+	if err != nil {
+		return vars, errors.Errorf("error creating test session: %v", err)
+	}
+
+	dc, err := s.Finder.Datacenter(ctx, vcsimhelpers.DatacenterName(datacenter))
+	if err != nil {
+		return vars, errors.Errorf("failed to locate datacenter reference: %v", err)
+	}
+
+	vars["VSPHERE_DATACENTER"] = dc.Reference().String()
```
rikatz, line 323: copying @dougm comment here:
The changed API doc says "name, inventory path or the MOID". But Reference().String() here (and so on below) is the MOR (Managed Object Reference). And Reference().Value is the MOID. The govmomi Finder changes support both, see: https://github.com/vmware/govmomi/blob/a2c22f0e33da798414846a99fd2f55522a405b13/object/common.go#L160-L166

test/e2e/quick_start_test.go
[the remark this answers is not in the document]
sbueringer: Fine for me to keep it as is if it makes sense in general to run this test for regula govmomi Ijust wasn't sure of that and if you're aware that this test is run there as well)

sbueringer:
Thank you very much!!

Feel free to merge once tests are green

/lgtm
/approve
/hold

dougm (commented):
lgtm

rikatz:
All tests passing, removing hold

Thanks all
/hold cancel

PR #3235, opened 2024-10-24
:seedling: Fix return messages to contain proper VM information

rikatz:
During the "context" usage refactoring, we have missed the error return messages/fields.

This way the return of an error message was containing the "context" field, and not the real "Virtual Machine context" (which is a stringer).


neolit123 (commented):
/lgtm

thanks
yes, i saw this was a problem from your other PR.
printing the stdlib context doesn't do much.

sbueringer:
Thx!

/lgtm
/approve

PR #3246, opened 2024-10-31
:bug: Proper ignore IPv6 announcements when IPv6 is not requested

rikatz:

On some networks IPv6 is announced via RA. Networkd (and maybe netplan) separates DHCPv6 and RA enablement, meaning that even when users set to ignore/disable DHCPv6, IPv6 is still configured on the Host VM if there is some Route Advertisement for IPv6. 

Turns out on some setups this may break the desired behavior, as DNSv6 and others may also end up being configured on the node using IPv6.

This PR intends to do the following changes:
* Organize cloud-init template to separate DHCPv, DHCPv6 and override flags as they are independent
* Configure cloud-init to pass accept-ra. This enforcement needs cloud-init on at least version 24.2 (see https://github.com/canonical/cloud-init/pull/4928 and https://github.com/canonical/cloud-init/pull/5060)

### DIFF
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -36,8 +36,9 @@ network:
       set-name: "eth{{ $i }}"
       {{- end }}
       wakeonlan: true
-      {{- if or $net.DHCP4 $net.DHCP6 }}
       dhcp4: {{ $net.DHCP4 }}
+      dhcp6: {{ $net.DHCP6 }}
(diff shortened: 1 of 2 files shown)


rikatz:
https://github.com/team-cluster-api/cluster-api-provider-vsphere/pull/35 this is passing :)

sbueringer:
Thx!

/lgtm
/approve

PR #3286, opened 2024-12-08
feat: add VM naming strategy for VMs in govami mode

viveksyngh:
- Add VM naming stragey for vsphere vms and VMs in govami mode

Fixes #


neolit123 (commented):
please give a ping once this is no longer a draft and is ready for review.

/cc

neolit123:
make sure you add a test.

apis/v1beta1/vspheremachine_types.go
```diff
@@ -68,6 +68,30 @@ type VSphereMachineSpec struct {
 	//
 	// +optional
 	GuestSoftPowerOffTimeout *metav1.Duration `json:"guestSoftPowerOffTimeout,omitempty"`
+
+	// NamingStrategy allows configuring the naming strategy used when calculating the name of the VSphereVM.
```
neolit123, line 72: calculating -> preparing
as it's not a math calculation.

> name of the VSphereVM.

this field is inside VSphereMachineSpec not  VSphereVMSpec

apis/v1beta1/vspheremachine_types.go
```diff
@@ -68,6 +68,30 @@ type VSphereMachineSpec struct {
...
+	// +optional
+	NamingStrategy *VSphereVMNamingStrategy `json:"namingStrategy,omitempty"`
+}
+
+// VSphereVMNamingStrategy defines the naming strategy for the VSphereVMs.
+type VSphereVMNamingStrategy struct {
```
neolit123, line 78: should this be named
VSphereMachineNamingStrategy

apis/v1beta1/vspheremachine_types.go
```diff
@@ -68,6 +68,30 @@ type VSphereMachineSpec struct {
...
+	// * `trunc`: truncates a string, e.g. `trunc 2 "hello"` or `trunc -2 "hello"`
+	// Notes:
+	// * While the template offers some flexibility, we would like the name to link to the Machine name
+	//   to ensure better user experience when troubleshooting
+	// * Generated names must be valid Kubernetes names as they are used to create a VSphereVM object
+	//   and usually also as the name of the Node object.
+	// * Names are automatically truncated at 63 characters. Please note that this can lead to name conflicts,
+	//   so we highly recommend to use a template which leads to a name shorter than 63 characters.
+	// +optional
+	Template *string `json:"template,omitempty"`
```
neolit123, line 94: no need for *string, just string?

apis/v1beta1/vspheremachine_types.go
```diff
@@ -68,6 +68,30 @@ type VSphereMachineSpec struct {
...
+	// Template defines the template to use for generating the name of the VSphereVM object.
+	// If not defined, it will fall back to `{{ .machine.name }}`.
+	// The templating has the following data available:
+	// * `.machine.name`: The name of the Machine object.
+	// The templating also has the following funcs available:
+	// * `trimSuffix`: same as strings.TrimSuffix
+	// * `trunc`: truncates a string, e.g. `trunc 2 "hello"` or `trunc -2 "hello"`
+	// Notes:
+	// * While the template offers some flexibility, we would like the name to link to the Machine name
+	//   to ensure better user experience when troubleshooting
```
neolit123, line 88:
```suggestion
	// * While the template offers some flexibility, it is best for the name to link to the Machine name
	//   to ensure better user experience when troubleshooting
```

personalization should not be present in official docs (ideally).

internal/webhooks/vspheremachinetemplate.go
```diff
@@ -111,10 +117,45 @@ func (webhook *VSphereMachineTemplateWebhook) ValidateUpdate(ctx context.Context
...
+	}
 	return nil, AggregateObjErrors(newTyped.GroupVersionKind().GroupKind(), newTyped.Name, allErrs)
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
 func (webhook *VSphereMachineTemplateWebhook) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
 	return nil, nil
 }
+
+func validateVSphereVMNamingTemplate(_ context.Context, vsphereMachineTemplate *infrav1.VSphereMachineTemplate) field.ErrorList {
```
neolit123, line 133: is this following some code that core CAPI already has or is this done from scratch as a use case for CAPV?
if the former, could you link to that code for reviewer's reference?

pkg/services/vimmachine.go
```diff
@@ -393,12 +410,72 @@ func (v *VimMachineService) createOrPatchVSphereVM(ctx context.Context, vimMachi
...
 	}
-	return machineName
+	return name, nil
+}
+
+const (
+	maxNameLength = 63
+)
+
+// Note: Inlining these functions from sprig to avoid introducing a dependency.
```
neolit123, line 429: sprig is a typo?

test/e2e/data/infrastructure-vsphere-govmomi/main/clusterclass/patch-vm-namingstrategy.yaml
```diff
@@ -0,0 +1,4 @@
+- op: add
+  path: /spec/template/spec/namingStrategy
+  value:
+    template: '{{ if le (len .machine.name) 20 }}{{ .machine.name }}{{else}}{{ trimSuffix "-" (trunc 14 .machine.name) }}-{{ trunc -5 .machine.name }}{{end}}'
```
neolit123, line 4: add missing new line at the end.

internal/webhooks/vspheremachinetemplate.go
viveksyngh: This is based on this MR which was added to support this in supervisor mode https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3099

In supervisor mode, the object is called VirtualMachine which drives the creation of vSphere VMs while in govami mode this object is called VSphereVM.

viveksyngh:
This MR is more or less similar code from this MR which added the functionality in the supervisor mode. https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3099 

I can make the requested changes but I have tried to align this as much as possible with supervisor mode changes.

internal/webhooks/vspheremachinetemplate.go
neolit123: looks like a lot comes from that PR.
why not also name it VirtualMachineNamingStrategy here? that seems more consistent despite the diff backend objects.

neolit123:
> This MR is more or less similar code from this MR which added the functionality in the supervisor mode. #3099
> 
> I can make the requested changes but I have tried to align this as much as possible with supervisor mode changes.

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3286#discussion_r1886744212

internal/webhooks/vspheremachinetemplate.go
viveksyngh: I can do that if we prefer that. I just to wanted to call it what it is being used for.

neolit123: that would be my preference. not sure what other reviewers will comment on this.

apis/v1beta1/vspheremachine_types.go
neolit123: EDIT: seems consistent with the supervisor PR. so you can leave it as is.

pkg/services/vimmachine.go
chrischdi: Its related to https://masterminds.github.io/sprig/

neolit123: you can just copy code from another project without attribution. also CNCF has special handling of MIT and some other licenses. see https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md#cncf-allowlist-license-policy

please add the entire dependency and let the go compiler optimize.

neolit123: /hold

chrischdi: Note: we don't want to simply import the whole sprig library and use the function from there to prevent the transitive dependencies. Instead we should better outline the license part here and propably move the inlined functions to a separate package.

neolit123: if importing is not an option, you could fork under third party and place the respective license files:
> It is either (A) stored unmodified in a designated third-party folder

but it needs to align with the options here:
https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md#cncf-allowlist-license-policy

third_part/forked/sprig:
- LICENSE
- some-code.go

example in k/k
https://github.com/kubernetes/kubernetes/tree/master/third_party/forked/shell2junit

sbueringer: Q: Where is the line between "this is a copy of 1 line of code" and "we implemented the same functionality but there is simply no other way to do it, so the implementation is identical"?

sbueringer: Frankly, before we copy over the entire sprig library I would prefer one of:
* let's import sprig instead of copying it
* let's implement whatever we actually need in a different way than sprig

neolit123: given, it will be a fork. the project can copy only the bits that it needs. a readme or just a code comment clarifing purpose and origin of the forked bits would suffice.

viveksyngh: This is already in the main and we are using it. https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/pkg/services/vmoperator/vmopmachine.go#L298 

I could not reuse it because they were ware private and not exported. Depending what we decide, I can make the change.

neolit123: > This is already in the main and we are using it. https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/pkg/services/vmoperator/vmopmachine.go#L298
> 
> I could not reuse it because they were ware private and not exported. Depending what we decide, I can make the change.

all copied code from a different license without attribution must be removed.
import the library and use it's exported functions.
if what you need from the library is not exported, go with the minimal fork approach.

viveksyngh: I have update the MR where I imported the function sprig library and use only those functions which are required. 

Please take a look.

apis/v1beta1/vspheremachine_types.go
chrischdi: +1, as this struct is used at the `VSphereMachine`

go.mod
```diff
@@ -45,6 +46,9 @@ require (
 )
 
 require (
+	dario.cat/mergo v1.0.1 // indirect
+	github.com/Masterminds/goutils v1.1.1 // indirect
+	github.com/Masterminds/semver/v3 v3.3.0 // indirect
```
chrischdi, line 51: These dependencies are something we might want to prevent having.

go.mod
```diff
@@ -58,8 +62,13 @@ require (
 	github.com/google/gnostic-models v0.6.8 // indirect
 	github.com/google/pprof v0.0.0-20241210010833-40e02aabc2ad // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.20.0 // indirect
+	github.com/huandu/xstrings v1.5.0 // indirect
+	github.com/mitchellh/copystructure v1.2.0 // indirect
+	github.com/mitchellh/reflectwalk v1.0.2 // indirect
```
chrischdi, line 67: These dependencies are even archived and not maintained anymore.

chrischdi: cc @sbueringer , not sure if we should instead do the third_party approach with "minimal" change 🤔

internal/webhooks/vspheremachinetemplate.go
chrischdi: 👍 to keep it as VSphereVM, because that's the object which gets created from a VSphereMachine.

internal/webhooks/vspheremachinetemplate_test.go
```diff
@@ -37,70 +37,96 @@ func TestVSphereMachineTemplate_ValidateCreate(t *testing.T) {
...
 		},
 		{
 			name:           "successful VSphereMachine creation with hardware version set",
-			vsphereMachine: createVSphereMachineTemplate("foo.com", "vmx-17", nil, "", []string{}, nil),
+			vsphereMachine: createVSphereMachineTemplate("foo.com", "vmx-17", nil, "", []string{}, nil, nil),
+		},
+		{
+			name:           "successful VSphereMachineTemplate creation with namingStrategy not set",
+			vsphereMachine: createVSphereMachineTemplate("foo.com", "vmx-17", nil, "", []string{}, nil, nil),
+		},
```
chrischdi, line 108:
```suggestion
```

Duplicate test case to above, let's adjust the name above though.

pkg/services/vimmachine.go
```diff
@@ -215,9 +223,14 @@ func (v *VimMachineService) GetHostInfo(ctx context.Context, machineCtx capvcont
...
 	vm := &infrav1.VSphereVM{}
+	name, err := generateVMObjectName(vimMachineCtx, vimMachineCtx.Machine.Name)
+	if err != nil {
+		return nil, err
+	}
+
 	vmKey := types.NamespacedName{
 		Namespace: vimMachineCtx.VSphereMachine.Namespace,
-		Name:      generateVMObjectName(vimMachineCtx, vimMachineCtx.Machine.Name),
+		Name:      name,
```
chrischdi, line 233: This get lookup will never work with a generated name containing `{{.random}}` because the name never matches now.

So for me it does not make sense to use the new `generateVMObjectName` here inside `findVSphereVM`.

go.mod
viveksyngh: I think these came up while importing sprig library.

pkg/services/vimmachine.go
chrischdi: We may need to change findVSphereVM to find the VSphereVM by OnwerReferences instead?!

This e.g. breaks the deletion flow in some way.

chrischdi (commented):
/hold due to the breakage of findVSphereVM

go.mod
chrischdi: xref https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3286#discussion_r1916210331

pkg/services/vimmachine.go
viveksyngh: We are not supporting {{.random}} function .. we only support `trunc` and `trimSuffix`

chrischdi: ah then it makes sense 😂 , sorry context switching to a similar topic which includes the random part.

pkg/services/vimmachine.go
```diff
@@ -393,12 +411,63 @@ func (v *VimMachineService) createOrPatchVSphereVM(ctx context.Context, vimMachi
...
+
+var nameTemplateFuncs = map[string]any{
+	"trimSuffix": sprig.GenericFuncMap()["trimSuffix"],
+	"trunc":      sprig.GenericFuncMap()["trunc"],
+}
+
+var nameTpl = template.New("name generator").Funcs(nameTemplateFuncs).Option("missingkey=error")
+
+// GenerateVSphereVMName generates the name of a VSphereVM based on the naming strategy.
+func GenerateVSphereVMName(machineName string, namingStrategy *infrav1.VSphereVMNamingStrategy) (string, error) {
```
chrischdi, line 438: This function seems to be the same as we have for supervisor?

Let's move this to `pkg/util` to a separate go file and use it at both places.

chrischdi:
/hold cancel

As above noted https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3286#discussion_r1916449487

apis/v1beta1/vspheremachine_types.go
viveksyngh: I can update that but I did this to avoid confusion when someone looking at the code later .. this is used to decide the strategy for vspherevms objects.

pkg/services/vimmachine.go
viveksyngh: I kept different because we have different struct for each mode. 

If we want to combine, should we also use a common struct for both then it will be make the pkg/util function more simpler otherwise we might need to do type conversion or use an interface

chrischdi: Let's not use the same struct, as these are API types and we need to be very careful about them.

Instead use a common function which takes a `*string` as template (so we only have to check that the template is not nil).

test/e2e/data/infrastructure-vsphere-govmomi/main/clusterclass/patch-vm-namingstrategy.yaml
chrischdi: Open todo?

viveksyngh: I think I already did update it.

apis/v1beta1/vspheremachine_types.go
chrischdi: Let's keep as is to be equivalent to supervisor.

test/e2e/data/infrastructure-vsphere-govmomi/main/clusterclass/patch-vm-namingstrategy.yaml
chrischdi: image

This shows there's no linebreak at the end :-)

pkg/util/template.go
```diff
@@ -0,0 +1,75 @@
...
+
+	data := map[string]interface{}{
+		"machine": map[string]interface{}{
+			"name": machineName,
+		},
+	}
+
+	tpl, err := nameTpl.Parse(*nameTemplate)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to generate name: failed to parse template %q", *nameTemplate)
```
chrischdi, line 58:
```suggestion
		return "", errors.Wrapf(err, "unable to parse template %q", *nameTemplate)
```

(the `failed to generate name for VSphereVM / VirtualMachine` comes at above error wraping)

pkg/util/template.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	var buf bytes.Buffer
+	if err := tpl.Execute(&buf, data); err != nil {
+		return "", errors.Wrap(err, "failed to generate name")
```
chrischdi, line 63:
```suggestion
		return "", err
```

pkg/util/template_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	gomegatypes "github.com/onsi/gomega/types"
+	"k8s.io/utils/ptr"
+)
+
+func Test_GenerateVSphereVMName(t *testing.T) {
```
chrischdi, line 27:
```suggestion
func Test_GenerateMachineNameFromTemplate(t *testing.T) {
```

pkg/util/template.go
```diff
@@ -0,0 +1,75 @@
...
+var nameTemplateFuncs = map[string]any{
+	"trimSuffix": sprig.GenericFuncMap()["trimSuffix"],
+	"trunc":      sprig.GenericFuncMap()["trunc"],
+}
+
+var nameTpl = template.New("name generator").Funcs(nameTemplateFuncs).Option("missingkey=error")
+
+// GenerateNameFromMachineNameTemplate generate a name from machine name and a naming strategy template.
+// the template supports only `trimSuffix` and `trunc` functions.
+func GenerateNameFromMachineNameTemplate(machineName string, nameTemplate *string) (string, error) {
```
chrischdi, line 41:
```suggestion
func GenerateMachineNameFromTemplate(machineName string, nameTemplate *string) (string, error) {
```

pkg/util/template_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewWithT(t)
+
+			got, err := GenerateNameFromMachineNameTemplate(tt.machineName, tt.template)
+
+			if (err != nil) != tt.wantErr {
+				t.Errorf("GenerateVSphereVMName error = %v, wantErr %v", err, tt.wantErr)
```
chrischdi, line 92:
```suggestion
				t.Errorf("GenerateMachineNameFromTemplate error = %v, wantErr %v", err, tt.wantErr)
```

pkg/util/template_test.go
```diff
@@ -0,0 +1,105 @@
...
+func Test_GenerateVSphereVMName(t *testing.T) {
+	tests := []struct {
+		name        string
+		machineName string
+		template    *string
+		want        []gomegatypes.GomegaMatcher
+		wantErr     bool
+	}{
+		{
+			name:        "default template",
```
chrischdi, line 36:
```suggestion
			name:        "return machineName if template is nil",
```

chrischdi (commented):
Last findings, happy to get this merged once that's fixed.

Thanks for working on it and sorry for taking so long :-) 🎉

test/e2e/data/infrastructure-vsphere-govmomi/main/clusterclass/patch-vm-namingstrategy.yaml
viveksyngh: It should be fixed now.

go.mod
viveksyngh: Updated the dependency.

chrischdi (commented):
/lgtm

/assign @sbueringer for a final round

chrischdi:
For checking ci

/hold

chrischdi:
Let's see how it behaves at https://github.com/team-cluster-api/cluster-api-provider-vsphere/pull/57

chrischdi:
/retest

to see if its a flake or an issue

sbueringer:
CI green, PR looks good

/approve
/hold cancel

PR #3287, opened 2024-12-09
🐛 supervisor: create ClusterModule per MachineDeployment and re-reconcile VirtualMachineResourceSetPolicy to update VirtualMachineSetResourcePolicy

chrischdi:

* CAPV currently creates one single cluster module which is used for all machinedeployments of a cluster.
* This was added like this because there always was only a single machinedeployment used.
* This PR:
	* adds a clustermodule per machinedeployment
	* updates the clustermodule's
	* watches machinedeployments now, to trigger updates for clustermodules
	* removes the old cluster-module name used before
	* verifies that the cluster module got created by vm-operator before creating a virtual machine

Fixes #

### DIFF
--- a/pkg/services/vmoperator/resource_policy.go
+++ b/pkg/services/vmoperator/resource_policy.go
@@ -18,11 +18,14 @@ package vmoperator
 
 import (
 	"context"
+	"sort"
 
 	"github.com/pkg/errors"
 	vmoprv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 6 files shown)


pkg/services/vmoperator/resource_policy.go
```diff
@@ -106,3 +122,32 @@ func (s *RPService) createVirtualMachineSetResourcePolicy(ctx context.Context, c
...
 	return vmResourcePolicy, nil
 }
+
+func getClusterModuleGroups(ctx context.Context, ctrlClient client.Client, cluster *clusterv1.Cluster) ([]string, error) {
+	machineDeploymentNames, err := getMachineDeploymentNamesForCluster(ctx, ctrlClient, cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	clusterModuleGroups := append([]string{ControlPlaneVMClusterModuleGroupName}, machineDeploymentNames...)
```
fabriziopandini, line 132: Might be we should add also fmt.Sprintf("%s-workers-0", cluster.Name) if not already there, at least for the transition time

chrischdi:
/hold

```
failed to patch VirtualMachineSetResourcePolicy clusterclass-rollout-i3y0x6/clusterclass-rollout-jxgupz:
      admission webhook "default.validating.virtualmachinesetresourcepolicy.v1alpha2.vmoperator.vmware.com"
      denied the request: spec.clustermodules: Invalid value: []string{"clusterclass-rollout-jxgupz-md-md-0-wpd29",
      "control-plane-group"}: field is immutable
```

pkg/services/vmoperator/resource_policy.go
```diff
@@ -37,15 +40,31 @@ type RPService struct {
...
-		resourcePolicy, err = s.createVirtualMachineSetResourcePolicy(ctx, clusterCtx)
+		resourcePolicy, err = s.createVirtualMachineSetResourcePolicy(ctx, clusterCtx, clusterModuleGroups)
 		if err != nil {
 			return "", errors.Errorf("failed to create Resource Policy: %+v", err)
 		}
+		return resourcePolicy.Name, nil
+	}
+
+	// Ensure .spec.clusterModuleGroups is up to date.
+	helper, err := patch.NewHelper(resourcePolicy, s.Client)
```
fabriziopandini, line 61: Note: I prefer this approach (using the patch helper), but let's consider also using createOrUpdate for consistency with the rest of the codebase

pkg/services/vmoperator/resource_policy.go
```diff
@@ -106,3 +122,32 @@ func (s *RPService) createVirtualMachineSetResourcePolicy(ctx context.Context, c
...
+		return err
+	}
+
+	for _, cm := range resourcePolicy.Status.ClusterModules {
+		if cm.GroupName == clusterModuleGroupName {
+			return nil
+		}
+	}
+
+	return errors.Errorf("VirtualMachineSetResourcePolicy's .status.clusterModules does not yet contain %s", clusterModuleGroupName)
```
fabriziopandini, line 152: Note: when we look into v1beta2 conditions for CAPV, we should consider if/how to surface when machine creation is stuck due to corresponding module missing

fabriziopandini (commented):
first pass

pkg/services/vmoperator/resource_policy.go
chrischdi: The code did before use `createOrUpdate`, but only did run it on creation and never updated anything.

sbueringer:
What is the current state of this?

chrischdi:
Current state is: I did the changes required, still wanted to test with a real vCenter, not just vcsim. But in general it should already be in the target state.

chrischdi:
/close

Unable to drive this further.

PR #3236, opened 2024-10-25
🌱 Replace CAPI's clustercache tracker by the new clustercache

chrischdi:
**What this PR does / why we need it**:

Removes usage of the deprecated remote.ClusterCacheTracker and related functions.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #3296, opened 2024-12-13
🐛 vspherevm: don't requeue on deletenode when there is no ClusterCache connection

chrischdi:

Makes the best-effort node-deletion to not cause infinite requeue when there is no connection to the workload cluster.

Bug was introduced in:

- #3236

Note: this only affected pre-releases of v1.12 prior to this commit.

If kubeadm init failed, this would have lead to CAPV always requeue and never remove the finalizer. So deletion would have been stuck.

Fixes #


fabriziopandini (commented):
Thanks for looking into this!
/lgtm
/approve

PR #3344, opened 2025-02-06
🌱 test: shutdown clustercache when ctx is done

chrischdi:

Follow-up from bumping due to clustercache changes in https://github.com/kubernetes-sigs/cluster-api/pull/11757

This should get the unit tests back to green.

/assign @sbueringer
/assign @fabriziopandini 

The tests in:

* `controllers/vmware/serviceaccount_controller_intg_test.go`
* `controllers/vmware/servicediscovery_controller_intg_test.go`

Have a special requirement: They require an additional kube-apiserver instance to simulate the workload cluster.
To be able to properly shutoff this kube-apiserver instance (including etcd) we need to stop the clustercache.
This was only possible at the very end of the whole test suite to not break other tests.

Instead now these two test suites get their own management cluster including clustercache controller so everything can be shutdown after the tests.

Fixes #

### DIFF
--- a/controllers/controllers_suite_test.go
+++ b/controllers/controllers_suite_test.go
@@ -98,6 +98,10 @@ func setup() {
 	if err != nil {
 		panic(fmt.Sprintf("Unable to setup ClusterCache: %v", err))
 	}
+	go func() {
+		<-ctx.Done()
+		clusterCache.(interface{ Shutdown() }).Shutdown()
+	}()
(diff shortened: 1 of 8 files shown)


chrischdi:
Note: we start more apiservers then clustercaches, have to fix that...

sbueringer:
Not very stable atm 

/retest

PR #1743, opened 2023-01-17
⚠️ Allow specifying a different credentials per VSphereMachine

farodin91:
**What this PR does / why we need it**:
Allow multiple vcenter in a single cluster

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1720

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

I will try to test it.

**Release note**:

```release-note
Allow custom identity in vmmachine
```

Issue #1720, opened 2022-12-02
[WIP] ⚠️ Allow specifying a different credentials per VSphereMachine


farodin91:
/kind feature

**Describe the solution you'd like**
It would be great if CAPV would support multiple vCenter in one cluster.

Use Cases:

- Migration between vCenter's
- Blast reduce reduction

Any idea or other use cases?

farodin91:
My first implementation idea is to add an identityRef to https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/cc840c560ec33f378aa1ea9dc9294eceba52b4b9/apis/v1beta1/vspheremachine_types.go#L34 and then evaluate first spec identityRef, next cluster resource, and fallback to old secret here https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/3aaeecc4a972d38d3828415936f47d66cae969ef/controllers/vspherevm_controller.go#L526.

@srm09 What do you think?

srm09:
Using this as a note for some things to think about:
- How do we use the thumbprint value for multiple vCenters?
- How does cluster module setup work for multiple vCenters?

chrischdi:
/reopen

There's interest in reviving this.

chrischdi:
/retitle [WIP] ⚠️ Allow specifying a different credentials per VSphereMachine

chrischdi:
/retitle Support multiple vCenter in one cluster

Wrong issue

PR #3353, opened 2025-02-12
[WIP] ⚠️ Allow specifying a different credentials per VSphereMachine

GabQ-Bib:
… and defaulting to this identity if it exists.

This PR adds support for configuring a VsphereIdentity in the vspheremachine / vspherevm objects, which the controller will use when looking for credentials to contact the vcenter with. This appears to be a good solution to the issue mentioned below, because these objects already carry the server name & thumbprint so it makes sense they also carry the credentials.
This PR is a rewrite of the [PR originally proposed by farodin91](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/1743).

Fixes [#1720](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1720)


GabQ-Bib:
I can see several way of implementing the feature:
- Add a second "getCredential" method that is specifically called when trying to extract credentials from the VspereVM (instead of the cluster). This is the implementation you will find in the attached commit. It has the less impact on the codebase because the original method signature is kept throughout the rest of the controllers. As of right now, both method share a lot of code which is not ideal. Some of the code of the two methods could be factored in a 3rd method to help with code duplication.
- A second implementation would be to change the getCredential method signature to directly use a "VsphereIdentityRef" along with a "namespace" string to cover all use cases. It requires a modest refactoring of the controllers that already use this method but it feels ultimately a better approach. I have a commit ready with this implementation.
- I'm wondering if we could just change the method signature to expect an "object that has a property VsphereIdentityRef" which both cluster and vspherevm would comply to, removing the need to change the way the method is called in the rest of the code. My understanding of Go makes me think this would require creating such a type and having both cluster and vspherevm embed it, but It feels like a much more profound rework than the other two implementation proposed above.

I'm still in the process of testing whether this implementation fixes the issue in a satisfactory manner for my use case. In the meantime I would love if you could find the time to do a quick overview of my proposal and provide insights.

chrischdi:
/retitle [WIP] ⚠️ Allow specifying a different credentials per VSphereMachine

fabriziopandini:
/close
Due to lack of bandwidth/Lack of progress

PR #3368, opened 2025-03-03
[release-1.12] 🌱 Makefile: drop test file to ignore ssh vulnerability in test code

chrischdi:

Deletes the file `test/framework/log/collector.go` while running govulncheck.

We on purpose ignore this vulnerability because bumping to `golang.org/x/crypto@v0.35.0` would require us to also bump transitively to Go 1.23 in our go.mod file which would be a breaking change.

The vulnerability only affects test code.

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -530,12 +530,14 @@ verify-licenses: ## Verify licenses
 
 .PHONY: verify-govulncheck
 verify-govulncheck: $(GOVULNCHECK) ## Verify code for vulnerabilities
+	rm "test/framework/log/collector.go"
 	$(GOVULNCHECK) ./... && R1=$$? || R1=$$?; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -530,12 +530,14 @@ verify-licenses: ## Verify licenses
 
 .PHONY: verify-govulncheck
 verify-govulncheck: $(GOVULNCHECK) ## Verify code for vulnerabilities
+	rm "test/framework/log/collector.go"
```
sbueringer, line 533: I think we have to delete some more files: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/actions/runs/13630381432/job/38096661742

sbueringer: Talked to Christian. The other findings are transitively using this file

Issue #947, opened 2020-07-01
Set InternalIP machine status.addresses


andrewsykim:
/kind feature

**Describe the solution you'd like**
Almost always a VM will have an address internal to the datacenter. In the case where we don't know if an address is external or internal to the datacenter, we should assume it is internal and set the type as InternalIP. Today we always set it as an ExternalIP

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

andrewsykim:
/remove-lifecycle stale

This is still pretty important from my perspective, in general we shouldn't label private addresses spaces as `ExternalIP`. It's also not consistent with how we address node addresses.

yastij:
@andrewsykim - Agreed. As this is a breaking changing my thinking was to include this in the set of breaking changes of v1a4

andrewsykim:
> As this is a breaking changing my thinking was to include this in the set of breaking changes of v1a4.

It's only a breaking change if we stop setting ExternalIP. I don't think it would be a breaking change to start setting InternalIP now though?

yastij:
The interesting bit is that we don't have a way to distinguish between InternalIP and ExternalIP. Are you suggesting to set both ?

andrewsykim:
If we can't distinguish between the two, we should set the type based on what is most common -- which from my understanding is InternalIP, since almost always a VM's machine IP is internal to a vCenter. In an ideal world we would just set InternalIP, but removing ExternalIP is a breaking change. 

My suggestion would be to set both InternalIP and ExternalIP for the remainder of v1alpha3 and remove ExternalIP starting v1alpha4. Or even better, set the type based on some well known heuristic like checking if the address is in one of the well-known private address spaces.

yastij:
that makes sense wrt timelines

srm09:
@yastij This still seems to be the case for Machine addresses, something that we should look into again?
/help

flyik:
Any updates? I can implement it if thoughts about it are still the same

CharlieR-o-o-t:
Have no idea why provider always sets ExternalIP , this is incorrect behavior. 
Have created fix PR: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3379

PR #3379, opened 2025-03-06
🐛  fix: set correct IP address type for machine, #947

CharlieR-o-o-t:
It's incorrect to always set ExternalIP type for all machine IP addresses. Need to fix that.

Fixes issue https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/947


chrischdi (commented):
This does not work like this.

There are places where explicitly MachineExternalIP get's filtered for.

Example:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/91d6fbb859bce9f21de565d3df4f525af4be89ba/pkg/util/machines.go#L69

Also we should make sure to not break other potential consumers and this would be a breaking change for them.

Do you know if there is a good definition in upstream documentation of the separation between ExternalIP and InternalIP? This could would rely on the definition of the private subnets.

In general I agree that this should be InternalIP, but we need to be careful.

cc @neolit123 , @rikatz

chrischdi:
I think a way forward would be to set both for some time and maybe deprecate setting the ExternalIP, maybe via a feature gate so we could go through normal deprecation lifecycle.

/ok-to-test

pkg/util/machines.go
```diff
@@ -258,3 +258,41 @@ func MachinesAsString(machines []*clusterv1.Machine) string {
...
+		return false
+	}
+
+	privateBlocks := []*net.IPNet{
+		{IP: net.IPv4(10, 0, 0, 0), Mask: net.CIDRMask(8, 32)},     // 10.0.0.0/8
+		{IP: net.IPv4(172, 16, 0, 0), Mask: net.CIDRMask(12, 32)},  // 172.16.0.0/12
+		{IP: net.IPv4(192, 168, 0, 0), Mask: net.CIDRMask(16, 32)}, // 192.168.0.0/16
+		{IP: net.IPv4(127, 0, 0, 1), Mask: net.CIDRMask(8, 32)},    // 127.0.0.0/8 (loopback)
+		{IP: net.IPv4(169, 254, 0, 0), Mask: net.CIDRMask(16, 32)}, // 169.254.0.0/16 (link-local)
+	}
```
neolit123, line 275: note that cluster local subnets are not the same as local network subnets.

neolit123 (commented):
> Do you know if there is a good definition in upstream documentation of the separation between ExternalIP and InternalIP? This could would rely on the definition of the private subnets.

node internal ip = ip of the node as accessed from the cluster
node external ip = 'real' node ip for external access

/cc @lubronzhan 
since i'm not a network expert

pkg/util/machines.go
neolit123: i don't think that would be the correct way to determine if the ip is internal. it would be an internal network ip, but not necessary the internal cluster ip allocated to a node. to check if the ip is internal as per k8s' definition, i think one would have to check how the controller manager cidr is configured.

CharlieR-o-o-t:
@chrischdi

Hello, thank you for quick reply, agree that we need to be carefull. 

About breaking change. 
I could be wrong, but I can see that type of IPAddress have no effect on  guestinfo.metadata or machine network configuration, it's just status in k8s object, on workload k8s cluster vsphere-cpi will deal with External/Internal IP address types.

This method is only used to check that we have at least one IP on control-plane bootstrap: 
https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/91d6fbb859bce9f21de565d3df4f525af4be89ba/pkg/util/machines.go#L59
IP output from the function is never used  (only in _test.go)

We are using cluster-api-provider-vsphere for long time and only now noticed this issue, when we decided to monitor machines with pubic network.

lubronzhan:
> > Do you know if there is a good definition in upstream documentation of the separation between ExternalIP and InternalIP? This could would rely on the definition of the private subnets.
> 
> node internal ip = ip of the node as accessed from the cluster node external ip = 'real' node ip for external access
> 
> /cc @lubronzhan since i'm not a network expert

Definition from k8s https://github.com/kubernetes/kubernetes/blob/cd060979bfae61630c829d4a88f1fba2c60b3ad1/pkg/apis/core/types.go#L5378-L5389
```
// NodeInternalIP identifies an IP address which is assigned to one of the node's
	// network interfaces. Every node should have at least one address of this type.
	//
	// An internal IP is normally expected to be reachable from every other node, but
	// may not be visible to hosts outside the cluster. By default it is assumed that
	// kube-apiserver can reach node internal IPs, though it is possible to configure
	// clusters where this is not the case.
	//
	// NodeInternalIP is the default type of node IP, and does not necessarily imply
	// that the IP is ONLY reachable internally. If a node has multiple internal IPs,
	// no specific semantics are assigned to the additional IPs.
```

The `internalIP` on the `Node.status` doesn't always mean it's from the private network cidr, don't think there is a strict limitation in the upstream that  where the IP range should be.

The InternalIP is could be from RFC 1918 private IP address ranges:
```
* 10.0.0.0/8 (10.x.x.x)
* 172.16.0.0/12 (172.16.x.x - 172.31.x.x)
* 192.168.0.0/16 (192.168.x.x)
```
Still, the the customer could decide what IP range to be used for their internalIP on the node when they configure vsphere-cpi. For example, some lazy customer just choose to use a public IP to pass to CPI.

But current code change is used for `MachineInternalIP`, which I don't think we use it somewhere in downstream at least. Usually we, or controllers use node's IntermalIP/ExternalIP


For this case, maybe the ideal solution would be, like what vsphere-cpi does, pass a range of IPs as internalMachineRange through cluster object, so this status could use that range to filter what IP to show for `MachineInternalIP`

CharlieR-o-o-t:
@lubronzhan  @neolit123 
Clearly got your point. 

I can rework this as it's done in vsphere-cpi. So user will define CIDRs for internal/external. 
The only one question - what to do if nothing defined in config for this?

sbueringer:
Usually, keep the behavior the same to not break anyone

Issue #3221, opened 2024-10-14
Adopt CAPI v1beta2 conditions


chrischdi:
**Describe the solution you'd like**

TODOs to adopt v1beta2 conditions in CAPV

- [x] add API types for v1beta2 to v1beta1 status (part of the other PRs)
- [x] Use the finalizer util like core CAPI
  - https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3388
- [x] Implement conditions
  - [x] (Govmomi & Supervisor) Paused: #3394 
  - [x] (Supervisor) VSphereCluster: #3383
  - [x] (Govmomi) VSphereCluster: #3396
  - [x] (Govmomi & Supervisor) VSphereMachine:	#3406
  - [x] (Govmomi) VSphereDeploymentZone & VSphereClusterIdentity: #3401 

cc @fabriziopandini @sbueringer 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-vsphere version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

sbueringer:
Thx!

sbueringer:
cc @chrischdi Let's close this tomorrow :)

chrischdi:
/close

We will do a round of manual validation though.

PR #3388, opened 2025-03-12
🌱 controllers: set finalizer in all controllers first to avoid race conditions between init and delete

chrischdi:

Part of 
- #3221

Always patch finalizers right at the beginning and use capi util for it to align with capi.

Fixes #


chrischdi:
/hold

for e2e tests

chrischdi:
mirror prow green
/hold cancel

PR #3394, opened 2025-03-14
[wip] 🌱 v1beta2: add v1beta2 paused condition for all objects

chrischdi:
**What this PR does / why we need it**:

Adds v1beta2's paused condition handling to all resources we have.

Also adjusts predicates accordingly.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #3396, opened 2025-03-17
🌱 v1beta2: implement VSphereCluster's Ready, FailureDomainsReady, ClusterModulesReady and VCenterAvailable conditions for govmomi

chrischdi:

Implements the following v1beta2 conditions for govmomi mode's VSphereCluster
- Ready
- FailureDomainsReady
- ClusterModulesReady
- VCenterAvailable

Part of #3221

Fixes #


chrischdi:
/hold

For rebase after

- #3394

merged

chrischdi:
/hold

Waiting for CI at:

- https://github.com/team-cluster-api/cluster-api-provider-vsphere/pull/70

controllers/vspherecluster_reconciler.go
```diff
@@ -117,18 +118,90 @@ func (r *clusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
 		return r.reconcileDelete(ctx, clusterContext)
 	}
```
sbueringer, line 120: Why did we remove this?

I think it's part of the contract: https://cluster-api.sigs.k8s.io/developer/providers/contracts/infra-cluster#typical-infracluster-reconciliation-workflow

controllers/vspherecluster_reconciler.go
```diff
@@ -117,18 +118,90 @@ func (r *clusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+			infrav1.VCenterAvailableCondition,
+		),
+	)
+
+	if !v1beta2conditions.Has(clusterCtx.VSphereCluster, infrav1.VSphereClusterVCenterAvailableV1Beta2Condition) {
+		v1beta2conditions.Set(clusterCtx.VSphereCluster, metav1.Condition{
+			Type:    infrav1.VSphereClusterVCenterAvailableV1Beta2Condition,
+			Status:  metav1.ConditionFalse,
+			Reason:  infrav1.VSphereClusterVCenterNotAvailableV1Beta2Reason,
+			Message: "did not observe VCenterAvailable status",
```
sbueringer, line 139: Should these messages start with upper case? (same below)

controllers/vspherecluster_reconciler.go
```diff
@@ -117,18 +118,90 @@ func (r *clusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+		})
+	}
+
+	if !v1beta2conditions.Has(clusterCtx.VSphereCluster, infrav1.VSphereClusterClusterModulesReadyV1Beta2Condition) {
+		v1beta2conditions.Set(clusterCtx.VSphereCluster, metav1.Condition{
```
sbueringer, line 144: Should this also be added when cluster modules are not enabled?

controllers/vspherecluster_reconciler.go
```diff
@@ -193,35 +266,82 @@ func (r *clusterReconciler) reconcileNormal(ctx context.Context, clusterCtx *cap
 
 	ok, err := r.reconcileDeploymentZones(ctx, clusterCtx)
 	if err != nil {
```
sbueringer, line 268: Seems like there are subtle changes to how this conditions is set. Can you give me the reasonsing behind it so I can check if the new logic is correct?

e.g. previously the condition was set to False if noteReady > 0 (l.475)

If we did not plan to change the logic, I would prefer adding the v1beta2 conditions in the same place as v1beta1. It's much easier to ensure the logic didn't change this way (or at least for the non-error cases)

controllers/vspherecluster_reconciler.go
```diff
@@ -193,35 +266,82 @@ func (r *clusterReconciler) reconcileNormal(ctx context.Context, clusterCtx *cap
...
+		})
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 	}
+	v1beta2conditions.Set(clusterCtx.VSphereCluster, metav1.Condition{
+		Type:   infrav1.VSphereClusterFailureDomainsReadyV1Beta2Condition,
+		Status: metav1.ConditionTrue,
+		Reason: infrav1.VSphereClusterFailureDomainsReadyV1Beta2Reason,
+	})
+
+	// TODO check all occurencies of infrav1.VCenterAvailableCondition and infrav1.ClusterModulesAvailable/ready condition
```
sbueringer, line 292: TODO?

pkg/session/session.go
```diff
@@ -268,13 +268,13 @@ func (s *Session) GetVersion() (infrav1.VCenterVersion, error) {
...
 	if err != nil {
-		return "", err
+		return "", errors.Wrapf(err, "failed to parse version %q", svcVersion)
 	}
 
 	if version.Major >= 6 {
 		return infrav1.NewVCenterVersion(svcVersion), nil
 	}
-	return "", unidentifiedVCenterVersion{version: svcVersion}
+	return "", errors.Errorf("version %q lower than 7", svcVersion)
```
sbueringer, line 277: I'm not following. How do we know the version is lower than 7?

Or should this be 6?

controllers/vspherecluster_reconciler.go
chrischdi: The reasoning was to ensure this condition always gets set.
I thought that'd be better especially around when calculating the summary condition. but okay to adapt to how it was earlier.

controllers/vspherecluster_reconciler.go
chrischdi: Oh, very good finding.

I did it to ensure the first condition at least always get set so we can rely on it when calculating Ready.

But then it probably is better to revert this and instead adjust the patch logic to be okay with none of the conditions ready depends on are set.

pkg/session/session.go
chrischdi: oh, should be lower than 6 😂

controllers/vspherecluster_reconciler.go
```diff
@@ -117,18 +118,90 @@ func (r *clusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+			Type:    infrav1.VSphereClusterClusterModulesReadyV1Beta2Condition,
+			Status:  metav1.ConditionFalse,
+			Reason:  infrav1.VSphereClusterClusterModulesNotReadyV1Beta2Reason,
+			Message: "did not observe cluster module status",
+		})
+	}
+
+	if err := v1beta2conditions.SetSummaryCondition(clusterCtx.VSphereCluster, clusterCtx.VSphereCluster, infrav1.VSphereClusterReadyV1Beta2Condition,
+		v1beta2conditions.ForConditionTypes{
+			infrav1.VSphereClusterFailureDomainsReadyV1Beta2Condition,
```
fabriziopandini, line 154: q: should we make this condition optional (IgnoreIfMissing) considering we don't have a fallback set above?

controllers/vspherecluster_reconciler.go
```diff
@@ -117,18 +118,90 @@ func (r *clusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+// patch updates the VSphereCluster and its status on the API server.
+func (r *clusterReconciler) patch(ctx context.Context, clusterCtx *capvcontext.ClusterContext) error {
+	// always update the readyCondition.
+	conditions.SetSummary(clusterCtx.VSphereCluster,
+		conditions.WithConditions(
+			infrav1.VCenterAvailableCondition,
+		),
+	)
+
+	if !v1beta2conditions.Has(clusterCtx.VSphereCluster, infrav1.VSphereClusterVCenterAvailableV1Beta2Condition) {
```
fabriziopandini, line 134: I'm wondering a little bit about this.
If the goal is to handle code paths where those conditions are not set after an upgrade, this could lead to a state where the condition is false and it is never moved to true, which is an issue.

What if instead, if the condition is missing, we set this condition to true if the vsphere cluster is ready, false otherwise

controllers/vspherecluster_reconciler.go
sbueringer: Yup. Or return before the defer. How does this work in the supervisor cluster reconciler?

controllers/vspherecluster_reconciler.go
chrischdi: ack, misinterpreted what was discussed :-)

Adjusting it to set them on ignore. The old summary only included VCenterAvailable, so we also will rely on that, and add the others with ignore.

controllers/vspherecluster_reconciler.go
chrischdi: This is gone now, we only set the condition now as we did in the past.

controllers/vspherecluster_reconciler.go
chrischdi: Changed now.

Only two changes are
* also set the condition if an error was returned.
* do not remove the condition

chrischdi:
/retest

Image pull failure while building kind image

controllers/vspherecluster_reconciler.go
```diff
@@ -126,9 +127,70 @@ func (r *clusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+				)),
+			),
+		},
+	); err != nil {
+		return pkgerrors.Wrapf(err, "failed to set %s condition", infrav1.VSphereClusterReadyV1Beta2Condition)
+	}
+
+	return clusterCtx.PatchHelper.Patch(ctx, clusterCtx.VSphereCluster,
+		patch.WithOwnedV1Beta2Conditions{Conditions: []string{
+			infrav1.VSphereClusterReadyV1Beta2Condition,
```
chrischdi, line 167: todo: ensure paused is here after rebase

controllers/vspherecluster_reconciler.go
```diff
@@ -346,15 +447,33 @@ func (r *clusterReconciler) reconcileDeploymentZones(ctx context.Context, cluste
 
 	clusterCtx.VSphereCluster.Status.FailureDomains = failureDomains
 	if readyNotReported > 0 {
+		log.Info("Waiting for failure domains to be reconciled")
 		conditions.MarkFalse(clusterCtx.VSphereCluster, infrav1.FailureDomainsAvailableCondition, infrav1.WaitingForFailureDomainStatusReason, clusterv1.ConditionSeverityInfo, "waiting for failure domains to report ready status")
+		v1beta2conditions.Set(clusterCtx.VSphereCluster, metav1.Condition{
+			Type:    infrav1.VSphereClusterFailureDomainsReadyV1Beta2Condition,
+			Status:  metav1.ConditionTrue,
```
sbueringer, line 454:
```suggestion
			Status:  metav1.ConditionFalse,
```
I think

controllers/vspherecluster_reconciler.go
```diff
@@ -346,15 +447,33 @@ func (r *clusterReconciler) reconcileDeploymentZones(ctx context.Context, cluste
...
+		})
 		return false, nil
 	}
 
 	if len(failureDomains) > 0 {
 		if notReady > 0 {
 			conditions.MarkFalse(clusterCtx.VSphereCluster, infrav1.FailureDomainsAvailableCondition, infrav1.FailureDomainsSkippedReason, clusterv1.ConditionSeverityInfo, "one or more failure domains are not ready")
+			v1beta2conditions.Set(clusterCtx.VSphereCluster, metav1.Condition{
+				Type:    infrav1.VSphereClusterFailureDomainsReadyV1Beta2Condition,
+				Status:  metav1.ConditionTrue,
```
sbueringer, line 466:
```suggestion
				Status:  metav1.ConditionFalse,
```

apis/v1beta1/vspherecluster_types.go
```diff
@@ -28,6 +28,78 @@ const (
...
+
+	// VSphereClusterVCenterUnreachableV1Beta2Reason surfaces when the vCenter for a VSphereCluster is unreachable.
+	VSphereClusterVCenterUnreachableV1Beta2Reason = "VCenterUnreachable"
+
+	// VSphereClusterVCenterNotAvailableV1Beta2Reason surfaces when the vCenter for a VSphereCluster is not available.
+	VSphereClusterVCenterNotAvailableV1Beta2Reason = clusterv1.NotAvailableV1Beta2Reason
+
+	// VSphereClusterVCenterAvailableDeletingV1Beta2Reason surfaces when the VSphereCluster is being deleted.
+	VSphereClusterVCenterAvailableDeletingV1Beta2Reason = clusterv1.DeletingV1Beta2Reason
+)
```
fabriziopandini, line 82: We should look into making this condition consistent across objects

controllers/clustermodule_reconciler.go
```diff
@@ -170,11 +177,23 @@ func (r Reconciler) Reconcile(ctx context.Context, clusterCtx *capvcontext.Clust
...
+			Reason:  infrav1.VSphereClusterClusterModulesNotReadyV1Beta2Reason,
+			Message: generateClusterModuleErrorMessage(modErrs),
+		})
+		return reconcile.Result{}, err
 	case len(modErrs) == 0 && len(clusterModuleSpecs) > 0:
 		conditions.MarkTrue(clusterCtx.VSphereCluster, infrav1.ClusterModulesAvailableCondition)
 	default:
 		conditions.Delete(clusterCtx.VSphereCluster, infrav1.ClusterModulesAvailableCondition)
 	}
+	v1beta2conditions.Set(clusterCtx.VSphereCluster, metav1.Condition{
```
fabriziopandini, line 192: q: why this is not after L188

chrischdi: Because the old condition did not get always set, but removed instead in that case.

apis/v1beta1/vspherecluster_types.go
sbueringer: In this PR or after we merged all currently open PRs?

sbueringer: Discussed, idea is to see if we need the additional NotAvailableV1Beta2Reason here

sbueringer: Damn. Strange that I didn't see that in Intellij

sbueringer:
/lgtm
/approve

Keeping hold for CI

PR #3405, opened 2025-03-18
🌱 e2e: servicediscovery controller to have an ip to discover in tests

chrischdi:

Fixes red e2e tests (especially around upgrades).

Error is/was:

```
* ServiceDiscoveryReady: failed to create or patch service Endpoints: Endpoints "supervisor" is   
  invalid: [subsets[0].addresses[0].ip: Invalid value: "": must be a valid IP address, (e.g.      
  10.9.8.7 or 2001:db8::ffff), subsets[0].addresses[0].ip: Invalid value: "": must be a valid IP  
  address]                                                                                        
```

Fixes #

### DIFF
--- a/test/framework/vmoperator/vmoperator.go
+++ b/test/framework/vmoperator/vmoperator.go
@@ -402,8 +402,8 @@ func ReconcileDependencies(ctx context.Context, c client.Client, dependenciesCon
 	supervisorAPIServerVIPService.Status = corev1.ServiceStatus{
 		LoadBalancer: corev1.LoadBalancerStatus{Ingress: []corev1.LoadBalancerIngress{
 			// Note: this creates a unusable service. During test no application should try to reach out to this.
-			// 169.254.0.1 is part of the link-local subnet	169.254.0.0/16.
-			{IP: "169.254.0.1"},
+			// 192.0.2.2 is part of the link-local subnet 192.0.2.0/24 which is reserved for documentation and testing.
+			{IP: "192.0.2.2"},
 		}},
 	}
 	_ = wait.PollUntilContextTimeout(ctx, 250*time.Millisecond, 5*time.Second, true, func(ctx context.Context) (bool, error) {
--- a/test/framework/vmoperator/vmoperator.go
+++ b/test/framework/vmoperator/vmoperator.go
@@ -402,7 +402,8 @@ func ReconcileDependencies(ctx context.Context, c client.Client, dependenciesCon
 	supervisorAPIServerVIPService.Status = corev1.ServiceStatus{
 		LoadBalancer: corev1.LoadBalancerStatus{Ingress: []corev1.LoadBalancerIngress{
 			// Note: this creates a unusable service. During test no application should try to reach out to this.
-			{IP: "127.0.0.255"},
+			// 169.254.0.1 is part of the link-local subnet	169.254.0.0/16.
+			{IP: "169.254.0.1"},
 		}},
 	}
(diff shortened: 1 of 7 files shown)


sbueringer:
Probably unrelated: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/3405/pull-cluster-api-provider-vsphere-e2e-vcsim-supervisor-main/1901938553072390144 ?

Retrying..

/test pull-cluster-api-provider-vsphere-e2e-vcsim-supervisor-main

sbueringer:
@chrischdi Ping me once you're happy with it / it works :D (will ignore notifications until then)

test/framework/vmoperator/vmoperator.go
```diff
@@ -361,6 +362,60 @@ func ReconcileDependencies(ctx context.Context, c client.Client, dependenciesCon
 		return retryError
 	}
 
+	// Create the supervisor service in kube-system for the servicediscovery controller to discover and set an IP address
```
sbueringer, line 365: Let's create an issue to take a look at the non-IP code path in servicediscovery controller that simply doesn't work today (probably we should just return an error in that case, but TBD)

test/framework/vmoperator/vmoperator.go
```diff
@@ -361,6 +362,60 @@ func ReconcileDependencies(ctx context.Context, c client.Client, dependenciesCon
...
+			log.Info("Created vm-operator service", "Service", klog.KObj(supervisorAPIServerVIPService))
+		}
+		return true, nil
+	})
+	if retryError != nil {
+		return retryError
+	}
+	supervisorAPIServerVIPService.Status = corev1.ServiceStatus{
+		LoadBalancer: corev1.LoadBalancerStatus{Ingress: []corev1.LoadBalancerIngress{
+			{IP: "192.168.1.99"},
```
sbueringer, line 403: Let's add a godoc how we picked this IP (private IP space + it should not be used by anyone so it simply fails if someone tries to use the svc)

sbueringer (commented):
Just a nit (doesn't require retest, so let's wait for the tests on mirror to complete then fix it)

test/framework/vmoperator/vmoperator.go
chrischdi: I think the non-IP code is valid, but depends on how you setup k8s.

test/framework/vmoperator/vmoperator.go
chrischdi: hmm I was picking it because for vcsim we use `192.168.1.100` for machine IPs. But this code here also runs when not using vcsim.

Thinking about it again: let's use a different IP. Changed to 127.0.0.255 which should not be used anywhere in our test infra for sure.

test/framework/vmoperator/vmoperator.go
sbueringer: I think it's not. According to the errors we're getting IP is mandatory. So this code path can not work: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/393a16983d02d5a2b254e4182002f70329f9dd8f/controllers/vmware/servicediscovery_controller.go#L394

sbueringer: I'm not saying the code path to read kube-public can never work, just that if we don't get an IP there)

sbueringer:
/approve

/hold 
for tests

test/framework/vmoperator/vmoperator.go
sbueringer: Opened an issue: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/3409

PR #3413, opened 2025-03-25
🐛 Sets character limit for VMService name

gab-satchi:
If a supervisor backed cluster is created with a name > 41, it won't go into provisioned state. We attempt to create a VirtualMachineService resource with a name that's <cluster-name>-control-plane-service. This creation will fail as it needs to conform to RFC 1123 label names standard which among other things enforces a max character limit of 63. 

This PR adds a util function that will generate and append a hash suffix if the provided name is too long. Existing healthy clusters won't be affected as the name is returned unchanged if it's less than 63 chars.


gab-satchi:
Closing in favour of #3415 

/close

PR #3490, opened 2025-06-06
🌱 e2e: also collect pod logs

chrischdi:

Also collect logs from the running pods, but as tarball to save some space.

Fixes #

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -65,7 +65,8 @@ on_exit() {
     find "${ARTIFACTS}" -type f -name pod-logs.tar.gz | while IFS= read -r tarball; do
       echo "Unpacking ${tarball} for secrets replacement"
       mkdir -p "${tarball}-unpacked"
-      tar -xzf "${tarball}" -C "${tarball}-unpacked"
+      # on_exit should not fail due to broken tarballs
+      tar -xzf "${tarball}" -C "${tarball}-unpacked" || true
       rm "${tarball}"
     done
     # Delete non-text files from artifacts directory to not leak files accidentially
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -94,7 +94,7 @@ on_exit() {
     find "${ARTIFACTS}" -type d -name pod-logs.tar.gz-unpacked | while IFS= read -r tarballDirectory; do
       tarball="${tarballDirectory%-unpacked}"
       echo "Packing ${tarballDirectory} to ${tarball} after secrets replacement"
-      tar -czf "${tarball}" -C . "${tarballDirectory}"
+      tar -czf "${tarball}" -C "${tarballDirectory}" .
       rm -r "${tarballDirectory}"
     done
     # Move all artifacts to the original artifacts location.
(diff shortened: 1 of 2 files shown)


hack/e2e.sh
```diff
@@ -104,7 +118,7 @@ export VSPHERE_USERNAME="${GOVC_USERNAME:-}"
 export VSPHERE_PASSWORD="${GOVC_PASSWORD:-}"
 export E2E_CONF_FILE="${REPO_ROOT}/test/e2e/config/vsphere.yaml"
 export E2E_CONF_OVERRIDE_FILE=""
-export E2E_VM_OPERATOR_VERSION="${VM_OPERATOR_VERSION:-v1.8.6-0-gde75746a}"
+export E2E_VM_OPERATOR_VERSION="${VM_OPERATOR_VERSION:-v1.8.6-0-gde75746a-65e87004}"
```
sbueringer: Intentional?

sbueringer: Also. Shouldn't this break ci? Should we cleanup an image somewhere?

chrischdi: This is compatible to vsphere 8 and 7, but good catch, should not be in this PR.

sbueringer: Yeah. But we reverted the PR that introduced this, right?
So I assume we have no further plans to use this image (which is why I would suggest to delete the image from the registry to avoid us accidentally using it)

chrischdi: Yes.

chrischdi: Fixed now. Also deleting the image.

sbueringer: Thank you!

hack/e2e.sh
```diff
@@ -61,6 +61,13 @@ on_exit() {
 
   # Cleanup VSPHERE_PASSWORD from temporary artifacts directory.
   if [[ "${ORIGINAL_ARTIFACTS}" != "" ]]; then
+    # unpack pod-logs.tar.gz files to replace secrets in them
```
sbueringer, line 64: I think this results in a very deep folder hierarchy. In the interest of future us, can we reduce that a bit?

I get this after extract: ./tmp 2/tmp.BAZS1S16eR/clusters/quick-start-l3rsho/machines/quick-start-l3rsho-cp-5x2qt-wfsf5/pod-logs.tar.gz-unpacked/pods/kube-system_etcd-quick-start-l3-wfsf5_6501f15974bbf5a29b9e36b04db38957/etcd.

Ideally kube-system would be top-level in this archive (but just getting rid of *-unpacked and above would be fine as well)

chrischdi: oh 🤔 good one. Let me check how this came. I did not expect that either.

chrischdi: Should be good now.

Its now `pods/kube-system_etcd-quick-start-l3-wfsf5_6501f15974bbf5a29b9e36b04db38957/etcd`

chrischdi: (I broke it somewhere?!)

sbueringer: Perfect

sbueringer:
Thank you very much!

/lgtm
/approve

PR #3498, opened 2025-06-11
:bug: StoragePolicy should ignore datastores in maintenance mode

rikatz:
Make storagepolicy compatibility check ignore datastores in maintenance mode.

During cloning, when using a storage policy, CAPV gets a list of datastores connected to the desired cluster to, then, check if the datastore is compatible with the storage policy.

This operation doesn't account the fact a datastore can be in maintenance mode and may lead CAPV trying to clone a template to a datastore that is not accessible or is being drained, etc.

The change here makes the datastore to be removed from the list of compatibility, if they are not in normal mode, before sending the datastore list to SPBM check

Fixes #

### DIFF
--- a/pkg/services/govmomi/vcenter/clone.go
+++ b/pkg/services/govmomi/vcenter/clone.go
@@ -27,6 +27,7 @@ import (
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/pbm"
 	pbmTypes "github.com/vmware/govmomi/pbm/types"
+	"github.com/vmware/govmomi/property"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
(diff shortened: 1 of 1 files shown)


rgcostea (approved):

rikatz:
Hum, the errors doesn't seem related to my change

/cc @sbueringer @chrischdi

sbueringer:
Yeah. Don't worry about the errors :). We're in the process of finalizing the new community-owned CI env

/retest

rikatz:
thanks! should I cherry-pick it to 1.13.1? We kind of found this bug in the wild :)

rikatz:
approval pls? 🥺   (unless I missed something so I can fix it :). )

sbueringer:
Just didn't get around to it :)

Thx for the fix, feel free to backport

/lgtm
/approve

PR #3523, opened 2025-06-23
[WIP] 🌱 e2e: use antrea instead of calico

chrischdi:
Testing antrea instead of calico


chrischdi:
/close

In favor of #3525

PR #3530, opened 2025-06-26
✨Support multi-networking for NSX-VPC and vsphere-network providers

DanielXiao:
This PR enhances Node network management to support multi-networking for the NSX-VPC and vsphere-network providers, while preserving existing behavior for backward compatibility.

Current Behavior
CAPV manages Node network lifecycle based on the configured network provider:
* vsphere-network (VDS): Uses the single default Network in the Supervisor Namespace.
* NSX (NSX-T Tier-1): Automatically creates a VirtualNetwork for the Node network.
* NSX-VPC (NSX-T VPC): Automatically creates a SubnetSet for the Node network.

What's New
NSX-VPC:
* Primary network: Users can now specify an existing SubnetSet as the Kubernetes primary network (used for load balancing, service discovery, Pod traffic, management traffic, etc.).
* Secondary networks: Users can configure additional SubnetSets or Subnets to handle specialized traffic (e.g., dedicated Pod networking, NFS access). 

vsphere-network:
* Primary network: Continues to use the default Network in the Supervisor Namespace (no change).
* Secondary networks: Users can now specify additional Networks from the Supervisor Namespace as secondary networks.

Compatibility
If users do not specify custom networking configurations, the default Node network behavior remains unchanged.

Fixes #

Testing

CR captures in VDS network
VSphereMachine:
```
apiVersion: vmware.infrastructure.cluster.x-k8s.io/v1beta1
kind: VSphereMachine
metadata:
  annotations:
    cluster.x-k8s.io/cloned-from-groupkind: VSphereMachineTemplate.vmware.infrastructure.cluster.x-k8s.io
    cluster.x-k8s.io/cloned-from-name: coffee-vhczs
    run.tanzu.vmware.com/resolve-os-image: os-name=photon
  creationTimestamp: "2025-07-24T08:27:29Z"
  finalizers:
  - vspheremachine.infrastructure.cluster.x-k8s.io
  generation: 3
  labels:
    cluster.x-k8s.io/cluster-name: coffee
    cluster.x-k8s.io/control-plane: ""
    cluster.x-k8s.io/control-plane-name: coffee-dz7f8
    topology.cluster.x-k8s.io/owned: ""
  name: coffee-dz7f8-wrjjp
  namespace: tkgs-test
  ownerReferences:
  - apiVersion: cluster.x-k8s.io/v1beta2
    blockOwnerDeletion: true
    controller: true
    kind: Machine
    name: coffee-dz7f8-wrjjp
    uid: 95763ca4-2388-4620-b55c-cf29403ca829
  resourceVersion: "1443229"
  uid: b4e04d55-15b8-4583-8bdf-1f3c9bcb2cdf
spec:
  className: best-effort-small
  failureDomain: domain-c54
  imageName: vmi-6ce774c29ec50b536
  network:
    interfaces:
      secondary:
      - mtu: 1400
        name: eth1
        network:
          apiVersion: netoperator.vmware.com/v1alpha1
          kind: Network
          name: network-1
        routes:
        - to: 192.169.0.0/24
          via: 172.16.100.1
  powerOffMode: hard
  providerID: vsphere://09fbd57b-6cb8-49dc-909e-ecfb3f7db4b1
  storageClass: wcpglobal-storage-profile
status:
  addresses:
  - address: 192.168.128.16
    type: InternalIP
  conditions:
  - lastTransitionTime: "2025-07-24T08:36:54Z"
    status: "True"
    type: Ready
  - lastTransitionTime: "2025-07-24T08:36:54Z"
    status: "True"
    type: VMProvisioned
  network:
    interfaces:
    - deviceKey: 4000
      ip:
        addresses:
        - address: 192.168.128.16
          state: preferred
        - address: fe80::250:56ff:fe81:4ad0
          state: unknown
        macAddr: 00:50:56:81:4a:d0
      name: eth0
    - deviceKey: 4001
      ip:
        addresses:
        - address: 172.16.0.7
          state: preferred
        - address: fe80::250:56ff:fe81:b01e
          state: unknown
        macAddr: 00:50:56:81:b0:1e
      name: eth1
  ready: true
  v1beta2:
    conditions:
    - lastTransitionTime: "2025-07-24T08:36:54Z"
      message: ""
      observedGeneration: 3
      reason: Ready
      status: "True"
      type: Ready
    - lastTransitionTime: "2025-07-24T08:36:54Z"
      message: ""
      observedGeneration: 3
      reason: Provisioned
      status: "True"
      type: VirtualMachineProvisioned
    - lastTransitionTime: "2025-07-24T08:27:30Z"
      message: ""
      observedGeneration: 3
      reason: NotPaused
      status: "False"
      type: Paused
  vmID: 09fbd57b-6cb8-49dc-909e-ecfb3f7db4b1
  vmIp: 192.168.128.16
  vmstatus: ready
```
VirtualMachine:
```
apiVersion: vmoperator.vmware.com/v1alpha4
kind: VirtualMachine
metadata:
  annotations:
    virtualmachine.vmoperator.vmware.com/first-boot-done: "true"
    vmoperator.vmware.com/bootstrap-hash-configspec: 2e1472b57af294d1
    vmoperator.vmware.com/created-at-build-version: 0.1.0+99e27ec+1.8.6+4474c728+24861350
    vmoperator.vmware.com/created-at-schema-version: v1alpha4
    vmoperator.vmware.com/last-resized-vm-class: '{"name":"best-effort-small","uid":"481b313f-3e8d-4acb-bc61-caa960f14e63","generation":1}'
    vmoperator.vmware.com/manager-id: d97b29c1-8295-405a-8ebe-59ac2616a46f
    vmoperator.vmware.com/upgraded-to-build-version: 0.1.0+99e27ec+1.8.6+4474c728+24861350
    vmoperator.vmware.com/upgraded-to-schema-version: v1alpha4
    vsphere-cluster-module-group: control-plane-group
    vsphere-tag: CtrlVmVmAATag
  creationTimestamp: "2025-07-24T08:34:48Z"
  finalizers:
  - vmoperator.vmware.com/virtualmachine
  generation: 1
  labels:
    capv.vmware.com/cluster.name: coffee
    capv.vmware.com/cluster.role: controlplane
    capw.vmware.com/cluster.name: coffee
    capw.vmware.com/cluster.role: controlplane
    cluster.x-k8s.io/cluster-name: coffee
    topology.kubernetes.io/zone: domain-c54
  name: coffee-dz7f8-wrjjp
  namespace: tkgs-test
  ownerReferences:
  - apiVersion: vmware.infrastructure.cluster.x-k8s.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: VSphereMachine
    name: coffee-dz7f8-wrjjp
    uid: b4e04d55-15b8-4583-8bdf-1f3c9bcb2cdf
  resourceVersion: "1708307"
  uid: b148115a-cbeb-47eb-a8a3-c6157b470e88
spec:
  biosUUID: 09fbd57b-6cb8-49dc-909e-ecfb3f7db4b1
  bootstrap:
    cloudInit:
      instanceID: 09fbd57b-6cb8-49dc-909e-ecfb3f7db4b1
      rawCloudConfig:
        key: user-data
        name: coffee-dz7f8-wrjjp
  className: best-effort-small
  image:
    kind: ClusterVirtualMachineImage
    name: vmi-6ce774c29ec50b536
  imageName: vmi-6ce774c29ec50b536
  instanceUUID: 637c668f-47ae-474f-b75f-1e8044ee9a8c
  network:
    interfaces:
    - name: eth0
      network:
        apiVersion: netoperator.vmware.com/v1alpha1
        kind: Network
        name: primary
    - gateway4: None
      gateway6: None
      mtu: 1400
      name: eth1
      network:
        apiVersion: netoperator.vmware.com/v1alpha1
        kind: Network
        name: network-1
      routes:
      - to: 192.169.0.0/24
        via: 172.16.100.1
  powerOffMode: Hard
  powerState: PoweredOn
  promoteDisksMode: Online
  readinessProbe:
    tcpSocket:
      port: 6443
  reserved:
    resourcePolicyName: coffee
  restartMode: TrySoft
  storageClass: wcpglobal-storage-profile
  suspendMode: TrySoft
status:
  biosUUID: 09fbd57b-6cb8-49dc-909e-ecfb3f7db4b1
  changeBlockTracking: false
  class:
    apiVersion: vmoperator.vmware.com/v1alpha4
    kind: VirtualMachineClass
    name: best-effort-small
  conditions:
  - lastTransitionTime: "2025-07-24T08:40:10Z"
    message: ""
    reason: "True"
    status: "True"
    type: Ready
  - lastTransitionTime: "2025-07-24T08:36:31Z"
    message: ""
    reason: "True"
    status: "True"
    type: GuestCustomization
  - lastTransitionTime: "2025-07-24T08:34:48Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineBootstrapReady
  - lastTransitionTime: "2025-07-24T08:36:29Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineClassConfigurationSynced
  - lastTransitionTime: "2025-07-24T08:34:48Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineClassReady
  - lastTransitionTime: "2025-07-24T08:34:49Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineConditionPlacementReady
  - lastTransitionTime: "2025-07-24T08:34:48Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineConditionVMSetResourcePolicyReady
  - lastTransitionTime: "2025-07-24T08:36:27Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineCreated
  - lastTransitionTime: "2025-07-22T23:09:06Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineImageReady
  - lastTransitionTime: "2025-07-24T08:34:49Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineNetworkReady
  - lastTransitionTime: "2025-07-24T08:36:29Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineReconcileReady
  - lastTransitionTime: "2025-07-24T08:34:48Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineStorageReady
  - lastTransitionTime: "2025-07-24T08:36:54Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineTools
  hardwareVersion: 22
  instanceUUID: 637c668f-47ae-474f-b75f-1e8044ee9a8c
  network:
    config:
      dns:
        hostName: coffee-dz7f8-wrjjp
        nameservers:
        - 192.19.189.10
      interfaces:
      - dns:
          nameservers:
          - 192.19.189.10
        ip:
          addresses:
          - 192.168.128.16/16
          gateway4: 192.168.1.1
        name: eth0
      - dns:
          nameservers:
          - 192.19.189.10
        ip:
          addresses:
          - 172.16.0.7/16
        name: eth1
    hostName: coffee-dz7f8-wrjjp
    interfaces:
    - deviceKey: 4000
      ip:
        addresses:
        - address: 192.168.128.16
          state: preferred
        - address: fe80::250:56ff:fe81:4ad0
          state: unknown
        macAddr: 00:50:56:81:4a:d0
      name: eth0
    - deviceKey: 4001
      ip:
        addresses:
        - address: 172.16.0.7
          state: preferred
        - address: fe80::250:56ff:fe81:b01e
          state: unknown
        macAddr: 00:50:56:81:b0:1e
      name: eth1
    ipStacks:
    - dns:
        domainName: .
        hostName: coffee-dz7f8-wrjjp
        nameservers:
        - 192.19.189.10
        searchDomains:
        - .
      ipRoutes:
      - gateway:
          address: 192.168.1.1
          device: "0"
        networkAddress: 0.0.0.0/0
      - gateway:
          device: "1"
        networkAddress: 172.16.0.0/16
      - gateway:
          device: "0"
        networkAddress: 192.168.0.0/16
      - gateway:
          address: 172.16.100.1
          device: "1"
        networkAddress: 192.169.0.0/24
    primaryIP4: 192.168.128.16
  nodeName: 10.161.80.240
  powerState: PoweredOn
  storage:
    requested:
      disks: 20Gi
    total: 24Gi
    usage:
      disks: 11Gi
      other: 4Gi
  uniqueID: vm-120
  volumes:
  - attached: true
    diskUUID: 6000C29e-69c3-f1ac-85e1-c3fa70d249a0
    limit: 20Gi
    name: coffee-dz7f8-wrjjp
    requested: 20Gi
    type: Classic
    used: 11Gi
  zone: domain-c54
```
OS capture
```
vmware-system-user@coffee-dz7f8-wrjjp [ ~ ]$ sudo ip route
default via 192.168.1.1 dev eth0 proto static
172.16.0.0/16 dev eth1 proto kernel scope link src 172.16.0.7
192.0.2.0/24 via 192.0.2.1 dev antrea-gw0 onlink
192.0.3.0/24 via 192.0.3.1 dev antrea-gw0 onlink
192.0.5.0/24 dev antrea-gw0 scope link src 192.0.5.1
192.168.0.0/16 dev eth0 proto kernel scope link src 192.168.128.16
192.169.0.0/24 via 172.16.100.1 dev eth1 proto static
vmware-system-user@coffee-dz7f8-wrjjp [ ~ ]$ ip a
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc prio state UP group default qlen 1000
    link/ether 00:50:56:81:4a:d0 brd ff:ff:ff:ff:ff:ff
    altname eno1
    altname enp11s0
    altname ens192
    inet 192.168.128.16/16 brd 192.168.255.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::250:56ff:fe81:4ad0/64 scope link
       valid_lft forever preferred_lft forever
3: eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1400 qdisc prio state UP group default qlen 1000
    link/ether 00:50:56:81:b0:1e brd ff:ff:ff:ff:ff:ff
    altname eno2
    altname enp19s0
    altname ens224
    inet 172.16.0.7/16 brd 172.16.255.255 scope global eth1
       valid_lft forever preferred_lft forever
    inet6 fe80::250:56ff:fe81:b01e/64 scope link
       valid_lft forever preferred_lft forever
```

CR captures in VPC network
VSphereCluster:
```
apiVersion: vmware.infrastructure.cluster.x-k8s.io/v1beta1
kind: VSphereCluster
metadata:
  annotations:
    cluster.x-k8s.io/cloned-from-groupkind: VSphereClusterTemplate.vmware.infrastructure.cluster.x-k8s.io
    cluster.x-k8s.io/cloned-from-name: tkc-infrastructure-v3.5.0
  creationTimestamp: "2025-07-29T03:49:43Z"
  finalizers:
  - vspherecluster.vmware.infrastructure.cluster.x-k8s.io
  generation: 2
  labels:
    cluster.x-k8s.io/cluster-name: tea
    topology.cluster.x-k8s.io/owned: ""
  name: tea-scdmz
  namespace: tkgs-test
  ownerReferences:
  - apiVersion: cluster.x-k8s.io/v1beta2
    blockOwnerDeletion: true
    controller: true
    kind: Cluster
    name: tea
    uid: 95e7fdf0-c40b-4288-9044-68685619563e
  resourceVersion: "5177935"
  uid: 9209aadc-e952-49ac-ab15-571d1df1989e
spec:
  controlPlaneEndpoint:
    host: 192.168.0.8
    port: 6443
  network:
    nsxVPC:
      createSubnetSet: false
status:
  conditions:
  - lastTransitionTime: "2025-07-29T03:50:01Z"
    status: "True"
    type: Ready
  - lastTransitionTime: "2025-07-29T07:57:08Z"
    status: "True"
    type: ClusterNetworkReady
  - lastTransitionTime: "2025-07-29T03:50:01Z"
    status: "True"
    type: LoadBalancerReady
  - lastTransitionTime: "2025-07-29T05:32:13Z"
    status: "True"
    type: ProviderServiceAccountsReady
  - lastTransitionTime: "2025-07-29T03:49:45Z"
    status: "True"
    type: ResourcePolicyReady
  - lastTransitionTime: "2025-07-29T05:32:12Z"
    status: "True"
    type: ServiceDiscoveryReady
  failureDomains:
    domain-c11:
      controlPlane: true
  ready: true
  resourcePolicyName: tea
  v1beta2:
    conditions:
    - lastTransitionTime: "2025-07-29T07:57:08Z"
      message: ""
      observedGeneration: 2
      reason: Ready
      status: "True"
      type: Ready
    - lastTransitionTime: "2025-07-29T03:50:01Z"
      message: ""
      observedGeneration: 2
      reason: Ready
      status: "True"
      type: LoadBalancerReady
    - lastTransitionTime: "2025-07-29T07:57:08Z"
      message: ""
      observedGeneration: 2
      reason: Ready
      status: "True"
      type: NetworkReady
    - lastTransitionTime: "2025-07-29T05:32:12Z"
      message: ""
      observedGeneration: 2
      reason: Ready
      status: "True"
      type: ProviderServiceAccountsReady
    - lastTransitionTime: "2025-07-29T03:49:45Z"
      message: ""
      observedGeneration: 2
      reason: Ready
      status: "True"
      type: ResourcePolicyReady
    - lastTransitionTime: "2025-07-29T05:32:12Z"
      message: ""
      observedGeneration: 2
      reason: Ready
      status: "True"
      type: ServiceDiscoveryReady
    - lastTransitionTime: "2025-07-29T03:49:43Z"
      message: ""
      observedGeneration: 2
      reason: NotPaused
      status: "False"
      type: Paused
```
VSphereMachine:
```
apiVersion: vmware.infrastructure.cluster.x-k8s.io/v1beta1
kind: VSphereMachine
metadata:
  annotations:
    cluster.x-k8s.io/cloned-from-groupkind: VSphereMachineTemplate.vmware.infrastructure.cluster.x-k8s.io
    cluster.x-k8s.io/cloned-from-name: tea-workers-xzcdw
    run.tanzu.vmware.com/resolve-os-image: os-name=photon
  creationTimestamp: "2025-07-29T03:50:19Z"
  finalizers:
  - vspheremachine.infrastructure.cluster.x-k8s.io
  generation: 2
  labels:
    cluster.x-k8s.io/cluster-name: tea
    cluster.x-k8s.io/deployment-name: tea-workers-hcblk
    cluster.x-k8s.io/set-name: tea-workers-hcblk-4sjjg
    machine-template-hash: 869863088-4sjjg
    topology.cluster.x-k8s.io/deployment-name: workers
    topology.cluster.x-k8s.io/owned: ""
  name: tea-workers-hcblk-4sjjg-4nlp4
  namespace: tkgs-test
  ownerReferences:
  - apiVersion: cluster.x-k8s.io/v1beta2
    blockOwnerDeletion: true
    controller: true
    kind: Machine
    name: tea-workers-hcblk-4sjjg-4nlp4
    uid: eabff8dc-786f-4c24-a06c-010f23360dd6
  resourceVersion: "5198126"
  uid: 221b7ac9-ff18-416e-904b-d9326fd4e66a
spec:
  className: best-effort-small
  imageName: vmi-9db3ea41fd94a92bb
  network:
    interfaces:
      primary:
        network:
          apiVersion: crd.nsx.vmware.com/v1alpha1
          kind: SubnetSet
          name: secondary-node-2
      secondary:
      - name: eth1
        network:
          apiVersion: crd.nsx.vmware.com/v1alpha1
          kind: SubnetSet
          name: secondary-node-1
  powerOffMode: hard
  providerID: vsphere://6ef42af8-4548-4673-a42b-6517b25f8240
  storageClass: wcpglobal-storage-profile
status:
  addresses:
  - address: 192.168.0.37
    type: InternalIP
  conditions:
  - lastTransitionTime: "2025-07-29T04:11:44Z"
    status: "True"
    type: Ready
  - lastTransitionTime: "2025-07-29T04:11:44Z"
    status: "True"
    type: VMProvisioned
  network:
    interfaces:
    - deviceKey: 4000
      ip:
        addresses:
        - address: 192.168.0.37
          state: preferred
        - address: fe80::650:56ff:fe00:3c00
          state: unknown
        macAddr: 04:50:56:00:3c:00
      name: eth0
    - deviceKey: 4001
      ip:
        addresses:
        - address: 10.246.0.5
          state: preferred
        - address: fe80::650:56ff:fe00:3c01
          state: unknown
        macAddr: 04:50:56:00:3c:01
      name: eth1
  ready: true
  v1beta2:
    conditions:
    - lastTransitionTime: "2025-07-29T04:11:44Z"
      message: ""
      observedGeneration: 2
      reason: Ready
      status: "True"
      type: Ready
    - lastTransitionTime: "2025-07-29T04:11:44Z"
      message: ""
      observedGeneration: 2
      reason: Provisioned
      status: "True"
      type: VirtualMachineProvisioned
    - lastTransitionTime: "2025-07-29T03:50:20Z"
      message: ""
      observedGeneration: 2
      reason: NotPaused
      status: "False"
      type: Paused
  vmID: 6ef42af8-4548-4673-a42b-6517b25f8240
  vmIp: 192.168.0.37
  vmstatus: ready
```
VirtualMachine:
```
apiVersion: vmoperator.vmware.com/v1alpha4
kind: VirtualMachine
metadata:
  annotations:
    virtualmachine.vmoperator.vmware.com/first-boot-done: "true"
    vmoperator.vmware.com/bootstrap-hash-configspec: 2e1472b57af294d1
    vmoperator.vmware.com/created-at-build-version: 0.1.0+05d6e95+1.8.6+0ee7ee38+87995934
    vmoperator.vmware.com/created-at-schema-version: v1alpha4
    vmoperator.vmware.com/last-resized-vm-class: '{"name":"best-effort-small","uid":"f4dec203-4f27-4983-848d-118c3b99cdb7","generation":1}'
    vmoperator.vmware.com/manager-id: 0216e57c-4233-461b-b78e-cdb6a8dd2f63
    vmoperator.vmware.com/upgraded-to-build-version: 0.1.0+05d6e95+1.8.6+0ee7ee38+87995934
    vmoperator.vmware.com/upgraded-to-schema-version: v1alpha4
    vsphere-cluster-module-group: tea-workers-0
    vsphere-tag: WorkerVmVmAATag
  creationTimestamp: "2025-07-29T04:04:54Z"
  finalizers:
  - vmoperator.vmware.com/virtualmachine
  generation: 1
  labels:
    capv.vmware.com/cluster.name: tea
    capv.vmware.com/cluster.role: node
    capw.vmware.com/cluster.name: tea
    capw.vmware.com/cluster.role: node
    cluster.x-k8s.io/cluster-name: tea
    topology.kubernetes.io/zone: domain-c11
  name: tea-workers-hcblk-4sjjg-4nlp4
  namespace: tkgs-test
  ownerReferences:
  - apiVersion: vmware.infrastructure.cluster.x-k8s.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: VSphereMachine
    name: tea-workers-hcblk-4sjjg-4nlp4
    uid: 221b7ac9-ff18-416e-904b-d9326fd4e66a
  resourceVersion: "5198124"
  uid: 3bf9a441-37fc-4087-b314-9a07618c447b
spec:
  biosUUID: 6ef42af8-4548-4673-a42b-6517b25f8240
  bootstrap:
    cloudInit:
      instanceID: 6ef42af8-4548-4673-a42b-6517b25f8240
      rawCloudConfig:
        key: user-data
        name: tea-workers-hcblk-4sjjg-4nlp4
  className: best-effort-small
  image:
    kind: ClusterVirtualMachineImage
    name: vmi-9db3ea41fd94a92bb
  imageName: vmi-9db3ea41fd94a92bb
  instanceUUID: 29f0b424-86b4-44a5-9119-6b6764f96695
  network:
    interfaces:
    - name: eth0
      network:
        apiVersion: crd.nsx.vmware.com/v1alpha1
        kind: SubnetSet
        name: secondary-node-2
    - gateway4: None
      gateway6: None
      name: eth1
      network:
        apiVersion: crd.nsx.vmware.com/v1alpha1
        kind: SubnetSet
        name: secondary-node-1
  powerOffMode: Hard
  powerState: PoweredOn
  promoteDisksMode: Online
  reserved:
    resourcePolicyName: tea
  restartMode: TrySoft
  storageClass: wcpglobal-storage-profile
  suspendMode: TrySoft
status:
  biosUUID: 6ef42af8-4548-4673-a42b-6517b25f8240
  changeBlockTracking: false
  class:
    apiVersion: vmoperator.vmware.com/v1alpha4
    kind: VirtualMachineClass
    name: best-effort-small
  conditions:
  - lastTransitionTime: "2025-07-29T04:11:00Z"
    message: ""
    reason: "True"
    status: "True"
    type: GuestCustomization
  - lastTransitionTime: "2025-07-29T04:04:55Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineBootstrapReady
  - lastTransitionTime: "2025-07-29T04:08:18Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineClassConfigurationSynced
  - lastTransitionTime: "2025-07-29T04:04:55Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineClassReady
  - lastTransitionTime: "2025-07-29T04:05:00Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineConditionPlacementReady
  - lastTransitionTime: "2025-07-29T04:04:55Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineConditionVMSetResourcePolicyReady
  - lastTransitionTime: "2025-07-29T04:08:18Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineCreated
  - lastTransitionTime: "2025-07-23T10:08:58Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineImageReady
  - lastTransitionTime: "2025-07-29T04:05:00Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineNetworkReady
  - lastTransitionTime: "2025-07-29T04:08:18Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineReconcileReady
  - lastTransitionTime: "2025-07-29T04:04:55Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineStorageReady
  - lastTransitionTime: "2025-07-29T08:31:38Z"
    message: ""
    reason: "True"
    status: "True"
    type: VirtualMachineTools
  hardwareVersion: 22
  instanceUUID: 29f0b424-86b4-44a5-9119-6b6764f96695
  network:
    config:
      dns:
        hostName: tea-workers-hcblk-4sjjg-4nlp4
        nameservers:
        - 192.19.189.10
        - 192.19.189.20
      interfaces:
      - dns:
          nameservers:
          - 192.19.189.10
          - 192.19.189.20
        ip:
          addresses:
          - 192.168.0.37/27
          gateway4: 192.168.0.33
        name: eth0
      - dns:
          nameservers:
          - 192.19.189.10
          - 192.19.189.20
        ip:
          addresses:
          - 10.246.0.5/27
        name: eth1
    hostName: tea-workers-hcblk-4sjjg-4nlp4
    interfaces:
    - deviceKey: 4000
      ip:
        addresses:
        - address: 192.168.0.37
          state: preferred
        - address: fe80::650:56ff:fe00:3c00
          state: unknown
        macAddr: 04:50:56:00:3c:00
      name: eth0
    - deviceKey: 4001
      ip:
        addresses:
        - address: 10.246.0.5
          state: preferred
        - address: fe80::650:56ff:fe00:3c01
          state: unknown
        macAddr: 04:50:56:00:3c:01
      name: eth1
    ipStacks:
    - dns:
        domainName: .
        hostName: tea-workers-hcblk-4sjjg-4nlp4
        nameservers:
        - 192.19.189.10
        - 192.19.189.20
        searchDomains:
        - .
      ipRoutes:
      - gateway:
          address: 192.168.0.33
          device: "0"
        networkAddress: 0.0.0.0/0
      - gateway:
          device: "1"
        networkAddress: 10.246.0.0/27
      - gateway:
          device: "0"
        networkAddress: 192.168.0.32/27
    primaryIP4: 192.168.0.37
  nodeName: 10.161.248.173
  powerState: PoweredOn
  storage:
    requested:
      disks: 20Gi
    total: 20Gi
    usage:
      disks: 17Gi
      other: "0"
  uniqueID: vm-87
  volumes:
  - attached: true
    diskUUID: 6000C290-07d2-ed82-d530-f6ef511dd295
    limit: 20Gi
    name: tea-workers-hcblk-4sjjg-4nlp4
    requested: 20Gi
    type: Classic
    used: 17Gi
  zone: domain-c11
```
OS capture
```
root@tea-workers-hcblk-4sjjg-4nlp4 [ ~ ]# ip a
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc prio state UP group default qlen 1000
    link/ether 04:50:56:00:3c:00 brd ff:ff:ff:ff:ff:ff
    altname eno1
    altname enp11s0
    altname ens192
    inet 192.168.0.37/27 brd 192.168.0.63 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::650:56ff:fe00:3c00/64 scope link
       valid_lft forever preferred_lft forever
3: eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc prio state UP group default qlen 1000
    link/ether 04:50:56:00:3c:01 brd ff:ff:ff:ff:ff:ff
    altname eno2
    altname enp19s0
    altname ens224
    inet 10.246.0.5/27 brd 10.246.0.31 scope global eth1
       valid_lft forever preferred_lft forever
    inet6 fe80::650:56ff:fe00:3c01/64 scope link
       valid_lft forever preferred_lft forever
root@tea-workers-hcblk-4sjjg-4nlp4 [ ~ ]# ip route
default via 192.168.0.33 dev eth0 proto static
10.246.0.0/27 dev eth1 proto kernel scope link src 10.246.0.5
192.0.0.0/24 via 192.0.0.1 dev antrea-gw0 onlink
192.0.1.0/24 dev antrea-gw0 scope link src 192.0.1.1
192.0.2.0/24 via 192.0.2.1 dev antrea-gw0 onlink
192.168.0.32/27 dev eth0 proto kernel scope link src 192.168.0.37
```


apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -34,7 +35,95 @@ type VSphereMachineVolume struct {
...
+// InterfacesSpec defines all the network interfaces of a VSphereMachine from Kubernetes perspective.
+// +kubebuilder:validation:XValidation:rule="has(self.primary) == has(oldSelf.primary)",message="field 'primary' cannot be added or removed after creation"
+type InterfacesSpec struct {
+	// Primary is the primary network interface.
+	//
+	// It is used to connect the Kubernetes primary network for Load balancer,
+	// Service discovery, Pod traffic and management traffic etc.
+	// Leave it unset if you don't want to customize the primary network and interface.
+	// Customization is only supported with network provider NSX-VPC.
+	// It should be set only when VSphereCluster spec.network.nsxVPC.createSubnetSet is set to false.
```
silvery1622, line 129: // VSphereCluster spec.network.nsxVPC.createSubnetSet should be set as false when this is configured.

config/manager/manager.yaml
```diff
@@ -21,7 +21,7 @@ spec:
         - "--diagnostics-address=${CAPI_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPI_INSECURE_DIAGNOSTICS:=false}"
         - --v=4
-        - "--feature-gates=NodeAntiAffinity=${EXP_NODE_ANTI_AFFINITY:=false},NamespaceScopedZones=${EXP_NAMESPACE_SCOPED_ZONES:=false},PriorityQueue=${EXP_PRIORITY_QUEUE:=false}"
+        - "--feature-gates=MultiNetworks=${MULTI_NETWORKS:=false},NodeAntiAffinity=${EXP_NODE_ANTI_AFFINITY:=false},NamespaceScopedZones=${EXP_NAMESPACE_SCOPED_ZONES:=false},PriorityQueue=${EXP_PRIORITY_QUEUE:=false}"
```
JackieLan: should it be true by default?

DanielXiao: This is alpha feature of upstream. We will enable it by default in downstream product.

JackieLan: thanks. good to know that it will be enabled by default in downstream.

apis/vmware/v1beta1/vspherecluster_types.go
```diff
@@ -123,10 +123,28 @@ const (
 	VSphereClusterServiceDiscoveryNotReadyV1Beta2Reason = clusterv1beta1.NotReadyV1Beta2Reason
 )
```
sbueringer, line 126: Some guidance for API fields (should make it easier to fix all the KAL findings)
* Every field:
  * must have an `+optional` or an `+required` marker (no implicit optional or required)
  * must have `omitempty`
  * must not use type `float`
  * must not use type `metav1.Duration`, use `int32` and unit suffix (e.g. `nodeDrainTimeoutSeconds int32`)
  * field comment must start with `JSON` name of the field (e.g. `CreateSubnetSet` => `// createSubnetSet`)
* Every field that has a struct type
  * must have `omitzero`
  * should not be pointers (in 99% of the cases it does not make sense to allow users to explicitly set `{}`)
    * If we need an easy way to check if the field is set, let's add a `IsDefined` method (e.g. https://github.com/kubernetes-sigs/cluster-api/blob/main/api/core/v1beta2/cluster_types.go#L571-L572)
  * the struct type should usually either have a required field or the `+kubebuilder:validation:MinProperties=1` marker (in 99% of the cases it does not make sense to allow users to explicitly set `{}`)
* Every array/slice field
  * must have `+kubebuilder:validation:MinItems` and `+kubebuilder:validation:MaxItems` markers (in 99% of the cases `MinItems` will be 1 as setting an array field explicitly to an empty array very rarely makes sense)
  * must have SSA tags, usually this will be `+listType=atomic` (let's talk about the cases, tracking fine-granular ownerwship with `map` takes up a lot of memory)
* Every bool field:
  * must be a pointer (type `*bool)`
* Every string field:
  * must have `+kubebuilder:validation:MinLength` and `+kubebuilder:validation:MaxLength` (in 99% of the cases `MinLength` will be 1 as setting a string field explicitly to an empty string very rarely makes sense)
  * If `MinLength=0` field must be `*string`, otherwise `string`
* Every int field:
  * must usually be of type `int32` (use `int64` if appropriate, never use `int`)
  * must have a `+kubebuilder:validation:Minimum` marker
  * should have a `+kubebuilder:validation:Maximum` marker if there is a maximum
  * If `Minimum=0` field must be `*int32`, otherwise `int32`

Additional infos here: https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md

apis/vmware/v1beta1/vspherecluster_types.go
```diff
@@ -123,10 +123,28 @@ const (
...
+	// CreateSubnetSet is a flag to indicate whether to create a SubnetSet or not as the primary network. If not set, the default is true.
+	// +kubebuilder:validation:Optional
+	CreateSubnetSet *bool `json:"createSubnetSet,omitempty"`
+}
+
+// Network defines the network configuration for the cluster with different network providers.
+// +kubebuilder:validation:XValidation:rule="has(self.nsxVPC) == has(oldSelf.nsxVPC)",message="field 'nsxVPC' cannot be added or removed after creation"
+type Network struct {
+	// +kubebuilder:validation:Optional
+	NsxVPC *NSXVPC `json:"nsxVPC,omitempty"`
```
sbueringer:
```suggestion
	NSXVPC *NSXVPC `json:"nsxVPC,omitempty"`
```
I think

(only checking for findings that KAL probably doesn't catch for this review)

apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -34,7 +35,95 @@ type VSphereMachineVolume struct {
...
+	//
+	// Please note this feature is available only with the following bootstrap
+	// providers: CloudInit.
+	//
+	// +kubebuilder:validation:Optional
+	Routes []RouteSpec `json:"routes,omitempty"`
+}
+
+// InterfaceSpec defines a network interface for a VSphereMachine.
+type InterfaceSpec struct {
```
sbueringer: What about calling this SecondaryInterfaceSpec and then InterfaceProperty => InterfaceSpec?

apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -34,7 +35,95 @@ type VSphereMachineVolume struct {
...
+	// Via is an IP4 address. IP6 is not supported yet.
+	//
+	// +kubebuilder:validation:Pattern=`^([0-9]{1,3}\.){3}[0-9]{1,3}$`
+	Via string `json:"via"`
+}
+
+// PartialObjectRef describes a reference to another object in the same
+// namespace as the referrer. The reference can be just a name but may also
+// include the referred resource's APIVersion and Kind.
+type PartialObjectRef struct {
```
sbueringer: Let's call this `InterfaceNetworkRefeference`

apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -34,7 +35,95 @@ type VSphereMachineVolume struct {
...
+	metav1.TypeMeta `json:",inline"`
```
sbueringer: Please add Kind and APIVersion directly

Please copy over the validation from: https://github.com/kubernetes-sigs/cluster-api/blob/3adb25101d845af093cc44975b67cc1b94838e2e/api/core/v1beta2/machinehealthcheck_types.go#L160-L183

apis/vmware/v1beta1/vspheremachinenetworkstatus_types.go
```diff
@@ -0,0 +1,216 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
zhanggbj: Please update the copyright with the current year.

apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -34,7 +35,95 @@ type VSphereMachineVolume struct {
...
+// InterfaceProperty defines properties of a network interface.
+type InterfaceProperty struct {
+	// Network is the name of the network resource to which this interface is
+	// connected.
+	Network PartialObjectRef `json:"network"`
+
+	// MTU is the Maximum Transmission Unit size in bytes.
+	//
+	// +kubebuilder:validation:Optional
+	MTU *int64 `json:"mtu,omitempty"`
```
sbueringer: Please add a Minimum validation marker

I assume Minimum is > 0, if yes then we can drop the pointer

I also assume we don't need int64, then let's make it a int32

apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -34,7 +35,95 @@ type VSphereMachineVolume struct {
 	StorageClass string `json:"storageClass,omitempty"`
 }
 
+// RouteSpec defines a static route for a guest.
```
sbueringer: nit: please order the types topologically. Just easier to read & review, i.e.
* VSphereMachineSpec
* VSphereMachineNetworkSpec
* InterfacesSpec
* InterfaceSpec
* InterfaceProperty
* PartialObjectRef
* RouteSpec

(let's only do this here, I think in the other file it makes sense to keep the order the same as in vm-operator to make diffing easier)

apis/vmware/v1beta1/vspheremachinenetworkstatus_types.go
```diff
@@ -0,0 +1,216 @@
...
+	// +optional
+	DNS *VSphereMachineNetworkDNSStatus `json:"dns,omitempty"`
+}
+
+// VSphereMachineNetworkStatus defines the observed state of a VM's
+// network configuration.
+//
+// This a mirror of the v1alpha2 VirtualMachineNetworkStatus. See
+// https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha2/virtualmachine_network_types.go
+// for more information. When vm-operator v1alpha2 is updated, this type need to be synchronized.
```
sbueringer: > When vm-operator v1alpha2 is updated, this type need to be synchronized

just fyi. This won't entirely be possible. We can make compatible changes, but not breaking ones (e.g. change types of fields)

If there are breaking changes we cannot bump our apiVersion frequently, also if eventually we might have to work against multiple vm-operator apiVersions at the same time. But for now it's fine, just wanted to share

zhanggbj: Looks like `VSphereMachineNetworkInterfaceStatus` is a full copy of `VirtualMachineNetworkInterfaceStatus` as you mentioned, I have a concern we may have drift in future and it's hard to notice if vm-operator v1alpha2 changes. And this will be covered in future  vm-operator API bump.

should we just reuse vmoprv1.VirtualMachineNetworkInterfaceStatus here?

CC @sbueringer 

```

vmoprv1 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"

type VSphereMachineNetworkStatus struct {
	// Interfaces describes the status of the VM's network interfaces.
	//
	// +optional
	Interfaces []vmoprv1.VirtualMachineNetworkInterfaceStatus `json:"interfaces,omitempty"`
}
```

internal/webhooks/vmware/vspherecluster.go
```diff
@@ -0,0 +1,91 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
zhanggbj: please fix the copyright year

internal/webhooks/vmware/vspherecluster_test.go
```diff
@@ -0,0 +1,278 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
zhanggbj: Same as above

apis/vmware/v1beta1/vspherecluster_types.go
```diff
@@ -123,10 +123,28 @@ const (
 	VSphereClusterServiceDiscoveryNotReadyV1Beta2Reason = clusterv1beta1.NotReadyV1Beta2Reason
 )
 
+// +kubebuilder:validation:XValidation:rule="has(self.createSubnetSet) == has(oldSelf.createSubnetSet) && self.createSubnetSet == oldSelf.createSubnetSet",message="CreateSubnetSet value cannot be changed after creation"
```
sbueringer:
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.createSubnetSet) == has(oldSelf.createSubnetSet) && self.createSubnetSet == oldSelf.createSubnetSet",message="createSubnetSet value cannot be changed after creation"
```
nit: so the user-facing error also shows the user-facing field name

config/manager/manager.yaml
sbueringer:
```suggestion
        - "--feature-gates=MultiNetworks=${EXP_MULTI_NETWORKS:=false},NodeAntiAffinity=${EXP_NODE_ANTI_AFFINITY:=false},NamespaceScopedZones=${EXP_NAMESPACE_SCOPED_ZONES:=false},PriorityQueue=${EXP_PRIORITY_QUEUE:=false}"
```
nit (same change in e.g. vsphere.yaml of course)

internal/webhooks/vmware/vspherecluster_test.go
```diff
@@ -0,0 +1,278 @@
...
+					g.Expect(err.Error()).To(ContainSubstring(tc.errMsg))
+				}
+			} else {
+				g.Expect(err).NotTo(HaveOccurred())
+			}
+		})
+	}
+}
+
+func TestVSphereCluster_ValidateUpdate(t *testing.T) {
```
sbueringer, line 153: @DanielXiao Let's do a follow-up PR after the RC release to add test coverage for the CEL expressions (we'll have to start using envtest for that)

internal/webhooks/vmware/vspheremachine.go
```diff
@@ -55,8 +62,15 @@ func (webhook *VSphereMachine) Default(_ context.Context, _ runtime.Object) erro
...
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type.
-func (webhook *VSphereMachine) ValidateCreate(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
-	return nil, nil
+func (webhook *VSphereMachine) ValidateCreate(_ context.Context, object runtime.Object) (admission.Warnings, error) {
+	objTyped, ok := object.(*vmwarev1.VSphereMachine)
+	if !ok {
+		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a VSphereMachine but got a %T", object))
+	}
+
+	allErrs := validateInterfaces(webhook.NetworkProvider, objTyped.Spec.Network, []string{})
```
sbueringer:
```suggestion
	allErrs := validateInterfaces(webhook.NetworkProvider, objTyped.Spec.Network, field.NewPath("spec", "network"))
```
nit: Let's do this instead please. You can then use append accordingly in the func

Also I would rename the func to validateNetwork as that is what we are passing into the func (even if we only validate the interfaces field for now)

(see the corresponding comment in the validateInterfaces func with an example on how to use fldPath then)

internal/webhooks/vmware/vspheremachine.go
```diff
@@ -96,10 +110,88 @@ func (webhook *VSphereMachine) ValidateUpdate(_ context.Context, oldRaw runtime.
...
+
 	return nil, webhooks.AggregateObjErrors(newTyped.GroupVersionKind().GroupKind(), newTyped.Name, allErrs)
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
 func (webhook *VSphereMachine) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
 	return nil, nil
 }
+
+func validateInterfaces(networkProvider string, network *vmwarev1.VSphereMachineNetworkSpec, pathPrefix []string) field.ErrorList {
```
sbueringer:
```suggestion
func validateInterfaces(networkProvider string, network *vmwarev1.VSphereMachineNetworkSpec, fldPath *field.Path) field.ErrorList {
```

internal/webhooks/vmware/vspheremachine.go
```diff
@@ -96,10 +110,88 @@ func (webhook *VSphereMachine) ValidateUpdate(_ context.Context, oldRaw runtime.
...
+	var allErrs field.ErrorList
+
+	interfacesPath := append(pathPrefix, "spec", "network", "interfaces")
+	primaryPath := append(interfacesPath, "primary")
+	primaryNetworkPath := append(primaryPath, "primary", "network")
+
+	if network != nil && network.Interfaces != nil {
+		if !feature.Gates.Enabled(feature.MultiNetworks) {
+			allErrs = append(allErrs, field.Forbidden(
+				field.NewPath(interfacesPath[0], interfacesPath[1:]...),
```
sbueringer:
```suggestion
				fldPath.Child("interfaces"),
```
same for other paths below.

Or alternatively define the following above and then use `.Child` when necessary
```go
	interfacesPath := fldPath.Child("interfaces")
```

(I personally would pick the first option, because it's easier to check if the paths are all correct)

apis/vmware/v1beta1/vspheremachine_types.go
sbueringer: You probably want to add a GroupVersionKind method as it is used by the webhook code

internal/webhooks/vmware/vspheremachine.go
```diff
@@ -96,10 +110,88 @@ func (webhook *VSphereMachine) ValidateUpdate(_ context.Context, oldRaw runtime.
...
 func (webhook *VSphereMachine) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
 	return nil, nil
 }
+
+func validateInterfaces(networkProvider string, network *vmwarev1.VSphereMachineNetworkSpec, pathPrefix []string) field.ErrorList {
+	var allErrs field.ErrorList
+
+	interfacesPath := append(pathPrefix, "spec", "network", "interfaces")
+	primaryPath := append(interfacesPath, "primary")
+	primaryNetworkPath := append(primaryPath, "primary", "network")
```
sbueringer: this path has primary twice (but see my comment below first)

internal/webhooks/vmware/vspheremachine.go
```diff
@@ -96,10 +110,88 @@ func (webhook *VSphereMachine) ValidateUpdate(_ context.Context, oldRaw runtime.
...
+		} else {
+			// Validate network type is supported
+			switch networkProvider {
+			case manager.NSXVPCNetworkProvider:
+				primary := network.Interfaces.Primary
+				if primary != nil && primary.Network.TypeMeta.GroupVersionKind() != pkgnetwork.NetworkGVKNSXTVPCSubnetSet {
+					allErrs = append(allErrs, field.Invalid(
+						field.NewPath(primaryNetworkPath[0], primaryNetworkPath[1:]...),
+						primary.Network.TypeMeta.GroupVersionKind(),
+						fmt.Sprintf("only support %s", pkgnetwork.NetworkGVKNSXTVPCSubnetSet)))
```
sbueringer:
```suggestion
						fmt.Sprintf("only supports %s", pkgnetwork.NetworkGVKNSXTVPCSubnetSet)))
```
nit: same below

internal/webhooks/vmware/vspheremachine.go
sbueringer: Also below we should use .Index(i) instead of building a string ourselves. Here a full example to show how clean this can look:
```go
			// Validate network type is supported
			switch networkProvider {
			case manager.NSXVPCNetworkProvider:
				primary := network.Interfaces.Primary
				if primary != nil && primary.Network.TypeMeta.GroupVersionKind() != pkgnetwork.NetworkGVKNSXTVPCSubnetSet {
					allErrs = append(allErrs, field.Invalid(
						fldPath.Child("interfaces", "primary", "network"),
						primary.Network.TypeMeta.GroupVersionKind(),
						fmt.Sprintf("only support %s", pkgnetwork.NetworkGVKNSXTVPCSubnetSet)))
				}
				for i, secondaryInterface := range network.Interfaces.Secondary {
					if secondaryInterface.Network.TypeMeta.GroupVersionKind() != pkgnetwork.NetworkGVKNSXTVPCSubnetSet &&
						secondaryInterface.Network.TypeMeta.GroupVersionKind() != pkgnetwork.NetworkGVKNSXTVPCSubnet {
						allErrs = append(allErrs, field.Invalid(
							fldPath.Child("interfaces", "secondary").Index(i).Child("network"),
							secondaryInterface.Network.TypeMeta.GroupVersionKind(),
							fmt.Sprintf("only support %s or %s", pkgnetwork.NetworkGVKNSXTVPCSubnetSet, pkgnetwork.NetworkGVKNSXTVPCSubnet)))
					}
				}
			case manager.VDSNetworkProvider:
				if network.Interfaces.Primary != nil {
					allErrs = append(allErrs, field.Forbidden(
						fldPath.Child("interfaces", "primary"),
						"primary interface can not be set when network provider is vsphere-network"))
				}
				for i, secondaryInterface := range network.Interfaces.Secondary {
					if secondaryInterface.Network.TypeMeta.GroupVersionKind() != pkgnetwork.NetworkGVKNetOperator {
						allErrs = append(allErrs, field.Invalid(
							fldPath.Child("interfaces", "secondary").Index(i).Child("network"),
							secondaryInterface.Network.TypeMeta.GroupVersionKind(),
							fmt.Sprintf("only support %s", pkgnetwork.NetworkGVKNetOperator)))
					}
				}
			default:
				allErrs = append(allErrs, field.Forbidden(fldPath.Child("interfaces"), fmt.Sprintf("interfaces can not be set when network provider is %s", networkProvider)))
			}
```

I would suggest to follow that pattern everywhere (let's only modify code added in this PR of course)

internal/webhooks/vmware/vspheremachine_test.go
```diff
@@ -92,3 +97,327 @@ func createVSphereMachine(providerID *string, imageName, className, storageClass
 
 	return vSphereMachine
 }
+
+func TestVSphereMachine_ValidateCreate_MultiNetwork(t *testing.T) {
+	// Explicitly enable the feature gate for all tests except the negative one
+	featuregatetesting.SetFeatureGateDuringTest(t, feature.Gates, feature.MultiNetworks, true)
```
sbueringer: nit: I think instead of doing this here and then setting ptrToBool(false) or nil in every test case you can also just set true/false in every test case and enable/disable it based on that

(If you keep the current *bool, this is a case where you can use ptr.To(false))

(same for TestVSphereMachineTemplate_ValidateInterfaces)

Or you can follow what you did in TestVSphereMachineTemplate_ValidateInterfaces and not set nil in every test case, no strong opinion from my side

internal/webhooks/vmware/vspheremachine.go
sbueringer: In VSphereMachineTemplate accordingly
```go
	allErrs := validateInterfaces(webhook.NetworkProvider, newVSphereMachineTemplate.Spec.Template.Spec.Network, field.NewPath("spec", "template", "spec", "network"))
```

internal/webhooks/vmware/vspheremachinetemplate_test.go
```diff
@@ -97,3 +102,219 @@ func TestVSphereMachineTemplate_Validate(t *testing.T) {
...
+								Kind:       pkgnetwork.NetworkGVKNetOperator.Kind,
+								APIVersion: pkgnetwork.NetworkGVKNetOperator.GroupVersion().String(),
+							},
+							Name: "primary-wrong",
+						},
+					},
+				},
+			},
+			wantErr:    true,
+			wantErrMsg: "only support",
```
sbueringer: nit: maybe diff this against the test for VSphereMachine and check more of the error messages

internal/webhooks/vmware/vspheremachinetemplate_test.go
```diff
@@ -97,3 +102,219 @@ func TestVSphereMachineTemplate_Validate(t *testing.T) {
...
+			},
+			wantErr: false,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
+			if tc.featureGate != nil {
+				featuregatetesting.SetFeatureGateDuringTest(t, feature.MutableGates, feature.MultiNetworks, *tc.featureGate)
```
sbueringer:
```suggestion
				featuregatetesting.SetFeatureGateDuringTest(t, feature.Gates, feature.MultiNetworks, *tc.featureGate)
```

nit: Apparently both works, maybe let's go with feature.Gates as we use that almost everywhere atm

pkg/services/network/nsxt_vpc_provider.go
```diff
@@ -110,10 +110,26 @@ func (vp *nsxtVPCNetworkProvider) verifyNsxtVpcSubnetSetStatus(vspherecluster *v
...
+}
+
 // VerifyNetworkStatus checks if the given runtime object is of type SubnetSet.
 // If it is, then it calls verifyNsxVpcSubnetSetStatus with the SubnetSet to verify its status.
 // If it's not, it returns an error.
-func (vp *nsxtVPCNetworkProvider) VerifyNetworkStatus(_ context.Context, clusterCtx *vmware.ClusterContext, obj runtime.Object) error {
+func (vp *nsxtVPCNetworkProvider) VerifyNetworkStatus(ctx context.Context, clusterCtx *vmware.ClusterContext, obj runtime.Object) error {
+	log := ctrl.LoggerFrom(ctx)
+	if !createSubnetSet(clusterCtx) {
+		log.Info("Skipping SubnetSet status check as CreateSubnetSet is false")
```
sbueringer:
```suggestion
		log.V(5).Info("Skipping SubnetSet status check as CreateSubnetSet is false")
```
I think we are running CAPV with --v=4 per default. Accordingly I would use 5 here so this doesn't show up per default

(same in l.153)

pkg/services/vmoperator/vmopmachine_test.go
```diff
@@ -288,6 +327,48 @@ var _ = Describe("VirtualMachine tests", func() {
...
+			Expect(iface.IP).NotTo(BeNil())
+			Expect(*iface.IP.AutoConfigurationEnabled).To(BeTrue())
+			Expect(iface.IP.MACAddr).To(Equal("00:50:56:00:00:01"))
+
+			// Verify DHCP configuration
+			Expect(iface.IP.DHCP).NotTo(BeNil())
+			Expect(iface.IP.DHCP.IP4.Enabled).To(BeTrue())
+			Expect(iface.IP.DHCP.IP4.Config).To(HaveLen(2))
+			Expect(iface.IP.DHCP.IP4.Config[0].Key).To(Equal("1"))
+			Expect(iface.IP.DHCP.IP4.Config[0].Value).To(Equal("timeout 60;"))
```
sbueringer, line 347: nit: maybe also check the second entry

pkg/services/network/network_test.go
```diff
@@ -198,15 +207,118 @@ var _ = Describe("Network provider", func() {
...
+						Expect(vm.Spec.Network.Interfaces[2].Network.TypeMeta.APIVersion).To(Equal(netopv1.SchemeGroupVersion.String()))
+						Expect(vm.Spec.Network.Interfaces[2].Network.Name).To(Equal("another-secondary-network"))
+						Expect(vm.Spec.Network.Interfaces[2].Gateway4).To(Equal("None"))
+						Expect(vm.Spec.Network.Interfaces[2].Gateway6).To(Equal("None"))
+					})
+
+					It("should add primary and secondary network interfaces", func() {
+					})
+
+					It("should handle multiple secondary interfaces", func() {
```
sbueringer: let's align the name of the test case to what we are testing (that after multiple reconciles we don't end up with duplicate interfaces)

(Same for similar cases below)

pkg/services/network/network_test.go
```diff
@@ -748,6 +1055,42 @@ var _ = Describe("Network provider", func() {
 				Expect(v1beta1conditions.IsFalse(clusterCtx.VSphereCluster, vmwarev1.ClusterNetworkReadyCondition)).To(BeTrue())
 			})
 		})
+
+		Context("with NSX-VPC network provider and createSubnetSet is false", func() {
+			BeforeEach(func() {
+				// Set createSubnetSet to false on the cluster spec
+				vSphereCluster.Spec.Network = &vmwarev1.Network{
+					NsxVPC: &vmwarev1.NSXVPC{
+						CreateSubnetSet: &[]bool{false}[0],
```
sbueringer: nit: Interesting :) This is simpler
```suggestion
						CreateSubnetSet: ptr.To(false),
```

sbueringer (commented):
Nothing major

internal/webhooks/vmware/vspherecluster.go
```diff
@@ -0,0 +1,91 @@
...
+	return webhook.validateClusterNetwork(newTyped)
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
+func (webhook *VSphereCluster) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
+	return nil, nil
+}
+
+func (webhook *VSphereCluster) validateClusterNetwork(cluster *vmwarev1.VSphereCluster) (admission.Warnings, error) {
+	if cluster.Spec.Network != nil && cluster.Spec.Network.NsxVPC != nil {
```
zhanggbj: IIRC the original design is "If --feature-gate MultiNetworkSupport=false, "network" can not be set.", if this is still true, I think we should return error 
```
if cluster.Spec.Network != nil && !feature.Gates.Enabled(feature.MultiNetworks) 
```
Here it will return `nil, nil`, if cluster.Spec.Network != nil but cluster.Spec.Network.NsxVPC == nil

internal/webhooks/vmware/vspheremachine.go
```diff
@@ -96,10 +110,88 @@ func (webhook *VSphereMachine) ValidateUpdate(_ context.Context, oldRaw runtime.
...
+
+	if network != nil && network.Interfaces != nil {
+		if !feature.Gates.Enabled(feature.MultiNetworks) {
+			allErrs = append(allErrs, field.Forbidden(
+				field.NewPath(interfacesPath[0], interfacesPath[1:]...),
+				"interfaces can only be set when feature gate MultiNetworks is enabled"))
+		} else {
+			// Validate network type is supported
+			switch networkProvider {
+			case manager.NSXVPCNetworkProvider:
```
zhanggbj, line 138: Questions: are these two rules validated somewhere?

- createSubnetSet is not set or set to true but primary interface is set.
- createSubnetSet is false but primary interface is not set.

apis/vmware/v1beta1/vspheremachinenetworkstatus_types.go
sbueringer: No. We are intentionally not using the struct directly. We don't want to automatically have changes to our API if the type in vm-operator changes.

Basically we make an intentional choice about what we want to expose here and that will remain the same even if the upstream API changes

Also there is no way for us to be compatible with multiple vm-operator versions in the future if we have a Go dependency to a specific apiVersion in vm-operator.

Additionally we should not have to bump the CAPV apiVersion every time we adopt a new apiVersion of vm-operator (it would be a breaking change that we cannot do within an apiVersion if we switch from the v1alpha2 struct to the v1alpha3 struct)

internal/webhooks/vmware/vspheremachine.go
sbueringer: These are intentionally not validated in the webhooks.

The problem is that a VSphereMachineTemplate has no strong link to a Cluster object (it can exist e.g. being mapped to a ClusterClass instead of a Cluster).

If we would only validate this on the VSphereMachine then it could fail somewhere in KCP or MachineSet controllers when we cannot create a VSphereMachine object from a VSphereMachineTemplate.

So it is now validated in nsxt_vpc_provider.go ConfigureVirtualMachine during reconciliation. 

If we hit the error this should show up in the `VirtualMachineProvisioned` condition.

In general we should make sure that ClusterClass variables are designed or validated in a way that we never hit this error in CAPV and we basically avoid misconfiguration on the variable level on the Cluster object

apis/vmware/v1beta1/vspheremachinenetworkstatus_types.go
zhanggbj: > Basically we make an intentional choice about what we want to expose here and that will remain the same even if the upstream API changes

It makes sense to me, especially when we only expose a subcollection of what VMOP has

> Also there is no way for us to be compatible with multiple vm-operator versions in the future if we have a Go dependency to a specific apiVersion in vm-operator.

Sorry for the confusion, I understand CAPV is not compatible with multiple VMOP API versions. I mean if we are using VMOP struct, we could visit it during API Version bump, otherwise it's hard for us to notice VMOP API changes.  But I think above implementation is fine and best-efforts we have.

sbueringer: What I meant is that we might have to support multiple vm-operator apiVersions in the future. Today we can rely on v1alpha2 always being available. This might not always be the case, considering the wide range of vm-operator / supervisor versions we have to support.

Also the moment v1alpha2 is dropped in vm-operator we would have a problem because our v1beta1 could then immediately not embed the struct from vm-operator anymore. Sure we could then inline the struct, but this is also technically a breaking change.

Other problems are:
* if vm-operator adds a new field to the struct we would immediately also expose the field. Doesn't matter if we want or not
* we are not able to improve upon the vm-operator struct if we embed it. This means we cannot improve the godoc or customize it for CAPV, we cannot add additional validation markers (minimum/maximum, minLength/maxLength), we cannot add omitempty/omitzero, add or remove pointers etc.

zhanggbj: Understood, the current implementation makes sense, so CAPV have full control over its API surfacing and is isolated from VMOP API changes.

DanielXiao: Thanks. I removed the comment to sync

internal/webhooks/vmware/vspherecluster_test.go
DanielXiao: I created a downstream ticket for it.

internal/webhooks/vmware/vspherecluster.go
DanielXiao: I modified the design, If --feature-gate MultiNetworkSupport=false, createSubnetSet can not be set.

apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -93,6 +101,153 @@ type VSphereMachineSpec struct {
...
+
+	InterfaceSpec `json:",inline"`
+}
+
+// InterfaceSpec defines properties of a network interface.
+type InterfaceSpec struct {
+	// network is the name of the network resource to which this interface is
+	// connected.
+	// +required
+	Network InterfaceNetworkRefeference `json:"network,omitempty,omitzero"`
```
sbueringer:
```suggestion
	Network InterfaceNetworkReference `json:"network,omitempty,omitzero"`
```
Sorry my bad, typo came from my suggestion

internal/webhooks/vmware/vspherecluster_test.go
```diff
@@ -0,0 +1,282 @@
...
+		name            string
+		vsphereCluster  *vmwarev1.VSphereCluster
+		networkProvider string
+		featureGates    map[string]bool
+		wantErr         bool
+		errType         *apierrors.StatusError
+		errMsg          string // expected error message or substring
+	}{
+		{
+			name:            "successful VSphereCluster creation without network",
```
sbueringer, line 46: nit: Either this or the next test case can be removed now as they are the same (this is another nice thing about removing pointers it reduces the permutations on the API fields)

internal/webhooks/vmware/vspherecluster_test.go
```diff
@@ -0,0 +1,282 @@
...
+			name:            "successful VSphereCluster creation with network but no nsxVPC",
+			vsphereCluster:  createVSphereCluster("test-cluster", vmwarev1.Network{}),
+			networkProvider: manager.NSXVPCNetworkProvider,
+			featureGates:    map[string]bool{"MultiNetworks": false},
+			wantErr:         false,
+		},
+		{
+			name: "successful VSphereCluster creation with nsxVPC when network provider is NSX-VPC",
+			vsphereCluster: createVSphereCluster("test-cluster", vmwarev1.Network{
+				NSXVPC: vmwarev1.NSXVPC{},
```
sbueringer: nit: This now only occurs if nsxVPC is not set in the YAML (so test name does't fit anymore)
If it is set, it is enforced that it's not empty 

Same in l.106

internal/webhooks/vmware/vspheremachine.go
```diff
@@ -96,10 +110,83 @@ func (webhook *VSphereMachine) ValidateUpdate(_ context.Context, oldRaw runtime.
...
+
 	return nil, webhooks.AggregateObjErrors(newTyped.GroupVersionKind().GroupKind(), newTyped.Name, allErrs)
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
 func (webhook *VSphereMachine) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
 	return nil, nil
 }
+
+func validateNetwork(networkProvider string, network vmwarev1.VSphereMachineNetworkSpec, fildPath *field.Path) field.ErrorList {
```
sbueringer:
```suggestion
func validateNetwork(networkProvider string, network vmwarev1.VSphereMachineNetworkSpec, fldPath *field.Path) field.ErrorList {
```
nit: We were usually abbreviating it with fldPath

pkg/services/network/nsxt_vpc_provider.go
```diff
@@ -110,10 +111,24 @@ func (vp *nsxtVPCNetworkProvider) verifyNsxtVpcSubnetSetStatus(vspherecluster *v
 	return nil
 }
 
+func createSubnetSet(clusterCtx *vmware.ClusterContext) bool {
+	cluster := clusterCtx.VSphereCluster
+	isCreate := true
+	if cluster.Spec.Network.NSXVPC.CreateSubnetSet != nil {
+		isCreate = *cluster.Spec.Network.NSXVPC.CreateSubnetSet
+	}
+	return isCreate
```
sbueringer:
```suggestion
	return ptr.Deref(clusterCtx.VSphereCluster.Spec.Network.NSXVPC.CreateSubnetSet, true)
```
nit. Up to you. Just wanted to show you how much this can be simplified now :)

apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -93,6 +101,153 @@ type VSphereMachineSpec struct {
...
+	// It is used to connect the Kubernetes primary network for Load balancer,
+	// Service discovery, Pod traffic and management traffic etc.
+	// Leave it unset if you don't want to customize the primary network and interface.
+	// Customization is only supported with network provider NSX-VPC.
+	// It should be set only when VSphereCluster spec.network.nsxVPC.createSubnetSet is set to false.
+	//
+	// +optional
+	Primary InterfaceSpec `json:"primary,omitempty,omitzero"`
+
+	// secondary is the secondary network interface.
```
sbueringer:
```suggestion
	// secondary are the secondary network interfaces.
```
nit

apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -93,6 +101,153 @@ type VSphereMachineSpec struct {
...
+}
+
+// IsDefined returns true if the InterfaceSpec is defined.
+func (r *InterfaceSpec) IsDefined() bool {
+	return !reflect.DeepEqual(r, &InterfaceSpec{})
+}
+
+// InterfaceNetworkRefeference describes a reference to another object in the same
+// namespace as the referrer. The reference can be just a name but may also
+// include the referred resource's APIVersion and Kind.
```
sbueringer:
```suggestion
// InterfaceNetworkRefeference describes a reference to another object in the same
// namespace as the referrer.
```
nit

apis/vmware/v1beta1/vspheremachine_types.go
```diff
@@ -93,6 +101,153 @@ type VSphereMachineSpec struct {
...
+func (r *InterfaceNetworkRefeference) GroupVersionKind() schema.GroupVersionKind {
+	return schema.FromAPIVersionAndKind(r.APIVersion, r.Kind)
+}
+
+// RouteSpec defines a static route for a guest.
+type RouteSpec struct {
+	// to is either "default", or an IP4 address. IP6 is not supported yet.
+	//
+	// +kubebuilder:validation:Pattern=`^(default|([0-9]{1,3}\.){3}[0-9]{1,3}(\/[0-9]{1,2})?)$`
+	// +kubebuilder:validation:MinLength=9
```
sbueringer, line 237: I'm not sure how we end up with these min/max lengths here

EDIT: Ah I think the godoc is slightly wrong and this is not an IP4 address but a CIDR?

(If there are any non-godoc changes that we would like to make, absolutely fine to address in a follow-up from my side)

apis/vmware/v1beta1/vspheremachinenetworkstatus_types.go
```diff
@@ -0,0 +1,292 @@
...
+	// specific configuration command. For example on Linux and BSD systems
+	// using the file dhclient.conf output would be reported at system scope:
+	// key='1', value='timeout 60;' key='2', value='reboot 10;'. The output
+	// reported per interface would be:
+	// key='1', value='prepend domain-name-servers 192.0.2.1;'
+	// key='2', value='require subnet-mask, domain-name-servers;'.
+	//
+	// +optional
+	// +listType=map
+	// +listMapKey=key
```
sbueringer: I would also make this atomic and ensure key uniqueness e.g. with CEL instead

(we should not use map if we only want to ensure the key is unique, using map also leads to fine-granular tracking in managedFields wich requires more memory + allows multiple clients to own different fields, I think we want/need neither of that)

(Fine to address in a follow-up)

apis/vmware/v1beta1/vspheremachinenetworkstatus_types.go
```diff
@@ -0,0 +1,292 @@
...
+	// lifetime describes when this address will expire.
+	//
+	// +optional
+	Lifetime metav1.Time `json:"lifetime,omitempty,omitzero"`
+
+	// origin describes how this address was configured.
+	//
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=512
```
sbueringer:
```suggestion
```
MinLength / MaxLength should not be necessary on an enum

apis/vmware/v1beta1/vspheremachinenetworkstatus_types.go
```diff
@@ -0,0 +1,292 @@
...
+	// +kubebuilder:validation:Enum=dhcp;linklayer;manual;other;random
+	Origin string `json:"origin,omitempty"`
+
+	// state describes the state of this IP address.
+	//
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=20
```
sbueringer:
```suggestion
```
Same

apis/vmware/v1beta1/vspheremachine_types.go
DanielXiao: Nice catch, I was not aware of it.

apis/vmware/v1beta1/vspheremachine_types.go
DanielXiao: Good catch. We should not allow "default", so min should be length of 0.0.0.0/0 and max should be length of  255.255.255.255/32

apis/vmware/v1beta1/vspheremachinenetworkstatus_types.go
DanielXiao: I simply copied it from VirtualMachine. It 's just for mirroring IP status so I agree it should be atomic. We don't need to ensure key uniqueness, just show what is showing in the VirtualMachine status.

sbueringer: Makes sense!

sbueringer:
Thank you very much! Really nice work!

/lgtm
/approve

/hold
Feel free to merge when you're ready

sbueringer:
/retest
(independent flake)

/lgtm

PR #3576, opened 2025-08-14
fix: use vsphereMachine instead of machine for fd

### LINKED ISSUE #3575: Adding a VM to a vm-host group is racy
/kind bug

Adding the feature for vm-host topology within the openshift installer, capv:

- Creates virtual machine
- Power on virtual machine
- Configures cluster's vm-host group with vm

The problem occurs between the power on and the time it takes to add the 
vm to the vm-host group, then a potential vmotion to the correct ESXi host.

The virtual machine could be in the wrong vsphere topology at the time for the CCM to init the
node and the node has incorrect topology region and zone labels.

- Creates virtual machine
- Configure vCenter cluster vm-host group with vm
- Power on virtual machine

Already have a resolution, will add this issue to the PR.

**Environment:**

- Cluster-api-provider-vsphere version: v1.12.0-rc.0.0.20250203113257-38f6b67f9b7d 
- 
https://github.com/openshift/installer/blob/e3cbeafe717594ac0f619212c207963217869624/cluster-api/providers/vsphere/go.mod#L9

jcpowermac:

This commit changes the failureDomain check from
using capi machine to vsphereMachine.
The vsphereMachine object has immediate access
to the failure domain struct, which resolves
an issue where virtual machines are not placed
within vm-host groups in a timely
manner causing misplacement in the topology.

Fixes #3575


controllers/vspherevm_controller.go
```diff
@@ -234,10 +234,10 @@ func (r vmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.R
 	}
 
 	var vsphereFailureDomain *infrav1.VSphereFailureDomain
-	if failureDomain := machine.Spec.FailureDomain; failureDomain != "" {
+	if failureDomain := vsphereMachine.Spec.FailureDomain; failureDomain != nil && *failureDomain != "" {
```
chrischdi: Am I right that its always the way that:

Whatever is set on `vSphereMachine`'s `.spec.failureDomain` gets propagated up (by CAPI's machine controller) to the `Machine`'s `.spec.failureDomain` field.

Its just that the timing is bad (CAPI does this too late?!)

Otherwise: should we use a fallback to `machine.Spec.FailureDomain`, if its not set on vsphereMachine?

chrischdi: Reading the CAPI contract: https://main.cluster-api.sigs.k8s.io/developer/providers/contracts/infra-machine#inframachine-failure-domain

It should be the other way around (which means code as-is should be correct?!).

jcpowermac: used `rr` to [troubleshoot](https://github.com/openshift/installer/pull/9887), the machine object doesn't receive the failuredomain name before the guest is powered on - then the race is on...will the node ccm init prior to being added to the vm-host group and potentially vmotion'ed to the correct location in the vCenter cluster. 


https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/ade4688102e03daa56106359cd1fea55e67700b6/pkg/services/govmomi/service.go#L183-L195

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/ade4688102e03daa56106359cd1fea55e67700b6/pkg/services/govmomi/service.go#L726

chrischdi: I agree that potentially is the behavior when setting failuredomain only on the InfraMachine object (VSphereMachineTemplate?!).

According to https://github.com/kubernetes-sigs/cluster-api/issues/11232 the correct usage is setting FailureDomain on the Machine object instead. Which means it would also be there in the referenced code.

Setting it on the VSphereMachine / VSphereMachineTemplate is the old deprecated behavior.

jcpowermac: I understand that its deprecated behavior, what is the solution then? The machine failuredomain name is slow to become available in the reconcile loop, which causes this race.

chrischdi: The solution if I'm right is: Set the failuredomain on the `Machine` (or `MachineSet` or `MachineDeployment`) object instead of the InfraMachine / VSphereMachine.

It should be fine to still set it on the InfraMachine / VSphereMachine. The important piece is to set it on the Machine instead of relying on CAPI copy this information to the InfraMachine / VSphereMachine.

sbueringer: I think it's okay to read failureDomain from VSphereMachine if not set on Machine to make this more robust. Let's just not update the Machine object

jcpowermac:
@chrischdi @sbueringer updated per comments ptal, thanks!

chrischdi (commented):
/lgtm
/approve

chrischdi:
Thanks for bringing this up 🎉

chrischdi:
/retitle 🐛 fix: use machine's .spec.failureDomain with fallback to vspheremachine's .spec.failureDomain if not defined

PR #3585, opened 2025-08-21
🌱 Update InterfaceSpec MTU minimum value

silvery1622:
Update the value from 68 to 576

Update InterfaceSpec MTU minimum value.
68 bytes is the minimum size of IPv4 datagram every device must be able to forward without further fragmentation.
576 bytes is the minimum size of IPv4 datagram every device has to be able to receive.

Fixes #


sbueringer:
Thx!
/lgtm
/approve

PR #3607, opened 2025-08-26
✨ Add flags to allow customization of CPU and memory shares, reservations and limits

tommasopozzetti:

This PR adds flags to optionally customize CPU shares and reservations for cloned VMs as part of the vSphereMachine spec


smcallister-bc:
This seems incredibly useful, thank you for doing this! Any chance you could do something similar for memory reservations as well?

tommasopozzetti:
I’d be happy to add to the PR similar logic for memory reservation and shares as well as potentially memory pinning

I was hoping to first get a glance from a maintainer to see if this approach is reasonable given it’s my first contribution to this project!

apis/v1beta1/types.go
```diff
@@ -167,6 +167,16 @@ type VirtualMachineCloneSpec struct {
...
+	// CPUReservationMhz is the amount of CPU in MHz that is guaranteed available to the virtual machine.
+	// Defaults to the eponymous property value in the template from which the
+	// virtual machine is cloned.
+	// +optional
+	CPUReservationMhz int64 `json:"cpuReservationMhz,omitempty"`
+	// CPUShares are a relative priority to other virtual machines used in case of resource contention.
+	// Defaults to the eponymous property value in the template from which the
+	// virtual machine is cloned.
+	// +optional
+	CPUShares int32 `json:"cpuShares,omitempty"`
```
chrischdi: Does it make sense to have something like:

```yaml
resources:
  reservation:
    cpu: ...
    memory: ...
  shares:
    ...
```

Or maybe this should be modelled in the k8s wordings which has `limits` and `requests`? (might not match the things this PR currently sets).

This is e.g. done by the vm-operator APIs. However, vm-operator does not use shares.

Could someone research what the benefits are of setting shares? And should we also consider CPU Limit to be set?

tommasopozzetti: @chrischdi thanks for the review!

In terms of shares vs reservations, shares are a relative measure of prioritization while reservations are an absolute one. A VM with 2Ghz set for reservation will be guaranteed that even under host contention, it will always have 2Ghz of cpu power available to it. The vSphere admission controller will prevent a VM to power on if the total sum of reservations for VMs on a given host exceeds the total available cpu power of that host (no overprovisioning).
Shares on the other hand are just meaningful relative to shares of other VMs on that host. The host can be over provisioned and if it comes under cpu contention, VMs will be prioritized for cpu time relative to each other depending on their shares. So if a host has 3 VMs, one with 6000 shares, one with 3000 and one with 1000, if the host comes under cpu contention, the first will get 60% of cpu time, the second 30% and the third 10%. Each VM normally gets shares assigned by default proportional to its number of vCPUs but it is very useful to be able to tune that at will.

In terms of using cpu/memory limits, I have never had to implement these but they essentially would artificially cause the same effects as if the underlying host was under resource contention even if it is not, when the VM reaches said limit. More detailed info [here](https://knowledge.broadcom.com/external/article/326270/impact-of-virtual-machine-memory-and-cpu.html). I'd be happy to add to this PR the limit as well as an optional configurable if desired.

Finally, in terms of the syntax, I'm open to suggestions! I personally feel like using the same syntax as standard Kubernetes containers might be misleading since the practical implementation of using reservations, shares and limits on VMs is very different than memory and cpu requests and limits for k8s pods. 
I was going for a more flat mapping similar to the other properties that matches with the VM options and would look like
```yaml
cpuReservationMhz: xxx
cpuShares: xxx
cpuLimitMhz: xxx
memoryReservationMB: xxx
memoryShares: xxx
memoryLimitMB: xxx
reserveAllMemory: false
```

but, if preferred, we could also go for something nested like
```yaml
resourceManagement:
  cpu:
    reservationMhz:
    shares:
    limitMhz:
  memory:
    reservationMB:
    shares:
    limitMB:
    reserveAll:
```
or similar

tommasopozzetti:
Hi @chrischdi, following up on your comment, do you have any thoughts on my reply?
Thanks!

tommasopozzetti:
@chrischdi or @sbueringer do you have any opinions on the above discussion? I'd love to do any edits that make sense to push this forward! This feature is one that we really need
Thank you!

apis/v1beta1/types.go
chrischdi: I'd prefer to takeover the definition how it is in vm-operator to have a similar API.

Which comes down to:

```
resources:
  requests: # --> reservations
    cpu: ... # in mhz, documented on the godoc
    memory: ...
  limits: # --> limits
    cpu: ...
    memory: ...
```

And also use the types which are common for kubernetes.

https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha5/virtualmachineclass_types.go#L82-L86

The fields should have proper description to what this fields map at the end.

For shares: see my other comment.

chrischdi (commented):
First of all: sorry for the long run on this and thanks for reminding me.

I'm fine with adding api and implementation which properly sets reservation (= requests in kubernetes) and limits for CPU and memory.

I'd like to again think about the use-case of setting shares for a VM. I want to understand the use-case so we don't add an api which won't really get used at the later stage.

VM-Operator does not have them configurable. And I think there's a reason.

If I get it right, setting shares only take effect if there is a lack of resources affecting all the VMs of the vSphere host / cluster / resourcepool (depending on setup).
So I think there's no way, or it is pretty hard, to set shares to a value that makes sense across the whole infrastructure.
The real solution here should be adding more physical capacity.

> Shares can be used to prioritize resource availability for VMs at the time of contention. 

[[0]](https://www.vmware.com/docs/drs-cluster-mgmt-perf)

@tommasopozzetti : I'd like to better understand the use-case you have for setting shares?

Maybe @akutz has some thoughts here.

@vr4manta maybe you have some thoughts here too?

apis/v1beta1/types.go
chrischdi: Note: having the unit in the name does not make sense to me. At least not for memory, and I think also not for CPU.
If we have a proper godoc that explains that this field is in hz, 2GhZ should map to e.g. "2Gi".

tommasopozzetti:
@chrischdi thank you for your review!

I will make the changes to the PR to implement the syntax you suggested to match vm-operator.
I would like to include shares as well in the design though. I'm personally not familiar with vm-operator but I can share some of the thoughts behind the use.

First and foremost, shares are in use all the time out of the box even if you do not set them. vSphere will assign automatically shares to every VM proportional to the chosen priority of the VM (or the default one for the resource pool if not defined) and the number of cores of the VM. So shares will be used regardless of our implementation here. My proposed addition just gives more optional control over them allowing to map the corresponding available api to configure a custom value of shares if desired.

We use this heavily. Shares are the only way to properly set relative priorities among VMs to distinguish among more and less critical workloads and in our case, we have more "classes" of workloads then the simple default 3 levels of priorities available so the ability to customize them is imperative.

We have actually been recommended to use this mechanism from VMware engineers so I assume these are used and valuable.

While I definitely agree that as you run out of capacity, adding capacity is always the best solution, it is not one always readily available and, when running over provisioned (which we are significantly to make efficient use of resources), spikes will always at times cause temporary contention. That contention must somehow be resolved and the way vSphere does that is giving relative cpu time based on the shares. Only using reservations and limits does not address this properly. It just sets a guaranteed minimum and a ceiling maximum for the resource but it does not give any direction to the system of how to prioritize in between that min and max among VMs.
A concrete example: if I have 10Ghz available, and 2 VMs both requesting 4Ghz and with a limit of 8Ghz and both spiking in a moment of high load to try and use 6Ghz each, how does the system choose the allocation of the available 10? What if one of those VMs is a dev workload and one a qa one? How do I explain to the system to prioritize qa? I can give twice as many shares to qa and in that scenario, it will get twice as much of the remaining capacity than the dev one

I can start working on the edits to have the following syntax and ensure types and comments match the vm-operator with the shares addition, but please do let me know in the meantime if you have further thoughts about the shares discussion!

```yaml
resources:
  requests: # --> reservations
    cpu: ... # in mhz, documented on the godoc
    memory: ...
  limits: # --> limits
    cpu: ...
    memory: ...
  shares:
    cpu: ...
    memory: ...
```

scott-grimes:
👍 would love this - both the reservations/limits and also shares

chrischdi:
Sounds reasonable, thanks for explaining!

Also the example Looks good to me 👍

Let me know once you updated the PR

tommasopozzetti:
@chrischdi I have updated the PR to follow the proposed syntax!
Let me know your thoughts!
Thank you

sbueringer:
I'll also try to take a look soon

/ok-to-test

tommasopozzetti:
Thanks @sbueringer!
I fixed all the linting issues. There is still one check failing but I'm not sure that it is related to any change here. Any advice would be great!

apis/v1beta1/types.go
```diff
@@ -167,6 +168,10 @@ type VirtualMachineCloneSpec struct {
 	// virtual machine is cloned.
 	// +optional
 	NumCoresPerSocket int32 `json:"numCoresPerSocket,omitempty"`
+	// resources is the definition of the VM's cpu and memory
+	// reservations, limits and shares.
+	// +optional
+	Resources VirtualMachineResources `json:"resources,omitempty,omitzero"`
```
chrischdi, line 175:
```suggestion

	// resources is the definition of the VM's cpu and memory
	// reservations, limits and shares.
	// +optional
	Resources VirtualMachineResources `json:"resources,omitempty,omitzero"`

```

apis/v1beta1/types.go
```diff
@@ -211,6 +216,34 @@ type VirtualMachineCloneSpec struct {
...
+type VirtualMachineResources struct {
+	// requests is the definition of the VM's cpu (in hertz) and memory (in bytes) reservations
+	// +optional
+	Requests VirtualMachineResourceSpec `json:"requests,omitempty,omitzero"`
+	// limits is the definition of the VM's cpu (in hertz) and memory (in bytes) limits
+	// +optional
+	Limits VirtualMachineResourceSpec `json:"limits,omitempty,omitzero"`
+	// shares is the definition of the VM's cpu and memory shares
+	// +optional
+	Shares VirtualMachineResourceSpec `json:"shares,omitempty,omitzero"`
```
chrischdi: This should be a different struct. Shares do not match resource.Quantity

apis/v1beta1/types.go
```diff
@@ -211,6 +216,34 @@ type VirtualMachineCloneSpec struct {
...
+// given VM hardware policy.
+// +kubebuilder:validation:MinProperties=1
+type VirtualMachineResourceSpec struct {
+	// cpu is the definition of the cpu quantity for the given VM hardware policy
+	// +optional
+	Cpu resource.Quantity `json:"cpu,omitempty"` //nolint:revive
+
+	// memory is the definition of the memory quantity for the given VM hardware policy
+	// +optional
+	Memory resource.Quantity `json:"memory,omitempty"`
```
chrischdi, line 248: Sorry for the back and forth. Now I see why you suggested mhz initially.

However, I'd keep this types as is.

But I think we maybe should add validation, that these are full mhz or mb values.

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -212,6 +213,44 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
 		Snapshot: snapshotRef,
 	}
 
+	// Set CPU reservations, limits and shares if specified
+	cpuAllocation := types.ResourceAllocationInfo{}
+	if !vmCtx.VSphereVM.Spec.Resources.Requests.Cpu.IsZero() {
+		cpuReservationMhz := int64(math.Ceil(float64(vmCtx.VSphereVM.Spec.Resources.Requests.Cpu.Value()) / float64(1000000)))
```
chrischdi: Can we add helper funcs for convertQuantityToMhz and convertQuantityToMB

pkg/services/govmomi/vcenter/clone.go
```diff
@@ -212,6 +213,44 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
...
+		cpuAllocation.Limit = ptr.To(cpuLimitMhz)
+	}
+	if !vmCtx.VSphereVM.Spec.Resources.Shares.Cpu.IsZero() {
+		cpuShares := types.SharesInfo{
+			Shares: int32(vmCtx.VSphereVM.Spec.Resources.Shares.Cpu.Value()),
+			Level:  types.SharesLevelCustom,
+		}
+		cpuAllocation.Shares = ptr.To(cpuShares)
+	}
+	spec.Config.CpuAllocation = ptr.To(cpuAllocation)
```
chrischdi: Are there any effects if this is set empty vs nil (which it was before?)

I'd prefer to not change the existing behavior.

So if no requests, limits or shares are set, we should not set CPUAllocation to an empty struct. Same below.

chrischdi:
conversion also need to be fixed to be round-tripable.

See e.g. what we do for AdditionalDisksGiB: https://github.com/openshift/cluster-api-provider-vsphere/blob/393a16983d02d5a2b254e4182002f70329f9dd8f/apis/v1alpha4/vspheremachine_conversion.go#L39

apis/v1beta1/types.go
fabriziopandini: If I'm not mistaked, this struct is shared between govmomi mode and supervisor mode.
If this is correct, impacts of this changes must me considered for both modes.

tommasopozzetti:
@chrischdi thanks for he input!
I think I addressed all the concerns.

The only comment I have is on the quantities and the validation. Rather then hard failing if the quantity is not set as a precise multiple of Mhz or MiB, which would be an atypical behavior for `resources` specification on a k8s resource, I propose (and implemented) to automatically round up to the nearest Mhz or MiB. The behavior has also been documented in the field descriptions.

There look to still be a check failing and it is unclear to me if it's still related to the conversion webhooks or unrelated. The logs don't make it very easy to understand what those checks are doing and why they are failing. I see several errors related to sessions to vCenter and parsing of the APIVersion and Kind, both things that don't have anything to do with the contents of this PR. Any further guidance here is definitely appreciated!
Thank you!

chrischdi:
> The only comment I have is on the quantities and the validation. Rather then hard failing if the quantity is not set as a precise multiple of Mhz or MiB, which would be an atypical behavior for resources specification on a k8s resource, I propose (and implemented) to automatically round up to the nearest Mhz or MiB. The behavior has also been documented in the field descriptions.

That's fine for me.

The failing tests are fuzz tests. They try to ensure that the conversions are round-tripable. So the failing tests here are:

v1beta1.VSphereVM -> v1alpha(3/4).VSphereVM -> v1beta1.VSphereVM

And that causes a diff.

Same for VSphereMachineTemplate.

We also need to add the conversion change for VSphereVM and VSphereMachineTemplate, because they also get the additional fields.

tommasopozzetti:
@chrischdi thanks for the pointers!
Looks like those are solved now!
One more test is failing but seems like it's failing on setting up the test environment?
Appreciate any pointers there as well, and thanks for your continued guidance!

tommasopozzetti:
@chrischdi circling back here to see if there is anything else I can do!
Thank you!

tommasopozzetti:
@chrischdi yay! Looks like all tests passed!

Would it be possible to get a final review and potentially get this in to be included in the next release? We are very much looking forward to using this feature!
Thank you!

apis/v1beta1/types.go
```diff
@@ -211,6 +218,51 @@ type VirtualMachineCloneSpec struct {
...
+	Shares VirtualMachineResourceShares `json:"shares,omitempty,omitzero"`
+}
+
+// VirtualMachineResourceSpec is the numerical definition of memory and cpu quantity for the
+// given VM hardware policy.
+// +kubebuilder:validation:MinProperties=1
+type VirtualMachineResourceSpec struct {
+	// cpu is the definition of the cpu quantity for the given VM hardware policy
+	// +optional
+	Cpu resource.Quantity `json:"cpu,omitempty"` //nolint:revive
```
sbueringer: Sorry might have missed a previous discussion. Is there a reason why we are ignoring the finding instead of using CPU? (same below)

Ah maybe to align to vm-operator. Please drop the nolint and use CPU instead. I prefer getting this right over alignment to vm-operator :)

tommasopozzetti:
@sbueringer Thank you! I corrected the CPU field as requested!

sbueringer:
> @sbueringer Thank you! I corrected the CPU field as requested!

Thank you very much!

> Would it be possible to get a final review and potentially get this in to be included in the next release? We are very much looking forward to using this feature!
> Thank you!

Yup, let's get this merged early next week, so it's part of the next release. 
Thx for the patience, sorry for the long delays for review, was a way-too-busy release cycle in core CAPI :)

/assign @chrischdi

for a final review

PR #3621, opened 2025-09-01
Remove the CEL validation on VSphereMachineTemplate.spec.network

DanielXiao:
When the Cluster API topology controller does a dry-run patch on the VSphereMachineTemplate to figure out if there are changes to apply, it fails with below error:

server side apply dry-run failed for modified object: VSphereMachineTemplate.vmware.infrastructure.cluster.x-k8s.io "guestcluster1-rhhvs" is invalid: spec.template.spec: Invalid value: "object": field ''network'' cannot be added or removed after creation

This validation should be skipped since it is expected the topology controller rotates a new VSphereMachineTemplate if Runtime Extension configures secondary interface on day 2. The Webhook contains diff validation on spec.network.interfaces to ensure immutability in a non dryrun case.

Fixes #


internal/webhooks/vmware/vspheremachinetemplate.go
```diff
@@ -60,12 +62,31 @@ func (webhook *VSphereMachineTemplate) ValidateCreate(ctx context.Context, obj r
...
+	if err != nil {
+		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a admission.Request inside context: %v", err))
+	}
+	if !topology.IsDryRunRequest(req, newObj) {
+		equal, diff, err := util.Diff(oldObj.Spec.Template.Spec, newObj.Spec.Template.Spec)
+		if err != nil {
+			return nil, apierrors.NewBadRequest(fmt.Sprintf("failed to compare old and new VSphereMachineTemplate: %v", err))
+		}
+		if !equal {
+			return nil, field.Invalid(field.NewPath("spec", "template", "spec"), newObj, fmt.Sprintf("VSphereMachineTemplate spec.template.spec field is immutable. Please create a new resource instead. Diff: %s", diff))
```
sbueringer:
```suggestion
			return nil, apierrors.NewInvalid(vmwarev1.GroupVersion.WithKind("VSphereMachineTemplate").GroupKind(), newObj.Name, field.ErrorList{
				field.Invalid(field.NewPath("spec", "template", "spec"), newObj, fmt.Sprintf("VSphereMachineTemplate spec.template.spec field is immutable. Please create a new resource instead. Diff: %s", diff)),
			})
```

sbueringer: I think this should be "wrapped" like in other cases

sbueringer:
Thx!

/lgtm
/approve

PR #3622, opened 2025-09-01
📖 Update release support and compatibility info

fabriziopandini:
While preparing 1.14 release I noticed some outdated info in the readme; I also took the opportunity to document the release process/guarantee we are following since some time now

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,28 +29,67 @@ Check out the [getting started guide](./docs/getting_started.md) for launching a
 
 ------
 
+# CAPV release support
+
+A Cluster API provider vSphere release correspond to a release in the GitHub repository for this project, and the corresponding images published in the Kubernetes docker registry.
+The CAPV release calendar will be aligned as much as possible to the Kubernetes and the Cluster API release release dates plus an offset for 
+performing test/validation after K8s/CAPI release are available (tentatively one or two weeks after both K8s and CAPI release are available). 
+
+The CAPV team actively supports the latest two minor releases (N, N-1); support in this context means that we:
+
+- Have CI signal with E2E tests, unit tests, CVE scans etc.
+- Accept bug fixes, perform golang or dependency bumps, etc.
+- Periodically cut patch releases
+- On top of supporting the N and N-1 releases, the CAPV team also maintains CI signal for the CAPV N-2 releases in case we have to do an emergency patch release.
+
+If there is a need for an emergency patch, e.g. to fix a critical security issue, please bring this up to maintainers and it will be considered on a case-by-case basis.
+
+All considered, each CAPV minor release is supported for a period of roughly 12 months:
+
+- The first eight months of this timeframe will be considered the standard support period for a minor release.
+- The next four months the minor release will be considered in maintenance mode.
+- At the end of the four-month maintenance mode period, the minor release will be considered EOL (end of life) and cherry picks to the associated branch are to be closed soon afterwards.
+
+The table below documents support matrix for Cluster API provider vSphere versions (older versions omitted).
+
+| Minor Release | Status                  | Supported Until (including maintenance mode)                                         |
+|---------------|-------------------------|--------------------------------------------------------------------------------------|
+| v1.14.x       | Standard support period | in maintenance mode when v1.16.0 will be released, EOL when v1.17.0 will be released |
+| v1.13.x       | Standard support period | in maintenance mode when v1.15.0 will be released, EOL when v1.16.0 will be released |
+| v1.12.x       | Maintenance mode        | Maintenance mode v1.14.0 release date, EOL when v1.15.0 will be released             |
+| v1.11.x       | EOL                     | EOL since v1.14.0 release date                                                       |
+| v1.10.x       | EOL                     | EOL since v1.13.0 release date                                                       |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,28 +29,67 @@ Check out the [getting started guide](./docs/getting_started.md) for launching a
 
 ------
 
+# CAPV release support
+
+A Cluster API provider vSphere release correspond to a release in the GitHub repository for this project, and the corresponding images published in the Kubernetes docker registry.
+The CAPV release calendar will be aligned as much as possible to the Kubernetes and the Cluster API release release dates plus an offset for
```
sbueringer:
```suggestion
The CAPV release calendar will be aligned as much as possible to the Kubernetes and the Cluster API release dates plus an offset for 
```

README.md
```diff
@@ -29,28 +29,67 @@ Check out the [getting started guide](./docs/getting_started.md) for launching a
...
+
+A Cluster API provider vSphere release correspond to a release in the GitHub repository for this project, and the corresponding images published in the Kubernetes docker registry.
+The CAPV release calendar will be aligned as much as possible to the Kubernetes and the Cluster API release release dates plus an offset for 
+performing test/validation after K8s/CAPI release are available (tentatively one or two weeks after both K8s and CAPI release are available). 
+
+The CAPV team actively supports the latest two minor releases (N, N-1); support in this context means that we:
+
+- Have CI signal with E2E tests, unit tests, CVE scans etc.
+- Accept bug fixes, perform golang or dependency bumps, etc.
+- Periodically cut patch releases
```
sbueringer: We don't actually do that. Do we want to start doing that or should we write "Cut patch releases on demand?

README.md
```diff
@@ -29,28 +29,67 @@ Check out the [getting started guide](./docs/getting_started.md) for launching a
...
+| v1.10.x       | EOL                     | EOL since v1.13.0 release date                                                       |
+| v1.9.x        | EOL                     | EOL since v1.12.0 release date                                                       |
+
 ## Compatibility with Cluster API and Kubernetes Versions
 
-This provider's versions are compatible with the following versions of Cluster API:
+Compatibility between CAPV releases and Cluster API depends on which Cluster API contract versions supported by CAPI and by which contract version is implemented in CAPV.
+
+CAPV implements following Cluster API contract versions:
+
```
sbueringer: The old table intentionally only covered a 1:1 mapping. Because every CAPV version is only tested with a specific CAPI version. Looks like with this table we are considerably expanding the **supported** combinations of CAPI and CAPV, but we don't have any test coverage for it. Are we sure we want to do that?

I would prefer if we would only support what we can test.

I would also prefer not having to fix bugs if folks are using combinations that are not tested by us

fabriziopandini: I think this is what I tried to convey with the table below that documents what we test.
Might be it is ok adding some phrasing about the fact that it is up to the users to validated combinations that are not tested in CI + support will be provided best effort.

sbueringer: > I think this is what I tried to convey with the table below that documents what we test.

Yup. I also understood that is the idea.

I'm just not sure how clear that comes across if we defined compatibility around the contract. I would infer from this that all combinations are supported as long as the CAPI versions support the same contract and I'm just not sure if that is sufficient. I think we made some smaller changes here and there to keep CAPV working with the current CAPI version. So I have no idea which other permutations might work or not

sbueringer: Concrete example.

CAPV v1.14 implements v1beta1 contract
CAPI v1.10 supports v1beta1 contract

CAPI v1.10 + CAPV v1.14 will fail because CAPV v1.14 uses core CAPI v1beta2 objects.

I think it would be a much clearer and simpler message if we restrict our compatibility & support to a strict 1:1 mapping

(Which is why I redcued the compatibility matrix to a 1:1 mapping in the past. Before that we also claimed support for combinations that we never tested)

fabriziopandini: Rephrased, PTAL

sbueringer: Thx!

README.md
```diff
@@ -29,28 +29,86 @@ Check out the [getting started guide](./docs/getting_started.md) for launching a
...
+also please be aware that in case of a CAPI/CAPV combination not listed above, support will be provided at best effort.
+
+Following info should be used when identifying possible CAPV/CAPI version combinations outside the table above:
+
+Compatibility between CAPV and Cluster API depends on which Cluster API contract versions is supported by CAPI and by
+which contract version is implemented in CAPV.
+
+The following table documents the Cluster API contract versions implemented by supported CAPV versions:
+
+|            | Cluster API contract version v1beta1 | Cluster API contract version v1beta1 |
```
sbueringer:
```suggestion
|            | Cluster API contract version v1beta1 | Cluster API contract version v1beta2 |
```

sbueringer: Entirely up to you but maybe just merge the two columns and write v1beta1 / v1beta2 in the cell like in l.98++

But fine as is for me

sbueringer:
/lgtm
/approve

/hold
up to you

/assign @chrischdi @zhanggbj

sbueringer:
/hold cancel

If there any further findings let us know and we'll follow-up

PR #3645, opened 2025-09-29
✨ WIP: Add affinity/anti-affinity to VirtualMachine spec

srm09:

Fixes #

### DIFF
--- a/pkg/services/vmoperator/vmopmachine.go
+++ b/pkg/services/vmoperator/vmopmachine.go
@@ -41,6 +41,7 @@ import (
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/v1beta1"
 	vmwarev1 "sigs.k8s.io/cluster-api-provider-vsphere/apis/vmware/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-vsphere/feature"
 	capvcontext "sigs.k8s.io/cluster-api-provider-vsphere/pkg/context"
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/context/vmware"
 	infrautilv1 "sigs.k8s.io/cluster-api-provider-vsphere/pkg/util"
@@ -171,10 +172,6 @@ func (v *VmopMachineService) ReconcileNormal(ctx context.Context, machineCtx cap
 		return false, errors.New("received unexpected SupervisorMachineContext type")
 	}
 
-	if supervisorMachineCtx.Machine.Spec.FailureDomain != "" {
-		supervisorMachineCtx.VSphereMachine.Spec.FailureDomain = ptr.To(supervisorMachineCtx.Machine.Spec.FailureDomain)
-	}
-
 	// If debug logging is enabled, report the number of vms in the cluster before and after the reconcile
 	if log.V(5).Enabled() {
 		vms, err := v.getVirtualMachinesInCluster(ctx, supervisorMachineCtx)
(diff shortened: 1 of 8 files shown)


pkg/services/vmoperator/vmopmachine.go
```diff
@@ -188,6 +185,96 @@ func (v *VmopMachineService) ReconcileNormal(ctx context.Context, machineCtx cap
...
+			}
+			if apierrors.IsNotFound(err) {
+				log.V(4).Info("VirtualMachineGroup not found, requeueing")
+				return true, nil
+			}
+		}
+
+		// Check the presence of the node-pool label on the VirtualMachineGroup
+		nodePool := supervisorMachineCtx.Machine.Labels[clusterv1.MachineDeploymentNameLabel]
+		if zone, ok := vmOperatorVMGroup.Labels[fmt.Sprintf("capv/%s", nodePool)]; ok && zone != "" {
```
zhanggbj: As discussed in capv design doc, adding prefix may will fail badly when using long node pool names. I've dropped the prefix in VMG controller.

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -188,6 +185,96 @@ func (v *VmopMachineService) ReconcileNormal(ctx context.Context, machineCtx cap
...
+			if !apierrors.IsNotFound(err) {
+				return false, err
+			}
+			if apierrors.IsNotFound(err) {
+				log.V(4).Info("VirtualMachineGroup not found, requeueing")
+				return true, nil
+			}
+		}
+
+		// Check the presence of the node-pool label on the VirtualMachineGroup
```
zhanggbj: nit: let's use Machine Deployment in capv instead of node pool.

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -188,6 +185,96 @@ func (v *VmopMachineService) ReconcileNormal(ctx context.Context, machineCtx cap
...
+		nodePool := supervisorMachineCtx.Machine.Labels[clusterv1.MachineDeploymentNameLabel]
+		if zone, ok := vmOperatorVMGroup.Labels[fmt.Sprintf("capv/%s", nodePool)]; ok && zone != "" {
+			supervisorMachineCtx.VSphereMachine.Spec.FailureDomain = ptr.To(zone)
```
zhanggbj: For day 2 operations, we only need to create the VM with a label "topology.kubernetes.io/zone: Zone-x", but no need to set VSphereMachine.Spec.FailureDomain or Machine.Spec.FailureDomain. We should keep them consistent since they're empty from cluster specification. 

After create VM with lable, VM Service will place it into that zone.

controllers/vmware/virtualmachinegroup_reconciler.go
```diff
@@ -0,0 +1,482 @@
...
+	// 		if owner.Kind == "MachineSet" && owner.APIVersion == clusterv1.GroupVersion.String() {
+	// 			if _, ok := validMS[owner.Name]; ok {
+	// 				workerMachines[m.Name] = struct{}{}
+	// 				break
+	// 			}
+	// 		}
+	// 	}
+	// }
+	// log.V(5).Info("Identified worker Machines linked to MachineSets", "count", len(workerMachines))
+
```
zhanggbj, line 399: Hi @srm09 , to get the current vspheremachine, I think it works during initial creations.
However during an rolling upgrade

- The old MachineSet (based on the previous template) starts scaling down.
- The new MachineSet (based on the updated template) starts scaling up.

I'm afraid it's not accurate to get VSphereMachine objects by list and compare with the desired replica count, so we should filter out by the MachineSets and wait. WDYT?

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -188,6 +194,112 @@ func (v *VmopMachineService) ReconcileNormal(ctx context.Context, machineCtx cap
...
+								},
+							},
+						},
+						TopologyKey: corev1.LabelTopologyZone,
+					},
+				},
+			},
+		}
+	}
+
```
zhanggbj, line 291: @srm09 
I noticed this is removed from design doc "Is failureDomain specified? If yes, create VM with zone label."
Could you help to clarify? If failureDomain is specified, I think we could simply honor it and create VM directly with zone label. Why we are still adding AF/AAF configs here?

srm09: To keep the configuration of all VMs similar and avoiding code branching when not necessarily needed.

sbueringer:
@srm09 Can this PR be closed?

sbueringer:
/close

Feel free to reopen if necessary

PR #3685, opened 2025-11-21
:bug: Remove invalid kustomizeconfig

mdbooth:

This change has no effect on the output of this kustomization because
the removed configuration was redundant. However, it fixes a bug which
can be triggered when using this kustomization as a base for another
kustomization.

We remove 2 types of configuration. The first is redundant, the second
is incorrect and triggers the bug.

Firstly we remove nameReference configurations which specify references
which are already part of kustomize's defaults. These were found in:
* config/govmomi/webhook/kustomizeconfig.yaml
* config/supervisor/webhook/kustomizeconfig.yaml

Secondly we remove all namespace configurations as these were all
incorrect. These directed the namespace transformer to update
webhooks/clientConfig/service/namespace. However, this is not the
intended function of the namespace transformer: it should only set the
namespace directly on objects and allow references to be updated
automatically by nameReference. Configuring it to update a reference
directly leaves kustomize with inconsistent internal state. Depending on
execution order this can cause a subsequent transformation to fail to
update the reference when it makes further changes to the Service
object.

This issue is common amongst multiple providers, likely because it was present in an early version of the kubebuilder templates.

This PR is based on kubernetes-sigs/cluster-api-provider-azure#5982

To confirm that the removed configuration was redundant:

* Execute make release-manifests-all
* Copy ./out to /tmp
* Apply/revert this change
* Execute make release-manifests-all again
* Recursively diff the outputs and confirm that they are identical


chrischdi:
Tested via:

```sh
kustomize build --load-restrictor LoadRestrictionsNone config/default > $OUTDIR/default.yaml
kustomize build --load-restrictor LoadRestrictionsNone config/supervisor > $OUTDIR/supervisor.yaml
```

And comparing outputs, there was no diff 👍 

chrischdi:
/hold

Pending CI

sbueringer:
Thx!

Issue #3452, opened 2025-05-09
Tracking issue for v1Beta2


sbueringer:
Tracking issue for the work to bump to v1beta2.

This is very incomplete as of now, but we can already use it to collect tasks.

- [ ] !!! Go through Cluster API umbrella issues and check which tasks are applicable to CAPV as well

- [ ] Create v1beta2 CAPV API as a mirror copy of CAPV v1beta1 API (prior art https://github.com/kubernetes-sigs/cluster-api/pull/12037)
  - [ ] Implement conversions (no op at this stage)
  - [ ] Change all the infrav1 imports to use v1beta2
- [ ] Ensure we use `clusterv1.MachineAddress` instead of `corev1.NodeAddress` in `VSphereMachineStatus`
- [ ] Move v1beta2 conditions top level, v1beta1 conditions to deprecated  (prior art https://github.com/kubernetes-sigs/cluster-api/pull/12066)
  - [ ] Fix conversions 
- [ ] Rename conditions const (prior art https://github.com/kubernetes-sigs/cluster-api/pull/12091)
- [ ] Implement changes for the v1beta2 contract (prior art https://github.com/kubernetes-sigs/cluster-api/pull/12098, https://github.com/kubernetes-sigs/cluster-api/pull/12101)
  - [ ] Add initialization struct, move ready into it
  - [ ] Fix conversions
  - [ ] Annotate CAPD CRDs to declare they are implementing the v1beta2 contract
- [ ] Stop using v1beta1 conditions, old replica counters, failureReason and failureMessage in controllers
- [ ] Migrate E2E tests to v1beta2
- [ ] https://github.com/kubernetes-sigs/cluster-api/pull/12231
- [ ] Migrate vcsim to use v1beta2 when bumping to CAPI v1.14 / having minimum CAPI version v1.11 in upgrade tests/vsphere.yaml
   - We can't upgrade vcsim during tests because restarts would kill the vcsim instance and loose data during tests. As long as we run a test using CAPI < v1.11, we can't reconcile based on v1beta2 because the resources simply don't exist yet.

Misc:
* [ ] Consider using clusterv1.MachineAddress instead of corev1.NodeAddress

PatrickLaabs:
Greetings everyone,
I was checking, if I might help on any task on this.

The last one "Consider using clusterv1.MachineAddress instead of corev1.NodeAddress" seemed reasonable for me, to start with.

[...]

PR #3746, opened 2026-01-02
✨ Implement v1beta2 contract (govmomi)

sbueringer:

Part of #3452

### DIFF
--- a/pkg/services/vimmachine.go
+++ b/pkg/services/vimmachine.go
@@ -126,15 +126,15 @@ func (v *VimMachineService) ReconcileDelete(ctx context.Context, machineCtx capv
 }
 
 // SyncFailureReason returns true if the VSphere Machine has failed.
-func (v *VimMachineService) SyncFailureReason(ctx context.Context, machineCtx capvcontext.MachineContext) (bool, error) {
+func (v *VimMachineService) SyncFailureReason(ctx context.Context, machineCtx capvcontext.MachineContext) error {
 	vimMachineCtx, ok := machineCtx.(*capvcontext.VIMMachineContext)
 	if !ok {
-		return false, errors.New("received unexpected VIMMachineContext type")
+		return errors.New("received unexpected VIMMachineContext type")
 	}
 
 	vsphereVM, err := v.findVSphereVM(ctx, vimMachineCtx)
 	if err != nil {
-		return false, err
+		return err
 	}
 	if vsphereVM != nil {
 		// Reconcile VSphereMachine's failures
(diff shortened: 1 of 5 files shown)


pkg/services/vimmachine.go
```diff
@@ -259,8 +269,8 @@ func (v *VimMachineService) reconcileProviderID(ctx context.Context, vimMachineC
 	if providerID == "" {
 		return false, errors.Errorf("failed to reconcile providerID: invalid BIOS UUID %s for %s", biosUUID, vimMachineCtx)
 	}
-	if vimMachineCtx.VSphereMachine.Spec.ProviderID == nil || *vimMachineCtx.VSphereMachine.Spec.ProviderID != providerID {
-		vimMachineCtx.VSphereMachine.Spec.ProviderID = &providerID
+	if vimMachineCtx.VSphereMachine.Spec.ProviderID != providerID {
```
fabriziopandini, line 271: I'm wandering if we should only set if empty, because I think that CAPI doesn't handle at all changes of the ProviderID from one value to another (once node ref is set, providerID is ignored).

Note: I'm also ok in deferring this discussion, since this PR is not changing the current behaviour.

pkg/services/vimmachine.go
```diff
@@ -137,11 +138,20 @@ func (v *VimMachineService) SyncFailureReason(ctx context.Context, machineCtx ca
...
+				vimMachineCtx.VSphereMachine.Status.Deprecated.V1Beta1 = &infrav1.VSphereMachineV1Beta1DeprecatedStatus{}
+			}
+			vimMachineCtx.VSphereMachine.Status.Deprecated.V1Beta1.FailureReason = vsphereVM.Status.FailureReason
+			vimMachineCtx.VSphereMachine.Status.Deprecated.V1Beta1.FailureMessage = vsphereVM.Status.FailureMessage
+		}
 	}
 
-	return vimMachineCtx.VSphereMachine.Status.FailureReason != nil || vimMachineCtx.VSphereMachine.Status.FailureMessage != nil, err
+	return vimMachineCtx.VSphereMachine.Status.Deprecated != nil && vimMachineCtx.VSphereMachine.Status.Deprecated.V1Beta1 != nil &&
+		(vimMachineCtx.VSphereMachine.Status.Deprecated.V1Beta1.FailureReason != nil || vimMachineCtx.VSphereMachine.Status.Deprecated.V1Beta1.FailureMessage != nil), err
```
fabriziopandini: I'm wondering if we should align to CAPI were we are now ignoring failure reason and message (we only surface them), while instead in CAPV this return value leads to stop reconciling the vSphereMachine

pkg/services/vimmachine.go
sbueringer: Makes sense to me, but I think it should be a separate PR to make this change appropriately visible

sbueringer: Added to the umbrella issue

pkg/services/vimmachine.go
sbueringer: Nice catch, pushed a fix

fabriziopandini (commented):
/lgtm
/approve

PR #3750, opened 2026-01-08
✨ Support automatic CPU topology on vCenter 8+

### LINKED ISSUE #3664: Unable to set CPU Topology to "Assigned at power on"
/kind bug

We tried to set `spec.numCoresPerSocket: 0` value for a VSphereMachineTemplate. We wanted to assign that value to `0` because that's the same value vSphere interface assigns when we a VM's CPU Topology is configured to `Assigned at power on`. We confirmed that through the `Developer Center` -> `Code Capture` recording after manually cloning the machine from the template, the same flow the operator does. Using the Golang for the `Code Capture` output shows something like this:

```golang
//---------------ReconfigVM_Task---------------
var spec = types.VirtualMachineConfigSpec{}
spec.NumCoresPerSocket = 0
spec.DeviceChange = []types.BaseVirtualDeviceConfigSpec{}
var spec_VirtualNuma = &types.VirtualMachineVirtualNuma{}
spec.VirtualNuma = spec_VirtualNuma
var req = &types.ReconfigVM_Task{}
req.This = managedObject
req.Spec = spec
methods.ReconfigVM_Task(ctx, r, req)
```

Instead, every time we tried to create a new VM with the above template, it would instead set `numCoresPerSocket` to the same value as `numCPUs`.

We expected the value of `0` to be respected as that is the way vSphere does.

I've done a some local setup changing [the portion of the code that does the the incorrect assignments of `numCoresPerSocket`](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/pkg/services/govmomi/vcenter/clone.go#L175-L178). I've changed from 

```golang
.
.
if numCoresPerSocket == 0 {
.
.
```

to

```golang
.
.
if numCoresPerSocket < 0 {
.
.
```

and that made the CPU Topology from the Template I'm using be respected, instead of forcefully assigning to the same number of CPUs the machine has. I'm not sure if that's the correct way but for now it works for me, I can create a Template with the CPU Topology configured to `Assigned at power on` and that works for my use case.

**Environment:**

- Cluster-api-provider-vsphere version: 1.14.0
- Kubernetes version: (use `kubectl version`): 1.32.8
- OS (e.g. from `/etc/os-release`): Ubuntu 24.04.3 LTS

L30Bola:
**What this PR does / why we need it**: Allows for automatic CPU allocation on vCenter 8+. Support `numCoresPerSocket` to be set to `0` on vCenter 8+ to enable automatic CPU topology assignment. For older vCenter versions, a value of `0` or negative will default to `numCPUs`.

Fixes #3664

### DIFF
--- a/pkg/services/govmomi/vcenter/clone.go
+++ b/pkg/services/govmomi/vcenter/clone.go
@@ -24,6 +24,7 @@ import (
 	"math/rand"
 	"time"
 
+	"github.com/blang/semver"
 	"github.com/pkg/errors"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/pbm"
@@ -174,10 +175,25 @@ func Clone(ctx context.Context, vmCtx *capvcontext.VMContext, bootstrapData []by
 	if numCPUs < 2 {
 		numCPUs = 2
 	}
-	numCoresPerSocket := vmCtx.VSphereVM.Spec.NumCoresPerSocket
-	if numCoresPerSocket < 0 {
-		numCoresPerSocket = numCPUs
+
+	vCenterVersion, err := vmCtx.Session.GetVersion()
+	if err != nil {
+		return errors.Wrapf(err, "unable to get vCenter version for %q", vmCtx)
 	}
+	vCenterVersionSemver, _ := semver.New(string(vCenterVersion)) // ignore error, version was already validated on vmCtx.Session.GetVersion()
+
+	numCoresPerSocket := vmCtx.VSphereVM.Spec.NumCoresPerSocket
+
+	if vCenterVersionSemver.Major < 8 {
+		if numCoresPerSocket == 0 {
+			numCoresPerSocket = numCPUs
+		}
+	} else {
+		if numCoresPerSocket < 0 {
+			numCoresPerSocket = numCPUs
+		}
+	}
+
 	memMiB := vmCtx.VSphereVM.Spec.MemoryMiB
 	if memMiB == 0 {
 		memMiB = 2048


sbueringer:
We're currently working on v1beta2. I'll take a look in that context, maybe there's something that we can do

(I think branching by vCenter version is not good enough to avoid breaking changes)

L30Bola:
> We're currently working on v1beta2. I'll take a look in that context, maybe there's something that we can do

I'm not sure I followed. Isn't `clone.go` already using v1beta2? (it seems based on the imports)

> (I think branching by vCenter version is not good enough to avoid breaking changes)

I don't have much knowledge on that part, if you can suggest me what would be better to avoid breaking changes for backwards compatibility, I'd be glad to help :)

sbueringer:
> > We're currently working on v1beta2. I'll take a look in that context, maybe there's something that we can do
> 
> I'm not sure I followed. Isn't `clone.go` already using v1beta2? (it seems based on the imports)

Yes it's using v1beta2, but we just introduced v1beta2 on CAPV main so we can still make changes in that apiVersion to address this issue
 
> > (I think branching by vCenter version is not good enough to avoid breaking changes)
> 
> I don't have much knowledge on that part, if you can suggest me what would be better to avoid breaking changes for backwards compatibility, I'd be glad to help :)

No worries. Can you please take a look at this PR https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3803 I think it addresses the issue

Basically the idea is to make NumCoresPerSocket a *int32 in v1beta2 and then just pass it through to vSphere. That should make it possible to set "Assigned at power on" by simply setting `0` (the pointer makes it possible to distinguish between unset and 0)

L30Bola:
Oh, then I'll test that PR later today. Will let you know my status ASAP.

sbueringer:
> Oh, then I'll test that PR later today. Will let you know my status ASAP.

Thx. Please note that you'll have to use v1beta2 for this to work

sbueringer:
I'll close this one for now as the alternative implementation is merged. Let me know if there's something that doesn't work for you and I'll follow-up

/close

L30Bola:
No problem, @sbueringer. Sorry for the little hiatus on this, didn't have time to test this yet. Will be sure to let you know the results.

sbueringer:
No worries & no rush :)

L30Bola:
Hey, @sbueringer. I was starting to test this setup yesterday. How do I install the v1beta2 vSphere CRDs? Should I apply them from https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/tree/main/config/default/crd/bases? I downloaded the `infrastructure-components.yaml` from the 1.15.2 release and they don't have the v1beta2 version. But the ones from `config/default/crd/bases` do.

sbueringer:
Please see: https://storage.googleapis.com/k8s-staging-capi-vsphere/components/nightly_main_20260213/infrastructure-components.yaml

Corresponding templates are in the same folder: 
image

L30Bola:
Okay, everything seems to be working as expected!

Just one thing that me and a coworker bumped while testing this (a very minor one): On the `vsphereMachineTemplate` CRD description for the field `.spec.template.spec.cloneMode` still shows `FullClone` and `LinkedClone` but it only accepts `fullClone` and `linkedClone`. Aside from that, 100% working.

Thanks for the very quick reply.

sbueringer:
> Okay, everything seems to be working as expected!

Perfect!

> Just one thing that me and a coworker bumped while testing this (a very minor one): On the vsphereMachineTemplate CRD description for the field .spec.template.spec.cloneMode still shows FullClone and LinkedClone but it only accepts fullClone and linkedClone

Ah, nice catch. Feel free to open a PR for it if you want. Otherwise I can also take care of it

sbueringer:
PR: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3838

Issue #3758, opened 2026-01-14
Support for multiple VM operator versions


fabriziopandini:
/kind feature

**Describe the solution you'd like**
New versions of CAPV need to continue to work on older vSphere / supervisor versions. 

This implies e.g. to be able to interact with VM operator's v1alpha2 API version when CAPV is running on vSphere 8.x, while instead using VM operator's v1alpha5 API version on newer vSphere versions.

**Tracking issue**

- [ ] Start testing with vm operator v1.9 https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3756
- [ ] Introduce support for multiple VM operator versions
    - [ ] Create all the required machinery, change production code to use it, but still use v1alpha2 to ensure no-regression happens
    - [ ] Change production code to use v1alpha5, test
    - [ ] Introduce the dynamic switch from v1alpha2 to v1alpha5 + E2E test for both scenarios
- [ ] Other
  - [ ] pick up the govc version after v0.52.0 as soon as available, drop tmp change in Makefile https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3756#discussion_r2691011537
  - [ ] pick up the new vmoperator version as available and start using it for testing as well

sbueringer:
🎉 

/close

PR #3767, opened 2026-01-15
✨ Add support for multiple supervisor versions

fabriziopandini:

Part of  https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/3758

### DIFF
--- a/pkg/conversion/api/vmoperator/hub/virtualmachine_hardware_types.go
+++ b/pkg/conversion/api/vmoperator/hub/virtualmachine_hardware_types.go
@@ -14,7 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// nolint: revive
 package hub
 
 // +kubebuilder:validation:Enum=IDE;NVME;SCSI;SATA
--- a/controllers/vmware/servicediscovery_controller.go
+++ b/controllers/vmware/servicediscovery_controller.go
@@ -152,7 +152,7 @@ type serviceDiscoveryReconciler struct {
 	clusterCache clustercache.ClusterCache
 }
 
-func (r *serviceDiscoveryReconciler) Reconcile(ctx context.Context, req reconcile.Request) (_ reconcile.Result, reterr error) {
+func (r *serviceDiscoveryReconciler) Reconcile(ctx context.Context, req reconcile.Request) (ret reconcile.Result, reterr error) {
 	log := ctrl.LoggerFrom(ctx)
 
 	// Get the vspherecluster for this request.
@@ -197,6 +197,14 @@ func (r *serviceDiscoveryReconciler) Reconcile(ctx context.Context, req reconcil
 	defer func() {
 		if err := r.patch(ctx, clusterContext); err != nil {
 			reterr = kerrors.NewAggregate([]error{reterr, err})
+		}
+
+		if reterr != nil {
+			// Note: controller-runtime logs a warning if an error is returned in combination with
+			// RequeueAfter / Requeue. Dropping RequeueAfter and Requeue here to avoid this warning
+			// (while preserving Priority).
+			ret.RequeueAfter = 0
+			ret.Requeue = false //nolint:staticcheck // We have to handle Requeue until it is removed
 		}
 	}()
 
--- a/test/infrastructure/net-operator/controllers/virtualmachine_controller.go
+++ b/test/infrastructure/net-operator/controllers/virtualmachine_controller.go
@@ -89,8 +89,13 @@ func (r *VirtualMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 	annotations[VMKickTimeAnnotation] = reconcileTime.Add(15 * time.Second).Format(time.RFC3339)
 
+	patch, err := conversionclient.MergeFrom(ctx, r.Client, o)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrapf(err, "failed to create patch for VirtualMachine object")
+	}
+
 	virtualMachine.SetAnnotations(annotations)
-	if err := r.Client.Patch(ctx, virtualMachine, client.MergeFrom(o)); err != nil {
+	if err := r.Client.Patch(ctx, virtualMachine, patch); err != nil {
 		return ctrl.Result{}, errors.Wrapf(err, "failed to patch VirtualMachine %s", klog.KObj(virtualMachine))
 	}
 


controllers/vmware/virtualmachinegroup_controller.go
```diff
@@ -46,13 +47,19 @@ func AddVirtualMachineGroupControllerToManager(ctx context.Context, controllerMa
 		Recorder: mgr.GetEventRecorderFor("virtualmachinegroup-controller"),
 	}
 
+	// NOTE: use vm-operator native types for watches (the reconciler uses the internal hub version).
+	vmGroup, err := conversionclient.WatchObject(reconciler.Client, &vmoprvhub.VirtualMachineGroup{})
+	if err != nil {
+		return err
```
sbueringer: Let's add a bit of error wrapping here please (please also check other usages of WatchObject)

main.go
```diff
@@ -109,8 +114,17 @@ var (
 	defaultSyncPeriod       = manager.DefaultSyncPeriod
 	defaultLeaderElectionID = manager.DefaultLeaderElectionID
 	defaultPodName          = manager.DefaultPodName
+
+	scheme *runtime.Scheme
 )
 
+func init() {
+	scheme = runtime.NewScheme()
```
sbueringer: Can we move the creation of this scheme into the if where we use it?

If it only has v1alpha2 / v1alpha5 it's a bit dangerous if it's used elsewhere accidentally

pkg/services/vmoperator/control_plane_endpoint.go
```diff
@@ -249,7 +250,10 @@ func (s *CPService) createVMControlPlaneService(ctx context.Context, clusterCtx
 			return nil, err
 		}
 	} else if !reflect.DeepEqual(originalVMService, vmService) {
-		patch := client.MergeFrom(originalVMService)
+		patch, err := conversionclient.MergeFrom(s.Client, originalVMService)
```
sbueringer: Let's check that the errors that might occur here have enough context. Either by having additional error wrapping here or inside the func (maybe we need both)

test/infrastructure/vcsim/controllers/vspherevm_controller_test.go
```diff
@@ -202,31 +205,31 @@ func Test_Reconcile_VSphereVM(t *testing.T) {
...
 				UID:       "bar",
 			},
-			Spec: clusterv1.ClusterSpec{
-				InfrastructureRef: clusterv1.ContractVersionedObjectReference{
-					APIGroup: infrav1beta1.GroupVersion.Group,
-					Kind:     "VSphereCluster",
-					Name:     vsphereCluster.Name,
+			Spec: clusterv1beta1.ClusterSpec{
+				InfrastructureRef: &corev1.ObjectReference{
+					APIVersion: infrav1beta1.GroupVersion.Group,
```
sbueringer: nit: value should not just be the group

pkg/conversion/internal/test/api/hub/types.go
```diff
@@ -0,0 +1,55 @@
+/*
```
sbueringer, line 1: Would be great if we can move pkg/conversion/internal/api to one of
* pkg/conversion/internal/test/
* pkg/conversion/internal/test/api/

Just to make clear that this is just for tests (prior art core CAPI internal/topology/upgrade/test/t1)

pkg/conversion/converter.go
```diff
@@ -0,0 +1,356 @@
...
+		targetVersionSelector: func(_ schema.GroupKind) string {
+			panic("targetVersionSelector not set")
+		},
+	}
+	return s
+}
+
+// SetTargetVersion sets the target version to be used for all groups and kinds known by this converter.
+func (s *Converter) SetTargetVersion(v string) {
+	s.targetVersionSelector = func(_ schema.GroupKind) string { return v }
```
sbueringer, line 69: I assume there is a reason, but why is this not a parameter in NewConverter?

pkg/conversion/converter.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+// AddTypes adds to the converter types that require conversion.
+func (s *Converter) AddTypes(gv schema.GroupVersion, types ...runtime.Object) error {
+	if gv.Version == "" {
```
sbueringer, line 78: Would it make sense to also check Group?

pkg/conversion/converter.go
```diff
@@ -0,0 +1,356 @@
...
+
+// Recognizes returns true if the converter is able to handle the provided group,version,kind
+// of an object.
+func (s *Converter) Recognizes(gvk schema.GroupVersionKind) bool {
+	_, exists := s.gvkToType[gvk]
+	return exists
+}
+
+// AddConversion adds to the Converter functions to be used when converting objects from one version to another.
+func (s *Converter) AddConversion(src runtime.Object, version string, dst runtime.Object, srcToDst, dstToSrc any) error {
```
sbueringer: version seems to be dstVersion. If I'm correct let's rename the param

Let's add Func as a suffix to the last two parameter

pkg/conversion/converter.go
```diff
@@ -0,0 +1,356 @@
...
+	gvkSrc, err := s.GroupVersionKindFor(src)
```
sbueringer: nit: can we "flip" the var names in this func? e.g. instead of gvkSrc => srcGVK, tSrc => srcT

I think it reads much easier and I think usually we name vars this way around

(same in the Convert method)

pkg/conversion/converter.go
```diff
@@ -0,0 +1,356 @@
...
+		Group:   gvkSrc.Group,
+		Version: version,
+		Kind:    tDst.Name(),
+	}
+	if gvkDst.Version == "" {
+		return errors.New("invalid version, version cannot be empty")
+	}
+
+	if gvkDst.Version == gvkSrc.Version {
+		return errors.New("invalid version, target version for a conversion cannot be the same registered for the source object")
```
sbueringer: ""same registered for the source object" sounds a bit off

pkg/conversion/converter.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+
+	if gvkSrc.Kind != gvkDst.Kind {
+		return errors.New("invalid destination type, destination type for a conversion must be of the same kind of the source object")
```
sbueringer:
```suggestion
		return errors.New("invalid destination type, destination type for a conversion must be of the same kind as the source object")
```

pkg/conversion/converter.go
```diff
@@ -0,0 +1,356 @@
...
+	if oldC, found := s.conversionFuncs[gvkDst][gvkSrc]; found && oldC != dstToSrcV {
+		return errors.Errorf("double registration of conversion function from %v to %v: old function is different from the new function", gvkDst, gvkSrc.Version)
+	}
+	s.conversionFuncs[gvkDst][gvkSrc] = dstToSrcV
+
+	return nil
+}
+
+// Convert converts an object into another with the same kind, but a different version.
+func (s *Converter) Convert(src runtime.Object, dst runtime.Object) error {
```
sbueringer: I don't know but I hope all this reflection does not become a performance issue

controller-runtime conversions work without reflection like this during every conversion

In general I'm wondering if all of this has to be done with reflection, e.g. why are we not using interfaces to get/set SourceTypeMeta?

pkg/conversion/converter.go
```diff
@@ -0,0 +1,356 @@
...
+// IsConvertible return true if an object requires conversion before write and after read.
+func (s *Converter) IsConvertible(obj runtime.Object) bool {
+	gvk, err := s.GroupVersionKindFor(obj)
+	if err != nil {
+		return false
+	}
+	return s.gvkConvertibleTypes[gvk]
+}
+
+// TargetGroupVersionKindFor returns the GroupVersionKind for the given object converted to the target version.
```
sbueringer: Looks like this should only be called with hub types. Maybe let's mention that in the godoc

pkg/conversion/converter.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+// Recognizes returns true if the converter is able to handle the provided group,version,kind
+// of an object.
+func (s *Converter) Recognizes(gvk schema.GroupVersionKind) bool {
+	_, exists := s.gvkToType[gvk]
+	return exists
+}
+
+// AddConversion adds to the Converter functions to be used when converting objects from one version to another.
```
sbueringer, line 127: Let's extend the godoc with an example how this func should be called. `any` is not very intuitive

pkg/conversion/converter.go
```diff
@@ -0,0 +1,356 @@
...
+	gvk, ok := s.typeToGVK[t]
+	if !ok {
+		return schema.GroupVersionKind{}, errors.Errorf("no type registered for %s.%s", t.PkgPath(), t.Name())
+	}
+	return gvk, nil
+}
+
+var errorType = reflect.TypeFor[error]()
+
+func conversionFuncIsValid(tSrc, tDst reflect.Type, f any) error {
```
sbueringer: Let's please add a godoc that shows how a valid func looks like. That makes it easier to read the reflection code below

pkg/conversion/converter_builder.go
```diff
@@ -0,0 +1,60 @@
...
+
+// AddTypes adds to the Converter types that require conversion.
+func (sb *ConverterBuilder) AddTypes(gv schema.GroupVersion, types ...runtime.Object) {
+	*sb = append(*sb, func(s *Converter) error {
+		return s.AddTypes(gv, types...)
+	})
+}
+
+// AddConversion adds to the Converter functions to be used when converting objects from one version to another.
+func (sb *ConverterBuilder) AddConversion(hub runtime.Object, version string, spoke runtime.Object, srcToDst, dstToSrc any) {
```
sbueringer: Maybe let's also add an example call here to show how any looks like

pkg/conversion/converter_test.go
```diff
@@ -0,0 +1,814 @@
...
+
+func init() {
+	v1alpha5ConverterBuilder.AddConversion(
+		&hub.A{},
+		v1alpha5GroupVersion.Version, &v1alpha5.A{},
+		v1alpha5.ConvertAFromHubToV1alpha5, v1alpha5.ConvertAFromV1alpha5ToHub,
+	)
+}
+
+type A struct {
```
sbueringer, line 75: I assume this is v1alpha2 here, can we please also move this into a separate package like v1alpha5 and hub?

pkg/conversion/converter_test.go
```diff
@@ -0,0 +1,814 @@
...
+			name:      "Fails when convertible object has a different source version than the target version",
+			converter: c,
+			src: &hub.A{
+				Foo: "bar",
+				Source: conversionmeta.SourceTypeMeta{
+					APIVersion: v1alpha2GroupVersion.String(),
+				},
+			},
+			dst:     &v1alpha5.A{},
+			wantDst: &v1alpha5.A{Foo: "bar"},
```
sbueringer:
```suggestion
```
This seems wrong?

pkg/conversion/converter_test.go
```diff
@@ -0,0 +1,814 @@
...
+			name: "Return true for a convertible object",
+			obj:  &hub.A{},
+			want: true,
+		},
+		{
+			name: "Return false for a type a convertible object can convert to",
+			obj:  &v1alpha5.A{},
+			want: false,
+		},
+		{
```
sbueringer: duplicate

pkg/conversion/converter_test.go
```diff
@@ -0,0 +1,814 @@
...
+			name: "Fails when conversions to the target version are not registered",
+			converter: func() *Converter {
+				c := NewConverter()
+				c.SetTargetVersion(v1alpha5GroupVersion.Version)
+				utilruntime.Must(addHubToConverter(c))
+				return c
+			}(),
+			obj:     &hub.A{},
+			wantGvk: v1alpha5GroupVersion.WithKind("A"),
```
sbueringer:
```suggestion
```
I think you can just check against tt.wantGVK in l.604 (then this would have showed up)

Same in l.671

pkg/conversion/test/roundtrip.go
```diff
@@ -0,0 +1,100 @@
...
+
+	FuzzerFuncs []any
+}
+
+// RoundTripTest returns a new testing function to be used in tests to make sure conversions between
+// the Hub version of an object and an the corresponding Spoke version aren't lossy.
+func RoundTripTest(input RoundTripTestInput) func(*testing.T) {
+	return func(t *testing.T) {
+		t.Helper()
+		t.Run("hub-spoke-hub", func(t *testing.T) {
```
sbueringer, line 54: Why are we specifically only testing hub-spoke-hub?

pkg/conversion/test/roundtrip.go
```diff
@@ -0,0 +1,100 @@
...
+// RoundTripTestInput contains input parameters
+// for the RoundTripTest function.
+type RoundTripTestInput struct {
+	Scheme    *runtime.Scheme
+	Converter *conversion.Converter
+
+	Hub   client.Object
+	Spoke client.Object
+
+	FuzzerFuncs []any
```
sbueringer: Wondering if we should use the same type here as for the conversion test util we already have

pkg/conversion/test/roundtrip.go
```diff
@@ -0,0 +1,100 @@
...
+			fuzzer := conversionutil.GetFuzzer(input.Scheme, func(_ runtimeserializer.CodecFactory) []any {
+				return append(input.FuzzerFuncs, func(in *conversionmeta.SourceTypeMeta, _ randfill.Continue) {
+					// Ensure SourceTypeMeta is not set by the fuzzer.
+					in.APIVersion = ""
+				})
+			})
+
+			spokeGVK, err := input.Converter.GroupVersionKindFor(input.Spoke)
+			if err != nil {
+				t.Fatal(err.Error())
```
sbueringer, line 74: Q: Why is this using t.Fatal instead of gomega like on our other conversion test util?

pkg/conversion/test/roundtrip.go
```diff
@@ -0,0 +1,100 @@
...
+				}
+
+				// Convert spoke back to hub and check if the resulting hub is equal to the hub before the round trip
+				hubAfter := input.Hub.DeepCopyObject()
+				if err := input.Converter.Convert(spoke, hubAfter); err != nil {
+					t.Fatalf("error calling ConvertTo: %v", err)
+				}
+
+				if source, err := conversionmeta.GetSource(hubAfter); err != nil || source.APIVersion != spokeGVK.GroupVersion().String() {
+					t.Fatal("ConvertTo is expected to set Convertible.APIVersion")
```
sbueringer:
```suggestion
					t.Fatal("Convert is expected to set Convertible.APIVersion")
```

pkg/conversion/converter.go
fabriziopandini: The actual implementation of the dynamic version selection will come as a follow up step in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/3758, and then I will have all the info to make an informed decision about it to move this as a New parameter or else.

What if have now is mostly a placeholder, and this func comes very handy for controlling the target version when testing the conversion client

pkg/conversion/converter.go
fabriziopandini: No strong opinions, but in theory empty group is valid in K8s (even though I don't think we will ever use them as a hub version).

pkg/conversion/converter.go
fabriziopandini: rephrased, PTAL

pkg/conversion/converter.go
fabriziopandini: I went back on forth on using reflection and having an interface for get/set SourceTypeMeta, and then sort of converged on solutions that are minimizing the toils required when implementing hub types and conversions

I have a few ideas on how to improve this in follow up iterations, we can chat about this at the first chance

pkg/conversion/converter.go
sbueringer: Makes sense, that's fine

I was only asking because I would expect that a converter created with a constructor doesn't panic during usage.
But it's fine if that's changed in a follow-up PR

pkg/conversion/converter.go
sbueringer: Good point. Fine either way for me. Do we deal with an empty Group string everywhere correctly? (otherwise the easy option would be to just block it)

pkg/conversion/converter.go
sbueringer: Sounds good. Definitely not a blocker for this PR. Probably changes around this later would only require small changes elsewhere

pkg/conversion/converter_test.go
fabriziopandini: I intentionally had to duplicate some types here to test e.g. when same type is already registered (added a not)

pkg/conversion/converter_test.go
fabriziopandini: Dropped the line (the test case return an error)

pkg/conversion/test/roundtrip.go
fabriziopandini: Because if we had to start from spoke, then we will need to implement a lot of custom fuzz func to set all the subset of fields that are relevant for CAPV.

This way we are still testing both the conversion directions, but without additional toil

pkg/conversion/test/roundtrip.go
fabriziopandini: I'm not entirely sure what you mean but this test is specific for the converter, and I don't think we should create dependencies with other tests to avoid duplication of one type.

pkg/conversion/test/roundtrip.go
fabriziopandini: I did this to non liked the idea of having an import for gomega in "prod" code, but no strong opinion

pkg/conversion/test/roundtrip.go
sbueringer: Makes sense, got it

pkg/conversion/test/roundtrip.go
sbueringer: I'm wondering if this should be []fuzzer.FuzzerFuncs instead of []any like in https://github.com/kubernetes-sigs/cluster-api/blob/24ba35ab15fc16d7c851c9a0a7454d842b85f6d7/util/conversion/conversion.go#L144

Might be nice for folks if this is as close as possible to how conversion tests are implemented

pkg/conversion/test/roundtrip.go
sbueringer: But this is not prod code right, it's a unit test util? (isn't it basically the same as https://github.com/kubernetes-sigs/cluster-api/blob/24ba35ab15fc16d7c851c9a0a7454d842b85f6d7/util/conversion/conversion.go#L149)

fabriziopandini: I'm not 100% sure, but it is not in a _Test file, so it might end up in the manager binary (if this is correct or we have some doubts we should probably fix in CAPI too)

pkg/conversion/converter.go
fabriziopandini: I think we are ok, but blocked for extra safety

pkg/conversion/test/roundtrip.go
sbueringer: > I'm not 100% sure, but it is not in a _Test file, so it might end up in the manager binary 

I think it's okay. This is the same situation as for all exported code of test libraries.

In this case it's even relatively safe as folks would have to construct a *testing.T manually to use it
Also usage of t.Fatal would be bad in production controller code

(that being said no super strong opinion about t.Fatal vs Gomega, just thought it would be nice to be consistent with our existent util so it behaves ~ the same)

pkg/conversion/test/roundtrip.go
```diff
@@ -0,0 +1,105 @@
...
+			}
+
+			for range 10000 {
+				// Create the hub and fuzz it
+				hubBefore := input.Hub.DeepCopyObject()
+				fuzzer.Fill(hubBefore)
+
+				// First convert hub to spoke
+				spoke := input.Spoke.DeepCopyObject()
+				if err := input.Converter.Convert(context.TODO(), hubBefore, spoke); err != nil {
```
sbueringer: Let's use t.Context() here and in l.88

pkg/conversion/converter_builder.go
```diff
@@ -0,0 +1,112 @@
...
+
+type conversionBuilder[hubObject, spokeObject runtime.Object] struct {
+	convertHubToSpokeFunc func(ctx context.Context, src hubObject, dst spokeObject) error
+	convertSpokeToHubFunc func(ctx context.Context, src spokeObject, dst hubObject) error
+}
+
+// Build a func that adds a conversion to a Converter.
+func (c conversionBuilder[hubObject, spokeObject]) Build(version string) func(converter *Converter) error {
+	return func(converter *Converter) error {
+		convertHubToSpokeAnyFunc := func(ctx context.Context, hub runtime.Object, spoke runtime.Object) error {
```
sbueringer:
```suggestion
		convertHubToSpokeFunc := func(ctx context.Context, hub runtime.Object, spoke runtime.Object) error {
```
I think, same in l.102. Or just inline the funcs into 105

pkg/conversion/converter.go
```diff
@@ -0,0 +1,317 @@
...
+// converter.AddConversion(
+//
+//	&vmoprvhub.VirtualMachine{},
+//	vmoprv1alpha2.GroupVersion.Version, &vmoprv1alpha2.VirtualMachine{},
+//	convert_hub_VirtualMachine_To_v1alpha2_VirtualMachine, convert_v1alpha2_VirtualMachine_To_hub_VirtualMachine,
+//
+// )
+//
+// AddConversionBuilder provides a convenient and typed way to perform this operation. More examples can be found in pkg/conversion/api/vmoperator.
+func (s *Converter) AddConversion(src runtime.Object, dstVersion string, dst runtime.Object, srcToDstFunc, dstToSrcFunc ConvertFunc) error {
```
sbueringer: Just realized that "dst to src" sounds pretty confusing as in that case dst is our source and src is our dest?

Maybe we should switch to hub/spoke naming, but entirely up to you, I'm okay with the current state

pkg/conversion/converter_test.go
```diff
@@ -0,0 +1,727 @@
...
+			src:       &corev1.Node{},
+			dst:       &hub.A{},
+			wantErr:   true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewWithT(t)
+
+			err := tt.converter.Convert(context.TODO(), tt.src, tt.dst)
```
sbueringer: Let's use t.Context()

pkg/conversion/api/meta/types.go
```diff
@@ -0,0 +1,37 @@
...
+*/
+
+// Package meta defines metadata for the hub version of supervisor's API objects.
+package meta
+
+// SourceTypeMeta defines type meta of the source of this object.
+type SourceTypeMeta struct {
+	// APIVersion defines the versioned schema of this representation of an object.
+	// Servers should convert recognized schemas to the latest internal value, and
+	// may reject unrecognized values.
```
sbueringer: Not sure if this part of the comment is helpful

pkg/conversion/api/vmoperator/hub/doc.go
```diff
@@ -0,0 +1,30 @@
...
+// - When CAPV reads and write vm-operator resources, it will use vm-operator's preferred version for the specific environment.
+// - vm-operator's preferred version depends primarily on the version of vm-operator that exists in the environment.
+// - Apart from the read and write operations, CAPV will use its own hub types for processing.
+//
+// Notably:
+// - CAPV hub types are a subset of vm-operator's types, with only the fields that are relevant for CAPV.
+//
+// +k8s:openapi-gen=true
+// +kubebuilder:object:generate=true
+// +groupName=infrastructure.cluster.x-k8s.io
```
sbueringer: We should probably use vmoperator.vmware.com

EDIT: not sure if this marker and the one in l.27 is used

pkg/conversion/api/vmoperator/hub/virtualmachine_types.go
```diff
@@ -0,0 +1,405 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   VirtualMachineSpec   `json:"spec,omitempty"`
+	Status VirtualMachineStatus `json:"status,omitempty"`
+
+	Source conversionmeta.SourceTypeMeta `json:"source,omitempty,omitzero"`
+}
+
+// GetConditions returns the set of conditions for this object.
+func (in *VirtualMachine) GetConditions() []metav1.Condition {
```
sbueringer: nit: for diff'ing it would be nice if we would not change the methods copied from vm-operator

(added godoc, changed method receiver, also changed receiver from non-pointer to pointer)


(maybe same for other types)

pkg/conversion/api/vmoperator/hub/virtualmachinegroup_types.go
```diff
@@ -0,0 +1,172 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   VirtualMachineGroupSpec   `json:"spec,omitempty"`
+	Status VirtualMachineGroupStatus `json:"status,omitempty"`
+
+	Source conversionmeta.SourceTypeMeta `json:"source,omitempty,omitzero"`
+}
+
+// GetConditions returns the set of conditions for this object.
+func (in *VirtualMachineGroupMemberStatus) GetConditions() []metav1.Condition {
```
sbueringer: Should this and the next method be on VirtualMachineGroup? (also here I would prefer if we don't modify copied methods)

(If we don't need the methods, let's just not copy them)

pkg/conversion/api/vmoperator/hub/virtualmachinesetresourcepolicy_types.go
```diff
@@ -0,0 +1,87 @@
...
+type VirtualMachineSetResourcePolicySpec struct {
+	ResourcePool        ResourcePoolSpec `json:"resourcePool,omitempty"`
+	Folder              string           `json:"folder,omitempty"`
+	ClusterModuleGroups []string         `json:"clusterModuleGroups,omitempty"`
+}
+
+// VirtualMachineSetResourcePolicyStatus defines the observed state of
+// VirtualMachineSetResourcePolicy.
+type VirtualMachineSetResourcePolicyStatus struct {
+	/*
```
sbueringer: Why is this commented out?

pkg/conversion/client/watches.go
```diff
@@ -0,0 +1,38 @@
...
+package client
+
+import (
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// WatchObject creates an object to be used for building controller watches.
+// The generated object is the converted version of the input object.
+func WatchObject(c client.Client, obj client.Object) (client.Object, error) {
```
sbueringer, line 25: Maybe let's also add a unit test for this one

pkg/conversion/client/client.go
```diff
@@ -0,0 +1,415 @@
...
+	vmoprvhub "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api/vmoperator/hub"
+	vmoprv1alpha2conversion "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api/vmoperator/v1alpha2"
+	vmoprv1alpha5conversion "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api/vmoperator/v1alpha5"
+)
+
+// DefaultConverter is a converter aware of the API types and the conversions defined in sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api.
+var DefaultConverter *conversion.Converter
+
+func init() {
+	DefaultConverter = conversion.NewConverter()
```
sbueringer: Just to mention it. Having this here in this file means the client package has transitive dependencies on our vm-operator API packages (see hub/v1alpha2/v1alpha5 below) (+ the vm-operator API packages of the vm-operator repo).

I don't know if this might lead to compile issues for someone importing this client down the line if they don't use the exact same vm-operator commit that we do

Maybe the default converter should be somewhere over there and not here in the client package

pkg/conversion/client/client.go
```diff
@@ -0,0 +1,415 @@
...
+	converter      *conversion.Converter
+	internalClient client.Client
+}
+
+// conversionClient must implement client.Client.
+var _ client.Client = &conversionClient{}
+
+// Get retrieves an obj for the given object key from the Kubernetes Cluster.
+func (c *conversionClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	if !c.converter.IsConvertible(obj) {
```
sbueringer: Just a general note. Across these packages we use a lot of different wording:
* src / dst
* target
* hub / spoke
* convertible

I think it would be easier to understand if wouldn't use that many different words to refer to the same (if possible)

pkg/conversion/client/client.go
```diff
@@ -0,0 +1,415 @@
...
+		return c.internalClient.Get(ctx, key, obj, opts...)
+	}
+
+	targetVersionObj, err := c.newTargetVersionObjectFor(obj)
+	if err != nil {
+		return err
```
sbueringer, line 90: Might be worth to do one round of double checking in this file to see if the error context we provide is enough, e.g. most of the errors below don't surface what "action" (Get/Apply/Create) we are doing when the error occurs

(I'm being relatively nitpicky on this because this code will be run in a lot of places)

pkg/conversion/client/client.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+
+	listObjs := []runtime.Object{}
+	for _, targetItemRaw := range targetItems {
+		targetItem, ok := targetItemRaw.(client.Object)
+		if !ok {
+			return errors.Errorf("%T does not implement client.Object", targetItemRaw)
+		}
```
sbueringer:
```suggestion
	for _, targetItem := range targetItems {
		targetItem, ok := targetItem.(client.Object)
		if !ok {
			return errors.Errorf("%T does not implement client.Object", targetItem)
		}
```
Raw seems a bit misleading to me

pkg/conversion/client/client.go
```diff
@@ -0,0 +1,415 @@
...
+		listObjs = append(listObjs, listItem)
+	}
+
+	return meta.SetList(list, listObjs)
+}
+
+// Apply applies the given apply configuration to the Kubernetes cluster.
+func (c *conversionClient) Apply(ctx context.Context, obj runtime.ApplyConfiguration, opts ...client.ApplyOption) error {
+	cObj, ok := obj.(client.Object)
```
sbueringer: I think regular ApplyConfigurations do not implement client.Object. So this check here will break Apply 

I think we also only have to cast to runtime.Object for IsConvertible.

I would change this to:
* if we can cast and if it is convertible => panic
* otherwise => internalClient.Apply

Let's add unit test coverage for the case that currently doesn't work, all assuming I"m correct. Not sure how familiar you are with ApplyConfigurations, here is an example: https://github.com/kubernetes-sigs/controller-runtime/blob/7fb375a4991a8745478768b600a6687da8203002/pkg/client/client_test.go#L1072

pkg/conversion/client/client.go
```diff
@@ -0,0 +1,415 @@
...
+
+	if !c.converter.IsConvertible(cObj) {
+		return c.internalClient.Apply(ctx, obj, opts...)
+	}
+
+	// conversionClient only implements methods used in CAPV.
+	panic("implement me")
+}
+
+// Create saves the object obj in the Kubernetes cluster.
```
sbueringer:
```suggestion
// Create creates the object obj in the Kubernetes cluster.
```
nit

pkg/conversion/client/client.go
```diff
@@ -0,0 +1,415 @@
...
+	if !ok {
+		return errors.Errorf("%T does not implement client.Object", obj)
+	}
+
+	if !c.converter.IsConvertible(cObj) {
+		return c.internalClient.Apply(ctx, obj, opts...)
+	}
+
+	// conversionClient only implements methods used in CAPV.
+	panic("implement me")
```
sbueringer: Maybe we should return errors instead of panic'ing (same below)

I think either way folks will notice that nothing works if they hit this

(but not sure)

pkg/conversion/client/client.go
```diff
@@ -0,0 +1,415 @@
...
+}
+
+// Patch patches the given obj in the Kubernetes cluster.
+func (c *conversionClient) Patch(ctx context.Context, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
+	if !c.converter.IsConvertible(obj) {
+		return c.internalClient.Patch(ctx, obj, patch, opts...)
+	}
+
+	if _, ok := patch.(*conversionMergePatch); !ok {
+		return errors.Errorf("patch must be created using sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/client.MergeFrom")
```
sbueringer: nit: or MergeFromWithOptions?

pkg/conversion/client/client.go
```diff
@@ -0,0 +1,415 @@
...
+}
+
+// DeleteAllOf deletes all objects of the given type matching the given options.
+func (c *conversionClient) DeleteAllOf(ctx context.Context, obj client.Object, opts ...client.DeleteAllOfOption) error {
+	if !c.converter.IsConvertible(obj) {
+		return c.internalClient.DeleteAllOf(ctx, obj, opts...)
+	}
+
+	// conversionClient only implements methods used in CAPV.
+	panic("implement me")
```
sbueringer: Might be worth implementing this one anyway. I think it should be straightforward

pkg/conversion/client/client.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+
+	if err := c.converter.Convert(ctx, targetVersionObj, obj); err != nil {
+		return errors.Wrapf(err, "failed to convert %s from target version while patching status on it", klog.KObj(targetVersionObj))
+	}
+	return nil
+}
+
+func (c conversionSubResourceClient) Apply(ctx context.Context, obj runtime.ApplyConfiguration, opts ...client.SubResourceApplyOption) error {
+	if !c.converter.IsConvertible(obj.(runtime.Object)) {
```
sbueringer: Similar to the other Apply, just wondering on top if this one panics

pkg/conversion/client/patches.go
```diff
@@ -0,0 +1,125 @@
...
+		from:          obj,
+		client:        cc,
+		options:       opts,
+	}, nil
+}
+
+// MergeFrom creates a Patch that patches using the merge-patch strategy with the given object as base.
+// When required, the generated patch performs conversion for both/one of the original or the target object.
+func MergeFrom(ctx context.Context, c client.Client, obj client.Object) (client.Patch, error) {
+	cc, ok := c.(*conversionClient)
```
sbueringer: Can we reuse MergeFromWithOptions here internally?

pkg/conversion/client/patches.go
```diff
@@ -0,0 +1,125 @@
...
+	return &conversionMergePatch{
+		conversionCtx: ctx,
+		from:          obj,
+		client:        cc,
+		options:       opts,
+	}, nil
+}
+
+// MergeFrom creates a Patch that patches using the merge-patch strategy with the given object as base.
+// When required, the generated patch performs conversion for both/one of the original or the target object.
```
sbueringer, line 56: I'm not entirely sure I understand from that sentence what we support

From the code it looks like:
* obj must be a hub
* the obj passed into Data can be hub or spoke

I'm not sure if we should really support the combination of hub and spoke. It seems very error prone to write a controller that passes an object of hub type into client.MergeFrom and then produces it's own object of spoke type

pkg/conversion/client/patches.go
```diff
@@ -0,0 +1,125 @@
...
+
+type conversionMergePatch struct {
+	conversionCtx context.Context //nolint:containedctx
+	client        *conversionClient
+
+	from    client.Object
+	options []client.MergeFromOption
+}
+
+// conversionClient must implement client.Patch.
```
sbueringer:
```suggestion
// conversionMergePatch must implement client.Patch.
```

pkg/conversion/client/patches_test.go
```diff
@@ -0,0 +1,238 @@
...
+		obj       client.Object
+		wantPatch *conversionMergePatch
+		wantErr   bool
+	}{
+		{
+			name: "Creates a patch",
+			c:    cc,
+			obj:  obj,
+			wantPatch: &conversionMergePatch{
+				conversionCtx: context.TODO(),
```
sbueringer: let's use t.Context()

We should only use context.TODO() if there is no context available that we can use today but we are planning to add it in the future

pkg/conversion/client/client_test.go
```diff
@@ -0,0 +1,1140 @@
...
+
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion"
+	conversionmeta "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api/meta"
+	vmoprvhub "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api/vmoperator/hub"
+	vmoprv1alpha2conversion "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api/vmoperator/v1alpha2"
+	vmoprv1alpha5conversion "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api/vmoperator/v1alpha5"
+)
+
+var (
+	ctx       = context.TODO()
```
sbueringer: Let's use t.Context() instead of a global todo context

pkg/conversion/client/client_test.go
```diff
@@ -0,0 +1,1140 @@
...
+					Namespace: "test-ns",
+				},
+				Spec: vmoprvhub.VirtualMachineSpec{
+					ClassName: "test-class",
+				},
+				Source: conversionmeta.SourceTypeMeta{
+					APIVersion: vmoprv1alpha2.GroupVersion.String(),
+				},
+			},
+			modifyFunc: func(o client.Object) client.Object {
```
sbueringer, line 681: *If* we decide to keep supporting it, I think this is missing test coverage where we pass an object of spoke type into Patch

pkg/conversion/client/client_test.go
```diff
@@ -0,0 +1,1140 @@
...
+			targetVersion: vmoprv1alpha5.GroupVersion.Version,
+			obj: &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-vm",
+					Namespace: "test-n",
+				},
+			},
+			modifyFunc: func(o client.Object) client.Object {
+				n := o.(*corev1.Node)
+				n.Spec.ProviderID = "fpp"
```
sbueringer: This should write something in status

pkg/conversion/client/patches.go
```diff
@@ -0,0 +1,125 @@
...
+var _ client.Patch = &conversionMergePatch{}
+
+// Type is the PatchType of the patch.
+func (p *conversionMergePatch) Type() types.PatchType {
+	return types.MergePatchType
+}
+
+// Data is the raw data representing the patch.
+// Note: obj can be either an object to be converted or an object already converted.
+func (p *conversionMergePatch) Data(obj client.Object) ([]byte, error) {
```
sbueringer, line 80: I think we talked about this before, but I don't remember the details.

Could it be that there are cases where we losing data?

1. v1alpha5 object in the apiserver with all data
2. v1alpha5=>vhub object has a subset of data
3. patch from "vhub original" => "vhub modified" generates a patch that drops some data

Especially wondering about cases where we have arrays and our hub type is missing some fields in structs

pkg/conversion/client/patches.go
sbueringer: Discussed, I misunderstood how Data is used within Patch

pkg/conversion/client/client_test.go
sbueringer: Follow-up misunderstanding of how Data is used in Patch

pkg/conversion/test/roundtrip.go
fabriziopandini: Aligned to what we have in CAPI

pkg/conversion/client/client.go
fabriziopandini: - dropped src / dst except for Convert
- target is now used only for converter.TargetVersion
- convertible is used only when checking the corresponding interface

Everything else should now use hub / spoke consistently

pkg/conversion/converter.go
fabriziopandini: see https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3767#discussion_r2698980552

pkg/conversion/api/meta/types.go
fabriziopandini: reviewed, PTAL

pkg/conversion/api/vmoperator/hub/doc.go
fabriziopandini: dropped

pkg/conversion/client/client.go
fabriziopandini: Moved to conversion/api

pkg/conversion/client/client.go
fabriziopandini: We are surfacing the action when we are performing conversions.
Everything else is type checking (action is not relevant) or calls to controller runtime client methods.

pkg/conversion/client/client_test.go
fabriziopandini: We have to keep supporting it, user can call patch.Data with an hub type, client internally pass a spoke type to avoid one conversion.

pkg/conversion/client/patches.go
fabriziopandini: All the sub structs for the list fields now are included in the hub types.
This, combined with the fact that you can write only on the same version you read from, should protect from this issue, or at least reach the same level of protection you patch lists fields when using native types.

pkg/conversion/client/client.go
sbueringer: Thx!

pkg/conversion/client/client.go
sbueringer: Thank you very much. Really appreciate it!

sbueringer:
Changes look good. Lgtm pending linter & CI fixes

pkg/conversion/api/vmoperator/hub/virtualmachine_hardware_types.go
```diff
@@ -0,0 +1,29 @@
...
+*/
+
+// nolint: revive
+package hub
+
+// +kubebuilder:validation:Enum=IDE;NVME;SCSI;SATA
+
+type VirtualControllerType string
+
+const (
```
sbueringer, line 23: It looks like these constants are unused? Should we just delete this file? (IIRC it was just added in the last commit)

sbueringer:
/test pull-cluster-api-provider-vsphere-e2e-supervisor-blocking-main
(network peering flake)

sbueringer:
e2e-vcsim-supervisor-main has Machines stuck in "poweringOn"

pkg/conversion/api/vmoperator/hub/virtualmachine_hardware_types.go
fabriziopandini: As a rule of thumb, when I imported an enum type I also imported corresponding enum values even if not used in CAPV, because I think this helps in understanding what the field actually is.

But no strong opinions.

sbueringer: Makes sense, that's fine

PR #3772, opened 2026-01-16
✨ Promote v1beta2 conditions on VSphereVM, VSphereMachine, VSphereCluster

sbueringer:

Part of #3452

### DIFF
--- a/config/supervisor/crd/patches/cainjection_in_providerserviceaccounts.yaml
+++ b/config/supervisor/crd/patches/cainjection_in_providerserviceaccounts.yaml
@@ -1,0 +1,8 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
+# CRD conversion requires k8s 1.13 or later.
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
+  name: providerserviceaccounts.vmware.infrastructure.cluster.x-k8s.io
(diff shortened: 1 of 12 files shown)


apis/v1beta2/vspherevm_types.go
```diff
@@ -335,22 +339,28 @@ type VSphereVMStatus struct {
...
-	// +listMapKey=type
-	// +kubebuilder:validation:MaxItems=32
-	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	V1Beta1 *VSphereVMV1Beta1DeprecatedStatus `json:"v1beta1,omitempty"`
+}
+
+// VSphereVMV1Beta1DeprecatedStatus groups all the status fields that are deprecated and will be removed when support for v1beta1 will be dropped.
+// See https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20240916-improve-status-in-CAPI-resources.md for more context.
+type VSphereVMV1Beta1DeprecatedStatus struct {
+	// conditions defines current service state of the VSphereCluster.
```
fabriziopandini:
```suggestion
	// conditions defines current service state of the VSphereVM.
```

pkg/context/vm_context.go
```diff
@@ -45,14 +45,14 @@ func (c *VMContext) String() string {
 
 // Patch updates the object and its status on the API server.
 func (c *VMContext) Patch(ctx context.Context) error {
-	return c.PatchHelper.Patch(ctx, c.VSphereVM, patch.WithOwnedV1Beta2Conditions{Conditions: []string{
+	return c.PatchHelper.Patch(ctx, c.VSphereVM, patch.WithOwnedConditions{Conditions: []string{
```
fabriziopandini, line 48: I'm wondering if we need options both for new and old conditions...
(if yes, probably worth to audit the code base/other patch  under context/previous PR where we touched conditions)

sbueringer: Not sure as we never had the option for v1beta1 and we never had an issue these last few years. But I can add a follow-up to the umbrella issue if you want

sbueringer: I'm not sure if we even always set both options in core CAPI to be honest

fabriziopandini: Let's add a follow up task, thanks

sbueringer: Done

PR #3778, opened 2026-01-19
✨ Implement v1beta2 contract (supervisor)

sbueringer:

Part of #3452

### DIFF
--- a/apis/vmware/v1beta2/vspheremachine_types.go
+++ b/apis/vmware/v1beta2/vspheremachine_types.go
@@ -39,7 +39,7 @@ type VSphereMachineVolume struct {
 
 // VSphereMachineSpec defines the desired state of VSphereMachine.
 type VSphereMachineSpec struct {
-	// providerID is the virtual machine's BIOS UUID formated as
+	// providerID is the virtual machine's BIOS UUID formatted as
 	// vsphere://12345678-1234-1234-1234-123456789abc
 	// +optional
 	// +kubebuilder:validation:MinLength=1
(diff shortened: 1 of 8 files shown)


controllers/vmware/vspherecluster_reconciler_test.go
```diff
@@ -210,13 +208,15 @@ func TestClusterReconciler_getFailureDomains(t *testing.T) {
 				zone(namespace.Name, "ns-three", false),
 				zone(namespace.Name, "ns-four", true),
 			},
-			want:        failureDomains("ns-one", "ns-two", "ns-three"),
+			want:        failureDomains("ns-one", "ns-three", "ns-two"),
```
fabriziopandini: May be let's sort in the failureDomains func instead of relying on the input being sorted

apis/vmware/v1beta2/vspheremachine_types.go
```diff
@@ -39,11 +39,12 @@ type VSphereMachineVolume struct {
 
 // VSphereMachineSpec defines the desired state of VSphereMachine.
 type VSphereMachineSpec struct {
-	// providerID is the virtual machine's BIOS UUID formatted as
-	// vsphere://12345678-1234-1234-1234-123456789abc.
-	// This is required at runtime by CAPI. Do not remove this field.
+	// providerID is the virtual machine's BIOS UUID formated as
```
fabriziopandini:
```suggestion
	// providerID is the virtual machine's BIOS UUID formatted as
```

controllers/vmware/vspherecluster_reconciler_test.go
sbueringer: This is test code. I thought it's better if the table expresses exactly what we want

I'll add a comment highlighting that we expect a sorted list

sbueringer: I'll take that slightly back. Originally the func tested here did not have the sort. But I now moved it inside so the test is covering the sort code as well

fabriziopandini (commented):
Great work, especially in figuring out how to get test working!

/lgtm
/approve

Issue #3780, opened 2026-01-20
Implement OS and Architecture Awareness for autoscaling from zero


zhouke1991:
/kind feature

### Problem Statement
Currently, when a user creates a non-Linux or non-amd64 node pool with the Cluster Autoscaler enabled, the autoscaler is unaware of the underlying infrastructure.

- The Issue: 
In a "scale-from-zero" scenario (where the node pool has 0 current nodes), the autoscaler attempts to build a template node to see if pending pods can fit. Without explicit configuration, it defaults to Linux assumptions. Consequently, Pods requiring Windows (via nodeSelector or tolerations) fail to trigger a scale-up because the autoscaler believes they cannot fit the node pool.

- Current Workaround: 
Users must manually add specific annotations `capacity.cluster-autoscaler.kubernetes.io/labels: "kubernetes.io/os=windows` to the MachineDeployment to inform the autoscaler that the node pool is Windows. This is manual and error-prone, resulting in a poor User Experience.

### Describe the solution you'd like**
Add OS and architecture metadata to VsphereMachineTemplate so the autoscaler can correctly scale worker nodes from zero by selecting compatible OS and CPU architecture.

The parent issue: https://github.com/kubernetes-sigs/cluster-api/issues/11961
Proposal: https://github.com/kubernetes-sigs/cluster-api/pull/11962

PR #3788, opened 2026-01-22
✨ Add operating system and architecture information to VSphereMachineTemplate status

zhouke1991:
Retrieve the OS and Arch information from the ClusterVirtualMachineImage resource and expose them in the VsphereMachineTemplate.Statues. Then the cluster-autoscaler can automatically consume it in the scaling from zero scenario.

1. ClusterVirtualMachineImage (VM Operator Resource), referenced by the imageName in VsphereMachineTemplate
```
apiVersion: vmoperator.vmware.com/v1alpha2
kind: ClusterVirtualMachineImage
metadata:
  name: ubuntu-2204-kube-v1.28.0
status:
  name: ubuntu-2204-kube-v1.28.0
  vmwareSystemProperties:
    - key: vmware-system.tkr.os-type
      value: linux
    - key: vmware-system.tkr.os-arch
      value: amd64
  productInfo:
    fullVersion: "22.04"
  osInfo:
    type: linux
```
2. VSphereMachineTemplate (After - With Status Populated by Controller)
```
apiVersion: vmware.infrastructure.cluster.x-k8s.io/v1beta2
kind: VSphereMachineTemplate
metadata:
  name: worker-template
  namespace: default
spec:
  template:
    spec:
      className: best-effort-small
      imageName: ubuntu-2204-kube-v1.28.0
      storageClass: wcpglobal-storage-profile
status:
  capacity:
    cpu: "2"
    memory: "4Gi"
  nodeInfo:
    architecture: amd64
    operatingSystem: linux
```
### Operating Systems:
- linux 
- windows 

###  Architectures:
- amd64 
- arm64 
- s390x 
- ppc64le

Fixes  https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/3780

### DIFF
--- a/controllers/vmware/vspheremachinetemplate_controller.go
+++ b/controllers/vmware/vspheremachinetemplate_controller.go
@@ -156,13 +156,13 @@ func normalizeArchitecture(arch string) vmwarev1.Architecture {
 
 func getOSAndArchFromClusterVirtualMachineImage(ctx context.Context, c client.Client, imageName string) (string, string, error) {
 	if imageName == "" {
-		return "", "", errors.New("imageName is empty")
+		return "", "", nil
 	}
 	// Try to fetch the ClusterVirtualMachineImage with the given name
 	cvmi := &vmoprvhub.ClusterVirtualMachineImage{}
 	if err := c.Get(ctx, client.ObjectKey{Name: imageName}, cvmi); err != nil {
 		if apierrors.IsNotFound(err) {
-			return "", "", errors.Wrapf(err, "ClusterVirtualMachineImage %q not found", imageName)
+			return "", "", nil
 		}
 		return "", "", errors.Wrapf(err, "failed to get ClusterVirtualMachineImage %q", imageName)
 	}
--- a/controllers/vmware/vspheremachinetemplate_controller_test.go
+++ b/controllers/vmware/vspheremachinetemplate_controller_test.go
@@ -223,8 +223,14 @@ func Test_vSphereMachineTemplateReconciler_Reconcile(t *testing.T) {
 			objects: []client.Object{
 				virtualMachineClass(namespace.Name, "vm-class", &vmoprv1alpha5.VirtualMachineClassHardware{Cpus: 1, Memory: quantity(1024)}),
 			},
-			wantErr:    "ClusterVirtualMachineImage \"missing-image\" not found",
-			wantStatus: nil,
+			wantErr: "",
+			wantStatus: &vmwarev1.VSphereMachineTemplateStatus{
+				Capacity: corev1.ResourceList{
+					corev1.ResourceCPU:    quantity(1),
+					corev1.ResourceMemory: quantity(1024),
+				},
+				// NodeInfo should remain empty as the image doesn't exist
+			},
 		},
 		{
 			name:                   "ClusterVirtualMachineImage without vmwareSystemProperties",
@@ -240,6 +246,21 @@ func Test_vSphereMachineTemplateReconciler_Reconcile(t *testing.T) {
 					corev1.ResourceMemory: quantity(1024),
 				},
 				// NodeInfo should remain empty as properties don't exist
+			},
+		},
+		{
+			name:                   "Empty imageName",
+			vSphereMachineTemplate: vSphereMachineTemplate(namespace.Name, "with-empty-image", "vm-class", nil),
+			objects: []client.Object{
+				virtualMachineClass(namespace.Name, "vm-class", &vmoprv1alpha5.VirtualMachineClassHardware{Cpus: 1, Memory: quantity(1024)}),
+			},
+			wantErr: "",
+			wantStatus: &vmwarev1.VSphereMachineTemplateStatus{
+				Capacity: corev1.ResourceList{
+					corev1.ResourceCPU:    quantity(1),
+					corev1.ResourceMemory: quantity(1024),
+				},
+				// NodeInfo should remain empty as imageName is empty
 			},
 		},
 	}


zhouke1991:
Code to consume the nodeInfo in cluster-autoscaler: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/clusterapi/clusterapi_nodegroup.go#L395-L396

api/supervisor/v1beta2/vspheremachinetemplate_types.go
```diff
@@ -72,6 +116,19 @@ type VSphereMachineTemplate struct {
 	Status VSphereMachineTemplateStatus `json:"status,omitempty,omitzero"`
 }
 
+// NodeInfo contains information about the node's architecture and operating system.
```
sbueringer: nit: let's move this to directly below the VSphereMachineTemplateStatus struct please

api/supervisor/v1beta1/conversion_test.go
```diff
@@ -231,6 +232,13 @@ func hubVSphereMachineTemplateResource(in *vmwarev1.VSphereMachineTemplateResour
 	in.ObjectMeta = clusterv1.ObjectMeta{} // Field does not exist in v1beta1.
 }
 
+func hubVSphereMachineTemplateStatus(in *vmwarev1.VSphereMachineTemplate, c randfill.Continue) {
+	c.FillNoCustom(in)
+	// NodeInfo doesn't exist in v1beta1, so it will be lost during hub-spoke-hub conversion.
```
sbueringer: It should not be lost during hub-spoke-hub conversion. Let's revert the changes to this file and adjust conversion.go like this:

```go
func (src *VSphereMachineTemplate) ConvertTo(dstRaw conversion.Hub) error {
	dst := dstRaw.(*vmwarev1.VSphereMachineTemplate)
	if err := Convert_v1beta1_VSphereMachineTemplate_To_v1beta2_VSphereMachineTemplate(src, dst, nil); err != nil {
		return err
	}

	restored := &vmwarev1.VSphereMachineTemplate{}
	ok, err := utilconversion.UnmarshalData(src, restored)
	if err != nil {
		return err
	}

	if ok {
		dst.Status.NodeInfo = restored.Status.NodeInfo
	}

	return nil
}

func (dst *VSphereMachineTemplate) ConvertFrom(srcRaw conversion.Hub) error {
	src := srcRaw.(*vmwarev1.VSphereMachineTemplate)
	if err := Convert_v1beta2_VSphereMachineTemplate_To_v1beta1_VSphereMachineTemplate(src, dst, nil); err != nil {
		return err
	}

	if dst.Spec.Template.Spec.ProviderID != nil && *dst.Spec.Template.Spec.ProviderID == "" {
		dst.Spec.Template.Spec.ProviderID = nil
	}

	if dst.Spec.Template.Spec.FailureDomain != nil && *dst.Spec.Template.Spec.FailureDomain == "" {
		dst.Spec.Template.Spec.FailureDomain = nil
	}

	return utilconversion.MarshalData(src, dst)
}
```

pkg/conversion/api/vmoperator/v1alpha2/virtualmachineimage_conversion.go
```diff
@@ -53,6 +53,17 @@ func convert_v1alpha2_VirtualMachineImage_To_hub_VirtualMachineImage(_ context.C
 	}
 	dst.Status.ProviderItemID = src.Status.ProviderItemID
```
sbueringer, line 54: Let's add a test for this in conversion_test.go (~ duplicate what we have for VirtualMachineImage)

Same for v1alpha5

controllers/vmware/vspheremachinetemplate_controller.go
```diff
@@ -107,9 +108,73 @@ func (r *vSphereMachineTemplateReconciler) Reconcile(ctx context.Context, req ct
...
+	}
+}
+
+func getOSAndArchFromClusterVirtualMachineImage(ctx context.Context, c client.Client, imageName string) (string, string) {
+	if imageName == "" {
+		return "", ""
+	}
+	// Try to fetch the ClusterVirtualMachineImage with the given name
+	cvmi := &vmoprvhub.ClusterVirtualMachineImage{}
+	if err := c.Get(ctx, client.ObjectKey{Name: imageName}, cvmi); err != nil {
```
sbueringer, line 163: I think we should only ignore NotFound errors not all errors. So I would like to do something like this
```go
	if err := c.Get(ctx, client.ObjectKey{Name: imageName}, cvmi); err != nil {
		if apierrors.IsNotFound(err){
			return "", "", nil
		}
		return "", "", err
	}
```

controllers/vmware/vspheremachinetemplate_controller.go
```diff
@@ -107,9 +108,73 @@ func (r *vSphereMachineTemplateReconciler) Reconcile(ctx context.Context, req ct
...
+	default:
+		return ""
+	}
+}
+
+func getOSAndArchFromClusterVirtualMachineImage(ctx context.Context, c client.Client, imageName string) (string, string) {
+	if imageName == "" {
+		return "", ""
+	}
+	// Try to fetch the ClusterVirtualMachineImage with the given name
```
sbueringer, line 161: Q: Is it enough for our use case to only handle ClusterVirtualMachineImage? vm-operator also supports VirtualMachineImage and display name

controllers/vmware/vspheremachinetemplate_controller.go
```diff
@@ -107,9 +108,73 @@ func (r *vSphereMachineTemplateReconciler) Reconcile(ctx context.Context, req ct
...
+	cvmi := &vmoprvhub.ClusterVirtualMachineImage{}
+	if err := c.Get(ctx, client.ObjectKey{Name: imageName}, cvmi); err != nil {
+		return "", ""
+	}
+
+	// Extract OS type and architecture from vmwareSystemProperties
+	var osType, osArch string
```
sbueringer: Should this be `osType, archType` instead?

controllers/vmware/vspheremachinetemplate_controller_test.go
```diff
@@ -124,6 +240,14 @@ func Test_vSphereMachineTemplateReconciler_Reconcile(t *testing.T) {
 				WithScheme(scheme).
 				WithObjects(append([]client.Object{namespace}, tt.objects...)...)
 
+			// Register ClusterVirtualMachineImage objects with status subresource
+			// so the fake client properly handles Status field reads
+			for _, obj := range tt.objects {
+				if _, ok := obj.(*vmoprv1alpha5.ClusterVirtualMachineImage); ok {
+					fakeClientBuilder = fakeClientBuilder.WithStatusSubresource(obj)
```
sbueringer: I think you can simply always call WithStatusSubresource with an empty ClusterVirtualMachineImage

controllers/vmware/vspheremachinetemplate_controller.go
sbueringer: Discussed, we're good

controllers/vmware/vspheremachinetemplate_controller.go
```diff
@@ -107,9 +108,79 @@ func (r *vSphereMachineTemplateReconciler) Reconcile(ctx context.Context, req ct
...
+	case "ppc64le":
+		return vmwarev1.ArchitecturePpc64le
+	default:
+		return ""
+	}
+}
+
+func getOSAndArchFromClusterVirtualMachineImage(ctx context.Context, c client.Client, imageName string) (string, string, error) {
+	if imageName == "" {
+		return "", "", errors.New("imageName is empty")
```
sbueringer:
```suggestion
		return "", "", nil
```
Let's do nothing in this case. I'm not sure if we might have cases where imageName is empty and then we shouldn't produce a lot of errors (VSphereMachineTemplates referenced in ClusterClasses, even if not now, maybe in the future)

controllers/vmware/vspheremachinetemplate_controller.go
```diff
@@ -107,9 +108,79 @@ func (r *vSphereMachineTemplateReconciler) Reconcile(ctx context.Context, req ct
...
+	}
+	// Try to fetch the ClusterVirtualMachineImage with the given name
+	cvmi := &vmoprvhub.ClusterVirtualMachineImage{}
+	if err := c.Get(ctx, client.ObjectKey{Name: imageName}, cvmi); err != nil {
+		if apierrors.IsNotFound(err) {
+			return "", "", errors.Wrapf(err, "ClusterVirtualMachineImage %q not found", imageName)
```
sbueringer: Also wondering if we should return `"", "", nil` here. I think there might be cases where imageName is just set to a dummy value in templates referenced in ClusterClasses

sbueringer:
Thx!

/lgtm

/assign @fabriziopandini

PR #3792, opened 2026-01-22
⚠️ Adjust v1beta2 APIs to use correct ref types (govmomi)

sbueringer:

Part of #3452


internal/webhooks/vspherevm_test.go
```diff
@@ -129,74 +128,74 @@ func TestVSphereVM_ValidateUpdate(t *testing.T) {
...
+			vSphereVM:    createVSphereVM("vsphere-vm-1", "foo.com", biosUUID, "", []string{"192.168.0.1/32", "192.168.0.10/32"}, infrav1.VSphereVMBootstrapReference{}, infrav1.Linux, infrav1.VirtualMachinePowerOpModeTrySoft, 0),
 			wantErr:      false,
 		},
 		{
 			name:         "updating bootstrapRef can be done",
-			oldVSphereVM: createVSphereVM("vsphere-vm-1", "foo.com", "", "", []string{"192.168.0.1/32"}, nil, infrav1.Linux, infrav1.VirtualMachinePowerOpModeTrySoft, 0),
-			vSphereVM:    createVSphereVM("vsphere-vm-1", "foo.com", biosUUID, "", []string{"192.168.0.1/32", "192.168.0.10/32"}, &corev1.ObjectReference{}, infrav1.Linux, infrav1.VirtualMachinePowerOpModeTrySoft, 0),
+			oldVSphereVM: createVSphereVM("vsphere-vm-1", "foo.com", "", "", []string{"192.168.0.1/32"}, infrav1.VSphereVMBootstrapReference{}, infrav1.Linux, infrav1.VirtualMachinePowerOpModeTrySoft, 0),
+			vSphereVM:    createVSphereVM("vsphere-vm-1", "foo.com", biosUUID, "", []string{"192.168.0.1/32", "192.168.0.10/32"}, infrav1.VSphereVMBootstrapReference{Name: "bootstrap-secret-1"}, infrav1.Linux, infrav1.VirtualMachinePowerOpModeTrySoft, 0),
```
fabriziopandini, line 144: Out of curiosity, why this one is passing from &corev1.ObjectReference{} > infrav1.VSphereVMBootstrapReference{Name: "bootstrap-secret-1"}

sbueringer: The test case here is "updating bootstrapRef can be done"

Previously the test verified that bootstrapRef can be changed from nil to &corev1.ObjectReference{}. 
In general that didn't make a lot of sense, but additionally as the field is not a pointer anymore I can't use a transition like this to verify that the field can be changed anymore.

So now I'm changing it from unset to something with Name set

PR #3801, opened 2026-01-23
🌱 Implement type check for hub types

fabriziopandini:

Part of https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/3758

### DIFF
--- a/pkg/conversion/test/roundtrip.go
+++ b/pkg/conversion/test/roundtrip.go
@@ -23,7 +23,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/onsi/gomega"
+	. "github.com/onsi/gomega" //nolint:revive,staticcheck
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/api/apitesting/fuzzer"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
@@ -65,9 +65,12 @@ type RoundTripCheckTypesInput struct {
 	Skip bool
 
 	// Instruct the type checked about a field rename, e.g. "VirtualMachine.Status.NodeName": "Host",
+	// Note: currently this applies only to fields of type scalar (not to pointers, array, map, struct).
 	FieldNameMap map[string]string
+}
 
-	// Internal settings
+type roundTripCheckTypesOptions struct {
+	RoundTripCheckTypesInput
 
 	// fullMatchRequired is set to true to ensure that hub types nested under fields of type slice/array
 	// fully match with the corresponding spoke type, thus preventing data loss.
@@ -85,22 +88,22 @@ func RoundTripTest(input RoundTripTestInput) func(*testing.T) {
 			if input.CheckTypes.Skip {
 				t.Skip("skipping type check. Please check hub and spoke types manually to ensure no data loss happens when creating patches for spoke types")
 			}
-			g := gomega.NewWithT(t)
+			g := NewWithT(t)
 
 			hubT, err := objType(input.Hub)
-			g.Expect(err).NotTo(gomega.HaveOccurred(), "failed to get type of the hub object")
+			g.Expect(err).NotTo(HaveOccurred(), "failed to get type of the hub object")
 
 			spokeT, err := objType(input.Spoke)
-			g.Expect(err).NotTo(gomega.HaveOccurred(), "failed to get type of the spoke object")
+			g.Expect(err).NotTo(HaveOccurred(), "failed to get type of the spoke object")
 
-			inspectTypes(t, hubT, spokeT, field.NewPath(hubT.Name()), input.CheckTypes)
+			inspectTypes(t, hubT, spokeT, field.NewPath(hubT.Name()), roundTripCheckTypesOptions{RoundTripCheckTypesInput: input.CheckTypes})
 		})
 
 		// Perform round trip between types ensuring conversion are properly implemented.
 		// Note: The test is checking only the hub-spoke-hub round trip because hub types have only a subset of spoke types,
-		// and this would require specific fuzzer config for each type.
+		// and spoke-hub-spoke would require specific fuzzer config for each type.
 		t.Run("hub-spoke-hub", func(t *testing.T) {
-			g := gomega.NewWithT(t)
+			g := NewWithT(t)
 
 			if _, isConvertible := input.Hub.(conversionmeta.Convertible); !isConvertible {
 				g.Fail("Hub type must implement sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api/meta/Convertible")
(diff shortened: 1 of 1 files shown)


pkg/conversion/api/vmoperator/v1alpha2/conversion_test.go
```diff
@@ -42,6 +42,11 @@ func TestFuzzyConversion(t *testing.T) {
 		FuzzerFuncs: []fuzzer.FuzzerFuncs{
 			virtualMachineFuncs,
 		},
+		CheckTypes: test.RoundTripCheckTypesInput{
```
sbueringer, line 45: Q: (not a blocker) Could it be that we are missing issues when an array field is renamed?

pkg/conversion/test/roundtrip.go
```diff
@@ -40,22 +47,63 @@ type RoundTripTestInput struct {
...
+	Skip bool
+
+	// Instruct the type checked about a field rename, e.g. "VirtualMachine.Status.NodeName": "Host",
+	FieldNameMap map[string]string
+
+	// Internal settings
+
+	// fullMatchRequired is set to true to ensure that hub types nested under fields of type slice/array
+	// fully match with the corresponding spoke type, thus preventing data loss.
+	fullMatchRequired bool
```
sbueringer: It's a bit awkward to have an unexported field in an exported struct. Can we avoid this? (e.g. by having another internal struct that embeds this struct?

pkg/conversion/test/roundtrip.go
```diff
@@ -40,22 +47,63 @@ type RoundTripTestInput struct {
...
+
+		// Check types for differences that might lead to data loss when creating patches for spoke types,
+		t.Run("Check types", func(t *testing.T) {
+			if input.CheckTypes.Skip {
+				t.Skip("skipping type check. Please check hub and spoke types manually to ensure no data loss happens when creating patches for spoke types")
+			}
+			g := gomega.NewWithT(t)
+
+			hubT, err := objType(input.Hub)
+			g.Expect(err).NotTo(gomega.HaveOccurred(), "failed to get type of the hub object")
```
sbueringer: nit. Maybe use a `.` import for gomega like we do almost everywhere

pkg/conversion/test/roundtrip.go
```diff
@@ -40,22 +47,63 @@ type RoundTripTestInput struct {
...
+
+			spokeT, err := objType(input.Spoke)
+			g.Expect(err).NotTo(gomega.HaveOccurred(), "failed to get type of the spoke object")
+
+			inspectTypes(t, hubT, spokeT, field.NewPath(hubT.Name()), input.CheckTypes)
+		})
+
+		// Perform round trip between types ensuring conversion are properly implemented.
+		// Note: The test is checking only the hub-spoke-hub round trip because hub types have only a subset of spoke types,
+		// and this would require specific fuzzer config for each type.
```
sbueringer:
```suggestion
		// and spoke-hub-spoke would require specific fuzzer config for each type.
```
I think

pkg/conversion/test/roundtrip.go
```diff
@@ -100,3 +148,126 @@ func RoundTripTest(input RoundTripTestInput) func(*testing.T) {
...
+
+	switch hubKind := hubT.Kind(); hubKind {
+	case reflect.Map:
+		// If hub type is a map, check both key and value types.
+		// Note: Current logic assumes that when a field in hub is a map, also the corresponding spoke field is. This can be improved in the future.
+		g.Expect(spokeT.Kind()).To(gomega.Equal(reflect.Map), fmt.Sprintf("field %s is a map in hub, not in spoke", path.String()))
+		inspectTypes(t, hubT.Key(), spokeT.Key(), path.Key("$key"), input)
+		inspectTypes(t, hubT.Elem(), spokeT.Elem(), path.Key("$elem"), input)
+	case reflect.Array, reflect.Slice:
+		// If hub type is a map, check items type.
```
sbueringer: not a map

pkg/conversion/test/roundtrip.go
```diff
@@ -100,3 +148,126 @@ func RoundTripTest(input RoundTripTestInput) func(*testing.T) {
...
+			}
+			inspectTypes(t, hubField.Type, spokeField.Type, fieldPath, input)
+		}
+
+		// If a full match is required, also check the other way around: all the spoke fields have a corresponding field in hub.
+		if input.fullMatchRequired {
+			for i := 0; i < spokeT.NumField(); i++ {
+				spokeField := spokeT.Field(i)
+				spokeFieldName := spokeField.Name
+				if spokeFieldNames.Has(spokeFieldName) {
```
sbueringer: `spokeFieldNames` is more like `hubFieldNames` (+ mapping if necessary), right?

sbueringer:
Ver nice, just a few minor findings

pkg/conversion/api/vmoperator/v1alpha2/conversion_test.go
fabriziopandini: That's partially correct.

Skip applies to all fields.
FieldNameMap applies only to fields of type scalar (not to pointers, array, map, struct). Added a note to clarify

pkg/conversion/test/roundtrip.go
fabriziopandini: Even if spoke field name by default is assumed to be equal to the hub field name, I don't think that we should call this hubFieldNames (when there is a mapping is spokeFieldName is different from hubFieldName).

Also, I explicitly want to track the list spokeFieldNames encountered in this loop, because this is what i need to validate full match in the following loop

pkg/conversion/api/vmoperator/v1alpha2/conversion_test.go
sbueringer: Not sure I follow. I didn't mean that we should list array fields here. I meant would we detect issues if the name of the array field changes in an apiVersion.

I think the current code would not be able to figure out that this is still the "same" array

(as mentioned not a blocker, I was just thinking about that case)

pkg/conversion/test/roundtrip.go
sbueringer: Maybe allowedSpokeFieldNames would be a good name?

(I would prefer not using `spokeFieldNames` because it is definitely not the list of fields in spoke)

pkg/conversion/test/roundtrip.go
```diff
@@ -81,22 +132,145 @@ func RoundTripTest(input RoundTripTestInput) func(*testing.T) {
...
+		for i := 0; i < hubT.NumField(); i++ {
+			hubField := hubT.Field(i)
+			hubFieldName := hubField.Name
+			fieldPath := path.Child(hubFieldName)
+
+			spokeFieldName := hubField.Name
+			if mappedSpokeFieldName, ok := options.FieldNameMap[fieldPath.String()]; ok {
+				spokeFieldName = mappedSpokeFieldName
+			}
+			spokeFieldNames.Insert(hubFieldName)
```
sbueringer:
```suggestion
			spokeFieldNames.Insert(spokeFieldName)
```

PR #3803, opened 2026-01-26
⚠️ Align NumCPUs & NumCoresPerSocket to API conventions

### LINKED ISSUE #3664: Unable to set CPU Topology to "Assigned at power on"

sbueringer:

Part of #3452
Fixes #3664

### DIFF
--- a/main.go
+++ b/main.go
@@ -323,6 +323,11 @@ func main() {
 		setupLog.Error(err, "Unable to start manager")
 		os.Exit(1)
 	}
+	if isGovmomiCRDLoaded && isSupervisorCRDLoaded {
+		err := fmt.Errorf("both supervisor and govmomi CRDs detected (unsupported configuration)")
+		setupLog.Error(err, "Unable to start manager")
+		os.Exit(1)
+	}
 
 	var watchNamespaces map[string]cache.Config
 	if watchNamespace != "" {
--- a/.golangci-kal.yml
+++ b/.golangci-kal.yml
@@ -111,12 +111,6 @@ linters:
       linters:
         - kubeapilinter
 
-    ## TODO: Think about CPU configuration, context: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/3664
-    - path: "api/govmomi/v1beta2/types.go"
-      text: "optionalfields: field (NumCPUs|NumCoresPerSocket) has a valid zero value"
-      linters:
-        - kubeapilinter
-
 issues:
   max-same-issues: 0
   max-issues-per-linter: 0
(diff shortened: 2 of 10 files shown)


main.go
```diff
@@ -323,6 +323,11 @@ func main() {
 		setupLog.Error(err, "Unable to start manager")
 		os.Exit(1)
 	}
+	if isGovmomiCRDLoaded && isSupervisorCRDLoaded {
```
sbueringer, line 326: This configuration was already failing with a controller-runtime error message before. I just added this error to have a better error message

sbueringer:
/hold

for feedback

sbueringer:
/hold cancel

@L30Bola If there's any further feedback, just let us know. We still have time until the release

PR #3805, opened 2026-01-26
⚠️ Align failureDomain fields to CAPI contract

fabriziopandini:
This PR aligns CAPV failureDomain fields to CAPI contract:
- spec.failureDomain has been dropped from VSphereMachine both govimomi and supervisor mode (*)
- status.failureDomain has been added from VSphereMachine in supervisor mode only (**)

(*) Field was used by CAPI before v1alpha3, after that it was preserved only for the transition; with CAPI v1beta2 contract this is going to be finally dropped.
(**) Info is already available in VirtualMachine CR, so this was a quick win. Instead, doing something similar for govmomi require additional research/work

Fixes #

### DIFF
--- a/api/supervisor/v1beta2/vspheremachine_types.go
+++ b/api/supervisor/v1beta2/vspheremachine_types.go
@@ -55,13 +55,6 @@ type VSphereMachineSpec struct {
 	// +kubebuilder:validation:MaxLength=512
 	ProviderID string `json:"providerID,omitempty"`
 
-	// failureDomain is the failure domain the machine will be created in.
-	// Must match a FailureDomain name on the Cluster status.
-	// +optional
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=256
-	FailureDomain string `json:"failureDomain,omitempty"`
-
 	// imageName is the name of the base image used when specifying the
 	// underlying virtual machine
 	// +required
@@ -313,6 +306,12 @@ type VSphereMachineStatus struct {
 	// +optional
 	Addresses []clusterv1.MachineAddress `json:"addresses,omitempty"`
 
+	// failureDomain is the failure domain where the VirtualMachine has been scheduled.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=256
+	FailureDomain string `json:"failureDomain,omitempty"`
+
 	// biosUUID is the biosUUID of the virtual machine.
 	// +optional
 	// +kubebuilder:validation:MinLength=1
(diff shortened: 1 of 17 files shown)


api/supervisor/v1beta2/vspheremachine_types.go
```diff
@@ -418,6 +417,7 @@ type VSphereMachineV1Beta1DeprecatedStatus struct {
 // +kubebuilder:printcolumn:name="Cluster",type="string",JSONPath=".metadata.labels['cluster\\.x-k8s\\.io/cluster-name']",description="Cluster"
 // +kubebuilder:printcolumn:name="Class",type="string",JSONPath=".spec.className",description="VirtualMachineClass name"
 // +kubebuilder:printcolumn:name="Provider ID",type="string",JSONPath=".spec.providerID",description="Provider ID",priority=10
+// +kubebuilder:printcolumn:name="Failure domain",type="string",JSONPath=".status.failureDomain",description="The failure domain where the VSphereMachine has been scheduled"
```
sbueringer, line 420: I think it would be nice to use the same priority here as in core CAPI. So either both or none are showing up without -o wide. Just would look more consistent

Maybe let's show it for both CRs per default (even without -o wide)
Looks like this column doesn't take up a lot of space

controllers/vmware/virtualmachinegroup_reconciler_test.go
```diff
@@ -54,12 +54,12 @@ func Test_shouldCreateVirtualMachineGroup(t *testing.T) {
 				*createMD("md3", "test-cluster", "zone1", 1),
 			},
 			vSphereMachines: []vmwarev1.VSphereMachine{
```
sbueringer, line 56: Do you have an idea why we had all these failureDomain fields in these tests if they are apparently entirely irrelvant for the tested functionality?

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -790,6 +784,8 @@ func (v *VmopMachineService) reconcileProviderID(ctx context.Context, supervisor
 		supervisorMachineCtx.VSphereMachine.Status.BiosUUID = vm.Status.BiosUUID
```
sbueringer: nit: let's inline or rename the func please

I would inline it, setting these 3 fields in this misnamed func and some others one level higher is just misleading

api/supervisor/v1beta2/vspheremachine_types.go
fabriziopandini: Aligned in CAPI

controllers/vmware/virtualmachinegroup_reconciler_test.go
fabriziopandini: I think we simply used an existing createMachine func

PR #3850, opened 2026-02-24
[release-1.13] 🌱 Drop go vuln check

fabriziopandini:
There is a vulnerability in otel can we cannot fix without bumping the min go versions, which is a thing we can't do in a release branch.

The only option is to ignore, but govulncheck do not support ignoring vulnerabilities

So, considering the branch is near EOL, disabling the check

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ SHELL:=/usr/bin/env bash
 #
 # Go.
 #
-GO_VERSION ?= 1.24.12
+GO_VERSION ?= 1.24.13
 GO_DIRECTIVE_VERSION ?= 1.23.0
 GO_CONTAINER_IMAGE ?= docker.io/library/golang:$(GO_VERSION)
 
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/cluster-api-provider-vsphere
 
-go 1.23.0
+go 1.24.0
```
sbueringer: We can't bump the go minor version here

Makefile
```diff
@@ -542,7 +542,6 @@ verify-govulncheck: $(GOVULNCHECK) ## Verify code for vulnerabilities
 .PHONY: verify-security
 verify-security: ## Verify code and images for vulnerabilities
 	$(MAKE) verify-container-images && R1=$$? || R1=$$?; \
-	$(MAKE) verify-govulncheck && R2=$$? || R2=$$?; \
 	if [ "$$R1" -ne "0" ] || [ "$$R2" -ne "0" ]; then \
```
sbueringer: We also have to drop R2 from this if

sbueringer:
/lgtm
/approve
Thx!

Issue #3822, opened 2026-02-05
Scale testing


sbueringer:
We're going to make some improvements to improve performance at scale and make scale testing with CAPV easier

Scale:
* PriorityQueue
  * [x] [Reconcile latency] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3730
* ReconcilerRateLimiting
  * [x] [Memory] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3846
* Optimize cache configuration
  * [x] [Memory] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3861
* Admission webhook improvements:
  * [x] [Memory] CR: https://github.com/kubernetes-sigs/controller-runtime/pull/3463
* Conversion webhook improvements:
  * [x] [Memory] conversion/convert gets called a lot for unserved versions (even though we never use old apiVersions) https://github.com/kubernetes/kubernetes/pull/130704
  * [x] [Memory] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3874
* Optimize cache configuration
  * [x] [Memory] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3752
* Misc:
  * [x] [Reduce apiserver calls] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3857
  * [x] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3893

Scale testing:
* [x] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3820
* [x] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3821
* [x] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3827
* [x] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3842
* [x] https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3878

Scale testing: Potential follow-ups:
* [ ] Make durations in vcsim e.g. via annotations configurable (etcdStartupDuration, apiServerStartupDuration, ..)

PR #3857, opened 2026-03-02
🐛 Stop patching VSphereMachine on every reconcile

sbueringer:

Part of #3822


pkg/services/vmoperator/vmopmachine.go
```diff
@@ -707,7 +707,7 @@ func convertKeyValueSlice(pairs []vmoprvhub.KeyValuePair) []vmwarev1.KeyValuePai
 func (v *VmopMachineService) reconcileNetwork(supervisorMachineCtx *vmware.SupervisorMachineContext, vm *vmoprvhub.VirtualMachine) bool {
 	// Propagate VM status.network.interfaces to VSphereMachine.Status.NetworkInterfaces
 	if vm.Status.Network != nil {
-		interfaces := make([]vmwarev1.VSphereMachineNetworkInterfaceStatus, 0, len(vm.Status.Network.Interfaces))
+		var interfaces []vmwarev1.VSphereMachineNetworkInterfaceStatus
```
sbueringer, line 710: For context: setting this to an empty array lead to a diff in our patchHelper logic and then to a call to the apiserver with an empty patch.

This happened on every single reconcile

I'm going to follow-up with another PR to core CAPI to improve patchHelper as well to avoid this situation

PR #3917, opened 2026-04-02
🌱 Migrate away from custom GitHub action approval workflow

karimzakzouk:
Removes the custom GitHub Actions workflow (pr-gh-workflow-approve.yaml) that
was used to automatically approve workflow runs when /ok-to-test is added to a PR.

This is no longer needed because Prow now natively supports this via
trigger_github_workflows: true, which is already enabled for this repo in
the test-infra plugins.yaml config.

Part of kubernetes-sigs/cluster-api#13532


sbueringer:
/lgtm
/ok-to-test

/assign @fabriziopandini 

Let's wait for confirmation via a core CAPI PR that this still works before merging this here as well

(We can verify by checking that GitHub actions are able to run after adding /ok-to-test)

PR #3918, opened 2026-04-02
🌱 Improve vm-operator versioning semantic

fabriziopandini:

Fixes #


Makefile
```diff
@@ -241,12 +241,12 @@ VCSIM_CONTROLLER_IMG ?= $(REGISTRY)/$(VCSIM_IMAGE_NAME)
 # vmoperator controller
 VM_OPERATOR_DIR := test/infrastructure/vm-operator
 VM_OPERATOR_TMP_DIR ?= $(VM_OPERATOR_DIR)/vm-operator.tmp
```
sbueringer, line 243: I grepped for the versions, we might also have to adjust:
* vsphere.yaml: name: v1.8.6-0-gde75746a + value below
* test/infrastructure/vm-operator/vm-operator-8.0.yaml: image: gcr.io/k8s-staging-capi-vsphere/extra/vm-operator:v1.8.6-0-gde75746a
* hack/e2e.sh: export E2E_VM_OPERATOR_VERSION="${VM_OPERATOR_VERSION:-v1.9.0-567-g93918c59}"
* vsphere.yaml:   - name: gcr.io/k8s-staging-capi-vsphere/extra/vm-operator:v1.9.0-567-g93918c59
* vsphere.yaml: - name: v1.9.0-567-g93918c59 (+value below)
* test/infrastructure/vm-operator/vm-operator-9.1.yaml: image: ...

Please double check remaining occurrences of v1.8.6-0-gde75746a & v1.9.0-567-g93918c59 afterwards

sbueringer:
Apart from https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3918#discussion_r3028220326, all good from my side. Feel free to merge whenever ready

test/infrastructure/vm-operator/vm-operator-9.1.yaml
```diff
@@ -36391,7 +36391,7 @@ spec:
           value: "true"
         - name: FSS_STORAGE_QUOTA_M2
           value: "true"
-        image: gcr.io/k8s-staging-capi-vsphere/extra/vm-operator:v1.9.0-567-g93918c59
+        image: gcr.io/k8s-staging-capi-vsphere/extra/vm-operator:9.1
```
shutingm-kubernetes, line 36394: Quick question regarding future updates: 
If we update vm-operator in go.mod later, we'll need to build and push the new image with same tag. Is there a risk that this hardcoded versioning might break other concurrent CI jobs during the merge process?"

fabriziopandini: I don't think that this is a concern.

First of all, there is a difference from the vm op version in go mod (the one CAPV uses to import API types), and the vm op image version we are using for tests.  

This change is mostly about the latter, and it is mostly about the model we use in our test matrix to identify the vm op test version we care about.

The new model gives us the choice to 
- move the vm-op commit used for a give test version without expanding our test matrix, e.g. if we pick a commit early in the cycle, and then we want to move to a newer commit later in the cycle 
- introduce new test versions and expand our test matrix

Also, image building remains in the Makefile and thus with full history, and we are in control of image publishing. So if something happen we know in advance why or we can easily figure it out.

Note: Even if we try to keep consistency beteweem vmop version we import in a branch and the wmop versions we use for testing, we need to test and support cases when vm op "library" version and the vm op "running" version  are different, and this is because of the async release model.

Note. If a CI job is running, it either pulls the old or the new image with the same tag (it does not matter if the image changes after pull).

shutingm-kubernetes: Thanks for the explanation, Fab! I see the benefits in terms of flexibility and keeping the test matrix clean.

Just to make sure I’m on the same page: because I noticed some recent changes in vm-operator (e.g., remove **UnmanagedVolumeClaim** and add **Removable** in v1alpha5). If we push these to the static 9.1 tag, any other concurrent PRs relying on the older v1alpha5 schema might hit a race condition and fail their CI tests until rebase this commit.

Is my understanding correct that we are moving toward a model where we accept this temporary CI blocking during such transitions?

fabriziopandini: I would not expect vm op to perform breaking changes on older API versions, but if they do it and CAPV is using those fields, we might have some disruption (considering we bump VM op version infrequently and the average number of PR running test concurrently, it is very unlikely to happen)

config/supervisor-tilt/metrics/kustomization.yaml
```diff
@@ -1,4 +1,5 @@
 resources:
+  - ./namespace.yaml
```
fabriziopandini, line 2: without this, tilt up fails for me (probably a racy condition)

test/infrastructure/vm-operator/vm-operator-9.1.yaml
shutingm-kubernetes: Understood. Thanks for weighing this risks.

fabriziopandini:
@sbueringer all supervisor tests are green, this PR is ready for a final pass.

Also tested with tilt with a few permutations 
- 8.0 v1alpha2 / vcsim 
- 9.1 v1alpha2 / vcsim
- 9.1 v1alpha5 / vcsim
- 9.1 v1alpha5 / vcsim + vm op simulation (should

sbueringer:
/lgtm
/approve

Feel free to merge when ready

sbueringer:
Thx!

/lgtm
/approve

PR #3835, opened 2026-02-13
🐛 Set VM.spec.affinity only on create

sbueringer:
Signed-off-by: Stefan Büringer buringerst@vmware.com

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #3987, opened 2026-05-08
✨ Add policies support to supervisor VSphereMachine

zhanggbj:

Description:

Introduce `spec.policies` on supervisor VSphereMachine and VSphereMachineTemplate, enabling declarative attachment of infrastructure policy objects to node VMs.

When the `InfrastructurePolicies` feature gate is enabled, policies specified in `spec.policies` are propagated to the underlying vm-operator `VirtualMachine.spec.policies` during reconciliation. The list is treated as atomic and immutable after creation.

Key changes:

- Add Policies []PolicyRef to VSphereMachineSpec (supervisor v1beta1 & v1beta2).
- Add InfrastructurePolicies alpha feature gate.
- Gate reconcile-time policy propagation.
- Add unit tests covering policy propagation on create, preservation on update reconcile.

Fixes #


api/supervisor/v1beta2/vspheremachine_types.go
```diff
@@ -114,6 +115,39 @@ type VSphereMachineSpec struct {
...
 	Naming VirtualMachineNamingSpec `json:"naming,omitempty,omitzero"`
+
+	// policies specifies a list of optional infrastructure policies to be applied to the virtual machine.
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=1024
+	Policies []PolicyRef `json:"policies,omitempty"`
+}
+
+// PolicyRef identifies an optional infrastructure policy to specify for the virtual machine.
```
JackieLan, line 127: please check we have exactly the same definition like:
https://github.com/vmware-tanzu/vm-operator/blob/main/api/v1alpha6/virtualmachine_policy_types.go#L11

BTW, might be ok for now, but do we have any LocalObjectRef-like struct defined in capv which can be reused?

api/supervisor/v1beta2/vspheremachine_types.go
```diff
@@ -47,6 +47,7 @@ type VSphereMachineVolume struct {
 }
 
 // VSphereMachineSpec defines the desired state of VSphereMachine.
+// +kubebuilder:validation:XValidation:rule="has(self.policies) == has(oldSelf.policies) && (!has(self.policies) || self.policies == oldSelf.policies)",message="policies are immutable after creation"
```
JackieLan, line 50: This rule rejects both directions: unset → set and set → unset. So a tenant cannot attach policies to an existing VSphereMachine, nor remove them later. That's stricter than the typical "set on create only" idiom (which is usually !has(oldSelf.policies) || self.policies == oldSelf.policies).

Is the bidirectional immutability intentional? It matches spec.affinity semantics in vm-operator (CAPV PR #3835 codified the same "set only on create" intent for affinity), so likely yes — but it's worth one sentence in the field doc-comment, because the failure mode for users is surprising (kubectl edit to add policies later returns "policies are immutable").

Does vm-operator's VirtualMachine.spec.policies carry the same immutability? If vm-operator allows mutation of policies post-create but CAPV does not, brownfield upgrade paths get awkward.

pkg/conversion/api/vmoperator/v1alpha2/conversion_test.go
```diff
@@ -115,3 +117,17 @@ func hubVirtualMachineGroupPlacementDatastoreStatus(in *vmoprvhub.VirtualMachine
 	// Fields existing in hub but not in v1alpha2.VirtualMachineGroupPlacementDatastoreStatus
 	in.TopLevelDirectoryCreateSupported = false
 }
+
+func hubVirtualMachineSpec(in *vmoprvhub.VirtualMachineSpec, c randfill.Continue) {
+	c.FillNoCustom(in)
+	// Policies exists in hub and v1alpha5 but not in v1alpha2; zero it so the
+	// hub-spoke-hub round-trip test does not report spurious data loss.
+	in.Policies = nil
```
JackieLan, line 125: will it hide a  semantic gap: hub → v1alpha2 → hub does drop Policies in production

api/supervisor/v1beta2/vspheremachine_types.go
```diff
@@ -114,6 +115,39 @@ type VSphereMachineSpec struct {
...
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	Name string `json:"name,omitempty"`
+
+	// kind of the infrastructure policy object being referenced.
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z]([-a-zA-Z0-9]*[a-zA-Z0-9])?$`
+	Kind string `json:"kind,omitempty"`
```
JackieLan, line 143: Kind regex `^[a-zA-Z]([-a-zA-Z0-9]*[a-zA-Z0-9])?$` allows hyphens, which isn't valid Kubernetes Kind syntax (CamelCase, no separators). Consider `^[A-Z][a-zA-Z0-9]*$` to match upstream conventions.

JackieLan (commented):
Overall, looks good, thanks.

api/supervisor/v1beta2/vspheremachine_types.go
zhanggbj: > please check we have exactly the same definition

Yes, we have same fields definition with extra kubebuilder validation.

> BTW, might be ok for now, but do we have any LocalObjectRef-like struct defined in capv which can be reused?

I didn't see any existed object like this.

api/supervisor/v1beta2/vspheremachine_types.go
zhanggbj: Yes, this is intentional that updating vspheremachine spec.policies is not allowed, since the right way to update policies in day-2 is to update Cluster Variables to the keep consistent, but not updating the vspheremachine directly.

When Cluster variable for policies are updated, since vspheremachinetemplate is updated, this will trigger a rollout.

pkg/conversion/api/vmoperator/v1alpha2/conversion_test.go
zhanggbj: Didn't get it, could you clarify what is the specific scenario for "hub → v1alpha2 → hub"?

api/supervisor/v1beta2/vspheremachine_types.go
zhanggbj: That's more precise, will update it in next patch.

zhanggbj:
/assign @fabriziopandini 
/assign @sbueringer 
PR is ready for review, thanks!

JackieLan (approved):

api/supervisor/v1beta2/vspheremachine_types.go
fabriziopandini, line 50: When we add CEL validation or other complex open API validation rules we usually add a corresponding unit test; those tests must use test env machinery to install CRDs, setup webhooks etc, e.g. internal/webhooks/test in cluster API

@sbueringer do you remember some CAPV/supervisor example we can point to

api/supervisor/v1beta2/vspheremachine_types.go
```diff
@@ -114,6 +115,40 @@ type VSphereMachineSpec struct {
 	// naming allows configuring the naming strategy used when calculating the name of the VirtualMachine.
 	// +optional
 	Naming VirtualMachineNamingSpec `json:"naming,omitempty,omitzero"`
+
+	// policies specifies a list of optional infrastructure policies to be applied to the virtual machine.
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=1024
+	Policies []PolicyRef `json:"policies,omitempty"`
```
fabriziopandini, line 124: New fields must be added to both v1beta1 and v1beta2 versions to avoid edge case errors during conversions

api/supervisor/v1beta2/vspheremachine_types.go
```diff
@@ -114,6 +115,40 @@ type VSphereMachineSpec struct {
 	// naming allows configuring the naming strategy used when calculating the name of the VirtualMachine.
 	// +optional
 	Naming VirtualMachineNamingSpec `json:"naming,omitempty,omitzero"`
+
+	// policies specifies a list of optional infrastructure policies to be applied to the virtual machine.
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=1024
```
fabriziopandini: Do we really need 1024 (it doesn't make much a difference, but If a user has to add more than a few policies, I would consider we have more foundational problems)

api/supervisor/v1beta2/vspheremachine_types.go
fabriziopandini: Since not all the kind are allowed, can't we use an enum?

pkg/conversion/api/vmoperator/v1alpha2/virtualmachine_conversion.go
```diff
@@ -254,6 +256,7 @@ func convert_v1alpha2_VirtualMachine_To_hub_VirtualMachine(_ context.Context, sr
 	dst.Status.NodeName = src.Status.Host // Field renamed in hub
 	dst.Status.PowerState = vmoprvhub.VirtualMachinePowerState(src.Status.PowerState)
 	dst.Status.Zone = src.Status.Zone
+	// v1alpha2 does not have status.Policies; hub field remains nil.
```
fabriziopandini: nit: we try to keep handling or comments for fields in alphabetical order, so this should go before dst.Status.PowerState
```suggestion
	// v1alpha2 does not have status.policies; hub field remains nil.
```

pkg/conversion/api/vmoperator/v1alpha5/virtualmachine_conversion.go
```diff
@@ -258,6 +268,19 @@ func convert_v1alpha5_VirtualMachine_To_hub_VirtualMachine(_ context.Context, sr
 	dst.Status.NodeName = src.Status.NodeName
 	dst.Status.PowerState = vmoprvhub.VirtualMachinePowerState(src.Status.PowerState)
 	dst.Status.Zone = src.Status.Zone
+	if src.Status.Policies != nil {
```
fabriziopandini, line 269: please move it up before dst.Status.PowerState (same below)

pkg/conversion/api/vmoperator/v1alpha2/virtualmachine_conversion.go
```diff
@@ -125,6 +125,8 @@ func convert_v1alpha2_VirtualMachine_To_hub_VirtualMachine(_ context.Context, sr
 		}
 	}
 	dst.Spec.MinHardwareVersion = src.Spec.MinHardwareVersion
+	// v1alpha2 does not have Policies; hub field remains nil.
+
```
fabriziopandini:
```suggestion
```

pkg/conversion/api/vmoperator/v1alpha2/virtualmachine_conversion.go
```diff
@@ -125,6 +125,8 @@ func convert_v1alpha2_VirtualMachine_To_hub_VirtualMachine(_ context.Context, sr
 		}
 	}
 	dst.Spec.MinHardwareVersion = src.Spec.MinHardwareVersion
+	// v1alpha2 does not have Policies; hub field remains nil.
```
fabriziopandini:
```suggestion
	// v1alpha2 does not have spec.policies; hub field remains nil.
```
(we generally use json name when referring to API fields)

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -637,6 +637,15 @@ func (v *VmopMachineService) reconcileVMOperatorVM(ctx context.Context, supervis
...
 
+	// Assign policies when the feature gate is enabled; explicitly clear them otherwise
+	// so that create and update are always consistent regardless of whether the gate
+	// was toggled between reconciles.
+	if feature.Gates.Enabled(feature.InfrastructurePolicies) {
+		vmOperatorVM.Spec.Policies = getPolicies(supervisorMachineCtx)
+	} else {
+		vmOperatorVM.Spec.Policies = nil
+	}
+
```
fabriziopandini: Can't we move handling of the feature gate in getPolicies, so we have a unique place where all the relevant logic for this field exist

Also, I'm not entirely sure the current check covers changes after create, can't we make this more explicit by relying on vmExists or vmOperatorVM.CreationTimestamp.IsZero() like we do in other places.

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -1013,6 +1022,44 @@ func getTopologyLabels(supervisorMachineCtx *vmware.SupervisorMachineContext, fa
...
+		a, b := newRefs[i], newRefs[j]
+		if a.APIVersion != b.APIVersion {
+			return a.APIVersion < b.APIVersion
+		}
+		if a.Kind != b.Kind {
+			return a.Kind < b.Kind
+		}
+		return a.Name < b.Name
+	})
+	return newRefs
```
fabriziopandini: Why are we cloning a sorting references? Slices have a stable ordering we can rely on, and also, we are setting them only on create, so stable ordering should not be a concern

api/supervisor/v1beta2/vspheremachine_types.go
zhanggbj: As arbitrary policy types are in planning and could be introduced in the short future. If we use enum, it will be blocked and requires CAPV update, so I think we should keep it relaxed at least for now.

pkg/services/vmoperator/vmopmachine.go
zhanggbj: Thanks for catching this. Since slices has a stable ordering, removing this.

api/supervisor/v1beta2/vspheremachine_types.go
zhanggbj: Totally make sense. Will update it to a more reasonable value.

api/supervisor/v1beta2/vspheremachine_types.go
zhanggbj: Good to know. I thought we won't backport new fields to v1beta1.
Will add new fields in v1beta1.

pkg/services/vmoperator/vmopmachine.go
zhanggbj: > Can't we move handling of the feature gate in getPolicies, so we have a unique place where all the relevant logic for this field exist

Fixed.

> Also, I'm not entirely sure the current check covers changes after create, can't we make this more explicit by relying on vmExists or vmOperatorVM.CreationTimestamp.IsZero() like we do in other places.

Since for CAPV, spec.policies is immutable after creation, will only set policies during creation.

For E2E perspective, policies update should rely on VM rolling update.

api/supervisor/v1beta2/vspheremachine_types.go
zhanggbj: IIUC, added CEL validation test to controllers/vmware/test/controllers_test.go

zhanggbj:
@fabriziopandini @JackieLan comments are addressed, please help to take another look, thanks!

pkg/conversion/api/vmoperator/v1alpha2/virtualmachine_conversion.go
```diff
@@ -254,6 +255,7 @@ func convert_v1alpha2_VirtualMachine_To_hub_VirtualMachine(_ context.Context, sr
 	dst.Status.NodeName = src.Status.Host // Field renamed in hub
 	dst.Status.PowerState = vmoprvhub.VirtualMachinePowerState(src.Status.PowerState)
 	dst.Status.Zone = src.Status.Zone
+	// v1alpha2 does not have status.policies; hub field remains nil.
```
fabriziopandini: Must be moved before status.PowerState

pkg/conversion/api/vmoperator/v1alpha5/virtualmachine_conversion.go
fabriziopandini: please move one line up

pkg/conversion/api/vmoperator/v1alpha5/virtualmachine_conversion.go
```diff
@@ -490,6 +523,19 @@ func convert_hub_VirtualMachine_To_v1alpha5_VirtualMachine(_ context.Context, sr
 		dst.Status.Network.PrimaryIP6 = src.Status.Network.PrimaryIP6
 	}
 	dst.Status.PowerState = vmoprv1alpha5.VirtualMachinePowerState(src.Status.PowerState)
+	if src.Status.Policies != nil {
```
fabriziopandini, line 525: Please move one line up

pkg/services/vmoperator/vmopmachine.go
```diff
@@ -1013,6 +1018,35 @@ func getTopologyLabels(supervisorMachineCtx *vmware.SupervisorMachineContext, fa
...
+	// Assign policies when the feature gate is enabled.
+	if !feature.Gates.Enabled(feature.InfrastructurePolicies) {
+		return nil
+	}
+
+	refs := supervisorMachineCtx.VSphereMachine.Spec.Policies
+	if len(refs) == 0 {
+		return nil
+	}
+	// Deduplicate by the (APIVersion, Kind, Name) triple so that an authored
```
fabriziopandini: It looks a little bit strange having silent handling for duplicated policies....
Should this be blocked on web hooks / input validation (like other necessary checks of the correctness of ref)

api/supervisor/v1beta2/vspheremachine_types.go
sbueringer: I think that location is okay

pkg/services/vmoperator/vmopmachine.go
zhanggbj: Hmmm, actually this is planned for a backup deduplicate, the front layer like webhooks or input validations should handle it. Let me remove it.

fabriziopandini:
@sbueringer PTAL

api/supervisor/v1beta2/vspheremachine_types.go
sbueringer: We realized recently that if an object is written with SSA we have to backport the fields. Otherwise we run into some issues

api/supervisor/v1beta1/vspheremachine_types.go
```diff
@@ -67,6 +68,12 @@ type VSphereMachineSpec struct {
 	// +optional
 	Volumes []VSphereMachineVolume `json:"volumes,omitempty"`
 
+	// policies specifies a list of optional infrastructure policies to be applied to the virtual machine.
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=64
```
sbueringer, line 75:
```suggestion
	// +kubebuilder:validation:MinItems=1
	// +kubebuilder:validation:MaxItems=64
```

(same for v1beta2)

api/supervisor/v1beta2/vspheremachine_types.go
sbueringer: We usually do not reuse structs like this if the use case is not 100% identical.

The problem is that validation or godoc might drift over time and if you share the struct you can't adjust

pkg/conversion/api/vmoperator/v1alpha2/conversion_test.go
sbueringer: We never do hub => v1alpha2 => hub conversion in prod

We read v1alpha2, mutate hub and write v1alpha2

api/supervisor/v1beta2/vspheremachine_types.go
```diff
@@ -114,6 +115,40 @@ type VSphereMachineSpec struct {
 	// naming allows configuring the naming strategy used when calculating the name of the VirtualMachine.
 	// +optional
 	Naming VirtualMachineNamingSpec `json:"naming,omitempty,omitzero"`
+
+	// policies specifies a list of optional infrastructure policies to be applied to the virtual machine.
```
sbueringer, line 119: Discussed this with Fabrizio, we would like to have validation in the webhook that the field can only be set if the feature gate is on

(for VSphereMachine and VSphereMachineTemplate)

(similar to https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/4003#discussion_r3310246629)

api/supervisor/v1beta2/vspheremachine_types.go
sbueringer: In CAPI we are using the following everywhere: https://github.com/kubernetes-sigs/cluster-api/blob/f068d1ccb8bcb5f868a8e9d24c03fedde3152fa3/api/core/v1beta2/machinehealthcheck_types.go#L175

Briefly talked to Joel and he mentioned that this is the validation that the kube-apiserver is doing:
* https://kubernetes.slack.com/archives/C8TSNPY4T/p1779882107705609
* https://github.com/kubernetes/apiextensions-apiserver/blob/53415cfc7b737bf3fc2fd062a16aa2076da8bbb3/pkg/apis/apiextensions/validation/validation.go#L793-L795
* https://github.com/kubernetes/apimachinery/blob/91c7cf9e414fcd94027d02768e9bcd995bac1fde/pkg/util/validation/validation.go#L222

sbueringer: cc @zhanggbj @JackieLan

zhanggbj: Thanks @sbueringer for pointing it out. To keep consistency inside CAPV and also other k8s projects, will revert it.

api/supervisor/v1beta2/vspheremachine_types.go
zhanggbj: Sure, will add a validation for feature gate checking.

api/supervisor/v1beta1/vspheremachine_types.go
zhanggbj: fixed.

sbueringer:
/override pull-cluster-api-provider-vsphere-e2e-supervisor-9-2-v1alpha6-main

v1alpha6 is not implemented yet, so this is expected to fail

sbueringer:
Thank you very much!

/lgtm


/assign @fabriziopandini

PR #3988, opened 2026-05-11
:sparkles:[WIP] Add control plane zone selection support

shuang-men:
**What this PR does / why we need it**:
Description:
introduce 'spec.controlPlaneFailureDomains' on VSphereCluster and VSphereClusterTemplate as an optional explicit list of failure domain names where control plane nodes should be placed. 

Key changes:

- add ControlPlaneFailureDomains []string  to VSphereClusterSpec (supervisor v1beta2); the field is v1beta2-only and not carried into v1beta1;
- add Control Plane failure domains filter to getFailureDomains logic;
- add unit tests covering ControlPlaneFailureDomains explicit list check and failure domains filtering function.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #4003, opened 2026-05-12
✨Add control plane zone selection support

ShuangMen:

Description:
introduce 'spec.failureDomains.controlPlane.selector' on VSphereCluster and VSphereClusterTemplate as an optional filter of failure domains where control plane nodes should be placed.

Key changes:

- Define new types: FailureDomainsSpec and FailureDomainsControlPlaneSpec
- Add FailureDomains FailureDomainsSpec to VSphereClusterSpec (supervisor v1beta2); the field is v1beta2-only and not carried into v1beta1;
- Add Control Plane failure domains filter to getFailureDomains logic;
- Add conditions for get failure domains step;
- Add unit tests covering failure domains filtering function.

Fixes #


ShuangMen:
@zhanggbj, I reuse my git account and get the EasyCLA, Thanks for your review comments in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/3988, let's move to this MR to continue the review.

sbueringer:
/test pull-cluster-api-provider-vsphere-e2e-govmomi-blocking-main

(just a quick test if CI is working)

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -508,12 +537,13 @@ func (r *ClusterReconciler) ZoneToVSphereClusters(ctx context.Context, o client.
 	return requests
 }
 
-// Returns the failure domain information discovered on the cluster
-// hosting this controller.
-func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace string) ([]clusterv1.FailureDomain, error) {
+// getFailureDomains returns the failure domain information discovered on the cluster
+// hosting this controller. The spec's FailureDomains.ControlPlane.Selector
+// field is used to set the ControlPlane boolean on each domain.
+func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace string, spec vmwarev1.VSphereClusterSpec) ([]clusterv1.FailureDomain, error) {
```
JackieLan: The selector is applied only in the NamespaceScopedZones (Zone) path. The legacy AvailabilityZone path hard-codes ControlPlane: true for every zone without calling applyControlPlaneFilter. A user on an older Supervisor (without NamespaceScopedZones enabled) who sets spec.failureDomains.controlPlane.selector gets silently ignored behavior — the selector has zero effect, and there is no warning condition or log message.

Document clearly (in the field doc-comment and API type comment) that the feature requires NamespaceScopedZones to be enabled, and emit a condition message if spec.failureDomains.controlPlane.selector != nil but the AvailabilityZone path is active.

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -237,14 +246,34 @@ func (r *ClusterReconciler) reconcileNormal(ctx context.Context, clusterCtx *vmw
...
+		}
+
+		conditions.Set(clusterCtx.VSphereCluster, metav1.Condition{
+			Type:    vmwarev1.VSphereClusterFailureDomainsReadyCondition,
+			Status:  metav1.ConditionFalse,
+			Reason:  reason,
+			Message: err.Error(),
+		})
+
 		return errors.Wrapf(
```
JackieLan, line 251: When the selector matches zero zones (e.g., user typo in label value), applyControlPlaneLabelSelector returns "no failure domains matched". This error:

Causes FailureDomainsReady=False with FailureDomainNotFound reason.
Returns early from reconcileNormal — the rest of cluster reconciliation (control plane endpoint, resource policy, network, etc.) is skipped.
The condition rolls up into the cluster-level ReadyCondition summary, making the entire VSphereCluster report Ready=False.
The controller requeues immediately, creating an infinite retry loop with no backoff specific to this error.
For what is an optional configuration field with no admission-time validation of the label key/value, this is an overly aggressive failure mode. A user typo in the selector YAML leaves the cluster stuck and noticeably degraded.

Suggested remediation: When the selector matches zero zones, set FailureDomainsReadyCondition=False with a descriptive Message, but continue reconciliation using a fallback policy (e.g., all zones eligible). Or, at minimum, remove FailureDomainsReadyCondition from the SetSummaryCondition inputs so this optional feature does not cascade into cluster-level Ready=False.

JackieLan (commented):
The design intent is sound and aligns with the CP zone selection one-pager. The implementation is clean and tests for the happy path are solid.

Thanks.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: - **When the selector matches zero zones ... but continue reconciliation using a fallback policy (e.g., all zones eligible)** 
Why we can't fallback to all zones (Fail-Closed): For the use case that need to segregate Management Zones (which have CP-class storage policies and management portgroups) from Workload Zones. If we fallback to 'all zones eligible' on a typo, KCP will provision a CP node in a Workload Zone where it has no CP storage and cannot reach the management network. This causes a critical infrastructure breakage. We absolutely must fail-closed rather than risk placing a CP node in an invalid zone.

- **no admission-time validation of the label key/value**
In the design, the GCC is implementing a validating webhook specifically to catch these typos and stale references up-front before they reach CAPV.

- **The Ready Condition**
Because a misconfigured selector fundamentally prevents the safe provisioning of the Control Plane, the cluster is effectively broken. Surfacing FailureDomainsReady=False into the top-level Ready=False summary is the most accurate reflection of the cluster's actual state.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments. I make some update in the code to address it.

api/supervisor/v1beta2/vspherecluster_types.go
```diff
@@ -166,6 +184,28 @@ type VSphereClusterSpec struct {
...
+// FailureDomainsSpec defines the desired state of FailureDomains.
+// +kubebuilder:validation:MinProperties=1
+type FailureDomainsSpec struct {
+	// controlPlane defines the failure domains for controlPlane.
+	// +optional
+	ControlPlane FailureDomainsControlPlaneSpec `json:"controlPlane,omitempty,omitzero"`
+}
+
+// FailureDomainsControlPlaneSpec defines the control plane failure domains.
+// +kubebuilder:validation:MinProperties=1
```
JackieLan, line 202: might be ok if such validation has been used in other place.

One trivial comment:

MinProperties=1 on FailureDomainsSpec and FailureDomainsControlPlaneSpec: Since Selector is the only field and it is +optional, a user who sets failureDomains: {} or failureDomains.controlPlane: {} in their YAML will get a webhook rejection (MinProperties not met), which is surprising for an empty struct. Consider removing MinProperties from these structs, or simply not annotating them (they rely on omitzero to suppress the zero-value).

JackieLan (approved):

api/supervisor/v1beta2/vspherecluster_types.go
ShuangMen: make lint will report error without "// +kubebuilder:validation:MinProperties=1"

ShuangMen:
@zhanggbj, could you review this again ? Thanks.

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -508,12 +537,13 @@ func (r *ClusterReconciler) ZoneToVSphereClusters(ctx context.Context, o client.
 	return requests
 }
 
-// Returns the failure domain information discovered on the cluster
-// hosting this controller.
-func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace string) ([]clusterv1.FailureDomain, error) {
+// getFailureDomains returns the failure domain information discovered on the cluster
+// hosting this controller. The spec's FailureDomains.ControlPlane.Selector
+// field is used to set the ControlPlane boolean on each domain.
```
zhanggbj: Minor:
Good to use "failure domain" or "failureDomain" to be consistent through the godoc or comments, instead of "domain"

api/supervisor/v1beta2/vspherecluster_types.go
```diff
@@ -166,6 +184,28 @@ type VSphereClusterSpec struct {
...
+type FailureDomainsControlPlaneSpec struct {
+	// selector is a label selector to dynamically match failure domains.
+	// Note: This feature requires the NamespaceScopedZones feature gate to be enabled.
+	// If a selector is provided while the feature gate is disabled, the cluster will surface a degraded condition.
```
zhanggbj: could you clarify "the cluster will surface a degraded condition"?
Currently if NamespaceScopedZones is false, it will return error and requeue, so the cluster got stuck, refer to https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/4003/changes#diff-1df9d61be5152fcadd30e1056ca67b94404224e90d66a3062c376bb08d49b9f1R585-R587

api/supervisor/v1beta2/vspherecluster_types.go
```diff
@@ -127,6 +127,24 @@ const (
...
+	VSphereClusterFailureDomainsReadyCondition = "FailureDomainsReady"
+
+	// VSphereClusterFailureDomainsReadyReason surfaces when the FailureDomains are successfully discovered and matched.
+	VSphereClusterFailureDomainsReadyReason = clusterv1.ReadyReason
+
+	// VSphereClusterFailureDomainsNotReadyReason surfaces generic failures during FailureDomains discovery (e.g., API server list errors).
+	VSphereClusterFailureDomainsNotReadyReason = clusterv1.NotReadyReason
+
+	// VSphereClusterFailureDomainsNotFoundReason surfaces explicitly when no failure domains match the specified control plane label selector.
+	VSphereClusterFailureDomainsNotFoundReason = "FailureDomainNotFound"
```
zhanggbj: Not matched, is it in plural?
```suggestion
	VSphereClusterFailureDomainsNotFoundReason = "FailureDomainsNotFound"
```

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -237,14 +246,34 @@ func (r *ClusterReconciler) reconcileNormal(ctx context.Context, clusterCtx *vmw
 	log := ctrl.LoggerFrom(ctx)
 
 	// Get any failure domains to report back to the CAPI core controller.
-	failureDomains, err := r.getFailureDomains(ctx, clusterCtx.VSphereCluster.Namespace)
+	failureDomains, err := r.getFailureDomains(ctx, clusterCtx.VSphereCluster.Namespace, clusterCtx.VSphereCluster.Spec)
 	if err != nil {
+		reason := vmwarev1.VSphereClusterFailureDomainsNotReadyReason
+		// If the error explicitly mentions that no domains matched, use our targeted reason
+		if strings.Contains(err.Error(), "no failure domains matched") {
```
zhanggbj: if that error message in `applyControlPlaneLabelSelector` changes, the check breaks. Please use  a typed error.

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -570,3 +611,77 @@ func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace str
...
+	if err != nil {
+		// Return the unmodified domains alongside the error
+		return domains, fmt.Errorf("invalid controlPlaneFailureDomainSelector: %w", err)
+	}
+
+	// Verify at least one domain matches the selector.
+	// We do this to prevent mutating the domains array if the configuration is invalid or yields 0 zones.
+	hasMatch := false
+	for _, domain := range domains {
+		if labelSel.Matches(labelsAsSet(domainLabels[domain.Name])) {
```
zhanggbj: Suggest to use labelSel.Matches(labels.Set(domainLabels[domain.Name])) provided by "k8s.io/apimachinery/pkg/labels", no `labelsAsSet` is needed.

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -528,15 +558,20 @@ func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace str
...
+				Name: zone.Name,
 			})
 		}
 
 		if len(failureDomains) == 0 {
 			return nil, nil
 		}
 
+		var err error
+		failureDomains, err = applyControlPlaneFilter(failureDomains, zoneLabels(zoneList), spec)
```
zhanggbj: zoneLabels is called unconditionally even when no selector is set. Suggest to move the zoneLabels call inside applyControlPlaneLabelSelector

controllers/vmware/vspherecluster_reconciler.go
zhanggbj: for helper functions, please only pass required fields.The function only uses spec.FailureDomains, suggest to pass spec.FailureDomains instead of the whole `vmwarev1.VSphereClusterSpec`.

api/supervisor/v1beta2/vspherecluster_types.go
```diff
@@ -127,6 +127,24 @@ const (
 	VSphereClusterServiceDiscoveryNotReadyReason = clusterv1.NotReadyReason
 )
 
+// VSphereCluster's FailureDomainsReady condition and corresponding reasons that will be used in v1Beta2 API version.
```
zhanggbj, line 130: typo, 
```suggestion
// VSphereCluster's FailureDomainsReady condition and corresponding reasons that will be used in v1beta2 API version.
```

zhanggbj (changes requested):

zhanggbj:
Please also update godoc https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/api/supervisor/v1beta2/vspherecluster_types.go#L200-L202 accordingly if new condition is introduced.

api/supervisor/v1beta2/vspherecluster_types.go
ShuangMen: I see other comments also use  "... that will be used in v1Beta2 API version"

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments, code updated to spec.FailureDomains

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments, code updated, move call zoneLabels into function applyControlPlaneFilter.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments. removed labelsAsSet and use labels.Set()

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments, code updated, use ErrNoFailureDomainsMatched.

api/supervisor/v1beta2/vspherecluster_types.go
ShuangMen: Thanks for the comments. code updated. use
VSphereClusterFailureDomainsNotFoundReason = "FailureDomainsNotFound"

api/supervisor/v1beta2/vspherecluster_types.go
ShuangMen: Thanks for the comments, I update the words to make this clear.

	// If a selector is provided while the feature gate is disabled, cluster
	// reconciliation will fail and requeue.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments, updated to "failure domain"

ShuangMen:
> Please also update godoc https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/api/supervisor/v1beta2/vspherecluster_types.go#L200-L202 accordingly if new condition is introduced.

@zhanggbj , thanks a lot for your review comments.  I've updated the code based on the comments and added the new FailureDomainsReady condition in the related godoc.

ShuangMen:
/assign @fabriziopandini
/assign @sbueringer
PR is ready for review, thanks!

api/supervisor/v1beta2/vspherecluster_types.go
```diff
@@ -166,6 +184,29 @@ type VSphereClusterSpec struct {
 	// network defines the network configuration for the cluster with different network providers.
 	// +optional
 	Network Network `json:"network,omitempty,omitzero"`
+
+	// failureDomains defines the failure domains.
+	// +optional
+	FailureDomains FailureDomainsSpec `json:"failureDomains,omitempty,omitzero"`
```
fabriziopandini, line 190: New fields must be added to both v1beta1 and v1beta2 versions to avoid edge case errors during conversions

api/supervisor/v1beta2/vspherecluster_types.go
```diff
@@ -166,6 +184,29 @@ type VSphereClusterSpec struct {
...
+type FailureDomainsSpec struct {
+	// controlPlane defines the failure domains for controlPlane.
+	// +optional
+	ControlPlane FailureDomainsControlPlaneSpec `json:"controlPlane,omitempty,omitzero"`
+}
+
+// FailureDomainsControlPlaneSpec defines the control plane failure domains.
+// +kubebuilder:validation:MinProperties=1
+type FailureDomainsControlPlaneSpec struct {
+	// selector is a label selector to dynamically match failure domains.
```
fabriziopandini:
```suggestion
	// selector is a label selector to dynamically match supervisor's Zone CR.
```

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -237,14 +248,34 @@ func (r *ClusterReconciler) reconcileNormal(ctx context.Context, clusterCtx *vmw
 	log := ctrl.LoggerFrom(ctx)
 
 	// Get any failure domains to report back to the CAPI core controller.
-	failureDomains, err := r.getFailureDomains(ctx, clusterCtx.VSphereCluster.Namespace)
+	failureDomains, err := r.getFailureDomains(ctx, clusterCtx.VSphereCluster.Namespace, clusterCtx.VSphereCluster.Spec.FailureDomains)
```
fabriziopandini: I would pass the  clusterCtx.VSphereCluster object to the function, rename it to reconcileFailureDomains, and move both setting conditions and Status.FailureDomains into the func (so that everything related to this topic is in a single func)

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -570,3 +613,75 @@ func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace str
 	})
 	return failureDomains, nil
 }
+
+// applyControlPlaneFilter stamps the ControlPlane boolean on each failure domain according to
+// the VSphereClusterSpec's control-plane placement constraints:
+//
+//   - If FailureDomains.ControlPlane.Selector is set, ControlPlane is true only for domains matching the selector.
+//   - If FailureDomains.ControlPlane.Selector is not set, ControlPlane is true for all domains (backwards-compatible default).
+func applyControlPlaneFilter(
```
fabriziopandini: should this be markControlPlaneFailureDomains

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -570,3 +613,75 @@ func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace str
...
+	domains []clusterv1.FailureDomain,
+	zoneList *topologyv1.ZoneList,
+	spec vmwarev1.FailureDomainsSpec,
+) ([]clusterv1.FailureDomain, error) {
+	if spec.ControlPlane.Selector != nil {
+		if zoneList == nil {
+			return nil, ErrNoFailureDomainsMatched
+		}
```
fabriziopandini: I think we should simplify the logic that determines if at least one failure domains is set to host control plane machines, because currently is scattered between this function and applyControlPlaneLabelSelector.

What we can do is to get the failure domains, and then use something like

```suggestion
	hasFailureDomainForControlPlaneMachines := slices.ContainsFunc (failureDomains, func(failureDomain clusterv1.FailureDomain) bool {
		return ptr.Deref(failureDomain.ControlPlane, false)
	})
```

To check the resulting list (instead of checking while we compute the resulting list

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -570,3 +613,75 @@ func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace str
...
+		l := domainLabels[domains[i].Name]
+		matches := labelSel.Matches(labels.Set((l)))
+		domains[i].ControlPlane = ptr.To(matches)
+	}
+
+	return domains, nil
+}
+
+// zoneLabels returns a map of zone name → labels for namespaced Zone objects.
+func zoneLabels(zoneList *topologyv1.ZoneList) map[string]map[string]string {
```
fabriziopandini: I would avoid to build a map[object]labels.
Let's instead go trough the list of zones and check each one labels (object.Lables)

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -570,3 +613,75 @@ func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace str
...
 	return failureDomains, nil
 }
+
+// applyControlPlaneFilter stamps the ControlPlane boolean on each failure domain according to
+// the VSphereClusterSpec's control-plane placement constraints:
+//
+//   - If FailureDomains.ControlPlane.Selector is set, ControlPlane is true only for domains matching the selector.
+//   - If FailureDomains.ControlPlane.Selector is not set, ControlPlane is true for all domains (backwards-compatible default).
+func applyControlPlaneFilter(
+	domains []clusterv1.FailureDomain,
```
fabriziopandini: nit 
```suggestion
	failureDomains []clusterv1.FailureDomain,
```
Let's use consistently failureDomains (also in other places)

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -570,3 +613,75 @@ func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace str
...
+	for i := range domains {
+		domains[i].ControlPlane = ptr.To(true)
+	}
+
+	return domains, nil
+}
+
+// applyControlPlaneLabelSelector sets ControlPlane=true only for domains whose labels match
+// the given selector. If no domains match the selector, it returns an error without modifying the domains.
+func applyControlPlaneLabelSelector(
```
fabriziopandini: After cleaning up ErrNoFailureDomainsMatched and zoneLabels, what will be left of this func is only , build selector + for loop where we check if zone match/set control plane flag.

If this is correct, I would suggest to simplify and merge this code into applyControlPlaneFilter

api/supervisor/v1beta2/vspherecluster_types.go
```diff
@@ -127,6 +127,24 @@ const (
...
+// VSphereCluster's FailureDomainsReady condition and corresponding reasons that will be used in v1Beta2 API version.
+const (
+	// VSphereClusterFailureDomainsReadyCondition documents the status of the FailureDomains discovery and filtering for a VSphereCluster.
+	VSphereClusterFailureDomainsReadyCondition = "FailureDomainsReady"
+
+	// VSphereClusterFailureDomainsReadyReason surfaces when the FailureDomains are successfully discovered and matched.
+	VSphereClusterFailureDomainsReadyReason = clusterv1.ReadyReason
+
+	// VSphereClusterFailureDomainsNotReadyReason surfaces generic failures during FailureDomains discovery (e.g., API server list errors).
+	VSphereClusterFailureDomainsNotReadyReason = clusterv1.NotReadyReason
```
fabriziopandini, line 139: We usually use an InternalErrorReason (VSphereClusterFailureDomainsReadyInternalErrorReason in this case) for errors that should not happen under normal circumstances e.g. API server errors while reading from cache.

After moving the errors above to a separated reason, I would suggest to use NotReady for FailureDomainsNotFound (status false, reason NotReady, message "no zone is matching the selector for control plane failure domains", or we can even use the selector string)<

If this works out, then we can drop VSphereClusterFailureDomainsNotFoundReason

ShuangMen: Thanks for the comments, code updated.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments, code updated.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments, code updated.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments, code updated.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments, code updated.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments, code updated.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments, code updated.

api/supervisor/v1beta2/vspherecluster_types.go
ShuangMen: Thanks for the comments, code updated.

api/supervisor/v1beta2/vspherecluster_types.go
ShuangMen: Thanks for the comments, add this to both v1beta1 and v1beta2

api/supervisor/v1beta2/vspherecluster_types.go
```diff
@@ -127,6 +127,24 @@ const (
 	VSphereClusterServiceDiscoveryNotReadyReason = clusterv1.NotReadyReason
 )
 
+// VSphereCluster's FailureDomainsReady condition and corresponding reasons that will be used in v1Beta2 API version.
+const (
+	// VSphereClusterFailureDomainsReadyCondition documents the status of the FailureDomains discovery and filtering for a VSphereCluster.
+	VSphereClusterFailureDomainsReadyCondition = "FailureDomainsReady"
```
fabriziopandini, line 133: For sake of completeness, please let's add the new condition also on v1beta1; a few differences apply
- Const suffix must be V1Beta2Condition or V1Beta2Reason (instead of Condition/Reason)
- The Condition field with the list of know conditions is status.v1beta2.conditions (instead of status.conditions)
- Ready condition is VSphereClusterReadyV1beta2Condition

api/supervisor/v1beta2/vspherecluster_types.go
```diff
@@ -127,6 +127,24 @@ const (
...
+
+	// VSphereClusterFailureDomainsReadyReason surfaces when the FailureDomains are successfully discovered and matched.
+	VSphereClusterFailureDomainsReadyReason = clusterv1.ReadyReason
+
+	// VSphereClusterFailureDomainsNotReadyReason surfaces when no failure domains match the specified control plane label selector.
+	VSphereClusterFailureDomainsNotReadyReason = clusterv1.NotReadyReason
+
+	// VSphereClusterFailureDomainsReadyInternalErrorReason surfaces unexpected failures during FailureDomains discovery.
+	VSphereClusterFailureDomainsReadyInternalErrorReason = "InternalError"
```
fabriziopandini:
```suggestion
	VSphereClusterFailureDomainsReadyInternalErrorReason = clusterv1.InternalErrorReason
```

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -508,18 +517,38 @@ func (r *ClusterReconciler) ZoneToVSphereClusters(ctx context.Context, o client.
...
 	// Determine the source of failure domain based on feature gates NamespaceScopedZones.
-	// If NamespaceScopedZones is enabled, use Zone which is Namespace scoped,otherwise use
+	// If NamespaceScopedZones is enabled, use Zone which is Namespace scoped, otherwise use
 	// Availability Zone which is Cluster scoped.
 	if feature.Gates.Enabled(feature.NamespaceScopedZones) {
 		zoneList := &topologyv1.ZoneList{}
 		listOptions := &client.ListOptions{Namespace: namespace}
 		if err := r.Client.List(ctx, zoneList, listOptions); err != nil {
-			return nil, errors.Wrapf(err, "failed to list Zones in namespace %s", namespace)
+			return markFailure(errors.Wrapf(err, "failed to list Zones in namespace %s", namespace))
```
fabriziopandini:
```suggestion
			conditions.Set(clusterResourceSet, metav1.Condition{
			   Type:    vmwarev1.VSphereClusterFailureDomainsReadyCondition,
			   Status:  metav1.ConditionFalse,
			  Reason:  vmwarev1.VSphereClusterFailureDomainsReadyInternalErrorReason,
			  Message: "Please check controller logs for errors",
			})
			return markFailure(errors.Wrapf(err, "failed to list Zones in namespace %s", namespace))
```
We usually do not surface errors messages directly on conditions, because non-deterministic error message can lead to infinite reconcile. In this case, the error should never happen (we reading from cache) so one reason more to avoid duplicating go error handling on conditions.

Considering this, let's also drop the inline func markFailure and use conditions.Set where necessary, even if it is a little bit verbose.

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -528,45 +557,104 @@ func (r *ClusterReconciler) getFailureDomains(ctx context.Context, namespace str
...
-				ControlPlane: ptr.To(true),
+				Name: zone.Name,
 			})
 		}
 
-		if len(failureDomains) == 0 {
-			return nil, nil
+		if len(failureDomains) > 0 {
+			var err error
+			failureDomains, err = markControlPlaneFailureDomains(failureDomains, zoneList, spec)
```
fabriziopandini: I think that this markControlPlaneFailureDomains can be greatly simplified if it is called inside the 
for _, zone := range zoneList.Items above (L554) and works one zone at time.

The signature of the func should be something like

```
func markControlPlaneFailureDomain(failureDomain *clusterv1.FailureDomain, zone topologyv1.Zone, spec vmwarev1.FailureDomainsControlPlaneSpec) error
```

And the for loop above should look

```
for _, zone := range zoneList.Items {
			// Skip zones which are in deletion
			if !zone.DeletionTimestamp.IsZero() {
				continue
			}

			failureDomain := clusterv1.FailureDomain{
				Name: zone.Name,
			}
			if err := markControlPlaneFailureDomain(&failureDomain, zone, vsphereCluster.Spec.FailureDomains.ControlPlane); err != nil {
				conditions.Set(vsphereCluster, metav1.Condition{
					Type:    vmwarev1.VSphereClusterFailureDomainsReadyCondition,
					Status:  metav1.ConditionFalse,
					Reason:  vmwarev1.VSphereClusterFailureDomainsReadyInternalErrorReason,
					Message: "Please check controller logs for errors",
				})
				return err
			}

			failureDomains = append(failureDomains, failureDomain)
		}
```

The check for at least one control plane zone must existing, must be moved at the end of the func (after the if feature.Gates.Enabled(feature.NamespaceScopedZones {} else {})

ShuangMen: Thanks for the comments. function markControlPlaneFailureDomain is simplified and  "at least one control plane zone must existing" is moved to the end as suggest.

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments. 
func markFailure is removed and use conditions.Set where necessary. Messages are set as suggest.

api/supervisor/v1beta2/vspherecluster_types.go
ShuangMen: Thanks for the comments. 
code updated as suggest.

api/supervisor/v1beta2/vspherecluster_types.go
ShuangMen: Thanks for the comments.
conditions are added on v1beta1.

ShuangMen:
@fabriziopandini, Thanks a lot for your review comments. I've updated the code to address them.

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -508,65 +516,141 @@ func (r *ClusterReconciler) ZoneToVSphereClusters(ctx context.Context, o client.
...
+		// Default to false unless the individual zone labels match the selector criteria
+		failureDomain.ControlPlane = ptr.To(false)
+		if selector.Matches(labels.Set(zone.Labels)) {
+			failureDomain.ControlPlane = ptr.To(true)
+		}
+	} else {
+		// Backwards-compatible default
+		failureDomain.ControlPlane = ptr.To(true)
+	}
+
```
fabriziopandini: nit:
```suggestion
    	return nil
	} 
	
	// Backwards-compatible default
	failureDomain.ControlPlane = ptr.To(true)
	return nil
```

controllers/vmware/vspherecluster_reconciler_test.go
```diff
@@ -106,12 +106,22 @@ var _ = Describe("Cluster Controller Tests", func() {
...
-			c := conditions.Get(clusterCtx.VSphereCluster, vmwarev1.VSphereClusterResourcePolicyReadyCondition)
-			Expect(c).NotTo(BeNil())
-			Expect(c.Status).To(Equal(metav1.ConditionFalse))
-			Expect(c.Reason).To(Equal(clusterv1.DeletingReason))
+
+			// Verify ResourcePolicy condition
+			c1 := conditions.Get(clusterCtx.VSphereCluster, vmwarev1.VSphereClusterResourcePolicyReadyCondition)
+			Expect(c1).NotTo(BeNil())
+			Expect(c1.Status).To(Equal(metav1.ConditionFalse))
+			Expect(c1.Reason).To(Equal(clusterv1.DeletingReason))
```
fabriziopandini: nit
```suggestion
			Expect(c1.Reason).To(Equal(vmwarev1.VSphereClusterResourcePolicyReadyDeletingReason))
```

controllers/vmware/vspherecluster_reconciler_test.go
```diff
@@ -144,6 +154,7 @@ func TestClusterReconciler_getFailureDomains(t *testing.T) {
 	tests := []struct {
 		name        string
 		objects     []client.Object
+		spec        vmwarev1.VSphereClusterSpec
```
fabriziopandini: Looks like there is no test case using this info.
Let's add test cases at least for:
- selector set, CP domains found (happy path)
- selector set, CP domains not found

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -508,65 +516,141 @@ func (r *ClusterReconciler) ZoneToVSphereClusters(ctx context.Context, o client.
...
+					Reason:  vmwarev1.VSphereClusterFailureDomainsReadyInternalErrorReason,
+					Message: "Please check controller logs for errors",
+				})
+				return err
+			}
+
+			failureDomains = append(failureDomains, failureDomain)
+		}
+	} else {
+		// Prevent silent failure if the user tried to use a selector without the feature gate.
```
fabriziopandini: What is the use case when this scenario might happen? (I think we should defer to web hooks / validation to prevent invalid configurations to happen)

controllers/vmware/vspherecluster_reconciler_test.go
```diff
@@ -276,3 +294,271 @@ func failureDomains(names ...string) []clusterv1.FailureDomain {
...
+	const ns = "test-ns"
+	namespace := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}}
+
+	tests := []struct {
+		name                 string
+		objects              []client.Object
+		spec                 vmwarev1.VSphereClusterSpec
+		featureGate          bool
+		wantCP               map[string]bool // nil means skip per-domain assertion; use wantNil
+		wantNil              bool
```
fabriziopandini: Lets combine this two into 
```suggestion
		want        []clusterv1.FailureDomain
```

controllers/vmware/vspherecluster_reconciler_test.go
```diff
@@ -276,3 +294,271 @@ func failureDomains(names ...string) []clusterv1.FailureDomain {
...
+
+			// Assert the Condition Reasons and static Messages to satisfy the PR requirements
+			if tt.wantConditionReason != "" {
+				cond := conditions.Get(vsphereCluster, vmwarev1.VSphereClusterFailureDomainsReadyCondition)
+				g.Expect(cond).NotTo(BeNil(), "expected FailureDomainsReadyCondition to be set")
+				g.Expect(cond.Reason).To(Equal(tt.wantConditionReason))
+
+				if tt.wantConditionMessage != "" {
+					g.Expect(cond.Message).To(Equal(tt.wantConditionMessage))
+				}
```
fabriziopandini: Let's use g.Expect(*conditions.Get(vsphereCluster, vmwarev1.VSphereClusterFailureDomainsReadyCondition)).To(conditions.MatchCondition(tt.wantCondition, conditions.IgnoreLastTransitionTime(true)))

controllers/vmware/vspherecluster_reconciler_test.go
```diff
@@ -276,3 +294,271 @@ func failureDomains(names ...string) []clusterv1.FailureDomain {
...
+				return
+			}
+
+			g.Expect(got).To(HaveLen(len(tt.wantCP)))
+			for _, fd := range got {
+				expected, ok := tt.wantCP[fd.Name]
+				g.Expect(ok).To(BeTrue(), "unexpected domain %q in result", fd.Name)
+				g.Expect(ptr.Deref(fd.ControlPlane, false)).To(Equal(expected),
+					"domain %q: expected ControlPlane=%v", fd.Name, expected)
+			}
```
fabriziopandini: Let's check the entire sphereCluster.Status.FailureDomains equals wantFailureDomains

controllers/vmware/vspherecluster_reconciler_test.go
```diff
@@ -276,3 +294,271 @@ func failureDomains(names ...string) []clusterv1.FailureDomain {
...
+	tests := []struct {
+		name                 string
+		objects              []client.Object
+		spec                 vmwarev1.VSphereClusterSpec
+		featureGate          bool
+		wantCP               map[string]bool // nil means skip per-domain assertion; use wantNil
+		wantNil              bool
+		wantErr              bool
+		wantConditionReason  string
+		wantConditionMessage string
```
fabriziopandini: Let's combine this in wantCondition.

controllers/vmware/vspherecluster_reconciler_test.go
```diff
@@ -276,3 +294,271 @@ func failureDomains(names ...string) []clusterv1.FailureDomain {
...
+				return
+			}
+
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(ptr.Deref(fd.ControlPlane, false)).To(Equal(tt.wantCtrlPlane))
+		})
+	}
+}
+
+func TestClusterReconciler_reconcileFailureDomains_ControlPlaneFilter(t *testing.T) {
```
fabriziopandini: This seems very similar to TestClusterReconciler_reconcileFailureDomains, can we merge the two test?

fabriziopandini (commented):
Last round of nits from my side, mostly about test code

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments. code updated as suggest.

controllers/vmware/vspherecluster_reconciler_test.go
ShuangMen: Thanks for the comments. code updated as suggest.

controllers/vmware/vspherecluster_reconciler_test.go
ShuangMen: Thanks for the comments. test cases are updated and covers below path:
- selector set, CP domains found (happy path)
- selector set, CP domains not found

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks for the comments @fabriziopandini 
I have removed this check from the reconciler as suggested.
Since we are on the last round of nits, would you like me to add the webhook validation in this PR, or if we should handle it in a separate follow-up PR ? Thanks

controllers/vmware/vspherecluster_reconciler_test.go
ShuangMen: thanks for the comments, code updated as suggest.

controllers/vmware/vspherecluster_reconciler_test.go
ShuangMen: Thanks for the comments, code updated as suggest.

controllers/vmware/vspherecluster_reconciler_test.go
ShuangMen: Thanks for the comments, code updated as suggest.

controllers/vmware/vspherecluster_reconciler_test.go
ShuangMen: thanks for the comments, merge the test cases into TestClusterReconciler_reconcileFailureDomains and remove TestClusterReconciler_reconcileFailureDomains_ControlPlaneFilter.

controllers/vmware/vspherecluster_reconciler_test.go
ShuangMen: Thanks for the comments. code updated as suggest.

controllers/vmware/vspherecluster_reconciler.go
fabriziopandini: @sbueringer opinions?

fabriziopandini:
@ShuangMen FYI for future PRs. If you avoid including merge commits in the commit history review will be faster (because other changes are out of the picture)

ShuangMen:
> @ShuangMen FYI for future PRs. If you avoid including merge commits in the commit history review will be faster (because other changes are out of the picture)

Thanks for your reminder, I will pay attention

controllers/vmware/vspherecluster_reconciler.go
sbueringer: Sorry I don't have enough context about how the code previously looked

ShuangMen: There was a check on set selector when feature.NamespaceScopedZones disabled.
Screenshot 2026-05-27 at 18 14 13

sbueringer:
> > @ShuangMen FYI for future PRs. If you avoid including merge commits in the commit history review will be faster (because other changes are out of the picture)
> 
> Thanks for your reminder, I will pay attention

So basically **when** you hit a merge conflict, just rebase onto latest main and push force. Thx!

controllers/vmware/vspherecluster_reconciler.go
sbueringer: Discussed with Fabrizio.

Let's do a follow-up PR for validation of the selector. We would like to validate:

1. That the selector is only set if the feature gate is on

2. We should validate that the selector is valid, e.g.

```go
	selector, err := metav1.LabelSelectorAsSelector(&<selector field>)
	if err != nil {
		allErrs = append(
			allErrs,
			field.Invalid(field.NewPath(<selector field path>), <selector field>, err.Error()),
		)
	}

	if selector != nil && selector.Empty() {
		allErrs = append(
			allErrs,
			field.Invalid(field.NewPath(<selector field path>), <selector field>, "selector must not be empty"),
		)
	}
```

3. Let's also add the same validation for VSphereClusterTemplate (for that you'll have to introduced a new validating webhook for VSphereClusterTemplate)

But let's do this in a follow-up PR

sbueringer:
/approve

So you can already start testing, please follow-up for this one: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/4003#discussion_r3310246629

controllers/vmware/vspherecluster_reconciler.go
ShuangMen: Thanks @sbueringer , I will do this in a new PR.

PR #4009, opened 2026-05-18
🌱 Split supervisor and govmomi feature gates, add support for gates depending on vm-operator-version

fabriziopandini:

Fixes #

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -22,7 +22,6 @@ spec:
         - "--insecure-diagnostics=${CAPI_INSECURE_DIAGNOSTICS:=false}"
         - --v=4
         - "--feature-gates=PriorityQueue=${EXP_PRIORITY_QUEUE:=true},ReconcilerRateLimiting=${EXP_RECONCILER_RATE_LIMITING:=true},NodeAntiAffinity=${EXP_NODE_ANTI_AFFINITY:=false}"
-        - "--vm-operator-api-version=${VM_OPERATOR_API_VERSION:=v1alpha5}"
         image: controller:latest
         imagePullPolicy: IfNotPresent
         name: manager
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -21,7 +21,7 @@ spec:
         - "--diagnostics-address=${CAPI_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPI_INSECURE_DIAGNOSTICS:=false}"
         - --v=4
-        - "--feature-gates=MultiNetworks=${EXP_MULTI_NETWORKS:=false},NodeAntiAffinity=${EXP_NODE_ANTI_AFFINITY:=false},NamespaceScopedZones=${EXP_NAMESPACE_SCOPED_ZONES:=false},NodeAutoPlacement=${EXP_NODE_AUTO_PLACEMENT:=false},PriorityQueue=${EXP_PRIORITY_QUEUE:=true},ReconcilerRateLimiting=${EXP_RECONCILER_RATE_LIMITING:=true}"
+        - "--feature-gates=PriorityQueue=${EXP_PRIORITY_QUEUE:=true},ReconcilerRateLimiting=${EXP_RECONCILER_RATE_LIMITING:=true},NodeAntiAffinity=${EXP_NODE_ANTI_AFFINITY:=false}"
         - "--vm-operator-api-version=${VM_OPERATOR_API_VERSION:=v1alpha5}"
         image: controller:latest
         imagePullPolicy: IfNotPresent
--- a/feature/gates.go
+++ b/feature/gates.go
@@ -17,17 +17,210 @@ limitations under the License.
 package feature
 
 import (
+	"errors"
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/pflag"
+	vmoprv1alpha2 "github.com/vmware-tanzu/vm-operator/api/v1alpha2"
+	vmoprv1alpha5 "github.com/vmware-tanzu/vm-operator/api/v1alpha5"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/component-base/featuregate"
 )
 
 var (
-	// MutableGates is a mutable version of DefaultFeatureGate.
-	// Only top-level commands/options setup and the k8s.io/component-base/featuregate/testing package should make use of this.
-	// Tests that need to modify featuregate gates for the duration of their test should use:
-	//   defer featuregatetesting.SetFeatureGateDuringTest(t, utilfeature.DefaultFeatureGate, features.<FeatureName>, <value>)()
-	MutableGates = featuregate.NewFeatureGate()
+	allGates featuregate.MutableVersionedFeatureGate
 
 	// Gates is a shared global FeatureGate.
-	// Top-level commands/options setup that needs to modify this featuregate gate should use DefaultMutableFeatureGate.
-	Gates featuregate.FeatureGate = MutableGates
+	Gates featuregate.FeatureGate
 )
+
+func init() {
+	// Add all gates to avoid issue in test code that assumes gates are properly set up.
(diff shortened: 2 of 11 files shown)


feature/gates.go
```diff
@@ -17,17 +17,210 @@ limitations under the License.
...
+	}
+
+	known := sets.New[string]()
+	for _, gate := range allowedGates.KnownFeatures() {
+		known.Insert(gate[0:strings.Index(gate, "=")])
+	}
+
+	var values []string
+	for _, s := range strings.Split(value, ",") {
+		if s != "" {
```
zhanggbj: looks like the intention is to skip empty string? now it's the opposite.

feature/gates.go
```diff
@@ -17,17 +17,210 @@ limitations under the License.
...
+	}
+
+	description := "A set of key=value pairs that describe feature gates for alpha/experimental features.\n"
+	description += fmt.Sprintf("Common options are:\n%s\n", strings.Join(fix(commonMutableGates.KnownFeatures(), false), "\n"))
+	description += fmt.Sprintf("Options for govmomi mode are:\n%s\n", strings.Join(fix(govmomiMutableGates.KnownFeatures(), true), "\n"))
+	for _, v := range supportedVersions {
+		supervisorMutableGates := featuregate.NewVersionedFeatureGate(toFeatureVersion(v))
+		utilruntime.Must(supervisorMutableGates.Add(supervisorGates))
+		utilruntime.Must(supervisorMutableGates.AddVersioned(supervisorVersionedGates))
+		description += fmt.Sprintf("Options for supervisor mode when --vm-operator-api-version=%s\n%s are:\n", v, strings.Join(fix(supervisorMutableGates.KnownFeatures(), true), "\n"))
```
zhanggbj:
```suggestion
		description += fmt.Sprintf("Options for supervisor mode when --vm-operator-api-version=%s are:\n%s\n",
    v,
    strings.Join(fix(supervisorMutableGates.KnownFeatures(), true), "\n"),
)
```

config/manager/manager.yaml
```diff
@@ -21,7 +21,7 @@ spec:
         - "--diagnostics-address=${CAPI_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPI_INSECURE_DIAGNOSTICS:=false}"
         - --v=4
-        - "--feature-gates=MultiNetworks=${EXP_MULTI_NETWORKS:=false},NodeAntiAffinity=${EXP_NODE_ANTI_AFFINITY:=false},NamespaceScopedZones=${EXP_NAMESPACE_SCOPED_ZONES:=false},NodeAutoPlacement=${EXP_NODE_AUTO_PLACEMENT:=false},PriorityQueue=${EXP_PRIORITY_QUEUE:=true},ReconcilerRateLimiting=${EXP_RECONCILER_RATE_LIMITING:=true}"
+        - "--feature-gates=PriorityQueue=${EXP_PRIORITY_QUEUE:=true},ReconcilerRateLimiting=${EXP_RECONCILER_RATE_LIMITING:=true},NodeAntiAffinity=${EXP_NODE_ANTI_AFFINITY:=false}"
         - "--vm-operator-api-version=${VM_OPERATOR_API_VERSION:=v1alpha5}"
```
sbueringer: Does it still make sense to set the vm-operator-api flag here?

test/infrastructure/net-operator/main.go
```diff
@@ -101,10 +106,10 @@ func InitFlags(fs *pflag.FlagSet) {
 	logsv1.AddFlags(logOptions, fs)
 
 	fs.StringVar(
-		&apiVersionVMOperator,
+		&vmOperatorAPIVersion,
 		"vm-operator-api-version",
 		vmoprv1alpha5.GroupVersion.Version,
-		fmt.Sprintf("the API version to use when reading and writing VM Operator resources in supervisor mode. Valid values are: %s, %s", vmoprv1alpha2.GroupVersion.Version, vmoprv1alpha5.GroupVersion.Version),
+		fmt.Sprintf("the API version to use when reading and writing VM Operator resources in supervisor mode. Valid values are: %s", strings.Join(supportedVMOperatorAPIVersions, ",")),
```
sbueringer:
```suggestion
		fmt.Sprintf("the API version to use when reading and writing VM Operator resources in supervisor mode. Valid values are: %s", strings.Join(supportedVMOperatorAPIVersions, ", ")),
```
Maybe? (if yes, same in other main.go's)

main.go
```diff
@@ -266,6 +270,8 @@ func InitFlags(fs *pflag.FlagSet) {
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=vspheremachinetemplates;vsphereclustertemplates,verbs=get;list;watch;patch;update
 
 func main() {
+	time.Sleep(10 * time.Second)
```
sbueringer: :joy:

feature/gates.go
```diff
@@ -17,17 +17,210 @@ limitations under the License.
...
+	var values []string
+	for _, s := range strings.Split(value, ",") {
+		if s == "" {
+			continue
+		}
+
+		// Gets the name of the feature flag.
+		arr := strings.SplitN(s, "=", 2)
+		k := strings.TrimSpace(arr[0])
+		if len(arr) != 2 {
```
sbueringer, line 155: Would be nice to have an example here to explain when we are hitting this branch

In general it's a bit hard to understand l.146-168, maybe try to improve the godoc a bit (mostly all the continue cases, not sure when we hit them)

feature/gates_test.go
```diff
@@ -0,0 +1,219 @@
...
+const (
+	CommonFeature                      featuregate.Feature = "CommonFeature"
+	GovmomiFeature                     featuregate.Feature = "GovmomiFeature"
+	SupervisorFeature                  featuregate.Feature = "SupervisorFeature"
+	SupervisorFeatureEnabledOnV1Alpha2 featuregate.Feature = "SupervisorFeatureEnabledOnV1Alpha2"
+	SupervisorFeatureEnabledOnV1Alpha5 featuregate.Feature = "SupervisorFeatureEnabledOnV1Alpha5"
+)
+
+var (
+	commonTestGates = map[featuregate.Feature]featuregate.FeatureSpec{
```
sbueringer, line 41: Should we have some beta features with Default: true here?

feature/gates.go
fabriziopandini: I tried to improve the overall comments around this code / this func, PTAL

config/manager/manager.yaml
fabriziopandini: No, this will break CAPV (this is the manifest we are using for govmomi)

sbueringer: So why are we setting it? :)

feature/gates.go
```diff
@@ -17,17 +17,221 @@ limitations under the License.
...
+	for _, gate := range allowedGates.KnownFeatures() {
+		known.Insert(gate[0:strings.Index(gate, "=")])
+	}
+
+	// Note: the code in the following loop mimics the code that parses the comma separated list of feature gates
+	// in Kubernetes's SetWithLogger func.
+	//
+	// In this case, we parse the comma separated list of feature gates and drop from this list only
+	// unknown, versioned feature gates set to false, while we keep everything else, including
+	// invalid feature gates so that errors will surface when we call allowedGates.Set later in this func.
```
sbueringer, line 143:
```suggestion
	// invalid feature gates so that errors will surface when we call allGates.Set later in this func.
```
I think

feature/gates.go
sbueringer: Thank you very much! Much easier to understand now

sbueringer:
One nit + https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/4009#discussion_r3272329621 PTAL

feature/gates.go
fabriziopandini: allowedGates is correct.

We first call allowedGates.Set in L186, and we check error from this call (while instead we call allGates.Set in L220 only if there are no errors in the first call)

config/manager/manager.yaml
fabriziopandini: Oh, I was looking at the feature flags because the diff was highlighting this.
We can drop the vm-operator-api manifest, but when running in govmomi mode it is just ignored

feature/gates.go
sbueringer: Ups, yup makes sense.

sbueringer:
Very nice!

Thank you very much, also for the extensive godocs!

/lgtm
/approve

PR #4023, opened 2026-05-28
🌱 Add KCP adoption test

fabriziopandini:

Fixes #

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -316,10 +316,12 @@ var _ = SynchronizedAfterSuite(func() {
 		switch testTarget {
 		case VCenterTestTarget:
 			// Cleanup the in cluster address manager
-			vSphereFolderName := e2eConfig.MustGetVariable("VSPHERE_FOLDER")
-			err := inClusterAddressManager.Teardown(ctx, vsphereip.MachineFolder(vSphereFolderName), vsphereip.VSphereClient(vsphereClient))
-			if err != nil {
-				Byf("Ignoring Teardown error: %v", err)
+			if inClusterAddressManager != nil {
+				vSphereFolderName := e2eConfig.MustGetVariable("VSPHERE_FOLDER")
+				err := inClusterAddressManager.Teardown(ctx, vsphereip.MachineFolder(vSphereFolderName), vsphereip.VSphereClient(vsphereClient))
+				if err != nil {
+					Byf("Ignoring Teardown error: %v", err)
+				}
 			}
 
 		case VCSimTestTarget:
--- a/test/e2e/kcp_infra_adoption_test.go
+++ b/test/e2e/kcp_infra_adoption_test.go
@@ -1,0 +1,485 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strconv"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	vmoprv1alpha5 "github.com/vmware-tanzu/vm-operator/api/v1alpha5"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/utils/ptr"
+	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	runtimev1 "sigs.k8s.io/cluster-api/api/runtime/v1beta2"
+	clusterctlv1 "sigs.k8s.io/cluster-api/cmd/clusterctl/api/v1alpha3"
+	"sigs.k8s.io/cluster-api/cmd/clusterctl/client/config"
+	"sigs.k8s.io/cluster-api/cmd/clusterctl/client/repository"
+	capi_e2e "sigs.k8s.io/cluster-api/test/e2e"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
+	"sigs.k8s.io/cluster-api/util/secret"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	vmwarev1 "sigs.k8s.io/cluster-api-provider-vsphere/api/supervisor/v1beta2"
+	conversionapi "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api"
+	vmoprvhub "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api/vmoperator/hub"
+	conversionclient "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/client"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/vmoperator"
+	vsphereframework "sigs.k8s.io/cluster-api-provider-vsphere/test/framework"
+)
+
+const (
+	kcpAdoptionSpecName       = "kcp-infra-adoption" // copied from CAPI
+	extensionServiceNamespace = "capv-test-extension"
+	extensionServiceName      = "capv-test-extension-webhook-service"
+)
+
+var _ = Describe("When testing KCP infra adoption 1CP [vcsim] [supervisor]", func() {
+	Setup(kcpAdoptionSpecName+"-1-cp", KCPInfraAdoptionTest(kcpAdoptionSpecName+"-1-cp", 1), SkipIf(testMode == GovmomiTestMode))
+})
+
+var _ = Describe("When testing KCP infra adoption 3CP [vcsim] [supervisor]", func() {
+	Setup(kcpAdoptionSpecName+"-3-cp", KCPInfraAdoptionTest(kcpAdoptionSpecName+"-3-cp", 3), SkipIf(testMode == GovmomiTestMode))
+})
+
+func KCPInfraAdoptionTest(specName string, controlPlaneMachineCount int64) func(testSpecificSettingsGetter func() testSettings) {
+	return func(testSpecificSettingsGetter func() testSettings) {
+		// Use stable vm-operator API version or the API version defined in the VM_OPERATOR_API_VERSION env var.
+		// Note: e2eConfig is null during initial ginkgo traversal.
+		apiVersionVMOperator := vmoprv1alpha5.GroupVersion.Version
+		if e2eConfig != nil {
+			if v := e2eConfig.GetVariableOrEmpty("VM_OPERATOR_API_VERSION"); v != "" {
+				apiVersionVMOperator = v
+			}
+		}
+
+		if e2eConfig != nil && testTarget == VCenterTestTarget {
+			Expect(e2eConfig.Variables).To(HaveKey("DOCKER_IMAGE_TAR"))
+		}
+
+		// Setup variables to be used later in the test
+		var virtualMachineSetResourcePolicy *vmoprvhub.VirtualMachineSetResourcePolicy
+		var virtualMachines []*vmoprvhub.VirtualMachine
+		var workloadClusterProxy framework.ClusterProxy
+		converter := conversionapi.DefaultConverterFor(
+			schema.GroupVersion{Group: vmoprvhub.GroupVersion.Group, Version: apiVersionVMOperator},
+		)
+
+		capi_e2e.KCPInfraAdoptionSpec(ctx, func() capi_e2e.KCPInfraAdoptionSpecInput {
+			return capi_e2e.KCPInfraAdoptionSpecInput{
+				E2EConfig:                e2eConfig,
+				ClusterctlConfigPath:     testSpecificSettingsGetter().ClusterctlConfigPath,
+				BootstrapClusterProxy:    bootstrapClusterProxy,
+				ArtifactFolder:           artifactFolder,
+				SkipCleanup:              skipCleanup,
+				Flavor:                   ptr.To(testSpecificSettingsGetter().FlavorForMode("topology-runtimesdk-self-hosted")),
+				PostNamespaceCreated:     testSpecificSettingsGetter().PostNamespaceCreatedFunc,
+				ControlPlaneMachineCount: new(controlPlaneMachineCount),
+				BeforeForceDelete: func(ctx context.Context, managementClusterProxy framework.ClusterProxy, cluster *clusterv1.Cluster, objects capi_e2e.ClusterObjects) error {
+					By("Saving objects required for adoption of a CAPV Cluster")
+
+					// Create a client to get read from the management cluster.
+					c, err := conversionclient.NewWithConverter(managementClusterProxy.GetClient(), converter)
+					if err != nil {
+						return err
+					}
+
+					// Drop owner reference from VirtualMachineSetResourcePolicy so it sticks around when force deleting the cluster and cluster objects.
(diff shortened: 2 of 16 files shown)


packaging/go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	k8s.io/apiextensions-apiserver v0.36.2
 	k8s.io/apimachinery v0.36.2
 	k8s.io/utils v0.0.0-20260210185600-b8788abfbbc2
-	sigs.k8s.io/cluster-api v0.0.0-00010101000000-000000000000
+	sigs.k8s.io/cluster-api v1.13.0-rc.0.0.20260527155447-3e41a576ab2e
```
sbueringer: Let's keep this at 000..

packaging/go.mod
```diff
@@ -2,7 +2,7 @@ module sigs.k8s.io/cluster-api-provider-vsphere/packaging
 
 go 1.26.0
 
-replace sigs.k8s.io/cluster-api => sigs.k8s.io/cluster-api v1.13.0-rc.0.0.20260526071518-a4a42bdc39a8
+replace sigs.k8s.io/cluster-api => sigs.k8s.io/cluster-api v1.13.0-rc.0.0.20260527155447-3e41a576ab2e
```
sbueringer: Why are we picking this specific commit from 27th May?

EDIT: Ah is that to pick up a first change to the KCP adoption test and then we did another bump later?

Maybe let's just simplify and bump all to the same version?

(I know I said a few days let's just bump test, but maybe we don't want to have a full and a partial bump in one PR? up to you. I can also fix that up in my next bump PR)

test/extension/handlers/topologymutation/handler.go
```diff
@@ -498,7 +498,19 @@ func patchSupervisorMachineTemplate(_ context.Context, vsphereMachineTemplate ru
 		vsphereMachineTemplate.Spec.Template.Spec.ImageName = imageName
 	}
 
-	return err
+	className, err := topologymutation.GetStringVariable(templateVariables, "virtualMachineClass")
+	if err != nil || className == "" {
+		return nil //nolint:nilerr
+	}
```
sbueringer:
```suggestion
	if err != nil && !topologymutation.IsNotFoundError(err) {
		return errors.Wrap(err, "could not set diskGiB")
	}
	if className == "" {
		return nil
	}
```

test/go.mod
```diff
@@ -35,9 +35,9 @@ require (
 	k8s.io/component-base v0.36.2
 	k8s.io/klog/v2 v2.140.0
 	k8s.io/utils v0.0.0-20260210185600-b8788abfbbc2
-	sigs.k8s.io/cluster-api v0.0.0-00010101000000-000000000000
+	sigs.k8s.io/cluster-api v1.13.0-rc.0.0.20260527155447-3e41a576ab2e
 	sigs.k8s.io/cluster-api-provider-vsphere v0.0.0-00010101000000-000000000000
-	sigs.k8s.io/cluster-api/test v0.0.0-00010101000000-000000000000
+	sigs.k8s.io/cluster-api/test v1.13.0-rc.0.0.20260617104334-7988dafd4518
```
sbueringer: Let's keep 0000

go.mod
```diff
@@ -37,7 +37,7 @@ require (
 	k8s.io/component-base v0.36.2
 	k8s.io/klog/v2 v2.140.0
 	k8s.io/utils v0.0.0-20260210185600-b8788abfbbc2
-	sigs.k8s.io/cluster-api v0.0.0-00010101000000-000000000000
+	sigs.k8s.io/cluster-api v1.13.0-rc.0.0.20260527155447-3e41a576ab2e
```
sbueringer: Same here

test/e2e/e2e_suite_test.go
```diff
@@ -316,10 +316,12 @@ var _ = SynchronizedAfterSuite(func() {
 		switch testTarget {
 		case VCenterTestTarget:
 			// Cleanup the in cluster address manager
-			vSphereFolderName := e2eConfig.MustGetVariable("VSPHERE_FOLDER")
-			err := inClusterAddressManager.Teardown(ctx, vsphereip.MachineFolder(vSphereFolderName), vsphereip.VSphereClient(vsphereClient))
-			if err != nil {
-				Byf("Ignoring Teardown error: %v", err)
+			if inClusterAddressManager != nil {
```
sbueringer, line 319: Just to double check. This is needed because now sometimes with the vCenter target we don't have this IP address manager?

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,490 @@
...
+	conversionapi "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api"
+	vmoprvhub "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/api/vmoperator/hub"
+	conversionclient "sigs.k8s.io/cluster-api-provider-vsphere/pkg/conversion/client"
+	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services/vmoperator"
+	vsphereframework "sigs.k8s.io/cluster-api-provider-vsphere/test/framework"
+)
+
+const (
+	kcpAdoptionSpecName             = "kcp-infra-adoption" // copied from CAPI
+	kcpAdoptionSExtensionConfigName = kcpAdoptionSpecName
```
sbueringer:
```suggestion
	kcpAdoptionExtensionConfigName = kcpAdoptionSpecName
```

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,490 @@
...
+	Setup(kcpAdoptionSpecName, KCPInfraAdoptionTest(3), SkipIf(testMode == GovmomiTestMode))
+})
+
+func KCPInfraAdoptionTest(controlPlaneMachineCount int64) func(testSpecificSettingsGetter func() testSettings) {
+	return func(testSpecificSettingsGetter func() testSettings) {
+		if testTarget == VCenterTestTarget {
+			Expect(os.Getenv("DOCKER_IMAGE_TAR")).ToNot(BeEmpty())
+		}
+
+		// Use latest vm-operator API version or the API version defined in the VM_OPERATOR_API_VERSION env var.
```
sbueringer: nit: latest is probably the wrong word (latest is v1alpha6, maybe latest stable?)

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,490 @@
...
+		// Note: e2eConfig is null during initial ginkgo traversal.
+		apiVersionVMOperator := vmoprv1alpha5.GroupVersion.Version
+		if e2eConfig != nil {
+			if v := e2eConfig.GetVariableOrEmpty("VM_OPERATOR_API_VERSION"); v != "" {
```
sbueringer, line 72: When is this code executed if e2eConfig is null during ginkgo traversal? (comment in l.75)

Is this executed multiple times?

Would be good to extend the godoc so it's easier to understand when we're ending up with v1alpha5 and when with the variable value

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,490 @@
...
+						return err
+					}
+
+					// Drop owner reference from VirtualMachineSetResourcePolicy so it sticks around when force deleting the cluster and cluster objects.
+					// "Orphaned" VirtualMachineSetResourcePolicy, will be re-adopted at a later stage of the test.
+					virtualMachineSetResourcePolicy = &vmoprvhub.VirtualMachineSetResourcePolicy{}
+					if err := c.Get(ctx, ctrlclient.ObjectKey{Namespace: cluster.Namespace, Name: cluster.Name}, virtualMachineSetResourcePolicy); err != nil {
+						return err
+					}
+
```
sbueringer: nit: Let's drop this new line here and the other one in l.123 so the godoc comment "applies" to the whole thing and it's easier to correlate

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,490 @@
...
+					//	This is not a problem for the boostrap management cluster, because as soon as images are built locally, they are accessible
+					//	from the management cluster as well. Instead, making images available on the VMs being adopted running on GVCE
+					//	requires to load the images into the VMs (required images are made available in a tar file created before the test starts).
+					//
+					//	Please note that when running the test locally this problem might be even more complex, because boostrap management cluster
+					//	and the self-hosted cluster on the VMs might be running on different platforms (e.g. arm64 for the boostrap management cluster, amd64 for the VMs on GVCE).
+					//
+					//	One way to get around this is the following:
+					//
+					//	1. Prepare the images that should be deployed on the VMs on GVCE locally, tacking care of building for the target architecture.
```
sbueringer:
```suggestion
					//	1. Prepare the images that should be deployed on the VMs on GVCE locally, taking care of building for the target architecture.
```

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,490 @@
...
+					//	ARCH=amd64 make e2e-images
+					//
+					//	2. Revert the change on manifests that are performed by make e2e-images.
+					//
+					//	3. Tag the images to match the default name existing in the manifests:
+					//
+					//	docker tag "gcr.io/k8s-staging-capi-vsphere/cluster-api-vsphere-controller-amd64:dev" "gcr.io/k8s-staging-capi-vsphere/cluster-api-vsphere-controller:main"
+					//	docker tag "gcr.io/k8s-staging-capi-vsphere/cluster-api-vcsim-controller-amd64:dev" "gcr.io/k8s-staging-capi-vsphere/cluster-api-vcsim-controller:main"
+					//	docker tag "gcr.io/k8s-staging-capi-vsphere/cluster-api-net-operator-amd64:dev" "gcr.io/k8s-staging-capi-vsphere/cluster-api-net-operator:main"
+					//	docker tag "gcr.io/k8s-staging-capi-vsphere/cluster-api-vsphere-test-extension-amd64:dev" "gcr.io/k8s-staging-capi-vsphere/cluster-api-vsphere-test-extension:dev"
```
sbueringer, line 211: Just to check. It's correct that test extension has dev tag vs the others have main?

I'm a bit surprised as the e2e.sh is using :dev for all images for docker save

EDIT: Ah probably this is because we revert the change on manifests and there is some diff between test-extension and all the other ones? (fine for me to not fix that up, just want to understand it)

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,490 @@
...
+					//		 > /tmp/images/image.tar
+					//
+					//	5. tilt-up the boostrap management cluster; tilt will take care of replacing default image names
+					//     with temporary image names corresponding to images build for the boostrap management cluster architecture.
+					//     Important! the tilt-settings file must use the same vm-op-version used in CI.
+					//
+					//	6. Run the E2E with -e2e.use-existing-cluster=true, all the other settings required for running tests on GVCE, and also
+					//     the following env variable
+					//
+					//	DOCKER_IMAGE_TAR=/tmp/images/image.tar
```
sbueringer: I saw you added this also to vsphere.yaml. But looks like we consume it only with os.Getenv.

With the current state adding it in vsphere.yaml doesn't have any effect.

Alternatively we can get rid of os.GetEnv, read it from the e2e config and then we don't have to set the env var here.

I think I would change this to read for e2e config (also for consistency with everything else) and then pass the file location into LoadImagesFunc

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,490 @@
...
+						return nil, err
+					}
+					for _, s := range secrets.Items {
+						if !secret.HasPurposeSuffix(s.Name) {
+							continue
+						}
+						s.SetResourceVersion("")
+						s.SetUID("")
+						s.SetLabels(map[string]string{clusterv1.ClusterNameLabel: cluster.Name})
+						s.SetAnnotations(map[string]string{})
```
sbueringer, line 273: I'm a bit surprised that we overwrite the labels/annotations. Why not just keep them the same as in the mgmt cluster?

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,490 @@
...
+					if err != nil {
+						return nil, err
+					}
+
+					// Create the VirtualMachineSetResourcePolicy.
+					// Note: Cluster and vSphereCluster are recreated by KCPInfraAdoptionSpec; once vSphereCluster is created, it will adopt the VirtualMachineSetResourcePolicy.
+					virtualMachineSetResourcePolicy.SetResourceVersion("")
+					virtualMachineSetResourcePolicy.SetUID("")
+					virtualMachineSetResourcePolicy.SetLabels(map[string]string{})
+					virtualMachineSetResourcePolicy.SetAnnotations(map[string]string{})
```
sbueringer, line 292: Same here, why do we set labels/annotations to empty?

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,490 @@
...
+								continue
+							}
+							return nil, err
+						}
+					}
+
+					return workloadClusterProxy, nil
+				},
+				// The runtime extension gets deployed to the test-extension-system namespace and is exposed
+				// by the test-extension-webhook-service.
```
sbueringer: test-extension-system & test-extension-webhook-service are incorrect

packaging/go.mod
fabriziopandini: It gets reverted by go mod tidy

packaging/go.mod
fabriziopandini: Bumped all to the same version

test/go.mod
fabriziopandini: It gets reverted by go mod tidy

go.mod
fabriziopandini: It gets reverted by go mod tidy

test/e2e/kcp_infra_adoption_test.go
fabriziopandini: I think that this way it is more similar to what a user will do (put the minimal amount of stuff required), so the test can provide a stronger signal on how reliable adoption is.

packaging/go.mod
sbueringer: Never had this problem :) Also looks like it's solved now

test/infrastructure/vm-operator/config/9.2/kustomization.yaml
```diff
@@ -23,6 +23,9 @@ resources:
 - external-crds/cns.vmware.com_storagepolicyquotas.yaml
 - external-crds/cns.vmware.com_storagepolicyusages.yaml
```
sbueringer: Is it enough to only add this for 9.2?

test/e2e/kcp_infra_adoption_test.go
sbueringer: Got it

test/e2e/kcp_infra_adoption_test.go
```diff
@@ -0,0 +1,485 @@
...
+					}
+
+					// Create VirtualMachines.
+					// Note: Machines, KubeadmConfig and vSphereMachines are recreated by KCPInfraAdoptionSpec; once vSphereMachines are created, it will adopt the VirtualMachines.
+					// Important: VirtualMachines must have the ClusterSelectorKey label otherwise they are filter out the cache. Also, vm-operator
+					// assumes that a few labels or annotations are immutable, so preserving all the label/annotations except pause.
+					for _, virtualMachine := range virtualMachines {
+						virtualMachine := virtualMachine.DeepCopy()
+						virtualMachine.SetResourceVersion("")
+						virtualMachine.SetUID("")
```
sbueringer, line 306: I understood the explanation why we are dropping almost all labels/annotations for the objects above.

Should we do the same for VirtualMachine?

Heads up: cache config
```go
	req, _ = labels.NewRequirement(vmoperator.ClusterSelectorKey, selection.Exists, nil)
	virtualMachineCacheSelector := labels.NewSelector().Add(*req)
```

test/infrastructure/vm-operator/config/9.2/kustomization.yaml
sbueringer: I think most tests are using vm-op 9.1? (or at least they should)

fabriziopandini: No, I was just testing.
Now added for all the supported versions

test/e2e/kcp_infra_adoption_test.go
fabriziopandini: I have tried to explain why I'm doing different here in the note right above this for loop

> Important: VirtualMachines must have the ClusterSelectorKey label otherwise they are filter out the cache. Also, vm-operator
> assumes that a few labels or annotations are immutable, so preserving all the label/annotations except pause.

sbueringer: Oh sorry, missed that

PR #4026, opened 2026-06-02
🐛 Fix mismatched cluster identity usage in deploymentZoneToCluster watch handler

### LINKED ISSUE #4025: Mismatched cluster identity usage for VSphereDeploymentZone reconciliations
/kind bug

When deploying multiple VSphereCluster resources that point to the same vCenter server but have different identities and failure domains, the deploymentZoneToCluster watch handler enqueues clusters too broadly based only on Server field matching. This causes clusters with mismatched identities to attempt validation of deployment zones using incorrect credentials.

Scenario:
1. Created Cluster A with identity A@vsphere.local accessing datacenter DA
2. Created Cluster B with identity B@vsphere.local accessing datacenter DB
3. Both clusters point to the same vCenter server: myvcenter.local
4. Created VSphereDeploymentZone AZ1 with failureDomain selector matching Cluster B
5. When the DeploymentZone is created/updated, both Cluster A and Cluster B are reconciled
6. Cluster A attempts to validate the zone using its own credentials (A@vsphere.local)
7. Controller logs show: `Failed to get datacenter, will logout: datacenter '/DB' not found`

The root cause is that `deploymentZoneToCluster()` in `controllers/vspherecluster_reconciler.go` matches zones to clusters only by `Spec.Server` field, without checking if the cluster's failure domain selector actually matches the deployment zone's labels.

Only clusters whose failure domain selector labels match the VSphereDeploymentZone's labels should be enqueued for reconciliation. Cluster A should not be reconciled when Cluster B's deployment zone changes, since Cluster A's failure domain selector doesn't match the zone's labels.

The fix requires modifying the `deploymentZoneToCluster` function to:
1. Check if the cluster has a failure domain selector defined
2. Convert the selector to a label selector
3. Verify that the deployment zone's labels match the cluster's failure domain selector before enqueuing

This prevents wrong cluster identities from being used to validate deployment zones that belong to different clusters on the same vCenter.

**Environment:**

- Cluster-api-provider-vsphere version: v1.15.2
- Kubernetes version: 1.34
- OS (e.g. from `/etc/os-release`): Ubuntu 22.04

arno-pons:

This PR fixes an issue where the `deploymentZoneToCluster` watch handler enqueues clusters too broadly based only on vCenter server matching, without verifying that the cluster's failure domain selector actually matches the deployment zone's labels. This causes clusters with different identities and different datacenters but pointing to the same vCenter to attempt validation of deployment zones using incorrect credentials.

**Problem**:
When multiple VSphereCluster resources point to the same vCenter server but have different identities and access different datacenters, the watch handler would enqueue ALL clusters sharing that server when a VSphereDeploymentZone changes. This results in:
- Cluster A (with identity A@vsphere.local) being reconciled for Cluster B's deployment zone
- Cluster A attempting to validate the zone using credentials that don't have access to Cluster B's datacenter
- Controller errors like: `Failed to get datacenter, will logout: datacenter '/DB' not found`

**Solution**:
Enhanced the `deploymentZoneToCluster` function to:
1. Check if the cluster has a failure domain selector defined
2. Convert the selector to a label selector
3. Verify that the deployment zone's labels match the cluster's failure domain selector before enqueuing
4. Added comprehensive logging for debugging selector matching

Now only clusters whose failure domain selector labels match the deployment zone's labels are enqueued for reconciliation.

Fixes #4025

### DIFF
--- a/controllers/vspherecluster_reconciler.go
+++ b/controllers/vspherecluster_reconciler.go
@@ -28,6 +28,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
@@ -309,7 +310,7 @@ func (r *clusterReconciler) reconcileNormal(ctx context.Context, clusterCtx *cap
 	// Reconcile cluster modules.
 	err = r.reconcileVCenterVersion(clusterCtx, vcenterSession)
 	if err != nil || clusterCtx.VSphereCluster.Status.VCenterVersion == "" {
-		deprecatedv1beta1conditions.MarkFalse(clusterCtx.VSphereCluster, infrav1.ClusterModulesAvailableV1Beta1Condition, infrav1.MissingVCenterVersionV1Beta1Reason, clusterv1.ConditionSeverityWarning, "%v", err)
+		deprecatedv1beta1conditions.MarkFalse(clusterCtx.VSphereCluster, infrav1.ClusterModulesAvailableV1Beta1Condition, infrav1.MissingVCenterVersionV1Beta1Reason, clusterv1.ConditionSeverityWarning,[...]
 		conditions.Set(clusterCtx.VSphereCluster, metav1.Condition{
 			Type:    infrav1.VSphereClusterClusterModulesReadyCondition,
 			Status:  metav1.ConditionFalse,
@@ -321,7 +322,7 @@ func (r *clusterReconciler) reconcileNormal(ctx context.Context, clusterCtx *cap
 
 	affinityReconcileResult, err := r.reconcileClusterModules(ctx, clusterCtx)
 	if err != nil {
-		deprecatedv1beta1conditions.MarkFalse(clusterCtx.VSphereCluster, infrav1.ClusterModulesAvailableV1Beta1Condition, infrav1.ClusterModuleSetupFailedV1Beta1Reason, clusterv1.ConditionSeverityWarning, "%v", err)
+		deprecatedv1beta1conditions.MarkFalse(clusterCtx.VSphereCluster, infrav1.ClusterModulesAvailableV1Beta1Condition, infrav1.ClusterModuleSetupFailedV1Beta1Reason, clusterv1.ConditionSeverityWarni[...]
 		conditions.Set(clusterCtx.VSphereCluster, metav1.Condition{
 			Type:    infrav1.VSphereClusterClusterModulesReadyCondition,
 			Status:  metav1.ConditionFalse,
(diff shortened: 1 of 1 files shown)


controllers/vspherecluster_reconciler.go
```diff
@@ -309,7 +310,7 @@ func (r *clusterReconciler) reconcileNormal(ctx context.Context, clusterCtx *cap
 	// Reconcile cluster modules.
 	err = r.reconcileVCenterVersion(clusterCtx, vcenterSession)
 	if err != nil || clusterCtx.VSphereCluster.Status.VCenterVersion == "" {
-		deprecatedv1beta1conditions.MarkFalse(clusterCtx.VSphereCluster, infrav1.ClusterModulesAvailableV1Beta1Condition, infrav1.MissingVCenterVersionV1Beta1Reason, clusterv1.ConditionSeverityWarning, "%v", err)
+		deprecatedv1beta1conditions.MarkFalse(clusterCtx.VSphereCluster, infrav1.ClusterModulesAvailableV1Beta1Condition, infrav1.MissingVCenterVersionV1Beta1Reason, clusterv1.ConditionSeverityWarning,[...]
```
sbueringer, line 313: This is not valid go, same below

controllers/vspherecluster_reconciler.go
```diff
@@ -586,6 +587,10 @@ func (r *clusterReconciler) controlPlaneMachineToCluster(ctx context.Context, o
 	}}
 }
 
+// deploymentZoneToCluster is a handler.ToRequestsFunc to be used to enqueue requests
+// for reconciliation for VSphereCluster when a VSphereDeploymentZone is changed.
+// Only clusters that match both the deployment zone's server and have a failure domain selector
+// that matches the deployment zone's labels will be enqueued.
 func (r *clusterReconciler) deploymentZoneToCluster(ctx context.Context, o client.Object) []ctrl.Request {
```
sbueringer, line 594: Modifying this function will only influence which VSphereClusters are getting reconciled when the VSphereDeploymentZone is modified.

Independent of that VSphereCluster's are reconciled at least every syncPeriod (10m).

Reconciles have to work independent of why they are triggered (the Reconcile func has no knowledge about why a Reconcile was triggered). So I don't think that changing this func will fix your issue.

sbueringer: Where exactly is the `Failed to get datacenter, will logout: datacenter '/DB' not found` error coming from?

Can you provide the full logs? (obfuscation is fine of course)

arno-pons: Hello @sbueringer;

Sorry for the delay, here an example of a full error we have:
```shell
capv-controller-manager-84b8bcf65c-wfc8z manager E0602 09:46:22.720396       1 session.go:211] "Failed to get datacenter, will logout" err="datacenter '/xxx' not found" controller="vspheredeploymentzone" controllerGroup="[infrastructure.cluster.x-k8s.io](http://infrastructure.cluster.x-k8s.io/)" controllerKind="VSphereDeploymentZone" VSphereDeploymentZone="clusterA-az" namespace="" name="clusterA-az" reconcileID="2553c62f-4b19-496a-a9db-b4ee5eff0fa4" VSphereFailureDomain="clusterA-az" VSphereCluster="clusterB/clusterB-fpjh2" server="[vcenter.url.masked](http://vcenter.url.masked/)" datacenter="/xxx" username="clusterB@vsphere"
```

where `vcenter.url.masked` is the same for both clusters.

For now, we have found a workaround which is to configure a uniq local DNS alias in coredns for each cluster's vcenter endpoint and using it both in az definition and cluster cluster api attribute. By doing this, we are forcing capv to handle each AZ check with the corresponding cluster's identity credentials.

sbueringer: I think the error you are getting occurs in the VSphereDeploymentZone reconciler (see "controller" k/v pair in the log). Specifically in vsphereDeploymentZoneReconciler.getVCenterSession. I assume how we find the credentials for a given VSphereDeploymentZone there is not correct.

That code is not affected by the current change in this PR so it won't fix the issue.

PR #4036, opened 2026-06-08
Add --reconciliation-timeout flag for slow vCenter SSO login

tdakkota:
Adds an option to configure a login timeout for slow-responding vSphere instances.

### DIFF
--- a/main.go
+++ b/main.go
@@ -114,6 +114,7 @@ var (
 	vSphereDeploymentZoneConcurrency  int
 	virtualMachineGroupConcurrency    int
 	skipCRDMigrationPhases            []string
+	reconciliationTimeout             time.Duration
 
 	managerOptions = capiflags.ManagerOptions{}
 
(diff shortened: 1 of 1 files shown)


sbueringer:
How slow is your vCenter login and which controllers are affected?

tdakkota:
vCenter SSO login via an external AD/LDAP account takes 30–60 seconds in our environment. The default CAPV reconciliation timeout is short enough that it fires before the login completes, causing reconciliation failures on first attempt for every resource type. Effectively any controller that calls into vCenter on reconcile is affected.

sbueringer:
> vCenter SSO login via an external AD/LDAP account takes 30–60 seconds in our environment. The default CAPV reconciliation timeout is short enough that it fires before the login completes, causing reconciliation failures on first attempt for every resource type. Effectively any controller that calls into vCenter on reconcile is affected.

I'm okay increasing the timeout, but I would prefer not exposing a low-level option like this.

WDYT about setting the reconciliation timeout hard-coded to 5 or 10 minutes? (please only for the controllers in `setupVAPIControllers`, controllers in `setupSupervisorControllers` do not communicate with vCenter)

(context: in previous versions of CAPV we didn't have a timeout at all)

sbueringer:
/close

Feel free to reopen if there is still interest in this

PR #4061, opened 2026-06-18
🌱 Stop running make lint-api multiple times in GitHub action + align to CAPI

sbueringer:

Fixes #

### DIFF
--- a/.github/workflows/pr-golangci-lint.yaml
+++ b/.github/workflows/pr-golangci-lint.yaml
@@ -8,6 +8,24 @@ on:
 permissions: {}
 
 jobs:
+  golangci-api:
+    name: lint-api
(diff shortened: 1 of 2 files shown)


sbueringer:
/hold 
until the api module PR is merged

.github/workflows/pr-golangci-lint.yaml
```diff
@@ -36,5 +53,3 @@ jobs:
         with:
           version: v2.12.2
           working-directory: ${{matrix.working-directory}}
```
sbueringer, line 55: I think new-from-rev was only needed when we introduced the linter initially.

I think nowadays with our excludes in golangci-kal.yaml we can just run the linter the same way as in CAPI

PR #4072, opened 2026-06-25
✨Add per-cluster network provider support

DanielXiao:
Implement the CAPV portion of the VKS cluster transition (VDS / NSX T1 -> NSX VPC) by introducing per-cluster network provider resolution behind the ClusterNetworkProvider feature gate.
- Add VSphereCluster.spec.network.provider to v1beta1 and v1beta2 with enum validation (vsphere-distributed, nsx-tier1, vpc) and a CEL immutability rule.
- Add NetworkProviderFactory with per-cluster and static implementations.
- Resolve the network provider lazily per reconcile in controllers; requeue when spec.network.provider is empty.
- Resolve the provider per-cluster in VSphereMachine and VSphereMachineTemplate webhooks when the gate is enabled; skip provider-based checks on VSphereClusterTemplate.
- Move control-plane VM readiness probe configuration to per-cluster machine context instead of manager startup.
- Rename pkg/manager network provider constant values to the new names (vpc, nsx-tier1, vsphere-distributed).
With the gate disabled, behavior is unchanged: a static factory backed by --network-provider drives all resolution.

Fixes #

### DIFF
--- a/controllers/vmware/servicediscovery_controller.go
+++ b/controllers/vmware/servicediscovery_controller.go
@@ -18,6 +18,7 @@ package vmware
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/url"
@@ -76,17 +77,12 @@ const (
 // +kubebuilder:rbac:groups="",resources=configmaps/status,verbs=get
 
 // AddServiceDiscoveryControllerToManager adds the ServiceDiscovery controller to the provided manager.
-func AddServiceDiscoveryControllerToManager(ctx context.Context, controllerManagerCtx *capvcontext.ControllerManagerContext, mgr manager.Manager, clusterCache clustercache.ClusterCache, options controller.Options) error {
-	networkProvider, err := inframanager.GetNetworkProvider(ctx, controllerManagerCtx.Client, controllerManagerCtx.NetworkProvider)
-	if err != nil {
-		return pkgerrors.Wrap(err, "failed to create network provider")
-	}
-
+func AddServiceDiscoveryControllerToManager(ctx context.Context, controllerManagerCtx *capvcontext.ControllerManagerContext, mgr manager.Manager, clusterCache clustercache.ClusterCache, options controller.Options, networkProviderFactory inframanager.NetworkProviderFactory) error {
 	r := &serviceDiscoveryReconciler{
-		Client:          controllerManagerCtx.Client,
-		Recorder:        mgr.GetEventRecorderFor("servicediscovery/vspherecluster-controller"),
-		NetworkProvider: networkProvider,
-		clusterCache:    clusterCache,
+		Client:                 controllerManagerCtx.Client,
+		Recorder:               mgr.GetEventRecorderFor("servicediscovery/vspherecluster-controller"),
+		NetworkProviderFactory: networkProviderFactory,
+		clusterCache:           clusterCache,
 	}
 	predicateLog := ctrl.LoggerFrom(ctx).WithValues("controller", "servicediscovery/vspherecluster")
 
@@ -138,9 +134,9 @@ func clusterToSupervisorVSphereClusterFunc(ctrlclient client.Client) func(ctx co
 }
 
 type serviceDiscoveryReconciler struct {
-	Client          client.Client
-	Recorder        record.EventRecorder
-	NetworkProvider services.NetworkProvider
+	Client                 client.Client
+	Recorder               record.EventRecorder
+	NetworkProviderFactory inframanager.NetworkProviderFactory
 
 	clusterCache clustercache.ClusterCache
 }
@@ -206,11 +202,23 @@ func (r *serviceDiscoveryReconciler) Reconcile(ctx context.Context, req reconcil
 		return reconcile.Result{}, nil
 	}
 
+	// Resolve the network provider for the VSphereCluster.
+	np, err := r.NetworkProviderFactory.ForCluster(ctx, vsphereCluster)
+	if err != nil {
+		if errors.Is(err, inframanager.ErrNetworkProviderEmpty) {
+			log.Info("Network Provider is empty, wait for a valid value")
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, pkgerrors.Wrapf(err,
+			"failed to resolve network provider for VSphereCluster %s",
+			klog.KObj(vsphereCluster))
+	}
+
 	// We cannot proceed until we are able to access the target cluster. Until
 	// then just return a no-op and wait for the next sync.
 	guestClient, err := r.clusterCache.GetClient(ctx, client.ObjectKeyFromObject(cluster))
 	if err != nil {
-		if pkgerrors.Is(err, clustercache.ErrClusterNotConnected) {
+		if errors.Is(err, clustercache.ErrClusterNotConnected) {
 			log.V(5).Info("Requeuing because connection to the workload cluster is down")
 			return ctrl.Result{RequeueAfter: time.Minute}, nil
 		}
@@ -222,7 +230,7 @@ func (r *serviceDiscoveryReconciler) Reconcile(ctx context.Context, req reconcil
 	return reconcile.Result{}, r.reconcileNormal(ctx, &vmwarecontext.GuestClusterContext{
 		ClusterContext: clusterContext,
 		GuestClient:    guestClient,
-	})
+	}, np)
 }
 
 func (r *serviceDiscoveryReconciler) patch(ctx context.Context, clusterCtx *vmwarecontext.ClusterContext) error {
@@ -237,8 +245,8 @@ func (r *serviceDiscoveryReconciler) patch(ctx context.Context, clusterCtx *vmwa
 	)
 }
 
-func (r *serviceDiscoveryReconciler) reconcileNormal(ctx context.Context, guestClusterCtx *vmwarecontext.GuestClusterContext) error {
-	if err := r.reconcileSupervisorHeadlessService(ctx, guestClusterCtx); err != nil {
+func (r *serviceDiscoveryReconciler) reconcileNormal(ctx context.Context, guestClusterCtx *vmwarecontext.GuestClusterContext, np services.NetworkProvider) error {
+	if err := r.reconcileSupervisorHeadlessService(ctx, guestClusterCtx, np); err != nil {
 		deprecatedv1beta1conditions.MarkFalse(guestClusterCtx.VSphereCluster, vmwarev1.ServiceDiscoveryReadyV1Beta1Condition, vmwarev1.SupervisorHeadlessServiceSetupFailedV1Beta1Reason,
 			clusterv1.ConditionSeverityWarning, "%v", err)
 		conditions.Set(guestClusterCtx.VSphereCluster, metav1.Condition{
@@ -256,7 +264,7 @@ func (r *serviceDiscoveryReconciler) reconcileNormal(ctx context.Context, guestC
 // reconcileSupervisorHeadlessService sets up a local k8s service in the workload cluster that
 // proxies to the Supervisor Cluster API Server. The add-ons are depend on this local service
 // to connect to the Supervisor Cluster.
-func (r *serviceDiscoveryReconciler) reconcileSupervisorHeadlessService(ctx context.Context, guestClusterCtx *vmwarecontext.GuestClusterContext) error {
+func (r *serviceDiscoveryReconciler) reconcileSupervisorHeadlessService(ctx context.Context, guestClusterCtx *vmwarecontext.GuestClusterContext, np services.NetworkProvider) error {
 	log := ctrl.LoggerFrom(ctx)
 
 	// Create the headless service to the supervisor api server on the target cluster.
@@ -281,7 +289,7 @@ func (r *serviceDiscoveryReconciler) reconcileSupervisorHeadlessService(ctx cont
 
 	var supervisorHosts []string
 	var err error
-	supervisorHosts, err = r.getSupervisorAPIServerAddresses(ctx, guestClusterCtx.Cluster)
+	supervisorHosts, err = r.getSupervisorAPIServerAddresses(ctx, guestClusterCtx.Cluster, np)
 	if err != nil {
 		// Note: We have watches on the LB Svc (VIP) & the cluster-info configmap (FIP).
 		// There is no need to return an error to keep re-trying.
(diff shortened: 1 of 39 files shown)


internal/webhooks/vmware/network_provider.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	clusterName, ok := obj.GetLabels()[clusterv1.ClusterNameLabel]
+	if !ok || clusterName == "" {
+		return "", fmt.Errorf("missing %q label, cannot resolve the owning Cluster", clusterv1.ClusterNameLabel)
+	}
+
+	cluster := &clusterv1.Cluster{}
+	if err := c.Get(ctx, client.ObjectKey{Namespace: obj.GetNamespace(), Name: clusterName}, cluster); err != nil {
+		return "", fmt.Errorf("failed to get Cluster %s/%s: %w", obj.GetNamespace(), clusterName, err)
```
sbueringer:
```suggestion
		return "", fmt.Errorf("failed to get Cluster %s: %w", klog.KRef(obj.GetNamespace(), clusterName), err)
```
(we're trying to standardize on klog.KObj / KRef)

internal/webhooks/vmware/network_provider.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	if !cluster.Spec.InfrastructureRef.IsDefined() {
+		return "", fmt.Errorf("cluster %s/%s does not have a spec.infrastructureRef set", cluster.Namespace, cluster.Name)
```
sbueringer:
```suggestion
		return "", fmt.Errorf("Cluster %s does not have a spec.infrastructureRef set", klog.KObj(cluster))
```
(might need a nolint, but we should use upper case Cluster here)

internal/webhooks/vmware/network_provider.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	vsphereCluster := &vmwarev1.VSphereCluster{}
+	key := client.ObjectKey{Namespace: cluster.Namespace, Name: cluster.Spec.InfrastructureRef.Name}
+	if err := c.Get(ctx, key, vsphereCluster); err != nil {
+		return "", fmt.Errorf("failed to get VSphereCluster %s/%s: %w", key.Namespace, key.Name, err)
```
sbueringer:
```suggestion
		return "", fmt.Errorf("failed to get VSphereCluster %s: %w", klog.KRef(key.Namespace, key.Name), err)
```

internal/webhooks/vmware/network_provider.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	if vsphereCluster.Spec.Network.Provider == "" {
+		return "", fmt.Errorf("VSphereCluster %s/%s spec.network.provider is empty, wait for a valid value", vsphereCluster.Namespace, vsphereCluster.Name)
```
sbueringer:
```suggestion
		return "", fmt.Errorf("VSphereCluster %s spec.network.provider is empty, wait for a valid value", klog.KObj(vsphereCluster))
```

internal/webhooks/vmware/vspherecluster.go
```diff
@@ -74,10 +75,30 @@ func (webhook *VSphereCluster) validateClusterNetwork(cluster *vmwarev1.VSphereC
...
-	if cluster.Spec.Network.NSXVPC.IsDefined() && webhook.NetworkProvider != manager.NSXVPCNetworkProvider {
+
+	// When the ClusterNetworkProvider gate is enabled, the provider to validate against is the
+	// cluster's own spec.network.provider; otherwise it is the static flag value.
+	provider := webhook.NetworkProvider
+	if feature.Gates.Enabled(feature.ClusterNetworkProvider) {
+		provider = cluster.Spec.Network.Provider
+
+		if provider == "" {
+			allErrs = append(allErrs, field.Required(
```
sbueringer, line 89: I would prefer if we return in this error case and the one in l.93 instead of falling through to some further validation

pkg/manager/network.go
```diff
@@ -27,20 +27,55 @@ import (
...
+
+	// Legacy global network provider names (pre-rename values of --network-provider).
+	legacyNSXNetworkProvider    = "NSX"
+	legacyNSXVPCNetworkProvider = "NSX-VPC"
+	legacyVDSNetworkProvider    = "vsphere-network"
+
+	// New global network provider names (used by --network-provider flag).
+	globalNSXNetworkProvider    = "nsx-tier1"
+	globalNSXVPCNetworkProvider = "vpc"
+	globalVDSNetworkProvider    = "vsphere-distributed"
```
sbueringer: Would it make sense to reuse VPC / NSXTier1 / VSphereDistributed here?

controllers/vmware/servicediscovery_controller.go
```diff
@@ -18,6 +18,7 @@ package vmware
 
 import (
 	"context"
+	stderrors "errors"
```
sbueringer: let's change pkg/errors to use the pkgerrors alias and avoid the alias here

controllers/vmware/servicediscovery_controller.go
```diff
@@ -206,6 +202,18 @@ func (r *serviceDiscoveryReconciler) Reconcile(ctx context.Context, req reconcil
 		return reconcile.Result{}, nil
 	}
 
+	// Resolve the network provider for the VSphereCluster.
+	np, err := r.NetworkProviderFactory.ForCluster(ctx, vsphereCluster)
+	if err != nil {
+		if stderrors.Is(err, inframanager.ErrNetworkProviderEmpty) {
+			log.Info("Network Provider is empty, wait for a valid value")
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
sbueringer:
```suggestion
			return reconcile.Result{}, nil
```
We don't need requeue, because we watch VSphereCluster

controllers/vmware/servicediscovery_controller.go
```diff
@@ -206,6 +202,18 @@ func (r *serviceDiscoveryReconciler) Reconcile(ctx context.Context, req reconcil
...
+		}
+		return reconcile.Result{}, errors.Wrapf(err,
```
sbueringer: let's use klog.KObj

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -20,8 +20,10 @@ package vmware
 import (
 	"cmp"
 	"context"
+	stderrors "errors"
```
sbueringer: Same here let's change pkg/errors to pkgerrors alias

controllers/vmware/vspherecluster_reconciler.go
```diff
@@ -141,8 +144,20 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+	np, err := r.NetworkProviderFactory.ForCluster(ctx, clusterContext.VSphereCluster)
+	if err != nil {
+		if stderrors.Is(err, inframanager.ErrNetworkProviderEmpty) {
+			log.Info("Network Provider is empty, wait for a valid value")
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+		return reconcile.Result{}, errors.Wrapf(err,
+			"failed to resolve network provider for VSphereCluster %s/%s",
+			clusterContext.VSphereCluster.Namespace, clusterContext.VSphereCluster.Name)
+	}
```
sbueringer, line 156: Same 2 findings as in servicediscovery controller

controllers/vspheremachine_controller.go
```diff
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	stderrors "errors"
```
sbueringer: Same for the alias here

controllers/vspheremachine_controller.go
```diff
@@ -366,8 +363,29 @@ func (r *machineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_
...
+			return reconcile.Result{}, errors.New("received unexpected SupervisorMachineContext type")
+		}
+		np, err = r.networkProviderFactory.ForCluster(ctx, supervisorMachineCtx.VSphereCluster)
+		if err != nil {
+			if stderrors.Is(err, inframanager.ErrNetworkProviderEmpty) {
+				log.Info("Network Provider is empty, wait for a valid value")
+				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+			return reconcile.Result{}, errors.Wrapf(err,
+				"failed to resolve network provider for VSphereCluster %s/%s",
```
sbueringer: let's use klog.KObj

sbueringer:
/retest

(failures were independent of this PR)

pkg/manager/network.go
DanielXiao: Yes — I'll unify on VPC / NSXTier1 / VSphereDistributed across VKS.

CAPV lands in kubernetes-services first, so during the window before the kubernetes-services per-cluster network provider PR merges, --network-provider can still be the legacy values (NSX / NSX-VPC / vsphere-network). We still need to convert those to PascalCase.

I had also mapped nsx-tier1 / vpc / vsphere-distributed as a safeguard in case WCP switched the flag to those names mid-window and broke downstream pipelines. That's low risk per their plan, so I've removed that mapping (globalNSXNetworkProvider, etc.) and kept only the legacy conversion.

Once the kubernetes-services PR is in, --network-provider will always be PascalCase. I can follow up with a PR to remove the conversion code entirely.

sbueringer: Sounds good!

sbueringer:
Thx!

/lgtm
/assign @fabriziopandini

pkg/manager/network.go
```diff
@@ -27,20 +27,43 @@ import (
...
 
+var legacyToNetworkProviderName = map[string]string{
+	legacyNSXNetworkProvider:    NSXNetworkProvider,
+	legacyNSXVPCNetworkProvider: NSXVPCNetworkProvider,
+	legacyVDSNetworkProvider:    VDSNetworkProvider,
+}
+
+// ConvertNetworkProviderName converts a legacy network provider name to the
+// PascalCase name used by CAPV. Unknown names are returned unchanged.
+func ConvertNetworkProviderName(name string) string {
```
fabriziopandini, line 57: I would expect we need this conversion only in main.go to fix the value in managerOpts.NetworkProvider before calling manager.New.

The rest of the code base IMO should rely on new names only (and not rely on the fallback mechanism in this func). e.g. GetNetworkProvider should not call ConvertNetworkProviderName

pkg/manager/options.go
```diff
@@ -115,6 +115,10 @@ func (o *Options) defaults() {
 	} else {
 		o.PodNamespace = DefaultPodNamespace
 	}
+
+	if o.NetworkProvider != "" {
+		o.NetworkProvider = ConvertNetworkProviderName(o.NetworkProvider)
```
fabriziopandini: support for legacy provider should be scoped to main.go (remove ConvertNetworkProviderName from here)

internal/webhooks/vmware/network_provider.go
```diff
@@ -0,0 +1,75 @@
...
+// --network-provider flag (staticProvider) is returned and no cluster is loaded.
+//
+// When the gate is enabled, the provider is resolved from the owning Cluster's VSphereCluster
+// following the resolution order:
+//  1. read the cluster.x-k8s.io/cluster-name label, load the Cluster, then load the VSphereCluster
+//     via Cluster.spec.infrastructureRef.
+//  2. if the VSphereCluster cannot be loaded -> reject (surface the error).
+//  3. if VSphereCluster.spec.network.provider is empty -> reject.
+//  4. otherwise return the provider value as-is.
+func resolveNetworkProvider(ctx context.Context, c client.Client, staticProvider string, obj metav1.Object) (string, error) {
```
fabriziopandini: I think we should make implementation in webhooks consistent to what we have 
in controllers
 
- Like controllers, web hooks should get in input a network provider factory, not a network provider name 
- Like for controllers, network provider factory should provide everything is needed to deal with network providers, in this case a method for resolving the provider starting from a cluster (Alternative is to get the provider as we usually do, and add to the provider interface .Name() )

On top of consistency between webhook and controllers, at the end of this change, all the logic for resolving network providers/network providers name will be in a single package (currently it is split in two packages)

DanielXiao: Resolved

pkg/manager/network.go
DanielXiao: resolved

pkg/manager/options.go
DanielXiao: Resolved

sbueringer:
/retest

(PTAL at the linter and verify-main)

internal/webhooks/vmware/vspheremachinetemplate.go
```diff
@@ -107,8 +111,13 @@ func (webhook *VSphereMachineTemplate) ValidateUpdate(ctx context.Context, oldOb
 	return webhook.validate(ctx, nil, newObj)
 }
 
-func (webhook *VSphereMachineTemplate) validate(_ context.Context, _, newVSphereMachineTemplate *vmwarev1.VSphereMachineTemplate) (admission.Warnings, error) {
-	allErrs := validateNetwork(webhook.NetworkProvider, newVSphereMachineTemplate.Spec.Template.Spec.Network, field.NewPath("spec", "template", "spec", "network"))
+func (webhook *VSphereMachineTemplate) validate(ctx context.Context, _, newVSphereMachineTemplate *vmwarev1.VSphereMachineTemplate) (admission.Warnings, error) {
+	np, err := webhook.NetworkProviderFactory.ForObject(ctx, webhook.Client, newVSphereMachineTemplate)
```
sbueringer, line 121: Does this work for a VSphereMachineTemplate referenced in a ClusterClass? (IIRC they should not have the cluster name label set)

DanielXiao: Good catch. There is no cluster name label on VSphereMachineTemplate referenced in a ClusterClass. How about skip validateNetwork() if the cluster name label is not present? 

- The VSphereMachineTemplate spec.network is usually not set and is copied and patched by the ClusterClass. 
- The ClusterClass's VSphereMachineTemplate does not belong to any Cluster so there is no need to validate it.

sbueringer: > How about skip validateNetwork() if the cluster name label is not present?

Yup, makes sense

sbueringer:
Thx!

/lgtm
/approve

PR #4096, opened 2026-07-20
✨ Support Externally Managed Network Provider

DanielXiao:
Allow clusters to attach VMs to networks provisioned outside CAPV, without CAPV managing network objects or load balancers, behind the ExternallyManagedProvider feature gate.

Fixes #

### DIFF
--- a/controllers/vmware/servicediscovery_controller.go
+++ b/controllers/vmware/servicediscovery_controller.go
@@ -214,6 +214,21 @@ func (r *serviceDiscoveryReconciler) Reconcile(ctx context.Context, req reconcil
 			klog.KObj(vsphereCluster))
 	}
 
+	// ExternallyManaged (and any provider that does not support the supervisor
+	// headless Service) skips ServiceDiscovery entirely. Mark ready so the skip
+	// is an explicit, observable success. Do this before waiting on guest-cluster
+	// connectivity since we will not touch the guest.
+	if !np.SupportsSupervisorService() {
+		log.Info("Skipping supervisor Service/Endpoints reconciliation; network provider does not support supervisor service")
+		deprecatedv1beta1conditions.MarkTrue(vsphereCluster, vmwarev1.ServiceDiscoveryReadyV1Beta1Condition)
+		conditions.Set(vsphereCluster, metav1.Condition{
+			Type:   vmwarev1.VSphereClusterServiceDiscoveryReadyCondition,
+			Status: metav1.ConditionTrue,
+			Reason: vmwarev1.VSphereClusterServiceDiscoveryReadyReason,
+		})
+		return reconcile.Result{}, nil
+	}
+
 	// We cannot proceed until we are able to access the target cluster. Until
 	// then just return a no-op and wait for the next sync.
 	guestClient, err := r.clusterCache.GetClient(ctx, client.ObjectKeyFromObject(cluster))
--- a/controllers/vmware/servicediscovery_controller.go
+++ b/controllers/vmware/servicediscovery_controller.go
@@ -214,6 +214,20 @@ func (r *serviceDiscoveryReconciler) Reconcile(ctx context.Context, req reconcil
 			klog.KObj(vsphereCluster))
 	}
 
+	// ExternallyManaged (and any provider that does not support the supervisor
+	// headless Service) skips ServiceDiscovery entirely. Mark ready so the skip
+	// is an explicit, observable success. Do this before waiting on guest-cluster
+	// connectivity since we will not touch the guest.
+	if !np.SupportsSupervisorService() {
+		deprecatedv1beta1conditions.MarkTrue(vsphereCluster, vmwarev1.ServiceDiscoveryReadyV1Beta1Condition)
+		conditions.Set(vsphereCluster, metav1.Condition{
+			Type:   vmwarev1.VSphereClusterServiceDiscoveryReadyCondition,
+			Status: metav1.ConditionTrue,
+			Reason: vmwarev1.VSphereClusterServiceDiscoveryReadyReason,
+		})
+		return reconcile.Result{}, nil
+	}
+
 	// We cannot proceed until we are able to access the target cluster. Until
 	// then just return a no-op and wait for the next sync.
 	guestClient, err := r.clusterCache.GetClient(ctx, client.ObjectKeyFromObject(cluster))
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -57,6 +57,12 @@ const (
 	// alpha: v1.17
 	ClusterNetworkProvider featuregate.Feature = "ClusterNetworkProvider"
 
+	// ExternallyManagedProvider gates setting VSphereCluster.spec.network.provider to
+	// ExternallyManaged and registering the ExternallyManaged network provider.
+	//
+	// alpha: v1.17
+	ExternallyManagedProvider featuregate.Feature = "ExternallyManagedProvider"
+
 	// InfrastructurePolicies is a feature gate for the Support for Supervisor infrastructure policies.
 	// When enabled, VSphereMachine.spec.policies are validated on admission and
 	// mapped to the underlying VirtualMachine spec.
@@ -100,10 +106,11 @@ var (
 	}
 
 	supervisorGates = map[featuregate.Feature]featuregate.FeatureSpec{
-		NamespaceScopedZones:   {Default: false, PreRelease: featuregate.Alpha},
-		NodeAutoPlacement:      {Default: false, PreRelease: featuregate.Alpha},
-		ClusterNetworkProvider: {Default: false, PreRelease: featuregate.Alpha},
-		MultiNetworks:          {Default: false, PreRelease: featuregate.Alpha},
+		NamespaceScopedZones:      {Default: false, PreRelease: featuregate.Alpha},
+		NodeAutoPlacement:         {Default: false, PreRelease: featuregate.Alpha},
+		ClusterNetworkProvider:    {Default: false, PreRelease: featuregate.Alpha},
+		ExternallyManagedProvider: {Default: false, PreRelease: featuregate.Alpha},
+		MultiNetworks:             {Default: false, PreRelease: featuregate.Alpha},
 	}
 
 	supervisorVersionedGates = map[featuregate.Feature]featuregate.VersionedSpecs{
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -54,8 +54,14 @@ const (
 	// for supervisor. When enabled, the network provider is resolved per-cluster from
 	// VSphereCluster.spec.network.provider instead of the static --network-provider flag.
 	//
-	// alpha: v1.17
+	// alpha: v1.18
 	ClusterNetworkProvider featuregate.Feature = "ClusterNetworkProvider"
+
+	// ExternallyManagedProvider gates setting VSphereCluster.spec.network.provider to
+	// ExternallyManaged and registering the ExternallyManaged network provider.
+	//
+	// alpha: v1.18
+	ExternallyManagedProvider featuregate.Feature = "ExternallyManagedProvider"
 
 	// InfrastructurePolicies is a feature gate for the Support for Supervisor infrastructure policies.
 	// When enabled, VSphereMachine.spec.policies are validated on admission and
@@ -100,10 +106,11 @@ var (
 	}
 
 	supervisorGates = map[featuregate.Feature]featuregate.FeatureSpec{
-		NamespaceScopedZones:   {Default: false, PreRelease: featuregate.Alpha},
-		NodeAutoPlacement:      {Default: false, PreRelease: featuregate.Alpha},
-		ClusterNetworkProvider: {Default: false, PreRelease: featuregate.Alpha},
-		MultiNetworks:          {Default: false, PreRelease: featuregate.Alpha},
+		NamespaceScopedZones:      {Default: false, PreRelease: featuregate.Alpha},
+		NodeAutoPlacement:         {Default: false, PreRelease: featuregate.Alpha},
+		ClusterNetworkProvider:    {Default: false, PreRelease: featuregate.Alpha},
+		ExternallyManagedProvider: {Default: false, PreRelease: featuregate.Alpha},
+		MultiNetworks:             {Default: false, PreRelease: featuregate.Alpha},
 	}
 
 	supervisorVersionedGates = map[featuregate.Feature]featuregate.VersionedSpecs{
--- a/pkg/manager/network.go
+++ b/pkg/manager/network.go
@@ -19,7 +19,6 @@ package manager
 import (
 	"context"
 
-	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/cluster-api-provider-vsphere/pkg/services"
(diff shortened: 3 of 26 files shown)


controllers/vmware/servicediscovery_controller.go
```diff
@@ -214,6 +214,21 @@ func (r *serviceDiscoveryReconciler) Reconcile(ctx context.Context, req reconcil
 			klog.KObj(vsphereCluster))
 	}
 
+	// ExternallyManaged (and any provider that does not support the supervisor
+	// headless Service) skips ServiceDiscovery entirely. Mark ready so the skip
+	// is an explicit, observable success. Do this before waiting on guest-cluster
+	// connectivity since we will not touch the guest.
+	if !np.SupportsSupervisorService() {
+		log.Info("Skipping supervisor Service/Endpoints reconciliation; network provider does not support supervisor service")
```
sbueringer: I think if we log this on every reconcile this will produce a lot of logs (see the list of watches above).

I think it's okay to drop the log, I think it's unlikely we need it for debugging

feature/feature.go
```diff
@@ -57,6 +57,12 @@ const (
 	// alpha: v1.17
 	ClusterNetworkProvider featuregate.Feature = "ClusterNetworkProvider"
 
+	// ExternallyManagedProvider gates setting VSphereCluster.spec.network.provider to
+	// ExternallyManaged and registering the ExternallyManaged network provider.
+	//
+	// alpha: v1.17
+	ExternallyManagedProvider featuregate.Feature = "ExternallyManagedProvider"
```
sbueringer, line 64: Just to double check that I got it right.

This feature also works with v1alpha2 and v1alpha5 as there are no code paths that directly require v1alpha5 or v1alpha6

It only depends on other feature gates that also don't depend on a specific vm-op apiVersion (e.g. ClusterNetworkProvider)

The code path for IPAM modes is behind the IPv6DualStack feature gate.

feature/feature.go
```diff
@@ -57,6 +57,12 @@ const (
 	// alpha: v1.17
 	ClusterNetworkProvider featuregate.Feature = "ClusterNetworkProvider"
 
+	// ExternallyManagedProvider gates setting VSphereCluster.spec.network.provider to
+	// ExternallyManaged and registering the ExternallyManaged network provider.
+	//
+	// alpha: v1.17
```
sbueringer: Should be v1.18. Please also update l.57 above

pkg/manager/network.go
```diff
@@ -78,6 +81,9 @@ func GetNetworkProvider(ctx context.Context, client client.Client, networkProvid
 	case VDSNetworkProvider:
 		log.Info("Pick NetOp (VDS) network provider")
 		return network.NetOpNetworkProvider(client), nil
+	case ExternallyManagedNetworkProvider:
+		log.Info("Pick ExternallyManaged network provider")
```
sbueringer: Could it be that these logs are not especially useful anymore?

* For the static factory we nowadays log command line flags, so it will show up there
* For dynamic we will log for all providers

If I'm right, let's drop this log and the ones for the other providers as well

pkg/manager/network_factory_test.go
```diff
@@ -80,6 +82,28 @@ func TestPerClusterNetworkProviderFactory(t *testing.T) {
...
+
+	t.Run("ExternallyManaged is registered when gate is enabled", func(t *testing.T) {
+		g := NewWithT(t)
+		featuregatetesting.SetFeatureGateDuringTest(t, feature.Gates, feature.ExternallyManagedProvider, true)
+		factory, err := NewPerClusterNetworkProviderFactory(ctx, c)
+		g.Expect(err).ToNot(HaveOccurred())
+		np, err := factory.ForCluster(ctx, clusterWithProvider(ExternallyManagedNetworkProvider))
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(np).ToNot(BeNil())
+		g.Expect(np.SupportsSupervisorService()).To(BeFalse())
```
sbueringer: *nit* maybe check np.Name() instead

pkg/manager/network_test.go
```diff
@@ -77,3 +79,16 @@ func TestConvertNetworkProviderName(t *testing.T) {
 		})
 	}
 }
+
+func TestGetNetworkProviderExternallyManaged(t *testing.T) {
```
sbueringer: I think this test doesn't add much over the new test cases in pkg/manager/network_factory_test.go. I would drop it

pkg/services/network/externally_managed_provider.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (np *externallyManagedNetworkProvider) Name() string {
+	return ExternallyManagedNetworkProviderName
+}
+
+// SupportsIPv6DualStack is unreachable for this provider: the VMService LB path
+// and ServiceDiscovery address discovery that consume it are both skipped.
+// Dual-stack is instead derived per-interface from the referenced Subnet / SubnetSet.
+func (np *externallyManagedNetworkProvider) SupportsIPv6DualStack() bool {
+	return false
```
sbueringer: I understood the comment that the return value from this func technically does not matter.

Would it be more correct to return `feature.Gates.Enabled(feature.IPv6DualStack)` as the network provider technically does support IPv6DualStack?

pkg/services/network/externally_managed_provider.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func (np *externallyManagedNetworkProvider) HasLoadBalancer() bool {
```
sbueringer, line 66: Let's add a comment why, e.g. something like what you wrote in another place: "CAPV will not create a VirtualMachineService for the control plane VIP; the Cluster’s spec.controlPlaneEndpoint is pre-set"

pkg/services/network/externally_managed_provider.go
```diff
@@ -0,0 +1,200 @@
...
+	return false
+}
+
+func (np *externallyManagedNetworkProvider) SupportsVMReadinessProbe() bool {
```
sbueringer, line 73: Similar, "No TCP-based readiness probe is configured on control plane VMs, as this probe is typically used to perform endpoint health checks for the VirtualMachineService-managed Load Balancer"

pkg/services/network/externally_managed_provider.go
```diff
@@ -0,0 +1,200 @@
...
+	deprecatedv1beta1conditions.MarkTrue(clusterCtx.VSphereCluster, vmwarev1.ClusterNetworkReadyV1Beta1Condition)
+	conditions.Set(clusterCtx.VSphereCluster, metav1.Condition{
+		Type:   vmwarev1.VSphereClusterNetworkReadyCondition,
+		Status: metav1.ConditionTrue,
+		Reason: vmwarev1.VSphereClusterNetworkReadyReason,
+	})
+	return nil
+}
+
+func (np *externallyManagedNetworkProvider) GetClusterNetworkName(_ context.Context, _ *vmware.ClusterContext) (string, error) {
```
sbueringer: Let's please drop this method from the interface and delete it here.

Looks like it's only "externally" used in network_test.go (if I see correctly) and there we can do an additional cast, e.g. something like this:
```go
np.(interface{ GetClusterNetworkName(_ context.Context, _ *vmware.ClusterContext) (string, error) }).GetClusterNetworkName()
```

pkg/services/network/externally_managed_provider.go
```diff
@@ -0,0 +1,200 @@
...
+	return "", nil
+}
+
+func (np *externallyManagedNetworkProvider) GetVMServiceAnnotations(_ context.Context, _ *vmware.ClusterContext) (map[string]string, error) {
```
sbueringer, line 97: Let's document something like "CAPV does not create a VirtualMachineService, so these annotations are unused; an empty map keeps the call site uniform with other providers."

pkg/services/network/externally_managed_provider.go
```diff
@@ -0,0 +1,200 @@
...
+	return map[string]string{}, nil
+}
+
+func (np *externallyManagedNetworkProvider) VerifyNetworkStatus(_ context.Context, _ *vmware.ClusterContext, _ runtime.Object) error {
```
sbueringer: Also only used in network_test.go, let's drop it from the interface and here

pkg/services/network/externally_managed_provider.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, nil
+	}
+}
+
+func ipamModesFromIPAddressType(ipAddressType nsxvpcv1.IPAddressType) []corev1.IPFamily {
+	switch ipAddressType {
+	case nsxvpcv1.IPAddressTypeIPv6:
+		return []corev1.IPFamily{corev1.IPv6Protocol}
+	case nsxvpcv1.IPAddressTypeIPv4IPv6:
+		return []corev1.IPFamily{corev1.IPv4Protocol, corev1.IPv6Protocol}
```
sbueringer, line 193: If I understand correctly the Subnet/SubnetSets API cannot express IPv6 primary. I assume:
a) there's nothing we can do about it
b) this is also not a problem for us?

pkg/services/network/network_test.go
```diff
@@ -1464,3 +1464,432 @@ var _ = Describe("NSXT VPC Provider DualStack", func() {
...
+
+			err := np.ConfigureVirtualMachine(ctx, clusterCtx, machine, vm)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(vm.Spec.Network.Interfaces).To(HaveLen(2))
+
+			Expect(vm.Spec.Network.Interfaces[0].Name).To(Equal(PrimaryInterfaceName))
+			Expect(*vm.Spec.Network.Interfaces[0].MTU).To(Equal(int64(1600)))
+			Expect(vm.Spec.Network.Interfaces[0].Network.Name).To(Equal("workload-subnet"))
+			Expect(vm.Spec.Network.Interfaces[0].Network.Kind).To(Equal(NetworkGVKNSXTVPCSubnet.Kind))
+			Expect(vm.Spec.Network.Interfaces[0].Routes).To(BeEmpty())
```
sbueringer, line 1625: Looks like we intentionally don't propagate the routes. Can we block setting routes with this provider in the webhook? (also for secondary)

Similar for Gateway4 / Gateway6 on primary and secondary interfaces


Do we have other fields in the API that we don't propagate with the externally managed provider?

pkg/services/network/network_test.go
```diff
@@ -1464,3 +1464,432 @@ var _ = Describe("NSXT VPC Provider DualStack", func() {
...
+			}
+
+			err := np.ConfigureVirtualMachine(ctx, clusterCtx, machine, vm)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(vm.Spec.Network.Interfaces).To(HaveLen(2))
+
+			Expect(vm.Spec.Network.Interfaces[0].Name).To(Equal(PrimaryInterfaceName))
+			Expect(*vm.Spec.Network.Interfaces[0].MTU).To(Equal(int64(1600)))
+			Expect(vm.Spec.Network.Interfaces[0].Network.Name).To(Equal("workload-subnet"))
+			Expect(vm.Spec.Network.Interfaces[0].Network.Kind).To(Equal(NetworkGVKNSXTVPCSubnet.Kind))
```
sbueringer, line 1623: maybe also verify apiVersion, same below

pkg/services/network/network_test.go
```diff
@@ -1464,3 +1464,432 @@ var _ = Describe("NSXT VPC Provider DualStack", func() {
...
+					}},
+				},
+			}
+			err := np.ConfigureVirtualMachine(ctx, clusterCtx, machine, vm)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(vm.Spec.Network.Interfaces).To(HaveLen(2))
+			Expect(vm.Spec.Network.Interfaces[0].IPAMModes).To(Equal([]corev1.IPFamily{corev1.IPv4Protocol, corev1.IPv6Protocol}))
+			Expect(vm.Spec.Network.Interfaces[1].IPAMModes).To(BeNil())
+			Expect(vm.Spec.Network.Interfaces[1].Gateway4).To(Equal("None"))
+			Expect(vm.Spec.Network.Interfaces[1].Gateway6).To(Equal("None"))
```
sbueringer, line 1828: Let's also check these fields for primary

controllers/vmware/servicediscovery_controller.go
DanielXiao: Dropped it.

feature/feature.go
DanielXiao: Yes. It works with any vmop API versions and I tested it end to end on v1alpha5.
IPAM modes is behind the IPv6DualStack feature gate.

pkg/manager/network.go
DanielXiao: Yes, they are not useful and I dropped them.

feature/feature.go
DanielXiao: Done

pkg/manager/network_factory_test.go
DanielXiao: Good catch. Thanks.

pkg/manager/network_test.go
DanielXiao: Done

pkg/services/network/externally_managed_provider.go
DanielXiao: Done

pkg/services/network/externally_managed_provider.go
DanielXiao: Done

pkg/services/network/externally_managed_provider.go
DanielXiao: Done

pkg/services/network/externally_managed_provider.go
DanielXiao: Dropped it.

pkg/services/network/externally_managed_provider.go
DanielXiao: Done

pkg/services/network/externally_managed_provider.go
DanielXiao: Dropped.

pkg/services/network/externally_managed_provider.go
DanielXiao: Subnet/SubnetSet only has IPv4, IPv6, and IPv4IPv6. Dual-stack is always mapped as [IPv4, IPv6]. Nothing we can do in CAPV; it is not a problem for this provider.

pkg/services/network/network_test.go
DanielXiao: Yes, we intentionally don't propagate the routes but still propagate networkRef, mtu and vlans (does not require or can not do in-place update). Gateway4 / Gateway6 are not on the CAPV API (vmop API). I added webhook validation against routes.

pkg/services/network/network_test.go
DanielXiao: Done

pkg/services/network/network_test.go
DanielXiao: Done

pkg/services/network/network_test.go
sbueringer: Makes sense!

fabriziopandini:
Nice work!
Thanks @sbueringer for helping with this PR

/lgtm
/approve

PR #4106, opened 2026-07-22
🐛 Add conversion webhook wiring for VSphereClusterIdentity CRD

### LINKED ISSUE #4105: VSphereClusterIdentity CRD missing conversion webhook wiring, causing permanent reconcile failure after v1beta1→v1beta2 upgrade
/kind bug

Upgraded an existing CAPV management cluster from a pre-v1beta2 release to v1.16.1 (v1beta1 + v1beta2 dual-version CRDs). Pre-existing VSphereClusterIdentity objects, created and reconciled by the old controller before the upgrade, retain their pre-migration status shape: a flat status.conditions[] array with old-style Ready/CredentialsAvailable entries where reason was legally omitted (the old Condition type's Reason field is json:"reason,omitempty").

After the upgrade, the new v1.16.1 controller reads these objects through its v1beta2 typed client (metav1.Condition, which has no omitempty on Reason). The missing reason key deserializes to Go's zero value "". Any subsequent patch by the controller re-serializes that as an explicit "reason":"", and the API server rejects it against the CRD's own schema (minLength: 1 on reason). This repeats on every reconcile, forever, since the object can never be successfully rewritten:

controller.go:495 "Reconciler error" err="failed to patch VSphereClusterIdentity <name>: failed to patch status conditions: VSphereClusterIdentity.infrastructure.cluster.x-k8s.io \"<name>\" is invalid: [status.conditions[0].reason: Invalid value: \"\": conditions[0].reason in body should be at least 1 chars long, status.conditions[1].reason: Invalid value: \"\": conditions[1].reason in body should be at least 1 chars long]"

Root cause traced to the CRD manifest itself: VSphereClusterIdentity is the only CRD in the whole project missing conversion webhook wiring. Confirmed by running "kustomize build config/default/crd" from source — every sibling CRD (VSphereDeploymentZone, VSphereCluster, VSphereFailureDomain, VSphereClusterTemplate, VSphereMachine, VSphereMachineTemplate, VSphereVM) gets a spec.conversion.strategy: Webhook block; VSphereClusterIdentity gets none (strategy: None). config/default/crd/kustomization.yaml is missing patches/webhook_in_vsphereclusteridentities.yaml and patches/cainjection_in_vsphereclusteridentities.yaml from its patch lists, and those two patch files don't exist at all — every sibling has both.

This is not a missing implementation — ConvertTo/ConvertFrom for VSphereClusterIdentity are fully implemented in api/govmomi/v1beta1/conversion.go (including the status-mirroring logic for Deprecated.V1Beta1/V1Beta2), registered on the scheme, and pass TestFuzzyConversion/for_VSphereClusterIdentity (both spoke-hub-spoke and hub-spoke-hub) in the existing test suite. It's purely the CRD-manifest-level declaration and its two kustomize patch files that were never added, unlike every other type that went through the same v1beta1→v1beta2 migration.

VSphereClusterIdentity's CRD should declare spec.conversion.strategy: Webhook pointing at the manager's /convert endpoint, same as every sibling CRD. With that in place, the API server would call the (already-implemented, already-tested) conversion webhook whenever a pre-upgrade object is read/written, correctly synthesizing the new Available/Paused conditions and populating status.deprecated.v1beta1 — the same backward-compatible bridging every other CAPV CRD already gets across a version upgrade. Pre-existing objects should not become permanently un-patchable after an upgrade.

Verified the fix is limited to manifest wiring — no Go code changes needed. Added the two missing files (mirroring webhook_in_vspheredeploymentzones.yaml / cainjection_in_vspheredeploymentzones.yaml, changing only the CRD name) and the two missing entries in config/default/crd/kustomization.yaml, then rebuilt with "kustomize build config/default/crd": the generated VSphereClusterIdentity CRD now carries the identical conversion.strategy: Webhook block as its siblings, with no other CRD in the output changed.

Confirmed end-to-end on a live testbed: patched only the vsphereclusteridentities CRD (mirroring a working sibling's clientConfig/caBundle) with no changes to any other CRD. The previously permanently-failing reconcile succeeded on the very next attempt, status.conditions now correctly reads Available: True (reason: Available) with the stale pre-upgrade entries relocated into status.deprecated.v1beta1.conditions, and dependent VSphereDeploymentZone/VSphereFailureDomain reconciliation — which had been stalled behind the unreadable identity — resumed immediately.

**Environment:**

- Cluster-api-provider-vsphere version: v1.16.1
- Kubernetes version: 1.34
- OS (e.g. from /etc/os-release): Ubuntu 22.04 / Photon OS 5.0 (management cluster containers)

RenilRaj-BR:

`VSphereClusterIdentity` is the only CRD in the project missing conversion webhook wiring — every other CRD that supports both v1beta1 and v1beta2 (`VSphereCluster`, `VSphereClusterTemplate`, `VSphereMachine`, `VSphereMachineTemplate`, `VSphereVM`, `VSphereDeploymentZone`, `VSphereFailureDomain`) has a `patches/webhook_in_<type>.yaml` and `patches/cainjection_in_<type>.yaml` in `config/default/crd/`, referenced from `config/default/crd/kustomization.yaml`. `VSphereClusterIdentity` has neither the patch files nor the kustomization entries, so its CRD is generated with `spec.conversion.strategy: None`.

The `ConvertTo`/`ConvertFrom` implementations for `VSphereClusterIdentity` already exist in `api/govmomi/v1beta1/conversion.go` (including the `status.deprecated.v1beta1`/`status.v1beta2` mirroring logic), are registered on the scheme, and are already covered by `TestFuzzyConversion/for_VSphereClusterIdentity`. The manager's webhook server also already serves a generic `/convert` handler (mounted as a side effect of any other type's webhook setup calling `Complete()`), so no controller-side code change is needed — this PR only adds the two missing manifest patch files and their two kustomization entries, mirroring `VSphereDeploymentZone` exactly.

Without this, an existing management cluster upgraded across the v1beta1→v1beta2 migration ends up with `VSphereClusterIdentity` objects stuck holding pre-migration status data (old-style `Ready`/`CredentialsAvailable` conditions with `reason` omitted, which is valid under the old schema but not the new one). Because there's no conversion webhook to bridge the two shapes, the new controller can never successfully patch the object's status again — every reconcile fails permanently with:

failed to patch VSphereClusterIdentity <name>: failed to patch status conditions: ... status.conditions[0].reason: Invalid value: "": conditions[0].reason in body should be at least 1 chars long

which in turn blocks every `VSphereCluster`/`VSphereDeploymentZone`/`VSphereFailureDomain` that depends on that identity. Verified the fix resolves this: after patching only the `vsphereclusteridentities` CRD (no other CRD changed), the previously permanently-failing reconcile succeeded on the next attempt, `status.conditions` correctly populated the merged `Available` condition, and downstream `VSphereDeploymentZone`/`VSphereFailureDomain` reconciliation resumed immediately.

Fixes #4105

### DIFF
--- a/config/default/crd/kustomization.yaml
+++ b/config/default/crd/kustomization.yaml
@@ -24,6 +24,7 @@ patches:
 - path: patches/webhook_in_vspherevms.yaml
 - path: patches/webhook_in_vspherefailuredomains.yaml
 - path: patches/webhook_in_vspheredeploymentzones.yaml
+- path: patches/webhook_in_vsphereclusteridentities.yaml
 - path: patches/webhook_in_vsphereclustertemplates.yaml
 # +kubebuilder:scaffold:crdkustomizewebhookpatch
 
@@ -35,6 +36,7 @@ patches:
 - path: patches/cainjection_in_vspherevms.yaml
 - path: patches/cainjection_in_vspherefailuredomains.yaml
 - path: patches/cainjection_in_vspheredeploymentzones.yaml
+- path: patches/cainjection_in_vsphereclusteridentities.yaml
 - path: patches/cainjection_in_vsphereclustertemplates.yaml
 # +kubebuilder:scaffold:crdkustomizecainjectionpatch
 
--- a/config/default/crd/patches/cainjection_in_vsphereclusteridentities.yaml
+++ b/config/default/crd/patches/cainjection_in_vsphereclusteridentities.yaml
@@ -1,0 +1,8 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
+# CRD conversion requires k8s 1.13 or later.
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
+  name: vsphereclusteridentities.infrastructure.cluster.x-k8s.io
--- a/config/default/crd/patches/webhook_in_vsphereclusteridentities.yaml
+++ b/config/default/crd/patches/webhook_in_vsphereclusteridentities.yaml
@@ -1,0 +1,16 @@
+# The following patch enables conversion webhook for CRD
+# CRD conversion requires k8s 1.13 or later.
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: vsphereclusteridentities.infrastructure.cluster.x-k8s.io
+spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      conversionReviewVersions: ["v1"]
+      clientConfig:
+        service:
+          namespace: system
+          name: webhook-service
+          path: /convert


sbueringer:

Thx!
/lgtm
/approve

PR #4118, opened 2026-07-27
🐛 Apply VM-wide static routes per-device in cloud-init metadata

### LINKED ISSUE #2170: Specifying "Global Static Routes" causes cloud-init to fail to parse instance metadata
/kind bug

I tried to specify "global static routes" in a clusterclass as this field was available:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/a2f8e744a7f4f72cdd5c1d33256cae49b3aa9021/apis/v1beta1/types.go#L271-L274

Upon inspecting `instance-data.json`, it appears the following instance metadata template was rendered correctly:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/a2f8e744a7f4f72cdd5c1d33256cae49b3aa9021/pkg/util/constants.go#L141-L148

```
"network": {
   "config": {
       "ethernets": {
           "id0": {
               "addresses": [
                   "192.168.117.162/24"
               ],
               "gateway4": "192.168.117.1",
               "match": {
                   "macaddress": "00:50:56:81:da:4e"
               },
               "nameservers": {
                   "addresses": [
                       "192.168.117.1"
                   ]
               },
               "set-name": "eth0",
               "wakeonlan": true
           }
       },
       "routes": [
           {
               "metric": 222,
               "to": "8.8.8.8",
               "via": "192.168.117.1"
           }
       ],
       "version": 2
   }
```

But running `journalctl -u cloud-init` reveals that cloud-init threw an exception upon finding "routes" in the network config.

After reading https://cloudinit.readthedocs.io/en/latest/reference/network-config-format-v2.html#network-config-v2, it does not appear that "routes" was a valid key under "network".

Specifying "routes" per-device works as expected, however.

Specifying global static routes should either be not supported or translated to a per-device setting somehow.

I also tried this with Ubuntu, but the machine did not acquire an IP address so I was unable to SSH in to see the problem. I suspect that the network config was invalid and so it was unable to acquire the static IP set in the instance metadata.

**Environment:**

- Cluster-api-provider-vsphere version: v1.7.0
- Kubernetes version: (use `kubectl version`): v1.26.5
- OS (e.g. from `/etc/os-release`): Photon OS 3.0

pujitha24:

Motivation:
When `VSphereVM.Spec.Network.Routes` (VM-wide/global static routes) is
set, the generated cloud-init network-config v2 metadata rendered them
as a top-level `routes:` key directly under `network:`. That is not a
valid location for `routes` in the netplan v2 schema used by cloud-init
(routes are only valid nested under an individual device). As a result,
cloud-init fails during `init-local` with `RuntimeError: No handler
found for command 'routes'`, and the guest network never comes up for
any machine using this field. Per-device routes
(`NetworkDeviceSpec.Routes`) were unaffected and already worked
correctly, since they were already rendered in the valid per-device
location.

Approach:
Instead of rendering the VM-wide routes as an invalid top-level key,
merge them into each device's own `.Routes` slice before rendering, so
they end up in the same (valid) per-device `routes:` block that
per-device routes already use. The invalid top-level template block is
removed. When `Spec.Network.Routes` is empty (the common case), the
merge is a no-op and generated metadata is unchanged.

Fixes #2170

**Validation**:
Ran `go build ./...` (passes) and `go test ./pkg/util/...` (passes),
including a new test case ("2nets+global-static-routes" in
pkg/util/machines_test.go) that verifies a VM-wide route is merged into
every device's route list alongside any pre-existing per-device routes,
and an updated existing test case that verifies the previously-invalid
top-level `routes:` output is now rendered per-device instead.

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/util/machines.go
+++ b/pkg/util/machines.go
@@ -90,6 +90,11 @@ func GetMachineMetadata(hostname string, vsphereVM infrav1.VSphereVM, ipamState 
 		// not as a top-level network property, so apply the VM-wide routes to
 		// every device instead of rendering them as an invalid top-level key.
 		devices[i].Routes = append(devices[i].Routes, vsphereVM.Spec.Network.Routes...)
+		if i == len(vsphereVM.Spec.Network.Devices)-1 {
+			for j := len(vsphereVM.Spec.Network.Devices); j < len(devices); j++ {
+				devices[j].Routes = append(devices[j].Routes, vsphereVM.Spec.Network.Routes...)
+			}
+		}
 
 		// Add the MAC Address to the network device
 		if len(networkStatuses) > i {
--- a/pkg/util/machines.go
+++ b/pkg/util/machines.go
@@ -86,6 +86,11 @@ func GetMachineMetadata(hostname string, vsphereVM infrav1.VSphereVM, ipamState 
 	for i := range vsphereVM.Spec.Network.Devices {
 		vsphereVM.Spec.Network.Devices[i].DeepCopyInto(&devices[i])
 
+		// netplan (used by cloud-init) only supports static routes per-device,
+		// not as a top-level network property, so apply the VM-wide routes to
+		// every device instead of rendering them as an invalid top-level key.
+		devices[i].Routes = append(devices[i].Routes, vsphereVM.Spec.Network.Routes...)
+
 		// Add the MAC Address to the network device
 		if len(networkStatuses) > i {
 			devices[i].MACAddr = networkStatuses[i].MACAddr
@@ -140,13 +145,11 @@ func GetMachineMetadata(hostname string, vsphereVM infrav1.VSphereVM, ipamState 
 	if err := tpl.Execute(buf, struct {
 		Hostname    string
 		Devices     []infrav1.NetworkDeviceSpec
-		Routes      []infrav1.NetworkRouteSpec
 		WaitForIPv4 bool
 		WaitForIPv6 bool
 	}{
 		Hostname:    hostname, // note that hostname determines the Kubernetes node name
 		Devices:     devices,
-		Routes:      vsphereVM.Spec.Network.Routes,
 		WaitForIPv4: waitForIPv4,
 		WaitForIPv6: waitForIPv6,
 	}); err != nil {
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -138,12 +138,4 @@ network:
         {{- end }}
       {{- end }}
     {{- end }}
-  {{- if .Routes }}
-  routes:
-  {{- range .Routes }}
-  - to: "{{ .To }}"
-    via: "{{ .Via }}"
-    metric: {{ .Metric }}
-  {{- end }}
-  {{- end }}
 `
--- a/pkg/util/machines_test.go
+++ b/pkg/util/machines_test.go
@@ -485,10 +485,84 @@ network:
       addresses:
       - "192.168.4.21"
       gateway4: "192.168.4.1"
-  routes:
-  - to: "192.168.5.1/24"
-    via: "192.168.4.254"
-    metric: 3
+      routes:
+      - to: "192.168.5.1/24"
+        via: "192.168.4.254"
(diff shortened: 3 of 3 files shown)


sbueringer:
@pujitha24 Please see: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/4118#issuecomment-5100445166

pkg/util/machines.go
```diff
@@ -86,6 +86,16 @@ func GetMachineMetadata(hostname string, vsphereVM infrav1.VSphereVM, ipamState
...
 		vsphereVM.Spec.Network.Devices[i].DeepCopyInto(&devices[i])
 
+		// netplan (used by cloud-init) only supports static routes per-device,
+		// not as a top-level network property, so apply the VM-wide routes to
+		// every device instead of rendering them as an invalid top-level key.
+		devices[i].Routes = append(devices[i].Routes, vsphereVM.Spec.Network.Routes...)
+		if i == len(vsphereVM.Spec.Network.Devices)-1 {
+			for j := len(vsphereVM.Spec.Network.Devices); j < len(devices); j++ {
+				devices[j].Routes = append(devices[j].Routes, vsphereVM.Spec.Network.Routes...)
+			}
```
sbueringer, line 96: Let's extend the new test case to also cover this code

randomvariable:
Thanks for addressing the bug, but I'm wondering if we should deprecate the global route?

sbueringer:
> Thanks for addressing the bug, but I'm wondering if we should deprecate the global route?

Not sure, maybe it's convenient that routes don't have to be duplicated into devices? 

Hm, given it never worked it looks it was not really used.

randomvariable:
> > Thanks for addressing the bug, but I'm wondering if we should deprecate the global route?
> 
> Not sure, maybe it's convenient that routes don't have to be duplicated into devices?
> 
> Hm, given it never worked it looks it was not really used.

Feedback from consumers would tell us.

@pujitha24 do you know of good use cases where the on link route entries wouldn't be sufficient?

pujitha24:
@sbueringer Good call — I extended the `more-network-statuses-than-spec-devices` test case (renamed to `more-network-statuses-than-spec-devices+global-static-routes`) to also set VM-wide `Network.Routes`, so the assertion now covers the extra-device branch (`if i == len(vsphereVM.Spec.Network.Devices)-1 { ... }`) that applies global routes to devices created purely from `networkStatuses`. I verified the test fails without that block and passes with it, so it's actually exercising the code in question. `go test ./pkg/util/...` passes locally.

sbueringer:
@pujitha24 You have to drop copilot as a co-author (https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance)

randomvariable:
The second commit still has copilot in it.

pujitha24:
@randomvariable Thanks for catching that — you're right, the `Co-authored-by: Copilot Autofix powered by AI` trailer was still on the second commit. I've rewritten that commit's message to drop the trailer (keeping the code change itself untouched — verified the file diff between old and new history is empty) and added the missing DCO `Signed-off-by`. Just double-checked with `git log` that none of the three commits on this branch carry any AI co-author trailer now, and `go build ./...` / `go test ./pkg/util/...` still pass. This will need a force-push to update the PR branch.

pujitha24:
@randomvariable good catch — the second commit's message still said "per Copilot review feedback"; I've reworded it to just "per review feedback" (no code change, verified the diff is identical before/after). Sorry for missing that the first time.

On deprecating the global route: I don't have concrete production use cases where per-device on-link routes wouldn't be sufficient — this PR was just meant to fix the existing bug (global routes rendering as an invalid top-level netplan key) rather than change the field's design. Happy to defer to you both on whether deprecation is worth pursuing as a separate follow-up.

sbueringer:
Thx!

/lgtm
/approve

I'm fine with keeping the field

PR #4122, opened 2026-07-30
🌱 Add testctl util

fabriziopandini:

Fixes #

### DIFF
--- a/hack/tools/testctl/README.md
+++ b/hack/tools/testctl/README.md
@@ -1,0 +1,417 @@
+# testctl
+
+`testctl` is a CLI tool that runs a sequence of test actions against a Cluster API management
+cluster, driven by a YAML configuration file. It is built around a small, extensible
+plugin system: each step in a config file invokes a named *plugin* that either **selects**
+a set of objects to act on (e.g. Clusters, MachineDeployments) or **executes** an action on
+them (e.g. scale, delete, upgrade).
+
+- [Usage](#usage)
+- [Config file format](#config-file-format)
+  - [Top-level shape](#top-level-shape)
+  - [`runConfig`](#runconfig)
+  - [Importing another config file](#importing-another-config-file)
+- [Plugin reference](#plugin-reference)
+- [Writing a new plugin](#writing-a-new-plugin)
+- [Known limitations](#known-limitations)
+
+## Usage
+
+```shell
+go run ./hack/tools/testctl \
+  --kubeconfig ~/.kube/config \
+  --kubeconfig-context my-management-cluster \
+  --config ./my-test.yaml
+```
+
+Flags:
+
+| Flag | Description |
+|---|---|
+| `--kubeconfig` | Path to the kubeconfig file for the management cluster. If empty, default discovery rules apply. |
+| `--kubeconfig-context` | Context to use within the kubeconfig file. If empty, the current context is used. |
+| `--config`, `-c` | Path to the YAML file describing the test sequence to run (required). |
+
+`testctl` also exposes the standard `k8s.io/component-base/logs` flags (e.g. `--v` for log
+verbosity).
+
+## Config file format
+
+### Top-level shape
+
+The whole config file is a single YAML document describing what to run. At the top level it
+must have a `run` list, and may optionally have a `runConfig`:
+
+```yaml
+run:
+  - plugin: select.cluster.testctl
+    nameRegex: "^my-cluster-"
+    run:
+      - plugin: scale.controlPlane.topology.cluster.testctl
+        replicas: 3
+runConfig:
+  concurrency: 2
+```
+
+Every entry in a `run` (or `runOnError`/`runAfter`, see below) list is a map with:
+
+- `plugin` (required): the key of the plugin to invoke, e.g. `select.cluster.testctl`. See
+  [Plugin reference](#plugin-reference) for the full list.
+- any additional fields the plugin accepts as its own config (see the plugin's config type in
+  the reference below). These are parsed strictly — an unknown field is a configuration error.
+- `run`: a nested list of plugin invocations to run against every object selected by this
+  entry. **Required** if the plugin is a selector plugin (it defines what happens with the
+  selected objects), and not allowed for a plugin that only executes an action.
+- `runOnError` / `runAfter` (both optional): lists of plugin invocations, with the same shape
+  as `run`, executed after this entry's own action (and, for selectors, after its nested `run`
+  list) completes. `runOnError` only runs if this entry (or something under it) failed;
+  `runAfter` always runs, similar to a `defer`. Both are useful for cleanup steps.
+- `runConfig` (optional): overrides for this entry and everything nested under it, see below.
+
+### `runConfig`
+
+`runConfig` can be set at the top level and/or on any individual `run`/`runOnError`/`runAfter`
+entry. Fields:
+
+| Field | Type | Default | Description |
+|---|---|---|---|
+| `limit` | int | unlimited | Max number of selected objects to consider for a selector plugin, or objects to create for `create.machineDeployment.topology.cluster.testctl`. |
+| `concurrency` | int | `1` | Number of objects to process in parallel when a selector plugin returns more than one object. |
+| `failFast` | bool | `false` | If `true`, stop scheduling new work as soon as one parallel action fails (already in-flight actions are still allowed to complete). |
+| `dryRun` | bool | `false` | Run in dry-run mode: plugins log what they would do but don't mutate the cluster. Inherited from the top-level `runConfig` if not set locally. |
+| `skipWait` | bool | `false` | Don't wait for an action to converge (e.g. for a scale-up to complete) before moving on. |
+| `debug` | bool | `false` | Prompt for confirmation before running each plugin. Inherited from the top-level `runConfig` if not set locally. |
+| `debugOnError` | bool | `false` | Prompt for confirmation after a plugin fails. Inherited from the top-level `runConfig` if not set locally. |
+| `timeout` | duration | none | Timeout applied to the plugin call (e.g. `5m`). |
+| `intervals` | map of duration lists | none | Reserved for future use — merged with the top-level `runConfig`'s map, but not currently read by any of the shipped plugins. |
+
+### Importing another config file
+
+The built-in `import.config.testctl` plugin lets you split a config across multiple files. Use
+it in place of a regular plugin entry:
+
+```yaml
+run:
+  - plugin: import.config.testctl
+    path: ./common/scale-up.yaml
+```
+
+The referenced file's content replaces this entry entirely (its `plugin` field is dropped and
+the rest of the entry is merged in). Two rules apply:
+
+- The importing entry's own config **cannot** have a `run` field — that must come from the
+  imported file.
+- If the importing entry defines its own `runConfig`, it overrides the one from the imported
+  file (rather than being merged with it).
+
+## Plugin reference
+
+Every plugin is registered under a stable key of the form `<verb>.<noun>.testctl`. Selector
+plugins add an object to the shared `TestObjects` map (keyed as noted below) so that nested
+plugins can act on it; most non-selector plugins require a specific selector as an ancestor in
+the call stack (see "Requires").
+
+| Key | Kind | Sets | Requires |
+|---|---|---|---|
+| `select.cluster.testctl` | selector | `Cluster` | — |
+| `delete.cluster.testctl` | executor | — | `select.cluster.testctl` |
+| `upgrade.cluster.testctl` | executor | — | `select.cluster.testctl` |
+| `update.controlPlaneEndpoint.cluster.testctl` | executor | — | `select.cluster.testctl` |
+| `select.controlPlane.testctl` | selector | `ControlPlane` | `select.cluster.testctl` |
+| `scale.controlPlane.testctl` | executor | — | `select.controlPlane.testctl` |
+| `select.machineDeployment.testctl` | selector | `MachineDeployment` | `select.cluster.testctl` |
+| `create.machineDeployment.testctl` | executor | — | `select.cluster.testctl` |
+| `scale.machineDeployment.testctl` | executor | — | `select.machineDeployment.testctl` |
+| `delete.machineDeployment.testctl` | executor | — | `select.machineDeployment.testctl` |
+
+### `select.cluster.testctl`
+
+Go type: `plugins.ClusterSelectorPlugin`, config: `plugins.ClusterSelectorPluginConfig`.
+
+Selects the Clusters to act on. This is the entry point for every cluster-scoped chain of
+plugins and has no call-stack requirement of its own.
+
+```yaml
+- plugin: select.cluster.testctl
+  clusterSelectors:      # optional, list of label selectors; results across selectors are ORed
+    - matchLabels:
+        environment: test
+  nameRegex: "^my-cluster-"  # optional, further filters candidates by Cluster name
+  run: [...]
+```
+
+### `delete.cluster.testctl`
+
+Go type: `plugins.ClusterDeletePlugin` (no config).
+
+Deletes the selected Cluster and waits for it to be removed (unless `skipWait` is set).
+
+```yaml
+- plugin: select.cluster.testctl
+  run:
+    - plugin: delete.cluster.testctl
+```
+
+### `upgrade.cluster.testctl`
+
+Go type: `plugins.ClusterUpgradePlugin`, config: `plugins.ClusterUpgradePluginConfig`.
+
+Patches `spec.topology.version` on the selected Cluster and waits for the control plane and
+all MachineDeployments to roll out to the new version. Requires the Cluster to use
+ClusterClass (`spec.topology` must be defined).
+
+```yaml
+- plugin: select.cluster.testctl
+  run:
+    - plugin: upgrade.cluster.testctl
+      version: v1.31.0
+```
+
+### `update.controlPlaneEndpoint.cluster.testctl`
+
+Go type: `plugins.ClusterControlPlaneEndpointPlugin`, config:
+`plugins.ClusterControlPlaneEndpointPluginConfig`.
+
+Starts or stops the in-memory control plane endpoint listener for an in-memory-backed
+Cluster, by calling a local HTTP endpoint (`http://127.0.0.1:19000`). 
+
+```yaml
+- plugin: select.cluster.testctl
+  run:
+    - plugin: update.controlPlaneEndpoint.cluster.testctl
+      status: NotRunning   # "Running" (default) or "NotRunning"
+```
+
+### `select.controlPlane.testctl`
+
+Go type: `plugins.ControlPlaneSelectorPlugin` (no config).
+
+Selects the Cluster's control plane. Only `KubeadmControlPlane` is currently supported.
+
+```yaml
+- plugin: select.cluster.testctl
+  run:
+    - plugin: select.controlPlane.testctl
+      run: [...]
+```
+
+### `scale.controlPlane.testctl`
+
+Go type: `plugins.ControlPlaneScalePlugin`, config: `plugins.ControlPlaneScalePluginConfig`.
+
+Patches `spec.topology.controlPlane.replicas` on the Cluster and waits for the
+KubeadmControlPlane to reach the target replica count. Requires ClusterClass
+(`spec.topology` must be defined).
+
+```yaml
+- plugin: select.controlPlane.testctl
+  run:
+    - plugin: scale.controlPlane.testctl
+      replicas: 3
+```
+
+### `select.machineDeployment.testctl`
+
+Go type: `plugins.MachineDeploymentSelectorPlugin`, config:
+`plugins.MachineDeploymentSelectorPluginConfig`.
+
+Selects the Cluster's MachineDeployments.
+
+```yaml
+- plugin: select.cluster.testctl
+  run:
+    - plugin: select.machineDeployment.testctl
+      topologyNameRegex: "^workers-"   # optional, filters by topology name
+      run: [...]
+```
+
+### `create.machineDeployment.testctl`
+
+Go type: `plugins.MachineDeploymentCreatePlugin`, config:
+`plugins.MachineDeploymentCreatePluginConfig`.
+
+Appends one or more MachineDeployment topologies to `spec.topology.workers.machineDeployments`
+and waits for the resulting MachineDeployments to come up. Requires ClusterClass
+(`spec.topology` must be defined).
+
+```yaml
+- plugin: select.cluster.testctl
+  run:
+    - plugin: create.machineDeployment.testctl
+      count: 2                  # how many MachineDeploymentTopologies to create
+      generateName: workers-    # prefix used to build each topology's name (workers-1, workers-2, ...)
+                                 # only used if template.name is not set
+      template:                 # a clusterv1.MachineDeploymentTopology
+        class: default-worker
+        replicas: 1
+```
+
+### `scale.machineDeployment.testctl`
+
+Go type: `plugins.MachineDeploymentScalePlugin`, config:
+`plugins.MachineDeploymentScalePluginConfig`.
+
+Patches the replica count of the selected MachineDeployment's topology. `replicas` (absolute)
+takes precedence over `replicasDiff` (relative) if both are set.
+
+```yaml
+- plugin: select.machineDeployment.testctl
+  run:
+    - plugin: scale.machineDeployment.testctl
+      replicas: 5        # scale to this many replicas
+      # replicasDiff: 2  # or, scale by this delta relative to the current count
+```
+
+### `delete.machineDeployment.testctl`
+
+Go type: `plugins.MachineDeploymentDeletePlugin` (no config).
+
+Removes the selected MachineDeployment's topology entry from
+`spec.topology.workers.machineDeployments` and waits for the MachineDeployment object itself to
+be deleted.
+
+```yaml
+- plugin: select.machineDeployment.testctl
+  run:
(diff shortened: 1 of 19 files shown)


hack/tools/testctl/core/plugins.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, err
+	}
+
+	if config.Path == "" {
+		return nil, pkgerrors.New("path must be set when calling the importer plugin")
+	}
+
+	path, err := filepath.Abs(config.Path)
+	if err != nil {
+		return nil, pkgerrors.Errorf("failed to convert %s to an absolute path", config.Path)
```
sbueringer:
```suggestion
		return nil, pkgerrors.Wrapf(err, "failed to convert %s to an absolute path", config.Path)
```

hack/tools/testctl/core/runner.go
```diff
@@ -0,0 +1,572 @@
...
+	Concurrency *int32 `json:"concurrency,omitempty"`
+
+	// failFast defines the behavior in case a test action fails.
+	// If not set, it defaults to false.
+	// Note that if concurrency is greater than 1, the system will try to complete other in-flight tests before exiting.
+	// TODO: think about using this for failures that could happen to a plugin in the middle of a list of plugins (run field). Currently it is always fail fast
+	FailFast *bool `json:"failFast,omitempty"`
+
+	// dryRun instructs plugins to run in dry run mode (no change applied).
+	// Note: SelectorPlugins are expected to return TestObjects no matter of the value of this field.
```
sbueringer, line 66: Let's also document the default behavior if the field is not set, like above

(same for other fields below)

hack/tools/testctl/core/runner.go
```diff
@@ -0,0 +1,572 @@
...
+// Run the tests defined in the config using the given plugins.
+func Run(ctx context.Context, c client.Client, config []byte, plugins map[string]any) error {
+	for k, plugin := range plugins {
+		if _, ok := plugin.(SelectorPlugin); ok {
+			continue
+		}
+		if _, ok := plugin.(ExecutorPlugin); ok {
+			continue
+		}
+		return pkgerrors.Errorf("plugin %s must implement one of SelectorPlugin, ExecutorPlugin", k)
```
sbueringer: I think this validation logic allows to implement both interfaces in one plugin. Is this intentional? If not, should we block it?

hack/tools/testctl/main.go
```diff
@@ -0,0 +1,146 @@
...
+
+	restConfig, err := clientcmd.NewDefaultClientConfig(*config, &clientcmd.ConfigOverrides{
+		Context: *context,
+	}).ClientConfig()
+	if err != nil {
+		return pkgerrors.Wrapf(err, "failed to create rest config from %q", configLoadingRules.GetExplicitFile())
+	}
+	restConfig.UserAgent = "testctl"
+	restConfig.QPS = 20
+	restConfig.Burst = 100
```
sbueringer: Is that enough? (our controllers are running with 1k & 2k to avoid client-side rate-limiting in scale tests)

hack/tools/testctl/core/runner.go
```diff
@@ -0,0 +1,572 @@
...
+		configs: make(map[string]any),
+	}
+}
+
+// parseConfig parses the config file passed to the runner as input.
+// Note: during this operation, config fragments are stored in runner.configs so they can be retrieved later at run stage.
+func (r *runner) parseConfig(ctx context.Context, path *field.Path, plugin any, callStack []string, rawConfig []byte) error {
+	// TODO: in future we might want to injecting env variables / use templating
+
+	var err error
```
sbueringer:
```suggestion
```
nit. Let's define this var locally where it is needed?

hack/tools/testctl/core/runner.go
```diff
@@ -0,0 +1,572 @@
...
+	if validator, ok := plugin.(CallStackValidatorPlugin); ok {
+		if err := validator.ValidateCallStack(ctx, callStack); err != nil {
+			return pkgerrors.Wrapf(err, "failed to validate call stack for %s", path.String())
+		}
+	}
+
+	// If this is a configurable plugin, read the custom plugin config.
+	// Note: When invoking plugin's ParseConfig method, drop all the fields that are managed from the runner itself,
+	// so the plugin authors can use Unmarshal strict.
+	if configurable, ok := plugin.(ConfigurablePlugin); ok {
```
sbueringer: What about also failing here if the plugin is not configurable but has some config?

hack/tools/testctl/core/runner.go
```diff
@@ -0,0 +1,572 @@
...
+	if _, ok := plugin.(*importerPlugin); ok {
+		return
+	}
+
+	if strings.HasSuffix(msgPrefix, " ") {
+		msgPrefix += " "
+	}
+	if strings.HasPrefix(msgSuffix, " ") {
+		msgSuffix = " " + msgSuffix
+	}
```
sbueringer: Why is this needed? (maybe let's add a godoc)

hack/tools/testctl/core/runner.go
```diff
@@ -0,0 +1,572 @@
...
+			return nil, pkgerrors.Wrapf(err, "failed to marshal %s.runConfig", path)
+		}
+
+		if err := yaml.UnmarshalStrict(cfgByte, &runConfig); err != nil {
+			return nil, pkgerrors.Wrapf(err, "failed to unmarshal %s.runConfig", path)
+		}
+	}
+
+	// Use top level runConfig as a default when it makes sense.
+	if path.String() != rootPath.String() {
```
sbueringer, line 487: I don't follow why we are defaulting these fields, but not the other ones in the RunConfig struct

hack/tools/testctl/core/runner.go
```diff
@@ -0,0 +1,572 @@
...
+	if generator, ok := plugin.(MessageGeneratorPlugin); ok {
+		if generatedMessage, err := generator.GenerateMessage(objects, pluginConfig); err == nil {
+			msg = generatedMessage
+		}
+	}
+	r.prompter(ctx, fmt.Sprintf("%s%s%s", msgPrefix, msg, msgSuffix))
+}
+
+// getRunConfig parses a runConfig field if defined, otherwise returns a default RunConfig.
+func (r *runner) getRunConfig(_ context.Context, path *field.Path, rawConfig map[string]any) (*RunConfig, error) {
```
sbueringer, line 472: For my understanding. This is reading the runConfig on the current level and is defaulting with the root one, but it does not inherit the values from in between, right?

hack/tools/testctl/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+A more complete example, scaling up a Cluster's control plane and workers, then rolling back
+on failure, then upgrading:
+
+```yaml
+runConfig:
+  concurrency: 2
+  dryRun: false
+
+run:
```
sbueringer, line 293: Maybe slightly extend to show how runConfig can be set inside the plugin hierarchy

hack/tools/testctl/core/runner.go
```diff
@@ -0,0 +1,572 @@
...
+var rootPath = field.NewPath("$")
+
+// TestObjects is a map of key/objects to be passed to a plugin, e.g. "Cluster":..., "MachineDeployment":...
+// Note: We are using this approach instead of adding info to the context because it is easier to debug.
+type TestObjects map[string]any
+
+// RunConfig defines the run config to be used when invoking plugins.
+// Note: For the sake of simplicity there is a unique RunConfig for all the plugins; as a consequence some field in this struct
+// might not be relevant for a specific plugin. Some field in this struct are managed by the runner itself, e.g. concurrency.
+type RunConfig struct {
```
sbueringer, line 48: We probably don't want to change it as this point, but I assume the code in this file would be simpler if the config file would be ~ mostly typed.

Let's make sure the example YAML in README.md covers all relevant config fields/variants. That should help to make up for the lack of full config structs that could be used as reference when writing a config file

hack/tools/testctl/core/runner_test.go
```diff
@@ -0,0 +1,869 @@
...
+			g.Expect(plugin).ToNot(BeNil())
+
+			rawConfig, err := yaml.Marshal(tt.config)
+			g.Expect(err).ToNot(HaveOccurred())
+
+			// TODO: check error message
+			err = r.parseConfig(t.Context(), rootPath, plugin, nil, rawConfig)
+			g.Expect(err).ToNot(HaveOccurred())
+
+			// TODO: check error message
```
sbueringer: Seems implemented

hack/tools/testctl/plugins/cluster_delete.go
```diff
@@ -0,0 +1,120 @@
...
+	cluster, err := p.unwrapTestObjects(objects)
+	if err != nil {
+		return err
+	}
+
+	log := ctrl.LoggerFrom(ctx).WithValues("Cluster", klog.KObj(cluster))
+	ctx = ctrl.LoggerInto(ctx, log)
+
+	if !cluster.DeletionTimestamp.IsZero() {
+		log.Info("Upgrade Cluster action skipped, Cluster already is deleting")
```
sbueringer: Should this be "Delete cluster action skipped, ..."

hack/tools/testctl/plugins/cluster_delete.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil
+	}
+
+	log.Info("Deleting Cluster")
+	if ptr.Deref(runConfig.DryRun, false) {
+		return nil
+	}
+
+	if err := c.Delete(ctx, cluster); err != nil {
```
sbueringer, line 63: maybe handle NotFound errors and return directly if one occurs?

hack/tools/testctl/plugins/cluster_delete.go
```diff
@@ -0,0 +1,120 @@
...
+		return pkgerrors.Wrapf(err, "failed to delete Cluster %s", klog.KObj(cluster))
+	}
+
+	if ptr.Deref(runConfig.SkipWait, false) {
+		return nil
+	}
+
+	log.Info("Waiting for Cluster to be deleted")
+	if err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 5*time.Minute, false, func(ctx context.Context) (done bool, err error) {
```
sbueringer: should this use the timeout from the runConfig? (same in other files, e.g. cluster_upgrade.go)

hack/tools/testctl/plugins/cluster_selector.go
```diff
@@ -0,0 +1,121 @@
...
+func (p *ClusterSelectorPlugin) unwrapConfig(pluginConfigUntyped any) *ClusterSelectorPluginConfig {
+	config := &ClusterSelectorPluginConfig{}
+	if pluginConfigUntyped != nil {
+		config = pluginConfigUntyped.(*ClusterSelectorPluginConfig)
+	}
+	return config
+}
+
+// UnwrapClusterTestObject extracts the Cluster from TestObjects.
+func UnwrapClusterTestObject(objects core.TestObjects) (*clusterv1.Cluster, error) {
```
sbueringer, line 110: Maybe let's move this to util.go? It's used in a lot of places but not in this file

hack/tools/testctl/plugins/cluster_upgrade.go
```diff
@@ -0,0 +1,169 @@
...
+
+// ValidateCallStack validates the call stack for the plugin.
+func (p ClusterUpgradePlugin) ValidateCallStack(_ context.Context, callStack []string) error {
+	if !slices.Contains(callStack, clusterSelectorPluginKey) {
+		return pkgerrors.Errorf("this plugin can only be called as a child of %s", clusterSelectorPluginKey)
+	}
+	return nil
+}
+
+func (p ClusterUpgradePlugin) unwrapConfigAndTestObjects(objects core.TestObjects, pluginConfigUntyped any) (*ClusterUpgradePluginConfig, *clusterv1.Cluster, error) {
```
sbueringer: Up to you but I think instead of having duplicate methods we could probably have one generic func in util.go

hack/tools/testctl/plugins/cluster_delete.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return err
+	}
+
+	log := ctrl.LoggerFrom(ctx).WithValues("Cluster", klog.KObj(cluster))
+	ctx = ctrl.LoggerInto(ctx, log)
+
+	if !cluster.DeletionTimestamp.IsZero() {
+		log.Info("Upgrade Cluster action skipped, Cluster already is deleting")
+		return nil
```
sbueringer: Should this still wait if skipWait is false? (similar in other cases like cluster_upgrade.go)

I think controlplane_scale.go is already implemented this way.

Would be good to have this consistent one way or the other

hack/tools/testctl/plugins/controlplane_scale.go
```diff
@@ -0,0 +1,151 @@
...
+	log.Info(fmt.Sprintf("Scaling KubeadmControlPlane from %d to %d replicas", currentReplicas, replicas))
+	if ptr.Deref(runConfig.DryRun, false) {
+		return nil
+	}
+
+	original := cluster.DeepCopy()
+	cluster.Spec.Topology.ControlPlane.Replicas = ptr.To(replicas)
+	if err := c.Patch(ctx, cluster, client.MergeFrom(original)); err != nil {
+		return pkgerrors.Wrapf(err, "failed to patch Cluster %s", klog.KObj(cluster))
+	}
```
sbueringer: Should this Patch be skipped if replicas is already at the correct value?

hack/tools/testctl/plugins/controlplane_selector.go
```diff
@@ -0,0 +1,113 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"sigs.k8s.io/cluster-api-provider-vsphere/hack/tools/testctl/core"
+)
+
+const (
+	controlPlaneSelectorPluginKey = "select.controlPlane.testctl"
+
+	// controlPlaneObjectKey is the key used to identify the ControlPlane object in TestObjects.
+	controlPlaneObjectKey = "ControlPlane"
```
sbueringer: Probably export this as well as ClusterObjectKey is already exported? (or unexport both)

(same for machineDeploymentSelectorObjectKey)

hack/tools/testctl/plugins/machinedeployment_create.go
```diff
@@ -0,0 +1,190 @@
...
+	Template clusterv1.MachineDeploymentTopology `json:"template,omitempty"`
+}
+
+func init() {
+	registerPlugin(machineDeploymentCreatePluginKey, &MachineDeploymentCreatePlugin{})
+}
+
+// Exec executes the plugin for the given TestObjects.
+func (p MachineDeploymentCreatePlugin) Exec(ctx context.Context, c client.Client, objects core.TestObjects, pluginConfigUntyped any, runConfig core.RunConfig) error {
+	config, cluster, err := p.getInfo(objects, pluginConfigUntyped)
```
sbueringer: Let's rename getInfo to unwrapConfigAndTestObjects. If I see correctly in other plugins it is called that

hack/tools/testctl/plugins/machinedeployment_delete.go
```diff
@@ -0,0 +1,157 @@
...
+
+// Exec executes the plugin for the given TestObjects.
+func (p MachineDeploymentDeletePlugin) Exec(ctx context.Context, c client.Client, objects core.TestObjects, _ any, runConfig core.RunConfig) error {
+	cluster, machineDeployment, mdTopologyIndex, err := p.unwrapTestObjects(objects)
+	if err != nil {
+		return err
+	}
+
+	if mdTopologyIndex == -1 {
+		// TODO: Log? MD CR still exists, but it has been already removed from the cluster.
```
sbueringer: If the MD still exists we should maybe also wait for it to go away

hack/tools/testctl/plugins/machinedeployment_delete.go
```diff
@@ -0,0 +1,157 @@
...
+var _ core.ExecutorPlugin = &MachineDeploymentDeletePlugin{}
+var _ core.MessageGeneratorPlugin = &MachineDeploymentDeletePlugin{}
+var _ core.CallStackValidatorPlugin = &MachineDeploymentDeletePlugin{}
+
+func init() {
+	registerPlugin(machineDeploymentDeletePluginKey, &MachineDeploymentDeletePlugin{})
+}
+
+// Exec executes the plugin for the given TestObjects.
+func (p MachineDeploymentDeletePlugin) Exec(ctx context.Context, c client.Client, objects core.TestObjects, _ any, runConfig core.RunConfig) error {
```
sbueringer, line 52: Is this being run in parallel for multiple MDs of the same Cluster? If yes, is this concurrency safe? (specifically the part where we delete MD by index)

Are we getting conflict errors? If yes, how do we handle them? (same for the MD scale plugin)

hack/tools/testctl/plugins/machinedeployment_delete.go
```diff
@@ -0,0 +1,157 @@
...
+	return nil
+}
+
+// GenerateMessage generates a message text for the plugin call.
+func (p MachineDeploymentDeletePlugin) GenerateMessage(objects core.TestObjects, _ any) (string, error) {
+	cluster, machineDeployment, _, err := p.unwrapTestObjects(objects)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("Delete MachineDeployments %s in Cluster %s", klog.KObj(machineDeployment), klog.KObj(cluster)), nil
```
sbueringer:
```suggestion
	return fmt.Sprintf("Delete MachineDeployment %s in Cluster %s", klog.KObj(machineDeployment), klog.KObj(cluster)), nil
```

hack/tools/testctl/plugins/machinedeployment_scale.go
```diff
@@ -0,0 +1,178 @@
...
+	log.Info(fmt.Sprintf("Scaling MachineDeployment from %d to %d replicas", currentReplicas, replicas))
+	if ptr.Deref(runConfig.DryRun, false) {
+		return nil
+	}
+
+	original := cluster.DeepCopy()
+	cluster.Spec.Topology.Workers.MachineDeployments[mdTopologyIndex].Replicas = ptr.To(replicas)
+	if err := c.Patch(ctx, cluster, client.MergeFrom(original)); err != nil {
+		return pkgerrors.Wrapf(err, "failed to patch Cluster %s", klog.KObj(cluster))
+	}
```
sbueringer: Should we skip the patch if the replicas field is already set to the correct value?

hack/tools/testctl/plugins/machinedeployment_scale.go
```diff
@@ -0,0 +1,178 @@
...
+
+// GenerateMessage generates a message text for the plugin call.
+func (p MachineDeploymentScalePlugin) GenerateMessage(objects core.TestObjects, pluginConfigUntyped any) (string, error) {
+	config, cluster, machineDeployment, mdTopologyIndex, err := p.unwrapConfigAndTestObjects(objects, pluginConfigUntyped)
+	if err != nil {
+		return "", err
+	}
+
+	replicas := targetMachineDeploymentReplicas(cluster, mdTopologyIndex, config)
+	return fmt.Sprintf("Scale MachineDeployments %s to %d replicas in Cluster %s", klog.KObj(machineDeployment), replicas, klog.KObj(cluster)), nil
```
sbueringer:
```suggestion
	return fmt.Sprintf("Scale MachineDeployment %s to %d replicas in Cluster %s", klog.KObj(machineDeployment), replicas, klog.KObj(cluster)), nil
```

hack/tools/testctl/plugins/util.go
```diff
@@ -0,0 +1,223 @@
...
+			continue
+		}
+
+		clusters = append(clusters, &cluster)
+		if limit != nil && len(clusters) >= int(ptr.Deref(limit, 0)) {
+			break
+		}
+	}
+
+	sort.Slice(clusters, func(i, j int) bool {
```
sbueringer, line 57: I wonder if the sort should happen before we filter out in l.50-57

Same in getMachineDeployments

hack/tools/testctl/core/runner.go
fabriziopandini: Fair point.
Implementing only one is simpler to reason about, so I have blocked it for this first iteration

hack/tools/testctl/main.go
fabriziopandini: increased

hack/tools/testctl/core/runner.go
fabriziopandini: Cleaned up a bit, PTAL

hack/tools/testctl/core/runner.go
fabriziopandini: While using the tool, I found useful to be able to set some RunConfig flag globally, so I added them here (dry-run, debug, debug-on-error).

Setting other RunConfig flags globally seems to have less sense (limit, concurrency, failFast, timeout)

I have added a note to the field description to clarify ^^^, but open to suggestions

hack/tools/testctl/core/runner.go
fabriziopandini: that's correct.

Each plugin has it's own RunConfig, but I found also useful to be able to set some values globally (dry-run, debug, debug-on-error, intervals).

I think we should not make it more complex - aka introduce a fully hierarchical configuration model - in this first iteration.

hack/tools/testctl/core/runner.go
fabriziopandini: The config is intentionally extensible and only partially typed to support:
- adding plugins with their own custom configs
- combining plugins at will

The README has documentation for all the fields, I would avoid to make YAML files in README more complicated at this stage and instead have a long term goal to add a folder with concrete/reusable examples over time

hack/tools/testctl/plugins/cluster_selector.go
fabriziopandini: I think it makes more sense to have it here, near to where we set Cluster entry in the TestObjects

hack/tools/testctl/plugins/cluster_delete.go
fabriziopandini: That's a fair point, and I think we should always try to check if operation actually completed, and otherwise wait.

I have fixed for cluster delete, but if that's ok I would defer fixing for other actions because changes are more invasive and this PR is already big (I have added TODOs to keep track where changes are required)

hack/tools/testctl/plugins/controlplane_scale.go
fabriziopandini: It is already skipping patch if we are at the target state (replicas set to the right value + current replicas aligned). But ok

hack/tools/testctl/plugins/machinedeployment_delete.go
fabriziopandini: Tracked as a TODO (similar to other check if operation actually completed)

hack/tools/testctl/plugins/machinedeployment_delete.go
fabriziopandini: That's a great point.

I think that a reasonable solution for a first iteration could be to try enforce that some plugins (e.g. MD delete) can only be run without parallelism. 

This requires to extend ValidateCallStack, so I have tracked as a TODO.

WRT to MD scale plugin, I don't think the same issue apply.

Note: there will some edge case like a configuration where it first delete an MD and then it tries to scale it up or down, but also I don't think it is worth the effort to try to make this test utility config bullet proof.

hack/tools/testctl/plugins/machinedeployment_scale.go
fabriziopandini: It is already skipping patch if we are at the target state (replicas set to the right value + current replicas aligned). But ok

hack/tools/testctl/plugins/util.go
fabriziopandini: Reviewed, PTAL

hack/tools/testctl/core/runner.go
sbueringer: Thx!

hack/tools/testctl/core/runner.go
sbueringer: Alrighty! Thx!

hack/tools/testctl/core/runner.go
sbueringer: Got it, that's fine.

hack/tools/testctl/core/runner.go
sbueringer: Alright, that's okay

hack/tools/testctl/plugins/cluster_selector.go
sbueringer: Got it, makes sense

hack/tools/testctl/plugins/cluster_delete.go
sbueringer: Absolutely, that's fine

hack/tools/testctl/plugins/controlplane_scale.go
sbueringer: Yup. I was thinking about the case where replicas is set to the right value but currentReplicas != config.Replicas

hack/tools/testctl/plugins/machinedeployment_delete.go
sbueringer: I assume this was addressed with "// TODO: fail if this plugin is run in parallel" (fine for me)

sbueringer:
Thx, very nice!

/lgtm
/approve

PR #4147, opened 2026-08-13
🌱 Add tool to cleanup scale env

sbueringer:

Fixes #

### DIFF
--- a/hack/tools/cleanup-scale-env/main.go
+++ b/hack/tools/cleanup-scale-env/main.go
@@ -1,0 +1,377 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


hack/tools/cleanup-scale-env/main.go
```diff
@@ -0,0 +1,377 @@
...
+			}
+
+			if len(list.Items) == 0 {
+				continue
+			}
+
+			objectsStillExist = true
+			log.Printf("Deleting %s objects (%d still exist)", gvk.Kind, len(list.Items))
+
+			objectsToDelete := slices.DeleteFunc(list.Items, func(u unstructured.Unstructured) bool {
```
fabriziopandini, line 169: Q: should we check if the object being deleted is in one of the namespacesToDelete (maybe also in the scale namespace)?

Same below

sbueringer: We also have to delete objects in the default namespace (at least ControlPlaneEndpoint,EnvVar)

Another problem is that if we have Clusters in other namespaces (e.g. default) they will stop working anyway as we restart the vcsim Pod (at least if they are CAPV clusters that also use the vcsim mode)

I would prefer the "purge everything" approach that we have atm as it's simple and works with our current use case

fabriziopandini: ACK, let's go with the assumption that in a scale env there is only scale related stuff

sbueringer:
/override pull-cluster-api-provider-vsphere-e2e-vcsim-supervisor-main
/override pull-cluster-api-provider-vsphere-e2e-vcsim-govmomi-main

(new tool is not used in CI)

PR #4158, opened 2026-08-21
🐛 Default value for graceful shutdown timeout

### LINKED ISSUE #4156: guestSoftPowerOffTimeoutSeconds default is incorrect
/kind bug

When using `powerOffMode: trySoft` and the default value for `guestSoftPowerOffTimeoutSeconds` (which is 300), CAPV triggers soft delete but immediately after triggers a hard delete on the VM.

The soft delete should be honored and the polling should continue for 300 seconds (which is the default `guestSoftPowerOffTimeoutSeconds`. But actually the timeout is read as 300ns.

In power.go, we have this code snippet
```
	var timeout time.Duration
	if vm.Spec.GuestSoftPowerOffTimeoutSeconds != 0 {
		timeout = time.Duration(vm.Spec.GuestSoftPowerOffTimeoutSeconds) * time.Second
	} else {
		timeout = infrav1.GuestSoftPowerOffDefaultTimeoutSeconds
	}
```
The else part here is missing the `* time.Second` multiplication which results in it being interpreted as 300ns instead of 300s.

This is easily worked around by explicitly setting `guestSoftPowerOffTimeoutSeconds: 300` though.

**Environment:**

- Cluster-api-provider-vsphere version: 1.16.1 (but issue is still on main)
- Kubernetes version: (use `kubectl version`): 1.35.6
- OS (e.g. from `/etc/os-release`): photon

kunalkapoor:

The guestSoftPowerOffTimeoutSeconds value defaults to 300ns instead of 300s causing the trySoft mode of graceful shutdown to immediately timeout and cause a hard shutdown.

This commit fixes that to the correct 300s value.

Fixes #4156

### DIFF
--- a/pkg/services/govmomi/power.go
+++ b/pkg/services/govmomi/power.go
@@ -68,7 +68,7 @@ func (vms *VMService) isSoftPowerOffTimeoutExceeded(vm *infrav1.VSphereVM) bool 
 	if vm.Spec.GuestSoftPowerOffTimeoutSeconds != 0 {
 		timeout = time.Duration(vm.Spec.GuestSoftPowerOffTimeoutSeconds) * time.Second
 	} else {
-		timeout = infrav1.GuestSoftPowerOffDefaultTimeoutSeconds
+		timeout = infrav1.GuestSoftPowerOffDefaultTimeoutSeconds * time.Second
 	}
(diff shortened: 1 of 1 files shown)


sbueringer:
Nice catch, thank you very much!!

/lgtm
/approve

PR #4205, opened 2026-09-14
🌱 Improve dependabot workflow

fabriziopandini:

Fixes #


.github/workflows/pr-dependabot.yaml
```diff
@@ -13,10 +13,18 @@ on:
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
+    timeout-minutes: 15
+    # limit who can run this job to dependabot only (this job should not run accidentally in forks with branches matching the dependabot/** filter above).
+    if: github.actor == 'dependabot[bot]'
```
zhanggbj, line 27: Suggest to add `github.actor == 'dependabot[bot]' || github.event_name == 'workflow_dispatch'`, in case maintainer re-trigger the job manually.

zhanggbj (commented):
nit as above.

.github/workflows/pr-dependabot.yaml
sbueringer: We never retriggered this job manually in the past. So I would not do that (if really necessary maintainers can also push --force)

fabriziopandini: agreed, there are also dependabot commands

PR #4208, opened 2026-09-16
🌱 Add janitor allowed prefix

fabriziopandini:

Fixes #


zhanggbj (approved):
/lgtm

hack/tools/boskosctl/main.go
```diff
@@ -39,16 +39,18 @@ import (
 )
 
 var (
```
sbueringer, line 41: How does this change affect the janitor?

fabriziopandini: The e2e script fetch info from boskos, than call janitor eg to perform cleanup on the folders or resource pools specified in those info.

This change is a safeguard ensuring that janitor will perform cleanup only on well know folder or resource pools (and reject any wrong value)

fabriziopandini: discussed, we need the fix both in janitor and boskosctl

